Issue #2, opened 2017-07-18
FakeSignal Sleep


teddyrendahl:
Remove mandatory sleep from FakeSignal. Makes pswalker tests take forever

teddyrendahl:
I think we have vanquished this...

PR #5, opened 2017-07-21
Soft Limits, tests, and the phantom sleep

apra93:
Added soft limits to the mirrors along with accompanying tests. And as an added bonus, found the hidden sleep and silenced it, for good.. or until someone decides to re-enable it


hhslepicka:
Looks pretty good to me! 👍 Thank you @apra93 for digging into this rabbit hole of get_ctrlvars and pyepics!

teddyrendahl:
Closes #2!

PR #9, opened 2017-07-28
Added ImagePlugin as a simulated class and a generic SimDetector

apra93:
Relatively benign PR. Added a ImagePlugin class to the sim devices so I can have the detectors actually return images for testing psbeam. This PR also fixes all the issues with the image attribute of ImagePlugin not working properly by overriding some ophyd attributes with variants that enforce type casting.

### DIFF
--- a/pcdsdevices/epics/areadetector/plugins.py
+++ b/pcdsdevices/epics/areadetector/plugins.py
@@ -7,6 +7,7 @@ PCDS plugins and Overrides for AreaDetector Plugins.
 import logging
 
 import ophyd
+import numpy as np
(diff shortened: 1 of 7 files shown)


teddyrendahl:
Don't see any issues. `pswalker` just uses `stats2` so no way this breaks anything right?

apra93:
Additional commit was for a new class for a generic simulated detector that has an imageplugin and statsplugin component along with tests.

apra93:
And no it shouldn't affect anything. Nothing in the new commit should be touching any of the existing stuff.

Issue #8, opened 2017-07-27
HOMS X position needs to be staged


teddyrendahl:
The X and Pitch needs to be staged.

PR #10, opened 2017-07-28
ENH: Mirror motors stage to nominal position

teddyrendahl:
Stage any `OMMotor` that has a `nominal_position`. We will have to put an `@stage_decorator` on before this does anything for S.O.P

Closes: #8


hhslepicka:
👍

PR #12, opened 2017-07-28
ENH: Pim stage

ZLLentz:
Have PIMMotor move in on .stage() and out on .unstage() so we can simplify the pswalker code and more generically check for staged detectors in things like the recovery plan. We can stop relying on prep_img_motors (which is a complicated overoptimization) if we just wrap walk_to_pixel with the stage decorator. This will also help us make sure the pims come all the way out at the end of the alignment. Have stage wait for the pim to go in, but don't wait on unstage because these are basically out of the beam in half a second.

Also a small addition to the .gitignore so an egg-info directory won't show as modified.

### DIFF
--- a/tests/sim/test_sim_pim.py
+++ b/tests/sim/test_sim_pim.py
@@ -174,3 +174,12 @@ def test_PIM_noise():
     status = pim.move("IN")
     assert(status.success)
     assert(pim.sim_y.value != 5 and np.isclose(pim.sim_y.value, 5, atol=0.1))
+
+def test_PIMMotor_stage():
+    pim = PIMMotor("TEST", pos_in=5)
+    pim.move_out(wait=True)
+    pim.stage()
+    assert(pim.position == "IN")
+    pim.unstage()
+    time.sleep(0.2)
+    assert(pim.position == "OUT")
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ build
 htmlcov
 .coverage
 .cache
+
+*.egg-info
--- a/pcdsdevices/epics/pim.py
+++ b/pcdsdevices/epics/pim.py
@@ -388,6 +388,14 @@ class PIMMotor(Device):
         """
         return self.blocking
 
+    def stage(self):
+        self.move_in(wait=True)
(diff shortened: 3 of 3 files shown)


hhslepicka:
We just need to be cautious to stage only one PIM at a time, right? Otherwise for the second camera the first one will be in place... but I believe that this is something that pswalker needs to handle, right?

tests/sim/test_sim_pim.py
```diff
@@ -174,3 +174,12 @@ def test_PIM_noise():
...
     assert(status.success)
     assert(pim.sim_y.value != 5 and np.isclose(pim.sim_y.value, 5, atol=0.1))
+
+def test_PIMMotor_stage():
+    pim = PIMMotor("TEST", pos_in=5)
+    pim.move_out(wait=True)
+    pim.stage()
+    assert(pim.position == "IN")
+    pim.unstage()
+    time.sleep(0.2)
```
teddyrendahl: Why do you need this sleep?

ZLLentz: .unstage() does not wait, I have no callback for completion, and I'm not guaranteed that the simulated api won't change to do some sort of asynchronous movement later. The sleep in the sim move was 0.1s, so I picked 0.2s. If there's something clever I can do with subscribe I can switch to that instead.

teddyrendahl: Nope just curious

teddyrendahl:
@hhslepicka I think @ZLLentz is going to change `iterwalk` such that we just stage on every individual `walk_to_pixel`. Only concern is we probably don't want to `stage` / `unstage` all the AD stuff

hhslepicka:
@teddyrendahl I don't think that it would be a problem to unstage/stage the whole AD stuff...
You will stage them again when you need it... if you are not collecting data with that camera it does not really matter if the AD stuff is enabled... and I don't think that it will be a time consuming task to enable the AD plugins...

teddyrendahl:
Merging in. Not an issue until we start using `@stage_decorators` in the `iterwalk`. I trust that we will have that all worked out by then.

PR #15, opened 2017-08-02
ENH: Stage and unstage of stats2 and image1 for PIMPulnix.

### LINKED ISSUE #6: AreaDetector unstage should revert the plugins back to the original state
The plugins are being left enabled... we need to revert the AD to what it was before the data collection.

hhslepicka:
This closes #6

### DIFF
--- a/pcdsdevices/epics/areadetector/base.py
+++ b/pcdsdevices/epics/areadetector/base.py
@@ -27,4 +27,12 @@ class ADComponent(base.ADComponent, Component):
 
 
 class ADBase(base.ADBase, Device):
-    pass
+    def stage(self, *args, **kwargs):
+        ret = Device.stage(self, *args, **kwargs)
+        try:
+            self.validate_asyn_ports()
+        except RuntimeError as err:
+            self.unstage(*args, **kwargs)
+            raise err
+        return ret 
+
--- a/pcdsdevices/epics/areadetector/plugins.py
+++ b/pcdsdevices/epics/areadetector/plugins.py
@@ -8,7 +8,8 @@ import logging
 
 import ophyd
 import numpy as np
-from ophyd.device import GenerateDatumInterface
+from ophyd.device import GenerateDatumInterface, Component as C
+from ophyd.signal import EpicsSignal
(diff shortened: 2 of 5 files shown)


teddyrendahl:
This doesn't break anything for @apra93's  simulation code?

hhslepicka:
I don't think so! More tests with real devices are necessary but at
pcds-devices the tests are passing.
Tomorrow I will test with another camera and perform a complete test with
the simulation.

hhslepicka:
Tested with real cameras and also tests at pswalker are passing with the changes.
This is good for another round of reviews @teddyrendahl @apra93 @ZLLentz ...

hhslepicka:
This PR requires us to change the Gateway to accept caput to PVs with "IMAGE" on the name... otherwise stage of the whole detector will fail miserably... stage at stats2 works fine! 👍 
Attn @ZLLentz

teddyrendahl:
Looks good. Both this PR and #17 fix the issue in `run_tests`

pcdsdevices/epics/areadetector/base.py
```diff
@@ -27,4 +27,12 @@ class ADComponent(base.ADComponent, Component):
...
 class ADBase(base.ADBase, Device):
-    pass
+    def stage(self, *args, **kwargs):
+        ret = Device.stage(self, *args, **kwargs)
+        try:
+            self.validate_asyn_ports()
+        except RuntimeError as err:
+            self.unstage(*args, **kwargs)
+            raise err
+        return ret
```
apra93: Just curious, doesn't inheriting from base.ADBase already give you the stage method?

hhslepicka: It does give you the stage method but the problem is that by default the stage method at ADBase does the following:

1. validate_asyn_ports
2. super().stage(...)

This patch allows us to change the nd_array_port before checking to see if the chain of plugins is valid. Before that it was not possible.

hhslepicka: This will soon not be needed as my patch was accepted upstream: https://github.com/NSLS-II/ophyd/pull/415

PR #17, opened 2017-08-03
ENH: LightInterface and EPICS testing

teddyrendahl:
- All PCDS Devices are now using the LightInterface metaclass. This comes with an explicit dependency on lightpath, which is now in the `requirements.txt`. The defaults for the device are setup in such a way that we should never fail to have all the methods of the interface. The only thing we are risking is some incidental `NotImplementedError` 😄 
- I converted what `ophyd` was using for using_fake_epics_pv and made it more general. I wrote a ton of notes on some pitfalls I stumbled into while putting it together
- Finished a few things I missed cleanup tests
- Tested everything on the IPM device, should be ready for production use

### DIFF
--- a/pcdsdevices/epics/ipm.py
+++ b/pcdsdevices/epics/ipm.py
@@ -17,7 +17,7 @@ class IPM(IocDevice):
     diode and one for the target. This creates a scalar readback that is also
     available over EPICS.
     """
-    diode = Component(InOutStates, " :DIODE")
+    diode = Component(InOutStates,   ":DIODE")
     target = Component(TargetStates, ":TARGET")
     data = FormattedComponent(EpicsSignalRO, "{self._data}")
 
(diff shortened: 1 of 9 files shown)


pcdsdevices/epics/ipm.py
```diff
@@ -15,23 +17,29 @@ class IPM(IocDevice):
     diode and one for the target. This creates a scalar readback that is also
     available over EPICS.
     """
-    diode = Component(InOutStates, ":DIODE")
+    diode = Component(InOutStates, " :DIODE")
```
hhslepicka: Is this space before ```:DIODE``` a typo?

teddyrendahl: Nice find!

tests/conftest.py
```diff
@@ -1,106 +1,263 @@
...
+import random
+import logging
+import threading
+from functools import wraps
+
+import epics
+import numpy as np
+
+
+class FakeEpicsPV(object):
```
apra93: I like this a lot. Gives me some ideas for FakeSignal...

PR #19, opened 2017-08-08
ENH: add arg for default timeout on OffsetMirror

ZLLentz:
Subclasses of the ophyd PositionerBase class take a timeout argument that sets the default timeout on calls to move and set. Having this on our OffsetMirror and passing the argument down to pitch will allow us to make set a move timeout on a per-mirror basis in pswalker/iterwalk without picking a global timeout for all of the mirror move commands.


apra93:
All good

PR #26, opened 2017-08-24
New classes for Valves, Gauges and RVC300 (GATTN Controller)

apra93:
I wanted to put together the gas attenuator system for the PAMM but was only able to get as far as adding the gauges, reworking the valves and adding the controller for the gas attenuator. They each come with simulated counterparts and tests that run using those. 

They still need a good amount of work and testing but should be a good start for anyone that eventually wants to use them. This PR is primarily so these don't just sit in my fork until I find time to flesh them out.

Valves
- ValveBase - Base valve class that has an interlock and command to open and close
- PositionValve - Inherits from ValveBase with an added pos (position) component
- BypassValve - Inherits from Valvebase with an added bypass_mode component
- OverrideValve - Inherits from Valvebase with an added override_mode component
- N2CutoffValve - Inherits from OverrideValve, nothing else added
- ApertureValve - Inherits from PositionValve and BypassValve
- ReadbackValve - Inherits from PositionValve and OverrideValve
- GateValve - Inherits from ValveBase and IocDevice

Gauges
- GaugeBase - Base gauge class that has a pressure and status component
- Pirani - Inherits from GaugeBase with no added components
- ColdCathode - Inherits from GaugeBase with a pirani component

RVC300
- RVC300 - Base RVC 300 with the components that appear standard for the controller
- FeeRVC300 - Inherits from RVC300 with some modifications to include a couple hard IOC PVs

Lower level classes that still need to be made to complete the gas attenuator system are the pumps (turbo and kashiyama), baratron gauge, and the mks146cs (gas det controller). After that, all the high-level sections can be put together.


apra93:
I'll go through, remove sim devices that do not have any component logic and replace the tests with `using_fake_epics` which should cut this down considerably. Then I'll remake the PR

PR #27, opened 2017-08-24
FIX: Adding prettytable to requirements.txt

hhslepicka:
FIX: Defining Ophyd == 0.6.1


hhslepicka:
We can check later what to do to use the new release of Ophyd... for now we use the stable release.

hhslepicka:
Attn. @teddyrendahl @ZLLentz

PR #29, opened 2017-08-25
ENH: PVState improvements and Stoppers

teddyrendahl:
Changes
--------
* I changed the way the PVState looks at the signals to interpret the complete state. The way it was it just returned the first valid state it found. This way if multiple signals have conflicting states we will report "unknown" i.e a motor is unplugged both the `out` and `in` limit switches are tripped

* I added a generic `StateStatus` you can just give it one of the state devices and it will create a background thread that marks the status done when you reach the desired state. I felt a little like I was rewriting stuff but I couldn't really see an explicit implementation of what we wanted 

* `PPSStopper` and regular `Stopper`

Mostly requesting an explicit review because I want to make sure there is not a simpler way to do this with the device subscribes. Also @ZLLentz I wanted your opinion on whether I should switch the DeviceState to use the `StateStatus`.  It looks like it is all setup with Positioner but it is not obvious to me what is happening with the `timeout` on the `move` command? I'll let you advise.

### DIFF
--- a/pcdsdevices/epics/state.py
+++ b/pcdsdevices/epics/state.py
@@ -5,7 +5,6 @@ Module to define records that act as state getter/setters for more complicated
 devices.
 """
 import logging
-import threading
 from threading import RLock
 from keyword import iskeyword
 import time
@@ -300,6 +299,13 @@ class StateStatus(StatusBase):
     """
     Status produced by state request
 
+    The status relies on two methods of the device, first the attribute
+    `.value` should reflect the current state. Second, the status will call the
+    built-in method `subscribe` with the `event_type` explicitly set to
+    "device.SUB_STATE". This will cause the StateStatus to process whenever the
+    device changes state to avoid unnecessary polling of the associated EPICS
+    variables
+
     Parameters
     ----------
     device : obj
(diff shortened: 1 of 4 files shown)


pcdsdevices/epics/state.py
```diff
@@ -269,6 +292,83 @@ def statesrecord_class(classname, *states, doc="", has_ioc=False):
...
+        Wait for the state to reach the requested value
+        """
+        try:
+            #Starting time and value
+            current_value   = self.device.value
+            expiration_time = time.time() + timeout if timeout else None
+            #Repeatedly check value
+            while not current_value == self.desired_state:
+                logger.info("Waiting for %s to be at at state %s",
+                            self.device.name, self.desired_state)
```
ZLLentz: logger.info at 10Hz will make every log unreadable

pcdsdevices/epics/state.py
```diff
@@ -79,17 +81,38 @@ def __init__(self, prefix, *, read_attrs=None, name=None, **kwargs):
...
+                    #Update state
+                    state_value = signal_state
+
+            #Handle unaccounted readbacks 
+            except (KeyError, AssertionError):
+                state_value = 'unknown'
+                break
+
+        #If all states deferred, report as unknown
+        return state_value or 'unknown'
```
ZLLentz: I don't think this change was strictly necessary, you could do everything with the old schema, but I think it's a good change because it makes setting up the states more intuitive and it makes it easier to catch inconsistent states.

pcdsdevices/epics/state.py
teddyrendahl: Good catch. Meant for that to be above the `while`

pcdsdevices/epics/state.py
```diff
@@ -269,6 +292,83 @@ def statesrecord_class(classname, *states, doc="", has_ioc=False):
...
+            #Report success
+            self._finished(success=True)
+
+        #Catch exceptions and log
+        except Exception as exc:
+            logger.error(exc)
+
+        #Make sure thread gets garbage collected
+        finally:
+            self._state_thread = None
```
ZLLentz: I don't think looping and polling is a good approach for this. Maybe instead we can subscribe a function to SUB_STATE and mark the status as finished from inside the callback once we reach the correct state.

teddyrendahl: Thought about that. Just have to investigate if there is a good way to clear the callback at the end of thread

teddyrendahl:
Changed StateStatus to use the callback of the device. This was the right call, looks a lot simpler and more reliable than starting threads. Only downside is we assume a little more about what the device we are given looks like

teddyrendahl:
@ZLLentz Sounds good.

PR #30, opened 2017-08-25
ENH: LightInterface for slits

teddyrendahl:
Added tests for slits methods as well as making lightpath compliant. Used an interesting schema to handle the strange-ness of the slits interaction with the beam. See the notes in the class doc-string.

### DIFF
--- a/pcdsdevices/epics/slits.py
+++ b/pcdsdevices/epics/slits.py
@@ -208,7 +208,7 @@ class Slits(IocDevice):
 
     def open(self):
         """
-        Open the slits to have a 50mm aperature on each side
+        Open the slits to have a large aperature on each side
         """
         self.open_cmd.put(1)
 
--- a/pcdsdevices/epics/slits.py
+++ b/pcdsdevices/epics/slits.py
@@ -35,9 +35,15 @@ class Slits(IocDevice):
     """
     Beam slits with combined motion for center and width.
 
-    Attributes
+    Parameters
     ----------
-    nominal_aperature : float
+    prefix : str
+        The EPICS base of the motor
+
+    name : str, optional
+        The name of the offset mirror
+
+    nominal_aperature : float, optional
         Nominal slit size that will encompass the beam without blocking
 
     Notes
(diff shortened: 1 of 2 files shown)


pcdsdevices/epics/slits.py
```diff
@@ -44,13 +66,26 @@ class Slits(IocDevice):
     close_cmd = Component(EpicsSignal, ":CLOSE")
     block_cmd = Component(EpicsSignal, ":BLOCK")
 
+    #Nominal
+    nominal_aperature = 5.0
```
ZLLentz: Any thoughts on making this a parameter? Some of the scientists open the slits to 10mm. I'm not sure it really matters.

pcdsdevices/epics/slits.py
```diff
@@ -106,10 +207,28 @@ def unstage(self):
         return super().unstage()
 
     def open(self):
+        """
+        Open the slits to have a 50mm aperature on each side
```
ZLLentz: docstring typo?

pcdsdevices/epics/slits.py
```diff
@@ -44,13 +66,26 @@ class Slits(IocDevice):
...
+            read_attrs = ['xcenter', 'xwidth', 'ycenter', 'ywidth']
         super().__init__(prefix, ioc=ioc, read_attrs=read_attrs, name=name,
                          **kwargs)
 
+        #Subscribe to changes in aperature
+        self.xwidth.readback.subscribe(self._aperature_changed,
+                                       run=False)
+        self.ywidth.readback.subscribe(self._aperature_changed,
+                                       run=False)
+
```
ZLLentz: We need to go back at some point and clean up all of our classes that subscribe during the init, when they don't really need to subscribe unless someone else subscribes to them. I have a feeling it's bad for pyepics to be unnecessarily monitoring things.

teddyrendahl: Maybe correct. Not sure I want to worry about it until we actually see an issue.

pcdsdevices/epics/slits.py
teddyrendahl: MFX is all set to 50 (which is a terrible value). It is just the `OPEN.DOL*` fields that control the value so this is probably set to a more sane value in hutches where people use this record. I'll edit the docstring to not specify a value

pcdsdevices/epics/slits.py
teddyrendahl: 👍

PR #33, opened 2017-08-29
ENH: Adding base classes for lower level devices to be used in SnD

apra93:
This PR adds the base classes for the lower level devices to be used in the split and delay system. They are mostly motors and are currently mostly empty but should, in theory, provide basic functionality for their respective motors assuming they all use the motor class correctly.

The classes are currently unfinished but this seemed like a good place to make the initial PR before continuing to work on them since tests pass both here and in SnD. Below are the classes:

**Motors**

`aerotech.py`
- `AeroBase` - Base class for all the aerotech motors
- `RotationAero` - Rotation aerotech motor stage.
- `LinearAero` - Linear aerotech motor. stage.

`attocube.py`
- `EccMotor` - Base class for the attocube ECC motors
- `TranslationEcc` - Linear translation attocube motor.
- `GoniometerEcc` - ECC motor that moves the goniometer.
- `DiodeEcc` - ECC motor that inserts and removes the diodes.
- `EccController` - Class for the attocube motor controller.

`micronix.py`
- `MicronixBase` - Base class for all the micronix motors (Currenly subclasses EpicsMotor. Needs to be inplemented with real micronix IOC.).
- `VT50` - Class for the micronix VT-50 motor.

**Detectors**

`diode.py`
_Note: Deprecate or add to these classes when we implement the wave8_
- `DiodeBase` - Base class for all diodes
- `HamamatsuDiode` - Class for the Hamamatsu diode used in SnD
- `HamamatsuXMotionDiode` - Hamamatsu diode with a micronix motor for the X motion.
- `HamamatsuXYMotionCamDiode` - Hamamatsu diode with micronix motors for X and Y motion along with a GigE camera.

`cam.py`
- `GigeCam` - Cam class for the gigecam

`detectors.py`
- `GigeDetector` - Detector class for the gige camera

In addition to these classes, some changes were made to `conftest.py`. A new function, `get_classes_in_module` is added that returns the classes defined in an inputted module. This made it cleaner to do basic ophyd tests when paired with pytest's `parametrize`.

I also changed the default value for the `self._value` attribute of `FakeEpicsPV` to be `0` because `None` was causing the tests mentioned above to fail.

### DIFF
--- a/pcdsdevices/epics/aerotech.py
+++ b/pcdsdevices/epics/aerotech.py
@@ -1,0 +1,41 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Attocube devices
+"""
+############
+# Standard #
+############
+import logging
+
+###############
+# Third Party #
+###############
+
+##########
+# Module #
+##########
+from .epicsmotor import Epicsmotor
+
+logger = logging.getLogger(__name__)
+
+
+class AeroBase(EpicsMotor):
+    """
+    Base Aerotech motor class.
+    """
+    pass
+
+
+class RotationAero(AeroBase):
+    """
+    Class for the aerotech rotation stage.
+    """
+    pass
+
+
+class LinearAero(AeroBase):
+    """
+    Class for the aerotech linear stage.
+    """
+    pass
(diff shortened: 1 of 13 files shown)


teddyrendahl:
Seems harmless to merge in. I wonder how many of these objects we will actually need in the end, but I don't see any issue with instantiating them all, then deprecating what we don't need.

PR #36, opened 2017-08-30
MAINT: Moved the overrided Signal classes to pcdsdevices.signal

apra93:
I wanted to import our overridden signal classes but they were in `pcdsdevices.epics.signal`, so I moved them to `pcdsdevices.signal`. `pcdsdevices.epics.signal` now just imports the classes from  `pcdsdevices.signal` the same way the other lower level ophyd classes are structured in pcdsdevices.

I also added a couple things to `.gitignore` so it can ignore tilda files among others.


ZLLentz:
I understand moving the FakeSignal, but why are the EpicsSignal classes being moved out of the epics area?

apra93:
Hmm, I was looking at this again and I don't have a good answer. I think I initially did this to match the setup we have for the `device` and `component`, but now I'm not seeing the need to do it. I might just remove the PR entirely

PR #37, opened 2017-08-31
ENH: Make subscriptions on PIMMotor pass down to states. Needed for gui.

ZLLentz:
The subscriptions on PIMMotor weren't really doing anything meaningful. What I really wanted was to be able to subscribe to state changes. The quickest way to implement this was to simply pass the subscribe/clear_sub method calls down to the component states object.


teddyrendahl:
Seems harmless

PR #38, opened 2017-09-01
ENH: Filling out AeroBase motors. Added a more components to EpicsMotor

apra93:
`AeroBase`

Added a bunch of the aerotech-specific `components` like the fault, clear, and config PVs. There are also a number of convenience attributes and methods to make using the motors easier. Also, the `move` method now makes sure the motor is enabled first before attempting the move. I'm also setting the `high_limit_switch` and `low_limit_switch` signals to be of type `FakeSignal` for now since there are no limit switches on the motors (yet) and it is causing timeout issues. Will add them back later.

`EpicsMotor`

Added a few new `components` for the soft limits, direction, and offset along with attributes and methods to make using them easier. I also added a `move_rel` method to make relative moves along with the aliases `mv`, `mvr` and `wm`.

`micronix.py`

Deprecated this since we are actually using Aerotech controllers on them, so there is now a new class in `aerotech.py` called `DiodeAero` that will control the diode motors.


apra93:
Needed to implement a new `EccBase` class since we could no longer use the motor record.

Made a number of additions at the scientists' requests. They include adding limit setters, status method, experiment screen launchers, and spec aliases.

teddyrendahl:
Looks fine. I wonder how much of this stuff we could put in the HXRSnD repo if it is just requests from scientists but understand you are in a rush

apra93:
@teddyrendahl Yeah I spent a good amount of time debating whether I should continue doing development in this repo. 

I think the model I thought we should adopt was that devices should be implemented in the repo of the project it's being made for, and only towards the end of the project when the classes are fully developed should it be merged into `pcds-devices`. 

So I'll probably move all this stuff over before making more commits.

@ZLLentz Yeah this is done and should contain all the changes they wanted for tonight.

PR #39, opened 2017-09-05
ENH: Releasing pre-beam commits. Filling out AeroBase, EccBase. Additions to EpicsMotor

apra93:
Made a number of additions to the SnD classes for their beam time tonight. They want the system to stay stable until after they've finished their shift so it's a good time to push back to the repo.

The changes made were primarily with three classes:
 - `AeroBase`
 - `EccBase`
 - `EpicsMotor`

Changes to `AeroBase` and `EccBase` were mostly additions to actually make them useful. Signals specific to each motor were added along with convenience methods and properties to make things easier on the scientists (and us).

For `EpicsMotor` I only added things that are always present with the motor record (i.e. direction, offset). Also added some setters for the limits and a couple new convenience methods in addition to aliases to the spec commands the scientists are familiar with.

A minor change is that I deprecated `micronix.py` since we are using aerotech controllers for them and moved `VT50` to `aerotech,py` as `DiodeAero`.


apra93:
Overlaps a good amount with the previous PR, closing for now.

PR #49, opened 2017-09-12
No wait in stage

ZLLentz:
If device setup is not fast, we need to explicitly do it in our plan to avoid jamming the RunEngine. To this end, we're removing the move wait=True calls from all stage methods, replacing them with save/restore where appropriate.

### DIFF
--- a/pcdsdevices/epics/mirror.py
+++ b/pcdsdevices/epics/mirror.py
@@ -422,16 +422,6 @@ class OMMotor(Device, PositionerBase):
         """
         return (self.low_limit, self.high_limit)
 
-    def stage(self):
-        """
-        Stage the OMS motor to nominal position
-        """
-        if self.nominal_position is not None:
-            logger.debug("Moving {} to nominal aligned position"
-                         "".format(self.name))
-            self.move(self.nominal_position, wait=True)
-        super().stage()
-
     @limits.setter
     def limits(self, value):
         """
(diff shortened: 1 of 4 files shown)


teddyrendahl:
Does this need a corresponding PR In `pswalker`? We should still be doing things like starting our scans at the nominal mirror positions...

danielballan:
Per a conversation with @hhslepicka last night, I'm working on a bluesky PR to make stage asynchronous, which you could cherry-pick if you want it before we are ready to migrate to v0.10.0 (or v0.11.0, which is probably where this PR will land).

This was already in our mental roadmap. The plan is:

* Simple change that would solve your problem: Allow Devices to return a status object on `stage()`. Continue to support current (soon to be "old-style") stage methods forever.
* Related, more serious change: Make _all_ Device methods that invoke network calls into coroutines so that there are no lengthy blocking calls left anywhere on the library. This should make it much easier to write GUIs that never hang. Luckily, the RunEngine can support such a change transparently, in a non-breaking way. It will first try to treat the Device's method as a coroutine and, failing that, fall back on its current behavior of making a blocking call (a pattern used in curio). We aim to have the RunEngine support this kind of thing by 1.0 (November 1) but we probably won't start making chnages on the ophyd side until later.

ZLLentz:
Thanks for the update @danielballan, glad to hear you're already thinking about this! For the time being I don't think we're doing enough with stage/unstage, etc. to warrant picking from a pre-release branch, but these are the kinds of features that will make the RunEngine robust as a tool for user applications. I'm imagining the status object from stage will be used as something to wait on before we proceed with trigger/read messages, which would be great.

I'd also support the more drastic change of making all io calls into coroutines, I think your approach here is correct.

PR #54, opened 2017-09-14
ENH: Fully overrided the limits properties and check_value to use soft limits

apra93:
The default implementation of the limits was returning the lower and upper control limit for the `user_setpoint` signal. I changed it to actually return the limits from the `low_soft_limit` and 'high_soft_limit` signals. This makes interacting with the limits more transparent and easier to work with (when simulating at least).

I also overrode the `check_value` method to also check the soft limits of the motor rather than just the `user_setpoint` limits.

### DIFF
--- a/pcdsdevices/epics/epicsmotor.py
+++ b/pcdsdevices/epics/epicsmotor.py
@@ -32,8 +32,8 @@ class EpicsMotor(EpicsMotor, Device):
     """
     Epics motor for PCDS.
     """
-    llm = Component(EpicsSignal, ".LLM")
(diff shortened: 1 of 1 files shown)


teddyrendahl:
Is this object for our special LCLS motor record or motors in general? What inherits from it? Just curious. Code looks good.

apra93:
@teddyrendahl I believe this should work for the motor record in general. Just assumes that there are `.LLM` and `.HLM` fields in the record.

apra93:
Also, checking `pcds-devices` it seems that only `imsmotor.py` uses it.

PR #57, opened 2017-09-18
CI: Check for DOCTR deploy key before moving forward.

hhslepicka:
Attn. @teddyrendahl.
This should fix the issue that we were seeing...
It worked for at master: https://travis-ci.org/hhslepicka/pcds-devices/jobs/277045716


teddyrendahl:
Fire in the hole!

PR #58, opened 2017-09-18
CI: Fixing my mistake with the if and [[ ... ]]

hhslepicka:


teddyrendahl:
Fire in the hole!

hhslepicka:
💥 :feelsgood:

PR #62, opened 2017-09-27
ENH: Mirror branching

teddyrendahl:
Changes and tests for implementation of `BranchingInterface` for OffsetMirror


teddyrendahl:
@ZLLentz Still need optional component ... Just kind of feels out of scope for Skywalker component. I checked and the device loads fine with invalid PVs. However, tab-completion does get destroyed. Both tab and optional component should be done, but probably has to wait for full hutchPython redesign.

PR #67, opened 2017-09-28
FIX: Add state as a Fake Signal at sim/OffsetMirror

hhslepicka:
@teddyrendahl this makes pswalker tests pass...
We can think of a better approach for the optional state & mps soon.


teddyrendahl:
Want to fix `Licence` while you are at it?

hhslepicka:
Good idea! Done!

PR #68, opened 2017-09-28
FIX: Check if slits are staged before unstaging

ZLLentz:
Fixes an issue where if there is a problem in staging some other device, this one is unaffected on `unstage`.

### DIFF
--- a/pcdsdevices/epics/slits.py
+++ b/pcdsdevices/epics/slits.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 import logging
 
+from ophyd.device import Staged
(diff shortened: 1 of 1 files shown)


pcdsdevices/epics/slits.py
```diff
@@ -214,8 +215,9 @@ def unstage(self):
         @stage_wrapper to make plans involving the slits to return to their
         starting positions.
         """
-        self.xwidth.move(self.stage_cache_xwidth,wait=False)
-        self.ywidth.move(self.stage_cache_ywidth,wait=False)
+        if self._staged = Staged.yes:
```
hhslepicka: `==` ?

PR #69, opened 2017-10-02
FIX: Remove MPS and states record from OffsetMirror

teddyrendahl:
This should fix the bug that stopped the last run. The issue was I added `mps` and `states` components to the OffsetMirror. This was fine when loading the class and playing around, but when bluesky "stages" the mirror it goes through the tree of components and subcomponents calling stage on everything. This failed when attempting to stage these misconfigured components. The long term issue is not all the mirrors have MPS inputs or state summaries. 

We now have two separate classes for mirrors,`PointingMirror` and `OffsetMirror`. The ladder is for mirrors like the HOMS and XRT M3 where they are necessary for alignment but not necessarily steering between different beamlines. `PointingMirror` is for mirrors that when inserted or removed actually change which beamline gets photons delivered to it. It is expected that these mirrors have MPS inputs and an EPICS record that summarizes the state. Currently XRT M1 does not have MPS, but it really should, so it seems like that should be the next issue we address instead of making another class for the mirrors.

### DIFF
--- a/pcdsdevices/epics/mirror.py
+++ b/pcdsdevices/epics/mirror.py
@@ -607,7 +607,7 @@ class Piezo(Device, PositionerBase):
         return rep
 
 
-class OffsetMirror(Device, PositionerBase, metaclass=BranchingInterface):
+class OffsetMirror(Device, PositionerBase):
     """
     X-Ray offset mirror class for each individual mirror system used in the FEE
     and XRT. Controls for the pitch, and primary gantry x and y motors are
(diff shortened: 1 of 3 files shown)


ZLLentz:
This looks correct to me. To summarize:
- OffsetMirror would cover FEE-M1H, FEE-M2H, XRT-M3, the mirrors that stay in
- PointingMirror would cover XRT-M2, XRT-M3, the mirrors that go in and out

Is this correct?

PR #73, opened 2017-10-13
ENH: DAQ device

ZLLentz:
Configure and run the daq via pydaq, now in a Flyer object. This has been tested on the real pydaq (requires a full daq to test) and has unit tests with a simulated pydaq. This was requested by Diling for Split and Delay.

Also attached to this PR:
- Added bluesky to the testing environment because I needed to check if the Flyer worked
- Added a logging config to the run_tests.py file to help me debug. This creates a new log file for each test run and saves the most recent 10.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ htmlcov
 .cache
 
 *.egg-info
+debug.log*
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
   # Useful for debugging any issues with conda
   - conda info -a
   #Grab all dependencies
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION coverage pip wheel happi ophyd==0.6.1 pytest -c conda-forge -c lightsource2-tag -c skywalker-dev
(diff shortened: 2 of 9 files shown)


ZLLentz:
Also, is codecov glitched? The real coverage on these files are 82% for the module (omits the real connect method) and 90% for the sim. Codecov doesn't even notice that I added the module at all.

teddyrendahl:
Nice work 👍 

I added the class into the Sphinx docs, hope you don't mind. Also from what I can see, CodeCov is freaking out. @hhslepicka, as our resident CI expert, any thoughts? Maybe we need to start considering the move to Coveralls ...

Just a couple comments on the actual code:
1) Can the DAQ interface allocate? It seems to me that this is what makes the most sense for the `stage` method. 
2) For `collect` you aren't retuning any information. Maybe you should just simply return information on whether the DAQ is collecting or not? I could see a situation where a user would want to go back through event documents and see whether or not the DAQ was actually collecting at certain points along the scan.

ZLLentz:
1. The pydaq interface requires the DAQ to already be up and allocated. There is no hook to connect before allocation.
2. I like this idea. We have information such as whether or not we were recording, the DAQ run number, and the number of events recorded that could be useful to have in the document stream.

ZLLentz:
Also, for full-effort we need to figure out how to implement the feature where the DAQ stops and starts taking data at different points in a scan. Bonus points for the "calib cycle" concept that uses the "controls" argument in pydaq.

teddyrendahl:
Yeah. Put it as an issue and we'll get to it when we have a Python 3.x use case

PR #75, opened 2017-10-17
ENH: MPS Factory

teddyrendahl:
First time with the factory stuff so please let me know if this makes sense. Made a function called `mps_factory` that takes a regular pcdsdevices class, keywords -> `mps_prefix`, `veto`, `args` and `kwargs`. The function then creates a new instance of the class with a subcomponent `mps`, and passes args and kwargs to the device constructor. The idea is you can make an arbitrary mps device constructor from another class using this and `functools.partial`:

```python
MPSDevice = partial(mps_factory, 'MPSDevice', Device)
d = MPSDevice(prefix='Tst:Prefix', mps_prefix='Tst:Mps:Prefix')
```
**Also**
---
* I made `mps` versions of `Stopper` and` GateValve`. I left `PointingMirror` and `PPSStopper` with explicit `mps` components as these will always have MPS inputs
*  I changed the API so the `mps_prefix` kwarg is consistent throughout the devices. `mps` seemed too short and vague , so I unified behind the standard`mps_prefix`. This meant breaking API for the `PPSStopper` and `PointingMirror`

### DIFF
--- a/pcdsdevices/epics/mps.py
+++ b/pcdsdevices/epics/mps.py
@@ -18,7 +18,7 @@ from ..interface import MPSInterface
 ##########
 from ophyd      import Device
 from .signal    import EpicsSignal, EpicsSignalRO
-from .component import Component as C
+from .component import Component as C, FormattedComponent as FC
 
 logger = logging.getLogger(__name__)
 
@@ -26,7 +26,7 @@ logger = logging.getLogger(__name__)
 class MPS(Device, metaclass=MPSInterface):
     """
     Class to interpret MPS information
-  
+
     The intention of this class is to be used as a sub-component of a device.
     There are three major attributes of each MPS bit that are relevant to
     operations; :attr:`.faulted` , :attr:`.bypassed` and :attr:`.veto_capable`.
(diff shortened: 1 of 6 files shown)


pcdsdevices/epics/mps.py
```diff
@@ -101,3 +101,36 @@ def _fault_change(self, *args, **kwargs):
         """
         kwargs.pop('sub_type', None)
         self._run_subs(sub_type=self.SUB_FAULT_CH, **kwargs)
+
+
+def mps_factory(clsname, cls,  *args, mps_prefix=None, veto=False,  **kwargs):
```
ZLLentz: Any reason why mps_prefix is optional?

pcdsdevices/epics/mps.py
```diff
@@ -101,3 +101,36 @@ def _fault_change(self, *args, **kwargs):
...
+    veto : bool, optional
+        Whether the MPS bit is capable of veto
+
+    args :
+        Passed to device constructor
+
+    kwargs:
+        Passed to device constructor
+    """
+    comp = FC(MPS, mps_prefix, veto=veto)
```
ZLLentz: Does this need to be a FC?

pcdsdevices/epics/valve.py
```diff
@@ -279,6 +269,9 @@ def open(self, wait=False, timeout=None, **kwargs):
         return super().open(wait=wait, timeout=timeout, **kwargs)
 
 
+MPSGateValve = partial(mps_factory, 'MPSGateValve', GateValve)
+MPSStopper   = partial(mps_factory, 'MPSStopper', Stopper)
```
ZLLentz: I'm not 100% sure I like how this comes together down here, but I don't think I have any better alternatives. The other way would have been to have a subclassed `__init__` method that sets the `_mps_prefix` attribute and then have a fixed `{self._mps_prefix}` prefix in the FormattedComponent. This would let mps_factory be used directly to make a class but it makes the implementation of mpg_factory more messy.

pcdsdevices/epics/mps.py
teddyrendahl: Thought this was the way to not add the master `prefix`?

pcdsdevices/epics/valve.py
teddyrendahl: Yeah I wanted to avoid subclassing `__init__`. Seemed like it could get messy. If we find we need explicit class creation I can go back and change it, but for now I'd argue for the cleaner implementation

pcdsdevices/epics/mps.py
teddyrendahl: Wanted to make sure we don't catch a stray `arg` and put it as `mps_prefix`?  But maybe I still don't avoid that... I'll change it to mandatory.

PR #77, opened 2017-10-19
ENH: Add basic reflaser with lightinterface

ZLLentz:
In our quest to instantiate all the config entries for lightpath


teddyrendahl:
This ready for merge?

ZLLentz:
I expect that this is implemented as I'd like to the very minimum needed for the handful of reflaser-only stacks. The only changes since we first looked at this is that there is a generic inout light device that can be used for other things that we just put in/out of the beam and block the path. There still is no states IOC for the timetool/reflaser stack on the XPP mono line.

PR #80, opened 2017-10-19
MAINT: Devices Odds and Ends

teddyrendahl:
* Created an IPMMotors class. This is all the lightpath should need, the full IPM with option for `data` key still exists with IPMMotors as a subclass
* I changed the Attenuator prefix handling.  Having just prefix `XPP` or `MFX` in the `happi` config seems like a bad idea, so you must now specify `MFX:ATT:COM`. I also fixed it so that the filter_prefix was being instantiated correctly, as it was not previously
* PickerBlade had no leading `:` for the signal specifications

### DIFF
--- a/pcdsdevices/epics/attenuator.py
+++ b/pcdsdevices/epics/attenuator.py
@@ -490,7 +490,9 @@ class AttenuatorBase(BasicAttenuatorBase):
 
     def __init__(self, prefix, *, name=None, read_attrs=None,
                  stage_setting=0, **kwargs):
-        self._filter_prefix = prefix.rstrip(':COM')
+        #Modify prefixes
+        self._filter_prefix = prefix
+        prefix += ':COM'
         if read_attrs is None:
             read_attrs = ["transmission", "transmission_3rd"]
         super().__init__(prefix, name=name, read_attrs=read_attrs,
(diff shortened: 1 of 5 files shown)


pcdsdevices/epics/attenuator.py
```diff
@@ -491,8 +491,7 @@ class does not include any filters.
 
     def __init__(self, prefix, *, name=None, read_attrs=None, ioc="",
                  stage_setting=0, **kwargs):
-        prefix = prefix + ":ATT:COM"
-        self._filter_prefix = prefix + ":ATT"
+        self._filter_prefix = prefix.rstrip(':COM')
```
ZLLentz: Can we meet halfway on this one? It was set up as it was before because the ioc literally enforces the `ATT:COM` part of the prefix. I agree that passing in just `XPP` is a bad move, but can we make it `XPP:ATT`?

Excerpt from IOC template file:
record(calcout, $(HUTCH):ATT:10:_SET_IN) {
  field(OUT, "$(DEVSTATEBASE):ATT:10:GO_IN PP")
  field(CALC, "(A>>B)&1")
  field(DOPT, "Use CALC")
  field(OOPT, "When Non-zero")
  field(B, "9") 
}

record(longout, $(HUTCH):ATT:COM:_C_GO) {
  field(DESC, "BitVector-Ceiling")
}

teddyrendahl: Sounds good.

PR #82, opened 2017-10-24
ENH: LODCM

ZLLentz:
Please don't merge yet unless we need it fast, this branch still needs docs and tests.

This is the LODCM fit into the branching interface. It has achieved the minimal amount of features needed for lightpath while having room to expand the class in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
   # Useful for debugging any issues with conda
   - conda info -a
   #Grab all dependencies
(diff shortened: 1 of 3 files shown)


ZLLentz:
This is probably ok and ready to go now. It could use some robustness testing with LODCM states to make sure the destinations readbacks are correct, but I'm moderately confident.

ZLLentz:
Hold on, completely forgot about hooking the top-level subscribe method into the component updates. A few more tweaks needed.

teddyrendahl:
Can you add the `hypothesis` fix? Build is failing ...

PR #85, opened 2017-10-25
ENH: Add raw input value to states

ZLLentz:
Pull the raw source value into factory-created states classes. This can be compared against the setpoints to diagnose an Unknown state. It works by checking the CALC record value being used to output a state string.

I made this addition because all the LODCM states are out of whack and I didn't have a quick way to see why. This makes it very easy to check the raw position from an instantiated states object and may be useful for other devices.


teddyrendahl:
Seems useful and un-intrusive 👍

PR #86, opened 2017-10-26
FIX: Simplify LODCM pathing

ZLLentz:
Ignore h2n, the states are unreliable. Being an expert device, we assume that beam is passing through the mono entirely if the diagnostics are not blocking and a crystal is inserted at h1n.


teddyrendahl:
👍  I think we need to remain diligent about simplifying IOCs and keeping the Python clean to add higher level methods

PR #87, opened 2017-10-27
ENH: DAQ Calibcycles

ZLLentz:
Between 'create' and 'save' documents, the daq will run. There is an option to use original run-forever features. The 'controls' argument has been reworked to be very similar to the good parts of old python.

Misc bugfixes in cleanup, using correct syntax now instead of bad confluence API docs. Added a bunch of debug statements.

### DIFF
--- a/pcdsdevices/daq.py
+++ b/pcdsdevices/daq.py
@@ -89,6 +89,7 @@ class Daq(FlyerInterface):
     @property
     def state(self):
         if self.connected:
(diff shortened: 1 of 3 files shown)


ZLLentz:
This is failing, but it works in the real setup. Pulling so @apra93 can use in the SND tonight. I will take care of the failing checks in a PR soon. There is likely a problem with one of my test objects.

PR #88, opened 2017-10-27
FIX: Testing robustness and edge cases

ZLLentz:
Changed sim daq to fail in the new ways we discovered from real daq
Remove unnecessary checks from the calibcycle tests
Cover edge case where first calibcycle could fail if starts too soon


teddyrendahl:
DAQ Simulator complete with crashes and freezes! Does it report `Allocation Error: 0x02`!

PR #89, opened 2017-10-27
API: Light subscriptions

### LINKED ISSUE #31: ENH: Subscription Handling
Many of the devices subscribe to child EpicsSignals to form higher level subscription events. i.e if an attenuator blade is moved, the Attenuator object triggers their own subscriptions that alert users that any blade has moved. 

These child subscriptions should only be made if the parent device has a relevant higher level subscription to prevent unnecessary traffic through `pyepics`

teddyrendahl:
Lots of files touched, very minimal changes.

## Subscription Changes
I was having the issue that many of the lightpath subscription updates were too aggressive. This was largely because `lightpath` just uses the `default_sub`. I opted to make every lightpath device have a subscription called `SUB_STATE`, that we don't have to worry about keeping `default_sub` as a low-volume subscription and can use it for whatever is intuitive for the device.

Also, I took the opportunity to unify how the devices do subscriptions. When we first started making devices we sometimes did it as follows:

```python
class MyDevice:
      state = C(DeviceState, ':STATE')
    
      def subscribe(self, *args, **kwargs):
             self.state.subscribe(*args, **kwargs)
```
This seemed kind of hacky as that prevented us from having any other subscriptions on the device. Now I made everything use the model

```python
class MyDevice:
      state = C(DeviceState, ':STATE')
      SUB_STATE = 'sub_state_changed'
              self.state.subscribe(self._on_state_change, run=False)

      def _on_state_change(self, **kwargs):
             kwargs.pop('sub_type', None)
             self._run_subs(sub_type=self.SUB_STATE, **kwargs)
```

That being said. I had to rewrite this a TON of times. I think we need to revisit having a base `LightDevice` that implements a lot of this stuff. This way we can avoid such a large PR and not have to worry about uniformity so much

Closes #31 

## Attenuator 
Also, I was having trouble with the attenuator class. For all the functions that needed to iterate over the filter components, it was referring to the `:NATT` PV. I now just cache the value under `number_of_filters`, this way we don't spam that PV.

### DIFF
--- a/pcdsdevices/epics/attenuator.py
+++ b/pcdsdevices/epics/attenuator.py
@@ -445,6 +445,7 @@ class BasicAttenuatorBase(Device):
             #Subscribe to all child filter objects
             for filt in self.filters:
                 filt.state_sig.subscribe(self._blade_moved, run=False)
+            self._has_subscribed = True
         super().subscribe(cb, event_type=event_type, run=run)
 
 
--- a/pcdsdevices/epics/attenuator.py
+++ b/pcdsdevices/epics/attenuator.py
@@ -157,14 +157,12 @@ class BasicAttenuatorBase(Device):
                  stage_setting=0, **kwargs):
         self._set_lock = RLock()
         self._stage_setting = stage_setting
+        self._has_subscribed = False
+
         if read_attrs is None:
             read_attrs = ["transmission"]
         super().__init__(prefix, name=name, read_attrs=read_attrs,
                          **kwargs)
-
-        #Subscribe to all child filter objects
-        for filt in self.filters:
-            filt.state_sig.subscribe(self._blade_moved, run=False)
 
 
     def __call__(self, transmission=None, **kwargs):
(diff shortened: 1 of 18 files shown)


ZLLentz:
I'm going to read this very carefully later, but one quick thought: I think we should do one better in the subscription model and never subscribe to sub-devices in `__init__`. This can potentially call `_reinitialize_pv` in a signal, and we need to avoid starting EPICS connections in `__init__` statements.

We can work around this by only starting the subscription in `subscribe` when it is first needed. We don't need to close old subscriptions.

teddyrendahl:
@ZLLentz I agree. I'll move the `self.state.subscribe` out of `__init__` you see a way we can make a single object / metaclass for this stuff? Not incredibly obvious to me.

ZLLentz:
@teddyrendahl I see some ways to make a parent class but they don't make the implementation cleaner. Maybe we revisit this later.

teddyrendahl:
@ZLLentz This should be ready for review. Also added a fix for `PPSStopper`, you can pass in `in_state` and `out_state`  to accomodate for `OUT`, `REMOVED` and `NOT_IN` enums ....

pcdsdevices/epics/attenuator.py
```diff
@@ -427,13 +426,34 @@ def remove(self, wait=False, timeout=None):
...
+        event_type : str, optional
+            Type of event to run callback on
+
+        run : bool, optional
+            Run the callback immediatelly
+        """
+        if not self._has_subscribed:
+            #Subscribe to all child filter objects
+            for filt in self.filters:
+                filt.state_sig.subscribe(self._blade_moved, run=False)
```
ZLLentz: did you forget a `self._has_subscribed = True` here?

PR #91, opened 2017-10-31
FIX: DAQ Class Robustness

ZLLentz:
- Duration arg now works
- Class will wait to be ready to begin rather than attempting to begin early and stalling forever
- Prevent user from choosing duration < 1 because these cause internal issues in pydaq... sometimes
- Prevent user from configuring during a run because this breaks the session
- Updates to sim class to reflect some things I've learned about pydaq

### DIFF
--- a/pcdsdevices/daq.py
+++ b/pcdsdevices/daq.py
@@ -26,6 +26,9 @@ except ImportError:
     logger.warning(('bluesky not in environment. Will not have '
                     'make_daq_run_engine.'))
 
+# Wait up to this many seconds for daq to be ready for a begin call
+BEGIN_TIMEOUT = 2
+
 
 # Wrapper to make sure we're connected
 def check_connect(f):
@@ -168,7 +171,7 @@ class Daq(FlyerInterface):
                      events, duration, wait)
         begin_status = self.kickoff(events=events, duration=duration,
                                     use_l3t=use_l3t, controls=controls)
-        begin_status.wait()
+        begin_status.wait(timeout=BEGIN_TIMEOUT)
         if wait:
             self.wait()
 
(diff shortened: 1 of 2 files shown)


pcdsdevices/daq.py
```diff
@@ -168,7 +171,7 @@ def begin(self, events=None, duration=None, use_l3t=None, controls=None,
                      events, duration, wait)
         begin_status = self.kickoff(events=events, duration=duration,
                                     use_l3t=use_l3t, controls=controls)
-        begin_status.wait()
+        begin_status.wait(timeout=BEGIN_TIMEOUT)
```
teddyrendahl: Think there is a use case for having a `kwarg` for this wait period? Sometimes you might want to check quickly, do something else, check again e.t.c

pcdsdevices/sim/daq.py
```diff
@@ -118,14 +118,21 @@ def _pick_duration(self, events, l1t_events, l3t_events, duration):
                 else:
                     return ev / 120
         if duration is not None:
-            if isinstance(duration, list):
+            if not isinstance(duration, list):
```
teddyrendahl: Before you couldn't have a list and now you need a list? Not doubting you, just want to know why.

pcdsdevices/daq.py
ZLLentz: This is what the kickoff function is for, you can call it directly and check on the status object. Begin strives to fill the niche of "once I begin the daq I expect it to have been started." In practice, the status is marked done in about 0s in most cases, in about 0.5 seconds if you've just ended a really short run, and after the default timeout length if the daq is derping out. That is, if it will succeed it will succeed quickly.

pcdsdevices/sim/daq.py
ZLLentz: I originally observed pydaq accepting integer arguments and rejecting [seconds, nanoseconds]. This is because for some integer values this occasionally works, and that for very small values of [seconds, nanoseconds] the daq will spazz out. I originally misattributed the spazzing of the daq to the list argument rather than the magnitude of the list argument and completely missed the fact that duration=int only worked for a small number of cases.

pcdsdevices/daq.py
```diff
@@ -572,7 +603,12 @@ def wait(self, timeout=None):
         def cb(*args, **kwargs):
             self._wait_done.set()
         self.finished_cb = cb
-        self._wait_done.wait(timeout=timeout)
+        finished = self._wait_done.wait(timeout=timeout)
```
apra93: Is this method just a threaded version of `ophyd.status.wait`?

pcdsdevices/sim/daq.py
teddyrendahl: Interesting.

pcdsdevices/daq.py
teddyrendahl: Sounds good. `kickoff` it is!

pcdsdevices/daq.py
ZLLentz: I think you're on to something here: I didn't need to write this function at all. Not sure why I decided to implement it this way rather than just use the built-in wait. I also might need to change this when we update ophyd. I think I'll ignore it for now because it works and clean it up later.

PR #92, opened 2017-11-02
MAINT: epicsmotor inheritence clarification

shadowguy:
Changed epicsmotor import statement to clarify that the ophyd epicsmotor class is the parent class.  This is helpful because just having the class declaration as EpicsMotor(EpicsMotor, Device) is ambiguous.

### DIFF
--- a/pcdsdevices/epics/epicsmotor.py
+++ b/pcdsdevices/epics/epicsmotor.py
@@ -9,7 +9,7 @@ from enum import Enum
 ###############
 # Third Party #
 ###############
-from ophyd import EpicsMotor
+import ophyd
 from ophyd.utils import LimitError
 
 ##########
@@ -28,7 +28,7 @@ class DirectionEnum(Enum):
     negative = 1
     neg = 1
 
-class EpicsMotor(EpicsMotor, Device):
+class EpicsMotor(ophyd.EpicsMotor, Device):
     """
     Epics motor for PCDS.
     """


teddyrendahl:
Hi @shadowguy! Congrats on your first `pcdsdevices` PR.

Just a couple of things. We typically use [these](https://docs.scipy.org/doc/numpy/dev/gitwash/development_workflow.html#writing-the-commit-message) guidlines to tag our commits. It helps us keep track of the purpose for changes. In this case you would use `MAINT: Changed ...`. This way when we go back to write things like Release Notes we can quickly just look through the commit history and see what `BUG` were squashed, what `API` breaks did we have e.t.c

Also, its nice with a PR to include a little motivation for why we need the change above. We can see what you did with the code, but it is good practice to write down why you thought this was necessary. This helps us evaluate the PR, and also helps us when we forget all this in six months and can quickly go back and see who, why and when the changes were made.

Anyways, thanks for contributing and hope it is the first of many!

shadowguy:
No worries @teddyrendahl, thanks for the advice.  Could you take another look?

ZLLentz:
@shadowguy In addition to the PR title he meant the commit message itself, which you'd have to go back in git and amend the commit to change, etc.

That being said, I'm merging this in. :+1:

shadowguy:
Whoops.  Thanks @ZLLentz...this might be a job for git rebase!

PR #93, opened 2017-11-04
MAINT: Use latest ophyd/bluesky

ZLLentz:
This is now OK to merge once all tests pass.

A look-over at the things I had to change would be much appreciated, since I changed a lot of things really quickly and never looked back.


tests/conftest.py
```diff
@@ -16,5 +18,44 @@ def get_classes_in_module(module, subcls=None):
...
+    write_pv changes the read_pv
+    """
+    def make_put(original_put, read_pv):
+        def put(*args, **kwargs):
+            original_put(*args, **kwargs)
+            read_pv.put(*args, **kwargs)
+        return put
+    write_pv = epics_signal._write_pv
+    read_pv = epics_signal._read_pv
+    write_pv.put = make_put(write_pv.put, read_pv)
```
teddyrendahl: Yes! This can simplify quite a few tests

tests/test_epics_attenuator.py
```diff
@@ -1,60 +1,63 @@
...
 logger = logging.getLogger(__name__)
 
-@using_fake_epics_pv
-@pytest.fixture(scope='function')
-def attenuator():
+
+def fake_att():
+    """
+    using_fake_epics_pv does cleanup routines after the fixture and before the
+    test, so we can't make this a fixture without destabilizing our tests.
```
teddyrendahl: What about the cleanup destabilizes? The only reason we need the wrapper around the test is that some PVs are instantiated lazily. We could keep using fixtures and instead just make sure that the all the PVs are instantiated as `FakeEpicsPV`s by calling `wait_for_connection`?

tests/test_epics_ipm.py
```diff
@@ -49,11 +69,14 @@ def test_ipm_motion(ipm):
...
-def test_ipm_subscriptions(ipm):
+def test_ipm_subscriptions():
+    logger.debug('test_ipm_subscriptions')
+    ipm = fake_ipm()
     #Subscribe a pseudo callback
     cb = Mock()
     ipm.subscribe(cb, event_type=ipm.SUB_STATE, run=False)
     #Change the target state
     ipm.target.state._read_pv.put('OUT')
+    attr_wait_true(cb, 'called')
```
teddyrendahl: `assert cb.called` was running before the subscription could call it?

teddyrendahl:
I feel comfortable with this. Looks like it was largely testing changes and not much actually changed in the devices. (In fact, only real changes looked like the `add_callback` API and the necessity for an `egu` property on Positioners?

One thing I might add is `state.StatusSubscription` can be imported directly from `ophyd` now per https://github.com/NSLS-II/ophyd/pull/442. I might advocate that we just wait for `1.0.0` and then make any last changes and merge. Nice work!

tests/test_epics_attenuator.py
ZLLentz: It literally breaks the class with the new implementation of FakeEpicsPV with no way of recovery. The end of the `using_fake_epics_pv` wrapper does such a thorough cleanup that the object will never work again.

tests/test_epics_ipm.py
ZLLentz: The new FakeEpicsPV is not guaranteed to return a callback instantly, much like a real epics PV.

ZLLentz:
I had the same thoughts, it was mostly the tests that broke due to some API changes with instantiating objects, `control_layer`, etc.

I will make the final changes after 1.0.0, I forgot that SubscriptionStatus got merged.

ZLLentz:
One more thought, there was a lot of "optional" restructuring of the classes that I chose to forego (e.g. make name a required kwarg in every class-level def to mirror the base device requirements). We should find a time to do a big library-wide cleanup for stuff like this and other misc implementation details.

danielballan:
Belatedly.... hooray!

PR #94, opened 2017-11-05
FIX: Light Subscription Fixes  / Remove is optional

teddyrendahl:
## Description 
A number of smaller bugs I found when stress testing the lightpath.

* When devices use `_run_subs` they should pass along a reference to themselves with `obj`. Prior to this, our devices would pass the signal that tripped their own subscription to be run passed as `obj`.
* `InOutDevice` needs to use `SUB_STATE` to specify which event type the lightpath should subscribe to.
* Added tests for `Reflaser` lightpath compatibility
* Remove is no longer required by the `LightInterface`. The main reason being is we do not want `remove` buttons for devices that can be removed by a single button in the lightpath. However, if we have `def remove(self): return NotImplemented`, we have to actually call the method to see if we should create the button. 
* There were a few objects using `.get` in their `inserted`, `removed` properties
* Fixed `PointingMirror` tests (not sure how these ever worked ....)
* Added Github issue and PR templates

## Motivation
All bugs were found during instability testing as well as running with https://github.com/slaclab/lightpath/pull/30

## How has this been tested?
Added a few tests and run underneath a live instantiation of the `lightpath`

### DIFF
--- a/pcdsdevices/epics/pim.py
+++ b/pcdsdevices/epics/pim.py
@@ -245,8 +245,17 @@ class PIMMotor(Device, PositionerBase):
         return self.move("OUT", wait=wait, **kwargs)
 
     #Conform to lightpath interface
-    remove = move_out
-    insert = move_in
+    def insert(self, *args, **kwargs):
+        """
+        Alias for :meth:`.move_in` for lightpath interface
+        """
+        return self.move_in(*args, kwargs)
+
+    def remove(self, *args, **kwargs):
+        """
+        Alias for :meth:`.move_out` for lightpath interface
+        """
+        return self.move_out(*args, kwargs)
 
     def move_diode(self, wait=False, **kwargs):
         """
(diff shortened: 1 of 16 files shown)


pcdsdevices/epics/pim.py
```diff
@@ -246,7 +246,7 @@ def move_out(self, wait=False, **kwargs):
 
     #Conform to lightpath interface
     remove = move_out
-
+    insert = move_in
```
ZLLentz: This construct is dangerous if we ever want to use inheritance. If we override `move_in` or `move_out` in a subclass, `insert` and `remove` will still point to exactly these methods, not the new overrides. I used to put this in a lot but I didn't realize the trouble it could cause, so I favor the more wordy `def insert(self, *args, **kwargs): return self.move_in(*args, **kwargs)` pattern.

teddyrendahl: Interesting. I'll make the swap.

teddyrendahl: Just realize that exactly this would have happened in `epics/valve.py` . `GateValve` inherits from `Stopper` which had the shorthand re-aliasing 😮

PR #98, opened 2017-11-14
Minor Mirror Additions

teddyrendahl:
## Description

Two major changes:
1) `PointingMirror.inserted` and `PointingMirror.removed` now use top-level `PointingMirror.state.value` to determine the state. Not sure why I felt the need to use the low-level one
2) Before calling `PointingMirror.remove` or `PointingMirror.insert`, we check that the horizontal gantry is coupled. I had a terrifying thought that people would be pressing this button in the lightpath and not checking the state, even though it should never be decoupled.  In reality, the proper implementation is to make a `Gantry` object, but not sure it is an immediate priority. 

## Motivation and Context

Was crashing the lightpath application

## How Has This Been Tested?

I was naughty and added this change to the release area to get the `lightpath` up and running. Added tests for new methods. 

Notes added in docstrings

### DIFF
--- a/pcdsdevices/epics/mirror.py
+++ b/pcdsdevices/epics/mirror.py
@@ -669,8 +669,6 @@ class OffsetMirror(Device, PositionerBase):
     """
     # Pitch Motor
     pitch = FormattedComponent(OMMotor, "{self.prefix}")
-    # Piezo Motor
-    piezo = FormattedComponent(Piezo, "PIEZO:{self._area}:{self._mirror}")
     # Gantry motors
     gan_x_p = FormattedComponent(OMMotor, "{self._prefix_xy}:X:P")
     gan_y_p = FormattedComponent(OMMotor, "{self._prefix_xy}:Y:P")
@@ -942,7 +940,6 @@ class OffsetMirror(Device, PositionerBase):
         if tmo is not None:
             tmo = float(tmo)
         self.pitch.timeout = tmo
-        self.piezo.timeout = tmo
         self.gan_x_p.timeout = tmo
         self.gan_y_p.timeout = tmo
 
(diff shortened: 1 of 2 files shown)


pcdsdevices/epics/mirror.py
```diff
@@ -994,14 +991,17 @@ class PointingMirror(OffsetMirror, metaclass=BranchingInterface):
...
+                                            "GANTRY:{self._prefix_xy}:X:DECOUPLE")
 
     def __init__(self, *args, mps_prefix=None, state_prefix=None, out_lines=None,
                  in_lines=None, **kwargs):
         self._has_subscribed = False
         #Store MPS information
         self._mps_prefix = mps_prefix
         #Store State information
-        self._state_prefix = state_prefix 
+        self._state_prefix = state_prefix
```
ZLLentz: Thank you for eliminating the trailing space

pcdsdevices/epics/mirror.py
```diff
@@ -1012,26 +1012,38 @@ def inserted(self):
...
         Remove the PointingMirror from the beamline
+
+        If the horizontal gantry is not coupled, the request will raise a
+        `RuntimeError`
         """
-        return self.state.move("OUT", wait=wait, timeout=timeout)
+        if self.coupled:
+            return self.state.move("OUT", wait=wait, timeout=timeout)
+        else:
+            raise RuntimeError("Gantry is not coupled, can not move")
```
ZLLentz: Is this for redundancy, or will the mirror totally allow you to attempt to move one motor super far with the coupling disabled?

teddyrendahl: For redundancy, the mirror will not let you move past a certain gantry difference. Still. Don't want to test this rigorously. Also, the mirror WILL move UP to this gantry difference, so not an ideal situation anyways.

The mirror should never be decoupled though, this was just so for ease of mind.

PR #102, opened 2017-11-28
FIX: Reliable Components

ZLLentz:
## Description

Undo `Component` overrides. Call `get_ctrlvars` when we instantiate a signal.

## Motivation and Context

- `Component` override used currently broken lazy=True option.
- `FormattedComponent` is obsolete, was for our defunct `IOCDevice`
- `get_ctrlvars` cannot be called in some callbacks, calling it early prevents this issue from ever happening.

## How Has This Been Tested?

Skywalker works with these.

No need

### DIFF
--- a/pcdsdevices/component.py
+++ b/pcdsdevices/component.py
@@ -1,30 +1,3 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-import ophyd
-
-
-class Component(ophyd.Component):
-    """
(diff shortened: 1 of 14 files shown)


teddyrendahl:
Looks fine. I think we should probably not have to call `wait_for_connection` in our tests, and that this is an issue with the `FakeEpicsPV`, but we don't have to fix this now.

I would also like to see the `ctrl_vars` issue get solved at a lower level, but we should discuss this with the NSLS-II guys.

PR #108, opened 2017-12-07
MPS Limits

yashrao99:
Not sure if the test file I added with it properly tests it... it passes the call of function for various values of the fault value

### DIFF
--- a/pcdsdevices/epics/mps.py
+++ b/pcdsdevices/epics/mps.py
@@ -136,7 +136,7 @@ def mps_factory(clsname, cls,  *args, mps_prefix, veto=False,  **kwargs):
     return cls(*args, **kwargs)
 
 
-def mustBeOpenLogic(mps_A, mps_B):
+def must_be_open_logic(mps_A, mps_B):
     """
 
     This logic should analyze the two MPS classes of a device. This logic will only allow
@@ -170,7 +170,7 @@ def mustBeOpenLogic(mps_A, mps_B):
         return True
 
 
-def mustKnowPositionLogic(mps_A, mps_B):
+def must_know_position_logic(mps_A, mps_B):
     """
 
     This logic should analyze the two MPS classes of a device. This logic will only allow
@@ -206,7 +206,9 @@ class MPSLimits(Device):
 
     The MPSLimits class is to determine what action is to be taken based on the MPS values
     of a devicepertaining to a single device. If a device has two MPS values, there is 
-    certain logic that needs to be followed to determine whether or not the beam is allowed    through.
+    certain logic that needs to be followed to determine whether or not the beam is allowed
+    through.
+
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


pcdsdevices/epics/mps.py
```diff
@@ -134,3 +134,52 @@ def mps_factory(clsname, cls,  *args, mps_prefix, veto=False,  **kwargs):
     comp = FC(MPS, mps_prefix, veto=veto)
     cls  = type(clsname, (cls,), {'mps' : comp})
     return cls(*args, **kwargs)
+
+
+def mustBeOpenLogic(mps_A, mps_B):
```
teddyrendahl: Can you make these `must_be_open_logic`? We typically don't use camel case

pcdsdevices/epics/mps.py
```diff
@@ -134,3 +134,52 @@ def mps_factory(clsname, cls,  *args, mps_prefix, veto=False,  **kwargs):
...
+        return False
+
+    if mps_A.fault.value == 1 and mps_B.fault.value == 0:
+        return False
+
+    if mps_A.fault.value == 0 and mps_B.fault.value == 1:
+        return True
+
+
+def mustKnowPositionLogic(mps_A, mps_B):
```
teddyrendahl: Also these functions should have docstrings

pcdsdevices/epics/mps.py
```diff
@@ -134,3 +134,109 @@ def mps_factory(clsname, cls,  *args, mps_prefix, veto=False,  **kwargs):
...
+
+        self.MPSA = mps_A
+        self.MPSB = mps_B
+        self.name = name
+        self.logic = logic
+        super().__init__('', name=name)
+
+    @property
+    def faulted(self):
+
```
teddyrendahl: Add a docstring here that explains that we use the `logic` function to evaluate the faulted states of `mpsA` and `mpsB`

tests/test_mpslimits.py
```diff
@@ -0,0 +1,46 @@
...
+    mpsLimits.mps_A.bypass._read_pv.put(0)
+    assert mpsLimits.mps_A.faulted
+    
+    mpsLimits.mps_B.fault._read_pv.put(1)
+    mpsLimits.mps_B.bypass._read_pv.put(0)
+    assert mpsLimits.mps_B.faulted
+
+
+    mustBeOpenLogic(mpsLimits.mps_A, mpsLimits.mps_B)
+    mustKnowPositionLogic(mpsLimits.mps_A, mpsLimits.mps_B)
```
teddyrendahl: I would put the straight logic test in their own function `test_logical_combinations` or something

tests/test_mpslimits.py
```diff
@@ -0,0 +1,46 @@
...
+def test_mps_faults():
+    mpsLimits = fake_mps()
+    #Faulted
+    mpsLimits.mps_A.fault._read_pv.put(1)
+    mpsLimits.mps_A.bypass._read_pv.put(0)
+    assert mpsLimits.mps_A.faulted
+    
+    mpsLimits.mps_B.fault._read_pv.put(1)
+    mpsLimits.mps_B.bypass._read_pv.put(0)
+    assert mpsLimits.mps_B.faulted
```
teddyrendahl: Yeah. Your intuition is right, this isn't quite testing what we want. You setup the problem right, but your `assert` statement doesn't check `mpsLimits.faulted`. What you really want to check is that `mpsLimits` returns the correct evaluation of the `logic` function for two active MPSLimits.

pcdsdevices/epics/mps.py
ZLLentz: Chiming in to say that this is part of the pep8 specification for python style: https://www.python.org/dev/peps/pep-0008/#function-names

teddyrendahl:
Nice work @yashrao99

Issue #113, opened 2017-12-21
Device Representations


teddyrendahl:
## Expected Behavior

Investigate the best way to represent an overview of the device. This should use some combination of the `ophyd.Device.summary`. Maybe include in this in the built-in `__repr__`

ZLLentz:
Connected to https://github.com/pcdshub/hutch-python/pull/61

Commenting to remind us of the old feature set where some devices would have verbose, graphical statuses that can be helpful:
```
In  [1]: att
Out [1]:
filter# |0|1|2|3|4|5|6|7|8|9|
 OUT    |X|X|X|X|X|X|X|X|X|X|
 IN     | | | | | | | | | | |
Transmission for 1st harmonic (E=9.42 keV): 1.000e+00
Transmission for 3rd harmonic (E=28.27 keV): 1.000e+00

In  [2]: lom
Out [2]:
LODCM Status
crystal 1 is OUT
   1st crystal type: C '111'
   2nd crystal type: Unknown
   Energy 1st crystal = 	 9.4897 keV
   Energy 2nd crystal is Unknown. Unable to determine crystal reflection
   Horizontal elements status:	 OUT
   Vertical elements status:	 OUT
   Calibration foils status:	 OUT
   Diode 1 status:		 OUT

** C Position Status (0 is normal set position) **
  (x1C,y1C,z1C):       -0.0001,  -0.0000,-398.2910
  (h1nC,h1pC):        -10.0006,   9.1968
  (th1C,chi1C):        18.4939,  -0.0172
  (x2C,y2C,z2C):        1.2866,  -0.1662, 398.2872
  (h2nC):              -0.0003
  (th2C,chi2C):        18.4969,   0.0199

** Tower 1 (crystal 1) **
  (x,y,z,hn,hp)   [user]:  -0.000, +7.853, -395.111,-11.714,+9.197
  (x,y,z,hn,hp)   [dial]:  +3.508,+14.070, +376.910, -6.042,+0.025
  (x,y,z,hn,hp) [offset]:  +3.508,+21.923,  -18.201, -5.672,+9.222
  (th,ch,thf,chf)   [user]: +17.6389, -2.1907, +0.0000, +0.0000
  (th,ch,thf,chf)   [dial]: +11.6622, -2.1764, +0.0000, +0.0000
  (th,ch,thf,chf) [offset]:  +5.9766, -0.0143, +0.0000, +0.0000

** Tower 2 (crystal 2) **
[...]

PR #117, opened 2017-12-21
Remove overrides

### LINKED ISSUE #109: Deprecate Happi
## Expected Behavior

There should be no explicit dependency on `happi` inside `pcdsdevices` . Happi loads devices with no explicit dependencies.

### LINKED ISSUE #114: Deprecate all enforced interfaces
## Expected Behavior

This is not Java. We will duck-type

teddyrendahl:
## Description

We should no longer subclass Ophyd Devices and Components. This PR fully switches everything away from `pcdsdevice` specific reimplementations. I am confident this will not break our live devices. There are a few cases where we will have to be more careful about when we instantiate PVs e.t.c, but it shouldn't be an issue if we follow the conventions laid out in #113 

## Motivation and Context

The overrides were all things that have now been pushed back to ophyd or covered up for a mistake elsewhere in the library. This reduces the level of complication in the code (All the tests pass and I deleted 1371 lines of code)

Closes #109 
Closes #114 

## How Has This Been Tested?

Almost all tests remained the same. That pesky PIM test fails again though. By default the `sim` tests aren't run, but you can use the `--sim` argument with `run_tests.py`

```
    @property
    def _readback(self):
        if not self.use_string:
            try:
>               return self._get_readback() + self._noise_func() * self.noise
E               TypeError: unsupported operand type(s) for +: 'NoneType' and 'float'
```


ZLLentz:
Also, I suggest you remove `test_happi.py` to reflect the fact that this module won't be taking care of that any more.

teddyrendahl:
Alright next wave:
* Remove `live_tests`. I like this concept but it belongs at hutch-level integration test
* Clean out all happi references
* Update `requirements.txt` and the conda-recipe

I realize I didn't mention this above, there is a strong possibility this breaks the `pswalker` tests. I think the `PIM` might have issues.

conda-recipe/meta.yaml
```diff
@@ -18,7 +18,6 @@ requirements:
   run:
     - python {{PY_VER}}*,>=3
     - ophyd >=1.0.0,<1.1
```
ZLLentz: Perhaps this is a good opportunity to unlock this requirement to be `>=1.0.0,<2` until we actually see a problem with a `1.x` release

tests_live/test_structure.py
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env python
-# -*- coding: utf-8 -*-
-def test_explosions():
-    pass
```
ZLLentz: I hope you found the existence of this file as funny as I did

teddyrendahl: 😆

Issue #111, opened 2017-12-21
Device Cleanup


teddyrendahl:
1. **EVERY** device should have `hints`, `_default_read_attrs` and `_default_configuration_attrs`. These should use the class default variables like so:

```python
class MyDevice(ophyd.Device):
          _default_read_attrs = ['my_read']
          _default_configuration_attrs = ['my_config']
 ```

2. **NO** callbacks should make EPICS calls. They should instead use the information provided in the callback
3. **NO** EPICS connections on initialization (no subscribing, no gets, no puts)
4. **NO** overrides to base ophyd classes. 
5. **EVERY** device should adhere to http://nsls-ii.github.io/bluesky/hardware.html

PR #118, opened 2018-01-10
Slits Class

teddyrendahl:
## Description

My goals started pretty large for this class but I ended up greatly reducing the complexity I had due to a few misunderstandings about how the hardware worked. I tried to capture what I learned below, and what I left out. I'll add this as a new issue when we decide to merge.

### Improvements
* Status objects are created by the logical combinations of multiple status objects where needed. 
* Default `read_attrs`, `configuration_attrs`, and `hints`
* Improved documentation
* Move method takes either a tuple or float depending on whether you want a non-square aperture or not.
* Staging uses a built-in `ophyd` trick. I tried to put this on the individual `SlitPositioner` class but if you try and move both center and width of the Slits at the same time it hangs the IOC. (The move is never attempted, but the DMOV flag reads `Moving` indefinitely). Right now I have the staging logic implemented on the `Slits` class, the center is left untouched.

### Left Undone
* There is an ugly way to get the actual `EpicsMotor` objects into the class. However it requires a `signal.get()` to make a device which is very un-Ophyd like.  I wrote it. It worked, but seemed too messy to be left-in. 
* There are two very easy IOC changes that can be made to help this class out.
1) The `nominal_aperture` should be kept in autosaved records, then things like `OPEN` will make a lot more sense / actually be useful. We can keep this info in our database for now, but I believe the EPICS route is the correct one (also not that much work)
2) The `:STOP` flag should be exposed just like the `DMOV`, that way we can stop the Slits if necessary.
 
## Motivation and Context

Part of #111 

## How Has This Been Tested?

Updated and improved tests, also tested on MFX hardware (IOC was version `R2.0.0`)

Updated Sphinx documentation and docstrings


pcdsdevices/slits.py
```diff
@@ -0,0 +1,321 @@
...
+    def egu(self):
+        """Engineering units"""
+        return self._egu or self.readback._read_pv.units
+
+    def _setup_move(self, position):
+        # This is subclassed because we need `wait` to be set to False unlike
+        # the default PVPositioner method. `wait` set to True will not return
+        # until the move has completed
+        logger.debug('%s.setpoint = %s', self.name, position)
+        self.setpoint.put(position, wait=False)
```
ZLLentz: It bugs me that we have to do this. There's some weird stuff going on in the slits IOC.

pcdsdevices/slits.py
```diff
@@ -0,0 +1,321 @@
...
+        y_stat = self.ywidth.move(height, wait=False, timeout=timeout)
+        status = x_stat & y_stat
+        # Add our callback if one was given
+        if moved_cb is not None:
+            status.add_callback(moved_cb)
+        # Wait if instructed to do so. Stop the motors if interrupted
+        if wait:
+            try:
+                status_wait(status)
+            except KeyboardInterrupt:
```
ZLLentz: Are we going to be adding logic like this into every class's `move` method? This seems like another thing that will get repetitive that should have some shared code, though it isn't obvious to me how to share it.

pcdsdevices/slits.py
```diff
@@ -0,0 +1,321 @@
...
+        """
+        self.block_cmd.put(1)
+
+    def stage(self):
+        """
+        Store the initial values of the aperture position before scanning
+        """
+        self._original_vals[self.xwidth.setpoint] = self.xwidth.readback.value
+        self._original_vals[self.ywidth.setpoint] = self.ywidth.readback.value
+        return super().stage()
```
ZLLentz: Is this a workaround for not being able to declare `stage_sigs` without setting a value?

pcdsdevices/slits.py
```diff
@@ -0,0 +1,321 @@
...
+        """
+        # Avoid making child subscriptions unless a client cares
+        if not self._has_subscribed:
+            # Subscribe to changes in aperture
+            self.xwidth.readback.subscribe(self._aperture_changed,
+                                           run=False)
+            self.ywidth.readback.subscribe(self._aperture_changed,
+                                           run=False)
+            self._has_subscribed = True
+        super().subscribe(cb, event_type=event_type, run=run)
```
ZLLentz: This should return the callback id from `super().subscribe`

pcdsdevices/slits.py
teddyrendahl: Yeah that is the reason.

pcdsdevices/slits.py
teddyrendahl: I'll add it to the list of things to modify if we ever tweak this IOC

pcdsdevices/slits.py
teddyrendahl: I think this one is special because of the combined statuses. Hopefully we can use the built-in `move` method associated with `PVPositioner`

pcdsdevices/slits.py
teddyrendahl: Will do

Issue #115, opened 2017-12-21
State Cleanup


teddyrendahl:
## Expected Behavior

Re-examine how we are doing the `StatesRecord` class. This would be cleaner as a class that we inherit rather than a subcomponent. We should also see how we can generalize this further as to not be require as much re-writing for each specific device class.

ZLLentz:
@teddyrendahl I'm going to tackle this one first because it affects how we design the other device classes.

teddyrendahl:
Was thinking the same thing. Let me know if you need assistance @ZLLentz

ZLLentz:
I'm going to sum up my ideas here before diving in. Let me know what you think so we can be on the same page. The goal is to get to a point where we can write code like:

```
from enum import Enum

from pcdsdevices.states import StatesDevice


class PIM(StatesDevice):
    _states_enum = Enum('PIMStates', 'DIODE YAG OUT')


my_pim = PIM('HX2:SB1:PIM')
``` 

And then we have the following features for free:
- `my_pim.state` is an `EpicsSignal` for the state value and setter
- `my_pim.state_config` is a `Device` subclass with all the state position, delta information


So we avoid the awful `my_pim.states.state` constructs but can still change offset/delta from Python, and it is straightforward to create these devices by subclassing `StatesDevice` and providing an Enum. The implementation isn't too bad because we always know which components we need and their names, and only what the `state_config` component is may change.

Some questions:
- Is separating a `Signal` and a connected configuration `Device` a sustainable design choice?
- Do we need to support multiple states in one `Device`?
- Is this a sufficient feature set, or do we want to have things like `my_pim.state_motor` as well?
- Can this be (should this be) constructed in a such a way to include the generic PVState functionality too? We can look for a key like `_states_pvs={}` for example.

ZLLentz:
I suppose we also then have `StatesDevice` provide overrides for `set` and `position`. Might have also missed other things.

teddyrendahl:
I really like the implementation goal above. Still trying to process what the proper division of sub-devices should look like. Also absolutely think that StatesDevice should reimplement `set` and `position`

ZLLentz:
I'm going to get started and defer to doing the sub-device division later, but sticking with a single `state` signal for now. I agree that this part is less obvious without more thought.

ZLLentz:
Leaving this open with the following goals:

[...]

Issue #107, opened 2017-12-05
PIMMotor in 'unknown' mode registers as inserted


teddyrendahl:
## Expected Behavior

If the PIMMotor is `unknown` it should report neither inserted or removed. 

## Current Behavior

If PIMMotor is 'unknown' it registers as `inserted`.

## Possible Solution

```python
def inserted(self):
     return self.states.value in ('IN', 'DIODE')
```

ZLLentz:
Solved with #136

PR #119, opened 2018-01-10
ENH: States rework

ZLLentz:
## Description

Rework our StatesRecord infrastructure as a new StatesPositioner class that can be trivially subclassed for different states records. Delete many lines of code. Move out of the epics folder to indicate that things are mostly good to a first-glance, though I didn't touch the PVStates code much.

## Motivation and Context

See #115 

## How Has This Been Tested?

Passes all the tests, but I didn't check for coverage.

Only lightly in the docstrings


ZLLentz:
Might want to add hints and clean up PVStates before the merge

pcdsdevices/epics/inout.py
```diff
@@ -1,64 +1,29 @@
...
-
-        run : bool, optional
-            Run the callback immediatelly
-        """
-        if not self._has_subscribed:
-            self.state.subscribe(self._on_state_change, run=False)
-            self._has_subscribed = True
-        super().subscribe(cb, event_type=event_type, run=run)
+class Diode(InOutPositioner):
+    pass
```
teddyrendahl: What is this here for?

pcdsdevices/epics/lodcm.py
```diff
@@ -232,39 +243,19 @@ def lodcm_move(self, h1n=None, yag=None, dectris=None,
...
         if not done_statuses:
             lodcm_status._finished(success=True)
             return lodcm_status
 
-        events = []
-        for i, status in enumerate(done_statuses):
-            event = Event()
-            events.append(event)
-            status.add_callback(self._make_mark_event(event))
+        first_status = done_statuses[0]
```
teddyrendahl: I think you forgot to put in the `status_wait` here unless I'm missing. This will still return immediately with `wait` set to True.

run_tests.py
```diff
@@ -26,12 +27,18 @@
 
     root_logger = logging.getLogger()
     root_logger.setLevel(logging.DEBUG)
-    log_filename = os.path.join(os.path.dirname(__file__), 'debug.log')
-    if os.path.isfile(log_filename):
+    log_dir = Path(os.path.dirname(__file__)) / 'logs'
```
teddyrendahl: 👍

pcdsdevices/state.py
```diff
@@ -0,0 +1,376 @@
...
+
+        timeout: int or float, optional
+            Move timeout in seconds
+
+        wait: bool, optional
+            If True, do not return until the motion has completed.
+        """
+        status = self.set(position, moved_cb=moved_cb, timeout=timeout)
+        if wait:
+            status_wait(status)
```
teddyrendahl: This should return the status?

teddyrendahl:
Nice work! This accomplishes a lot:

1) Deleted 300 lines of code 👍 
2) `StateSignal` and `StatePositioner` seem way more explicit to me.
3) Everything looks easier to subclass

I think the next step would be to add some logic to `InOutPositioner` so that we can have `in_state` and `out_state` class variables, and then all of our `inserted` `removed` properties make themselves. Obviously, not in the scope of this PR, but just looking forward. It also seems like we need to think about that `state_config` object and finalize what we want.

Ready to approve once comments above are addressed.

teddyrendahl:
Worth closing #107 in this PR?

pcdsdevices/epics/inout.py
ZLLentz: There may be some desire to include a data readback on `Diode` in the future. I can remove this if you disagree.

pcdsdevices/epics/lodcm.py
ZLLentz: You are correct, I forgot to handle the `wait` keyword... but this was a bug before. Good catch, I'll handle this when we clean up the lodcm class.

pcdsdevices/state.py
ZLLentz: This was a concious choice. I'd rather create a convention where `move` and other convenience methods are designed for the command-line user, where having the status pop up is annoying as per scientist feedback. Then we reserve the `set` method for the raw "just do the thing and give me a status object". This creates a clear distinction as to what belongs in each method and creates a path to do status bars in move, etc.

I realize as typing this that I took this too far in `valve.py` and probably messed up part of the `lightpath` interface. This should be revisited before the merge.

ZLLentz:
Should probably save #107 for the PIM cleanup PR

pcdsdevices/epics/lodcm.py
teddyrendahl: 👍

pcdsdevices/epics/inout.py
teddyrendahl: Up to you. More of a curiosity, but I am tempted to say delete it until we have a use case. Don't feel particularly strongly

pcdsdevices/state.py
teddyrendahl: Yeah. We need those `insert`, `remove` methods to return a Status object. I'm fine with this convention.

pcdsdevices/state.py
```diff
@@ -0,0 +1,422 @@
...
+        """
+        self.state.put(state.name)
+
+    def _create_states_enum(self):
+        """
+        Create an enum that can be used to keep track of aliases, state names,
+        and integer enum values.
+        """
+        state_def = {}
+        for i, state in enumerate(self.states_list):
```
teddyrendahl: Be careful of state records that don't continuously use the `mbbi` i.e 1 is insert, 4 is removed. Maybe check if the item in the list exists so I can:
```python
_state_list = ['inserted', None, None, 'removed']
```

pcdsdevices/state.py
ZLLentz: This morning I realized that this crusade was kind of overkill. We should return `Status` objects from `move` because that's what `ophyd` does, but then reserve the `mv`, `mvr`, etc. methods for the bells and whistles and provide them with multi-inheritance from a simple interface class that references `set` or `move`. I still think `set` shouldn't have any special options like `wait` which belong in `move`.

pcdsdevices/epics/inout.py
ZLLentz: You're probably right, simpler is better. It's bad form to prepare for something that never happens.

pcdsdevices/state.py
ZLLentz: Do records like these exist? I wasn't aware of any. We can definitely interpret the `None` as no state and build the `enum` accordingly. The class will also defer to your special `states_enum` if you provide one, which might be a cleaner option.

ZLLentz:
All that's left to do here is check the code, merge, and open/close the relevant issue.

ZLLentz:
Yes, `states_enum` exists for comparing aliased values. It also provides a natural way to choose a state from an `int` input.

PR #123, opened 2018-01-19
ENH: Framework for In/Out devices

ZLLentz:
## Description

Implement the InOut parent device framework for states devices. This has not been spread to each of the individual device implementations yet.

## Motivation and Context

All devices with clear `states` positions function in essentially the same way with respect to inserted/removed, so we've consolidated this code in one place.
Continuation of #115

## How Has This Been Tested?

Added tests for full coverage.

Full docstrings, haven't touched the sphinx rst


pcdsdevices/inout.py
```diff
@@ -0,0 +1,90 @@
...
+    def _pos_in_list(self, state_list):
+        current_state = self.get_state(self.position)
+        for state in state_list:
+            if current_state == self.get_state(state):
+                return True
+        return False
+
+
+class InOutRecordPositioner(StateRecordPositioner, InOutPositioner):
+    pass
```
teddyrendahl: Might be nice to have a little docstring here at least. I realize that it will probably be redundant with the `StateRecordPositioner`, `StatePositioner`. but I'm actually guessing that more developer will interface with this class that the plain positioner class. I vote either right something short or reference a shared explanation

PR #130, opened 2018-01-23
ENH: Attenuator

ZLLentz:
## Description

Add cleaned-up `Attenuator` class implemented as a `PVPositioner`. Separate some of the inout code into it's own class `AggregateSignal` because I was going to use it for the `FeeAtt` status, but this didn't make it into the final cut.

## Motivation and Context

Using a standard base class makes the implementation simpler and more reliable. Accommodating the outdated FEE IOC is quickly becoming untenable.

## How Has This Been Tested?

Added tests up to 100% coverage.

Only docstrings for now.


pcdsdevices/signal.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def _run_sub_value(self, *args, **kwargs):
+        kwargs.pop('sub_type')
+        sig = kwargs.pop('obj')
+        kwargs.pop('old_value')
+        value = kwargs['value']
+        with self._lock:
+            old_value = self._readback
+            # Update just one value and assume the rest are cached
+            # This allows us to run subs without EPICS gets
```
teddyrendahl: 👍

teddyrendahl:
Looks good. `AggregatedSignal` looks interesting.

PR #131, opened 2018-01-23
API: remove iocadmin... for now

ZLLentz:
We aren't using this and have no near-term plans to use this. I suggest we remove it for now and revive it in a clean form later if we decide we need it. This is another one of those troublesome devices where there are multiple incompatible versions of the IOC module.


teddyrendahl:
Tempted to leave this. Currently works as a complete device. We can always get it back however.

ZLLentz:
@teddyrendahl This was my thought. It's not harmful to leave it in, but I'd rather not put time into cleaning it up before the tight release schedule. A cleanup would involve writing the factory function or w/e we use to attach it to a device object... probably some sort of proxy signal link so that devices on the same IOC don't duplicate the PVs a bunch of times.

PR #134, opened 2018-01-23
ENH: Lodcm

### LINKED ISSUE #122: LODCM remove command ignores wait parameter
See title

ZLLentz:
## Description

Clean up all the clutter. Simplify down to a states record with just the H1N (first crystal) state, plus some utilities for clearing the diagnostic path. Minor change to `inout` to cause no action on `remove` if already removed. This is because of the `OUTLOW` state in the diagnostics here.

closes #122 

## Motivation and Context

90% of the time you use the lodcm you just want to get it out of the way. This handles that in a simple, concise way without repeating code. I'd like to defer the rest for later, but it probably just needs some raw motors for alignment tweaking. I expect that an alignment expert just wants the motors, and everyone else just wants to know where H1N is and if the beam passes through the diagnostic stand. If there are no objections, I'll probably just hardcode the motors for the XPP and XCS LODCMS, since there are a ton of them with arbitrary prefixes and they are fixed installations.

## How Has This Been Tested?

Passes tests

Docstrings

### DIFF
--- a/pcdsdevices/lodcm.py
+++ b/pcdsdevices/lodcm.py
@@ -3,7 +3,7 @@ from threading import RLock
 from ophyd import Device, Component
 from ophyd.status import DeviceStatus, wait as status_wait
 
-from ..inout import InOutRecordPositioner
+from .inout import InOutRecordPositioner
 
 
 class H1NStates(InOutRecordPositioner):
--- a/pcdsdevices/lodcm.py
+++ b/pcdsdevices/lodcm.py
@@ -1,3 +1,11 @@
+"""
+LODCM: Large offset dual-crystal monochrometer.
+
+The scope of this module is to identify where the beam is intended to go after
+passing through the monochrometer, and to manipulate the diagnostics.
+
+This will not assist in the alignment procedure.
+"""
 from threading import RLock
 
 from ophyd import Device, Component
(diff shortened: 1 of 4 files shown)


pcdsdevices/lodcm.py
```diff
@@ -0,0 +1,158 @@
...
+
+        return status
+
+
+class LODCMXPP(LODCM):
+    foil = Cmp(FoilXPP, ":FOIL")
+
+
+class LODCMXCS(LODCM):
+    foil = Cmp(FoilXCS, ":FOIL")
```
teddyrendahl: Lets have these just live in the respective hutch directories.

teddyrendahl:
Yeah. I like this. Whatever we make for alignment belongs in an LODCM auto-alignment project ... We'll cross that giant bridge of backlash when we get there.

pcdsdevices/lodcm.py
ZLLentz: Should I move them out now, or later?

teddyrendahl: I would not include them in this PR.

ZLLentz: Where should I save them then, so I don't have to archaeology again?

ZLLentz: I'll spat them into an issue for later and remove them here

PR #139, opened 2018-01-25
OffsetMirror

teddyrendahl:
## Description

There are a number of classes in here. I can see the argument that this is too complicated for this library but I think we should leaved it here for now. The classes are:

`OMotor` -> Alex and I's butchered motor record implementation
`Pitch` -> Self-explanatory
`Gantry` -> For the horizontal and vertical axes
`OffsetMirror` -> Loads all of the above classes with minimal input from the user
`PointingMirror` -> An `OffsetMirror` mixed with `InOutPositioner` for XRT M1 and XRT M2

### Implementation Details
* The automated prefixing looks super messy, but I think it makes it nice for the user. All the XRT mirrors can be loaded like:
```python
xrt_m1 = OffsetMirror('XRT:M2H', name='MEC Mirror')
```
The reason I even have these as kwargs is the FEE has a terrible naming convention. This is probably my fault, but in the end, this is the reality we live in
* I put in an explicit check for NaN values. However, when I try `signal.put(math.nan)` this raises a `ValueError` as well. I know we did this in the past, so I left the check in there. Not sure how we managed to do it however.
* I removed a number of the assumptions about what you want to do with the main `OffsetMirror`. It is really only there to load everything nicely for you. For instance, `pswalker` will now have to scan the `pitch` subcomponent. I think this is better, explicit is better than implicit **AND** we really don't want people moving the wrong axis with the HOMS.

## Motivation and Context

Part of #111 

## How Has This Been Tested?

Use same tests as before with a few additional

Updated docstrings


teddyrendahl:
Have not tested on the live devices yet. We should wait until I do to merge. However, I figured I would submit to so you can review before then.

pcdsdevices/mirror.py
```diff
@@ -0,0 +1,336 @@
...
+    transmission = 1.0
+    SUB_STATE = 'sub_state_changed'
+
+    _default_read_attrs = ['pitch', 'xgantry.readback',
+                           'xgantry.gantry_difference']
+
+    _default_configuration_attrs = ['ygantry.setpoint']
+
+    def __init__(self, prefix, *, prefix_xy=None,
+                 xgantry_prefix=None, **kwargs):
```
ZLLentz: Design question: should we enforce `name` as a required `kwarg` here and at similar `__init__` statements, or ignore it and let the base `Device` class enfore the required `kwarg`?

pcdsdevices/mirror.py
```diff
@@ -0,0 +1,336 @@
+"""
+Offset Mirror Classes
+
+This script contains all the classes relating to the offset mirrors used in the
```
ZLLentz: This is a module, not a script

pcdsdevices/mirror.py
```diff
@@ -0,0 +1,336 @@
...
+FEE and XRT. Each offset mirror contains a stepper motor and piezo motor to
+control the pitch, two pairs of motors to control the horizontal and vertical
+gantries
```
ZLLentz: This is an incomplete sentence

pcdsdevices/mirror.py
```diff
@@ -0,0 +1,336 @@
...
+    Parameters
+    ----------
+    prefix : str
+        The EPICS base pv to use
+
+    name : str, optional
+        The name of the motor
+
+    kwargs:
+        All keyword arguments are passed onto PVPositioner
```
ZLLentz: Is documenting that `**kwargs` are not ignored needed here?

pcdsdevices/mirror.py
```diff
@@ -0,0 +1,336 @@
...
+class OMMotor(PVPositioner):
+    """
+    Base class for each motor in the LCLS offset mirror system.
+
+    Parameters
+    ----------
+    prefix : str
+        The EPICS base pv to use
+
+    name : str, optional
```
ZLLentz: name isn't an optional parameter any more

pcdsdevices/mirror.py
```diff
@@ -0,0 +1,336 @@
...
+    prefix : str
+        Base prefix for both stepper motors i.e XRT:M1H. Do not include the "P"
+        or "S" to indicate primary or secondary steppers
+
+    gantry_prefix : str, optional
+        Prefix for the shared gantry diagnostics if it is different than the
+        stepper motor prefix
+
+    kwargs:
+        All passed to OMMotor
```
ZLLentz: Is documenting that `**kwargs` are not ignored needed here?

pcdsdevices/mirror.py
```diff
@@ -0,0 +1,336 @@
...
+
+    read_attrs : sequence of attribute names, optional
+        The signals to be read during data acquisition (i.e., in read() and
+        describe() calls)
+
+    configuration_attrs : sequence of attribute names, optional
+        The signals to be returned when asked for the motor configuration (i.e.
+        in read_configuration(), and describe_configuration() calls)
+
+    name : str, optional
```
ZLLentz: name is no longer optional

pcdsdevices/mirror.py
```diff
@@ -0,0 +1,336 @@
...
+    # Pitch Motor
+    pitch = FC(Pitch, "MIRR:{self.prefix}")
+    # Gantry motors
+    xgantry = FC(Gantry, "{self._prefix_xy}:X",
+                 gantry_prefix="{self._xgantry}",
+                 add_prefix=['suffix', 'gantry_prefix'])
+    ygantry = FC(Gantry, "{self._prefix_xy}:Y")
+    # Transmission for Lightpath Interface
+    transmission = 1.0
+    SUB_STATE = 'sub_state_changed'
```
ZLLentz: This `SUB_STATE` definition is not used

tests/test_mirror.py
```diff
@@ -22,50 +23,97 @@ def fake_branching_mirror():
...
+    m.pitch.motor_egu._read_pv.put('urad')
     return m
 
 @using_fake_epics_pv
-def test_branching_mirror():
+def test_nan_protection():
+    branching_mirror = fake_branching_mirror()
+    with pytest.raises(ValueError):
+        branching_mirror.pitch.check_value(math.nan)
+
```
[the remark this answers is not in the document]
teddyrendahl: ✔️

pcdsdevices/mirror.py
```diff
@@ -0,0 +1,336 @@
...
+        # Unknown
+        else:
+            return []
+
+    @property
+    def branches(self):
+        """
+        Return all possible beamlines for mirror destinations
+        """
+        if self.in_lines and self.out_lines:
```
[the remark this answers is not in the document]
teddyrendahl: Good catch. This method was way more complicated before we just agreed to split `OffsetMirror` and `PointingMirror` this is a remanent.

pcdsdevices/mirror.py
```diff
@@ -0,0 +1,336 @@
...
+        self.out_lines = out_lines
+        super().__init__(*args, **kwargs)
+
+    @property
+    def destination(self):
+        """
+        Current list of destinations the mirror currently supports
+        """
+        # Inserted
+        if self.inserted and not self.removed:
```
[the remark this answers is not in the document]
teddyrendahl: Hmmm. If you don't like I can take it out. It is mostly repeating the MPS standards (checking that both limits make sense not just one). However, I don't think it is possible for the state record to make this mistake, so this method is just checking for an error in our Python code.

pcdsdevices/mirror.py
```diff
@@ -0,0 +1,336 @@
...
+        if self.in_lines and self.out_lines:
+            return self.in_lines + self.out_lines
+
+    def set(self, *args, **kwargs):
+        """
+        Check that our gantry is coupled before state moves
+        """
+        # Check the X gantry
+        if self.xgantry.decoupled.get():
+            raise PermissionError("Can not move the gantry is uncoupled")
```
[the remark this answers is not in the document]
teddyrendahl: First part ✔️ 
Second part - Yes. Both of the decoupled checks belong in `check_value`. I understand that this isn't inherently "checking the value" but it is where all the other safety level checks happen. It also means we don't have to worry about `set` vs `move` as check_value is called in both.

pcdsdevices/mirror.py
teddyrendahl: Leaving as a kwarg. Lets have Device handle this.

teddyrendahl:
Alright. All of the comments above have been addressed. Most of them were very obvious and fixed without comment.

I also tested this on XRT:M2H. All the signals read correctly and a move was completed successfully. Ready for final check and merge.

PR #141, opened 2018-01-25
Valve

teddyrendahl:
## Description

Updated `Stopper`, `GateValve`, `PPSStopper` to all use the `inout` module. This removed a lot of code. Everything was extremely straight forward.

## Motivation and Context

Addresses #111 

## How Has This Been Tested?

Minimal changes to existing tests. Added a few additional checks for completeness

Updated documentation


pcdsdevices/valve.py
```diff
@@ -0,0 +1,160 @@
...
+        opening
+        """
+        return bool(self.interlock.get())
+
+    def open(self, **kwargs):
+        """
+        Open the valve
+        """
+        if self.interlocked:
+            raise InterlockError('Valve is currently forced closed')
```
ZLLentz: This check is ignored on `valve.remove()`

pcdsdevices/valve.py
```diff
@@ -0,0 +1,160 @@
...
+
+    in_state : str, optional
+        String associatted with in enum value
+
+    out_state :str, optional
+        String associatted with out enum value
+    """
+    state = C(EpicsSignalRO, '', string=True)
+    # Default attributes
+    _default_read_attrs = ['summary']
```
ZLLentz: Where is this `summary` component?

pcdsdevices/valve.py
```diff
@@ -0,0 +1,160 @@
...
+
+    Because naming conventions for the states are non-uniform this class allows
+    you to enter the values at initialization.
+
+    Parameters
+    ----------
+    prefix : str
+        Full PPS Stopper PV
+
+    name : str, optional
```
ZLLentz: name is not optional

pcdsdevices/valve.py
```diff
@@ -0,0 +1,160 @@
...
+        return self.remove(**kwargs)
+
+    def close(self, **kwargs):
+        """
+        Close the stopper
+        """
+        return self.insert(**kwargs)
+
+
+class GateValve(Stopper):
```
ZLLentz: My non-vacuum brain finds this inheritance path very confusing, since stoppers block the beam robustly and I've heard stories of blowing through gate valves. It works in python because they behave basically the same, but I can't help but feel like there's a less confusing way to structure the OOP.

pcdsdevices/valve.py
```diff
@@ -0,0 +1,160 @@
...
+    prefix : str
+        Full PPS Stopper PV
+
+    name : str, optional
+        Alias for the stopper
+
+    in_state : str, optional
+        String associatted with in enum value
+
+    out_state :str, optional
```
ZLLentz: colon spacing

pcdsdevices/valve.py
```diff
@@ -0,0 +1,160 @@
...
+    A base class for a device with two limits switches controlled via an
+    external command PV. This full encompasses the controls Stopper
+    installations as well as un-interlocked GateValves
+
+    Parameters
+    ----------
+    prefix : str
+        Full PPS Stopper PV
+
+    name : str, optional
```
ZLLentz: name is not optional

pcdsdevices/valve.py
teddyrendahl: Good catch

pcdsdevices/valve.py
teddyrendahl: I realize it doesn't make sense physically, but it make sense inheritance-wise. To EPICS, a GateValve is just a Stopper with more restrictions. I don't really want to add a third class but I can if you want. `Stopper` can not inherit from `GateValve` because you would need to delete the interlock logic.

The OOP would then be `Valve`, which is then subclasses by `GateValve` and `Stopper`?

pcdsdevices/valve.py
teddyrendahl: Good catch.

pcdsdevices/valve.py
ZLLentz: OOP would be something more abstract like `OpenCloseLimitsDevice` (ok, maybe a shorter name) that feeds into both `GateValve` and `Stopper`. This is not important to do, unless later we decide that `Stopper` needs a feature that `GateValve` should definitely not have. We can of course just refactor if the need arises.

PR #142, opened 2018-01-25
ENH: Picker

ZLLentz:
## Description

Picks the pulses, as a state positioner. I also expanded the features of the pvstate logic to accommodate the inout picker.

## Motivation and Context

PP is relatively clean now. Brain is fried trying to debug, so I might have missed some things. There might be a subscription bug in `AggregateSignal`, but it seems to be very subtle... I might tackle it once we finish everything else unless it comes up again, in which case I'll tackle it sooner.

## How Has This Been Tested?

Passes tests

docstrings

### DIFF
--- a/pcdsdevices/pulsepicker.py
+++ b/pcdsdevices/pulsepicker.py
@@ -2,7 +2,7 @@ from ophyd import (Device, EpicsSignal, EpicsSignalRO, Component,
                    FormattedComponent)
 from ophyd.status import wait as status_wait, SubscriptionStatus
 
-from ..inout import InOutRecordPositioner
+from .inout import InOutRecordPositioner
 
 
 class PickerBlade(Device):
--- a/pcdsdevices/pulsepicker.py
+++ b/pcdsdevices/pulsepicker.py
@@ -122,6 +122,14 @@ class PulsePickerInOut(PulsePicker):
     """
     PulsePicker paired with a states record to control the Y position. This
     allows us to insert and remove the entire device from the beam.
+
+    The inout states record lives in a separate IOC from the main pulsepicker
+    due to versioning issues. The parent IOC is called 'device_states'. We're
+    expecting the resulting states record to have states 'Unknown', 'OUT',
+    and 'IN', in that order. The naming convention for the states is to take
+    the first two segments of the pulsepicker prefix and add 'PP:Y' to the end.
+    So therefore, if the picker is 'TST:DG1:MMS:03', the inout states should be
+    'TST:DG1:PP:Y'.
     """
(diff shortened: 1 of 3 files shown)


pcdsdevices/pulsepicker.py
```diff
@@ -0,0 +1,155 @@
...
+    cmd_follower = Cmp(EpicsSignal, ':RUN_FOLLOWERMODE')
+
+    states_list = ['OPEN', 'CLOSED']
+    in_states = ['CLOSED']
+    out_states = ['OPEN']
+    _states_alias = {'CLOSED': ['CLOSED', 'IN']}
+    _state_logic = {'blade': {0: 'OPEN',
+                              1: 'CLOSED',
+                              2: 'CLOSED'}}
+
```
teddyrendahl: Is `blade` a read_attr by default?

pcdsdevices/pulsepicker.py
```diff
@@ -0,0 +1,155 @@
...
+        self._log_request('FOLLOWER')
+        self.cmd_follower.put(1)
+        if wait:
+            self._wait(self.mode, 6, 'FOLLOWER')
+
+
+class PulsePickerInOut(PulsePicker):
+    """
+    PulsePicker paired with a states record to control the Y position. This
+    allows us to insert and remove the entire device from the beam.
```
teddyrendahl: Maybe add a note that the reason this isn't all one class is because these PVs come from a separate IOC

pcdsdevices/state.py
```diff
@@ -302,15 +308,22 @@ class PVStatePositioner(StatePositioner):
...
         This is for cases where the logic is simple. If there are more complex
         requirements, replace the `state` component.
-    """
+
+    _state_logic_mode: string
+        This should be 'ALL' (default) if the pvs need to agree for a valid
+        state. You can set this to 'FIRST' to instead use the first state
+        found while traversing the state_logic tree. This means an earlier
+        state definition can mask a later state definition.
+        """
```
teddyrendahl: 👍

teddyrendahl:
Seems reasonable. A little wary of the aforementioned bug but largely just because of vagueness. Not too worried.

I think we are acknowledging we'll have to come back through here when we start bluesky scanning with this thing.

pcdsdevices/pulsepicker.py
ZLLentz: I've only kept the `_default_read_attrs = ['state']` from the parent classes.

pcdsdevices/pulsepicker.py
ZLLentz: I will add a note, with some instructions on how to set it up maybe.

PR #143, opened 2018-01-25
MPS

teddyrendahl:
## Description

* Combined `test_mps` and `test_mps_limits`
* Both `MPS` and `MPSLimits` inherit from `MPSBase` now to avoid duplicate code
* To be in line with the lightpath terminology, `faulted` is the MPS bit is registering a fault. `tripped` is if the bit is registering a fault and it is not bypassed.
* Since we are now not including the MPS as a sub-component on devices we could probably deprecate the `mps_factory`. I left it in just in case it ends up being useful.

## Motivation and Context

Addresses #111 

## How Has This Been Tested?

Left most tests unmodified for `MPS`, added a few for `MPSLimits`


teddyrendahl:
This needs to be rebased on the result of #141. It will fail until then.

pcdsdevices/mps.py
```diff
@@ -0,0 +1,257 @@
...
+                   FormattedComponent as FC)
+
+logger = logging.getLogger(__name__)
+
+
+class MPSBase(OphydObject):
+    """
+    Base MPS class
+
+    Used for shared methods between the individual :class;`.MPS` bit class and
```
ZLLentz: semicolon

PR #147, opened 2018-01-26
ENH: mv_interface

### LINKED ISSUE #112: Status Displays
## Expected Behavior

Statuses should be displayed nicely for user methods, but hidden for built-in bluesky methods

ZLLentz:
## Description

Add an interface for shortcuts like `mv`, `wm`. Put it on everything that moves. Add progress bars to the update moves.

closes #112 

## Motivation and Context

Maintaining aliases like these is impossible unless managed centrally in one place. These will be asked for if they are missing. I have removed parameters that will be seen as confusing and omitted the status object returns as per what was asked for in the `snd` project. There is room to add more options and methods as they are requested, and they will spread themselves out among the devices if we update them here.

## How Has This Been Tested?

Added basic tests. Unfortunately, I missed 100% coverage this time, because I will not be generating a `KeyboardInterrupt` as part of a test.

Longer docstrings than I usually do, since I expect these are the front-interface methods.


teddyrendahl:
Looks cool. I think this makes `bluesky` an actual dependency. Can you remove it from `dev-requirements` and add it to `requirements`? Also make the corresponding change in the conda-recipe.

pcdsdevices/mv_interface.py
```diff
@@ -0,0 +1,117 @@
...
+        timeout: number, optional
+            If provided, the mover will throw an error if motion takes longer
+            than timeout to complete. If omitted, the mover's default timeout
+            will be use.
+
+        wait: bool, optional
+            If True, wait for motion completion before returning. Defaults to
+            False.
+        """
+        self.move(position, timeout=timeout, wait=wait)
```
teddyrendahl: I was under the impression that this interface would define `.move` as well, and all it would require was a `set` definition?

ZLLentz: I think the problem is we don't know how people are implementing their `move` and `set` methods. Some classes put all the smarts in `move` and some put them all in `set`. I can provide a `move` command too, and since this is inherited last it might be ok, but that runs the risk of getting hit with a `super().move` call from an arbitrary parent class.

ZLLentz: We can also do something ridiculous like provide both `move` and `set`, and then whichever one you omit gets provided.

teddyrendahl: Voting to table this discussion for now. This establishes the API and will work. Refinements can be made later

PR #155, opened 2018-02-16
ENH: Daq Features and Cleanup

### LINKED ISSUE #156: CI: Missing Files in Codecov
`daq.py`, `pim.py`, and `mirror.py` are omitted from `codecov`. I think this is a bug in `codecov`, because they show up in `coverage report -m` in the travis build.

Curiously, these are the files with the same names as files in the `sim` folder.

ZLLentz:
## Description

This is a big one. Do not merge until we do some field testing.
### Added:
- `manual` mode
- `calib_cycle` plan
- `daq_wrapper`, `daq_decorator`
- Concept of the "one daq object" that is stored to be used in `daq_wrapper` and `calib_cycle`
- Option to pass in `controls` as a `list` of `Device` objects
- Missing docstrings
### Fixed:
- `complete` now does the right thing: only calls for an immediate `stop` if we were configured to run forever, and never ends the run. This is the job of `collect` in the `fly_during_wrapper`. This means we can `kickoff` and `complete` several times in a run.
- `read_configuration` follows the `bluesky` spec correctly now
- miscellaneous bugs
### Changed:
- `Daq` now takes a `RE` argument to link it to a `RunEngine` instance. This chosen `RE` will be able to run correctly with `daq_wrapper`.
- `mode` is now an enum, with states `on`, `manual`, and `auto`
- Some objects that the user shouldn't mess with now have leading underscores
### Removed:
- Some useless arguments from the `Daq` constructor
- `DaqStatus`: reimplementation that didn't add anything useful
- `create_daq_RE`: we found this to be a dangerous approach
### CI:
- Rename a few files in the `sim` folder to have different names than in the `pcdsdevices` folder. This was somehow confusing `codecov` and dropping three files from the report.
Closes #156 

## Motivation and Context

This makes it much easier to set up `daq` scans through `bluesky` in `hutch-python`. Contributes toward https://github.com/pcdshub/hutch-python/issues/47

## How Has This Been Tested?

High coverage, but this isn't ready for a merge until I make it down to a hutch to do the "feel" tests.

Docstrings


teddyrendahl:
Okay thought about this more, and I think this will work well. Supports all the use cases, makes running the DAQ easy but also more explicit than just having it attached to the RunEngine.

There are two things I would like to see not necessarily included in this PR, but need to be done before completion.

1) Documentation. Duh. I think a Jupyter Notebook of how to to use this will be very helpful.
2) To get to the full implementation of the `per_step` hook I think we need one more plan that is `calib_at_step`, this performs the default `one_1d_step` or `one_nd_step` and then runs the `calib_cycle` plan you have already created.

Also just one more idea. I don't think it would be hard to look at a plan and see which mode the DAQ should be in. This might make things even more user friendly. For example, if you see multiple `kickoff` commands for the DAQ  with a pre-processor you could switch to `manual` mode.  Just a thought.

ZLLentz:
I think you are correct on all three points.

ZLLentz:
I'm going to move these three points into issues, and assume that I'll find some sort of bugs later that will also go into separate PRs.

PR #158, opened 2018-02-27
FIX: PPSStopper created state_list not states_list

teddyrendahl:
## Description

Unhappy discovery while creating a `PPSStopper`. The class constructor overwrites `state_list` and not `states_list` as it should. This means that the `states_enum` is not created correctly which causes a multitude of problems.


ZLLentz:
oops

PR #159, opened 2018-02-28
New component for enabling/disabling motors & Added new turbo device

yashrao99:


ZLLentz:
Yash, you should follow the pep 8 conventions for method names, etc. for code added to this module: https://www.python.org/dev/peps/pep-0008/.

You should also remove the turbo file from the `epics` folder and bring it up a level.

Does this `Turbo` class work for all of our turbo pumps on the beamline?

teddyrendahl:
Hi @yashrao99 can you separate the Turbo Pump and the `DISP` PRs. You should be able to do this by following https://github.com/k88hudson/git-flight-rules#i-made-several-commits-on-a-single-branch-that-should-be-on-different-branches and making a new PR for your new branch.


I'll have some comments there. Just to comment on this Turbo PR for a second. Part of the reason I don't want to merge this upstream is the discrepancy in the the PVs, from my understanding we don't have any Turbos with `turboStart` and `turboRun`, it is just that the EPICS Vacuum IOCs are inconsistent and they will have one of these. This is the kind of thing that needs to be fixed on the IOC level.  it is tempting to try and patch this at the Python level but if we don't put our foot down somewhere this repository becomes unmanageable by paying for the sins of lower level software. The first step in automation is standardization. Before merging this I would like to talk to Jason about changing the IOC so we have one signal with no `try / except` blocks.

PR #168, opened 2018-03-06
API: move daq out to the pcdsdaq module

ZLLentz:
## Description

Remove `Daq` class, it has been moved to the `pcdsdaq` module

## Motivation and Context

This class has always felt like an outlier in this module, and this became more clear when we started talking about Jungfrau objects for example. We needed a new module for non-epics daq-only classes.


requirements.txt
```diff
@@ -1,2 +1 @@
-ophyd==1.0.0
-bluesky==1.0.0
```
teddyrendahl: Bluesky is still a requirement right?

teddyrendahl:
Looks fine. What modules import this currently? What do we expect to break? 

From top of my head -> `HXRSnD` `hutch-python`... that it?

ZLLentz:
That's it, so far

PR #185, opened 2018-03-20
ENH: Clear flags, reinitialize, autosetup

### LINKED ISSUE #184: IMS Initialization
## Expected Behavior

IMSMotor class should be able to clear error, status and power-up codes.

teddyrendahl:
## Description

Posting this for discussion now, will add tests later if we decide this is the best way forward.
* Adds a general `_clear_flag` function used by `clear_powerup`, `clear_stall` and `clear_error`
* Add a `reinitialize` function
* Wraps these in a nice `auto_setup` method.

This is pretty much a straight port of the `auto_setup` functionality in the old `blbase.Motor` for setting up IMS motors with the notable exception of the `ParameterManager`. The minor improvements are that these methods all return status objects so they can be done asynchronously and they are pulled from a `bit_flags` dictionary instead of module variables.

Also, should `auto_setup` go in `stage`? Obviously not if we include the ParameterManager but as it is now it might make sense to run this at the start of scans.

## Motivation and Context

Closes #184 

## How Has This Been Tested?

Not yet.

Docstrings

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -304,12 +304,12 @@ class IMS(PCDSMotorBase):
         bit = flag_info['readback']
         mask = flag_info.get('mask', 1)
 
-        # Create a function to check the flag
-        def flag_is_present(msta):
-            return (int(msta) >> bit) & mask
+        # Create a callback function to check for bit 
+        def flag_is_cleared(value=None, **kwargs):
+            return not bool((int(value) >> bit) & mask)
 
         # Check that we need to actually set the flag
-        if not flag_is_present(self.bit_status.get()):
+        if flag_is_cleared(value=self.bit_status.get()):
             logger.debug("%s flag is not currently active", flag)
             return DeviceStatus(self, done=True, success=True)
 
@@ -317,9 +317,9 @@ class IMS(PCDSMotorBase):
         logger.info('Clearing %s flag ...', flag)
         self.seq_seln.put(flag_info['clear'])
         # Generate a status
-        st = SubscriptionStatus(self.msta, flag_is_present)
+        st = SubscriptionStatus(self.bit_status, flag_is_cleared)
         if wait:
-            status_wait(st)
+            status_wait(st, timeout=timeout)
         return st
 
 
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -223,11 +223,8 @@ class IMS(PCDSMotorBase):
         # Reinitialize if necessary
         if self.error_severity.get() == 3:
             self.reinitalize(wait=True)
-        # Issue flag commands and collect statuses
-        st = [func() for func in (self.clear_powerup,
-                                  self.clear_stall,
-                                  self.clear_error)]
-        status_wait(st[0] & st[1] & st[2])
+        # Clear all flags
+        self.clear_all_flags(wait=True)
 
     def reinitalize(self, wait=False):
         """Reinitialize the IMS motor"""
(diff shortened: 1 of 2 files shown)


ZLLentz:
Some thoughts:

- Should `bit_flag` be something more standard than a class attribute of `IMS`?
- `bit_flag` should be a hidden variable because it will only confuse a command line user
- I object to the name `status` for a signal that is just a series of bits. This will confuse the user.
- Rather than put `auto_setup` into `stage`, you should break it up into two methods. One will include the `pmgr` routines, and `clear_all` (name pending) or some equivalent will only check and clear the bits so it can go into `stage`.
- Can something like this realistically be tested? I remember that when I was doing the python 2 version of this the biggest problem was race conditions with the ims IOC.

ZLLentz:
One more suggestion: on `stage`, can you check the `PN` field, and if it's empty, reinitialize? This is how you avoid the critical crash bug.

teddyrendahl:
Because if you tell the motor to move before it's initialized it crashes the IOC? Crazy ...

ZLLentz:
Yes, I agree, this is crazy. Seems like an ad-hoc solution (checking PN) but it is a reliable way to identify and avoid the bug.

pcdsdevices/epics_motor.py
```diff
@@ -162,6 +163,121 @@ class IMS(PCDSMotorBase):
...
+        self.stage()
+
+    def auto_setup(self):
+        """
+        Automated setup of the IMS motor
+
+        If necessarry this command will:
+            * Reinitialize the motor
+            * Clear powerup, stall and error flags
+        """
```
ZLLentz: Does this RST look correct in the docs? I recall you need an extra space sometimes for the bulleted lists.

pcdsdevices/epics_motor.py
```diff
@@ -162,6 +163,121 @@ class IMS(PCDSMotorBase):
...
+        # Reinitialize if necessary
+        if self.error_severity.get() == 3:
+            self.reinitalize(wait=True)
+        # Clear all flags
+        self.clear_all_flags(wait=True)
+
+    def reinitalize(self, wait=False):
+        """Reinitialize the IMS motor"""
```
ZLLentz: Can you add parameters, etc. to this docstring?

ZLLentz:
@teddyrendahl should we include this in the release I'm about make?

teddyrendahl:
This is ready for a follow-up review. This has now been tested on a number of IMS 2.0.0 IOCs with multiple combination of error codes. I also could not resist and added some tests just to get the coverage up. In the process, of course, this uncovered some bugs.

pcdsdevices/epics_motor.py
```diff
@@ -197,6 +198,134 @@ class IMS(PCDSMotorBase):
...
+
+        This clears all present flags on the motor and reinitializes the motor
+        if we don't register a valid part number
+        """
+        # Check the part number to avoid crashing the IOC
+        if not self.part_number.get() or self.error_severity.get() == 3:
+            self.reinitialize(wait=True)
+        # Clear any pre-existing flags
+        self.clear_all_flags()
+        self.stage()
```
ZLLentz: recursion depth exceeded

pcdsdevices/epics_motor.py
```diff
@@ -197,6 +198,134 @@ class IMS(PCDSMotorBase):
...
+        AndStatus:
+            Combined status of clearing all flags
+        """
+        # Issue flag commands and collect statuses
+        st = [func(wait=True) for func in (self.clear_powerup,
+                                           self.clear_stall,
+                                           self.clear_error)]
+        # Combine statuses and wait for result
+        and_st = st[0] & st[1] & st[2]
+        return and_st
```
ZLLentz: Kind of wonky status object here

PR #187, opened 2018-03-27
ENH: Presets on FltMvInterface

ZLLentz:
## Description

Allow our special interface to save preset positions to yaml files. This is a port/rework of the old hutch python feature.

There will be one of these files per device that has presets, or multiple if we're running out of multiple paths. The file will be called `devicename.yml`.
```
pi:
  active: true
  comment: the constant
  history:
    26 Mar 2018 17:33:20: 3.1415926535
  last_updated: 26 Mar 2018 17:33:20
  value: 3.1415926535
uno:
  active: false
  history:
    26 Mar 2018 17:33:08: 1.023
    26 Mar 2018 17:33:11: 1
  last_updated: 26 Mar 2018 17:33:11
  value: 1
zero:
  active: true
  history:
    26 Mar 2018 17:32:42: 0
    26 Mar 2018 17:32:45: 0
    26 Mar 2018 17:32:46: 0
    26 Mar 2018 17:32:57: 0.0234
  last_updated: 26 Mar 2018 17:32:57
  value: 0.0234
```
If presets are configured using `setup_prefix_paths`, then there will be a `presets` object attached to the device. The intention is to call `setup_prefix_paths(xpp=<path>, lr13=<path>)` or some equivalent in the `hutch-python` setup, and then we call methods like:
```
motor.presets.add_xpp('my_position', 1.4, comment='hello there') 
motor.presets.add_here_lr13('sample')

motor.mv_sample()
```
And then by manipulating these paths we can specify which are permanent and which are temporary. I'm leaving this flexible because `pcdsdevices` needs to define the method calls but should not define where files are stored.

## Motivation and Context

This is used every day in xpp. The usage for experiment-specific positions is clear, the usage for permanent positions is dubious (they should become an IOC). Either way, it's good to allow flexibility.

## How Has This Been Tested?

Added tests up to high coverage, did a "feel" test with fake objects. Seems ok.

I have provided full docstrings and this should already be under the umbrella of the existing autodocs.

### DIFF
--- a/pcdsdevices/mv_interface.py
+++ b/pcdsdevices/mv_interface.py
@@ -227,12 +227,15 @@ class Presets:
         Works like threading.Rlock in that you can acquire it multiple times
         safely.
 
-        Raises an OSError if we cannot acquire the file lock.
-
         Parameters
         ----------
         fd: ``file``
             The file descriptor to lock on.
+
+        Raises
+        ------
+        OSError:
+            If we cannot acquire the file lock.
         """
         if self._fd is None:
             path = self._path(preset_type)
--- a/pcdsdevices/mv_interface.py
+++ b/pcdsdevices/mv_interface.py
@@ -208,7 +208,9 @@ class Presets:
         try:
             data = self._read(preset_type)
         except FileNotFoundError:
-            data = {name: {}}
+            data = {}
+        if name not in data:
+            data[name] = {}
         if value is not None:
             try:
                 old_value = data[name]['value']
@@ -253,9 +255,10 @@ class Presets:
             self._register_method(self, 'add_here_' + preset_type, add_here)
         for preset_type, data in self._cache.items():
             for name in data.keys():
-                mv = self._make_mv_pre(preset_type, name)
+                mv, umv = self._make_mv_pre(preset_type, name)
                 wm = self._make_wm_pre(preset_type, name)
                 self._register_method(self._device, 'mv_' + name, mv)
+                self._register_method(self._device, 'umv_' + name, umv)
                 self._register_method(self._device, 'wm_' + name, wm)
                 setattr(self.positions, name,
                         PresetPosition(self, preset_type, name))
@@ -288,9 +291,9 @@ class Presets:
             comment: ``str``, optional
                 A comment to associate with the preset position.
             """.format(preset_type)
-            self.presets._update(preset_type, name, value=value,
-                                 comment=comment)
-            self.presets.sync()
+            self._update(preset_type, name, value=value,
+                         comment=comment)
+            self.sync()
 
         def add_here(self, name, comment=None):
             """
@@ -304,13 +307,13 @@ class Presets:
             comment: ``str``, optional
                 A comment to associate with the preset position.
             """.format(preset_type)
-            add(name, self.wm(), comment=comment)
+            add(self, name, self._device.wm(), comment=comment)
         return add, add_here
 
     def _make_mv_pre(self, preset_type, name):
         """
-        Create a suitable version of ``mv`` for a particular preset type and
-        name e.g. ``mv_sample``.
+        Create a suitable versions of ``mv`` and ``umv`` for a particular
+        preset type and name e.g. ``mv_sample``.
         """
         def mv_pre(self, offset=0, timeout=None, wait=False):
             """
@@ -333,7 +336,26 @@ class Presets:
             """.format(name)
             pos = self.presets._cache[preset_type][name]['value']
             self.mv(pos+offset, timeout=timeout, wait=wait)
-        return mv_pre
+
+        def umv_pre(self, offset=0, timeout=None):
+            """
+            Update move to the {} preset position, or to a fixed offset from
+            it.
+
+            Parameters
+            ----------
+            offset: ``float``, optional
+                Offset from the preset position. If omitted, we'll move
+                as close to the preset position as possible.
+
+            timeout: ``float``, optional
+                If provided, the mover will throw an error if motion takes
+                longer than timeout to complete. If omitted, the mover's
+                default timeout will be use.
+            """
+            pos = self.presets._cache[preset_type][name]['value']
+            self.umv(pos+offset, timeout=timeout)
+        return mv_pre, umv_pre
 
     def _make_wm_pre(self, preset_type, name):
         """
@@ -420,25 +442,32 @@ class PresetPosition:
         self._presets.sync()
 
     @property
+    def info(self):
+        """
+        All information associated with this preset.
+        """
+        return self._presets._cache[self._preset_type][self._name]
+
+    @property
     def pos(self):
         """
         The set position of this preset.
         """
-        return self._presets._cache[self._preset_type][self._name]['value']
+        return self.info['value']
 
     @property
     def comment(self):
         """
         The comment associated with this preset.
         """
-        return self._presets._cache[self._preset_type][self._name]['comment']
+        return self.info.get('comment')
 
     @property
     def history(self):
         """
         This position history associated with this preset.
         """
-        return self._presets._cache[self._preset_type][self._name]['history']
+        return self.info.get('history')
 
     def __repr__(self):
         return self.pos
--- a/pcdsdevices/mv_interface.py
+++ b/pcdsdevices/mv_interface.py
@@ -305,7 +305,8 @@ class Presets:
         self._create_methods()
 
     def _log_flock_error(self):
-        logger.error('Cannot acquire file lock for %s', self.name)
+        logger.error(('Unable to acquire file lock for %s. '
+                      'File may be being edited by another user'), self.name)
         logger.debug('', exc_info=True)
 
     def _create_methods(self):
--- a/pcdsdevices/mv_interface.py
+++ b/pcdsdevices/mv_interface.py
@@ -207,7 +207,7 @@ class Presets:
         logger.debug('read presets for %s', self._device.name)
         with self._file_open_rlock(preset_type) as f:
             f.seek(0)
-            return yaml.load(f)
+            return yaml.load(f) or {}
 
     def _write(self, preset_type, data):
         """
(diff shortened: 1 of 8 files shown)


pcdsdevices/mv_interface.py
```diff
@@ -120,3 +134,352 @@ def umvr(self, delta, timeout=None):
...
+        Synchronize the presets with the database.
+        """
+        logger.debug('call %s presets.sync()', self._device.name)
+        self._remove_methods()
+        self._cache = {}
+        for preset_type in self._paths.keys():
+            try:
+                self._cache[preset_type] = self._read(preset_type)
+            except FileNotFoundError:
+                pass
```
teddyrendahl: Might be nice to put a `debug` message here. Catching exceptions with only a pass makes me nervous, even if we except that not every device has every preset type.

pcdsdevices/mv_interface.py
```diff
@@ -64,6 +73,11 @@ class FltMvInterface(MvInterface):
 
     This lets us do more with the interface, such as relative moves.
     """
+    def __init__(self, *args, **kwargs):
+        if Presets._paths:
+            self.presets = Presets(self)
```
teddyrendahl: I'm not sure I like the ordering that this implies. It means this attribute won't exist if we mess up the order of `setup_preset_paths`. Is there a way we can attach an empty `Presets` object that is updated by `setup_preset_paths`?

pcdsdevices/mv_interface.py
```diff
@@ -120,3 +134,352 @@ def umvr(self, delta, timeout=None):
...
+        Presets._paths[k] = Path(v)
+
+
+class Presets:
+    """
+    Manager for device preset positions. This provides methods for adding new
+    presets, checking which presets are active, and related utilities.
+
+    It will install the ``mv_presetname`` and ``wm_presetname`` methods onto
+    the associated device.
```
teddyrendahl: I think the format for the class docstrings should be one line summary then longer paragraph explanation.

ZLLentz: Yeah, you're right, that's the standard. I always mess this up on the first pass, I'll go back and fix it.

pcdsdevices/mv_interface.py
```diff
@@ -120,3 +134,352 @@ def umvr(self, delta, timeout=None):
...
+        with self._path(preset_type).open('r') as f:
+            return yaml.load(f)
+
+    def _write(self, preset_type, data):
+        """
+        Utility function to overwrite a particular preset's datum dictionary.
+        """
+        logger.debug('write presets for %s', self._device.name)
+        with self._path(preset_type).open('w') as f:
+            yaml.dump(data, f, default_flow_style=False)
```
teddyrendahl: Consider using [flock](https://docs.python.org/3.6/library/fcntl.html#fcntl.flock) to make sure that we don't have two sessions try and write to the file simultaneously?

pcdsdevices/mv_interface.py
ZLLentz: I understand. The caveat here is that this will be called for every device that doesn't have a preset defined.

pcdsdevices/mv_interface.py
ZLLentz: Yeah, we can do this. I assumed there would be some use cases for not having a presets object sneaking into your session, so that was the motivation for the original implementation.

pcdsdevices/mv_interface.py
ZLLentz: Is it possible for `flock` to clean up poorly and leave the file in a locked state? As implemented it will only cause issues if two sessions in one hutch are writing to the same object's presets at the same time +/- a second or two. I'd be wrapping the `_update` method in some sort of `flock` wrapper so that each client would wait for the flock, then read/update/write.

teddyrendahl:
Okay internals all look good. Good job making something messy like adding a bunch of methods with arbitrary string names look reasonable. I have two high-level thoughts:

1. It appears you have a `comment` which is actually a description of the state itself, I found this confusing because upon first reading I thought that this was intended to be a comment for each update of the position. I think we might want to consider both; adding a description that summarizes the state and a comment for every time it is updated that means your YAML file might look like:
```yaml
sam_x:
  active: true
  comment: Sample Position
  history:
    26 Mar 2018 17:33:08: 1.023 Aligned with ref. laser
    26 Mar 2018 17:33:11: 1 Updated after mirror pointing
  last_updated: 26 Mar 2018 17:33:11
  value: 1
```
2. I understand the benefit of making things general but I'm not in love with the methods for each arbitrary datatype. I'd kind of like to make the assumption that we will only have beamline presets and experiment presets and these are toggled on/off with boolean kwargs:
```python
sam_x.add_here(beamline=True)
```
We can argue about this. My first counter example would be let's write a script that aligns the XPP BE lens stack then saves it as a preset. I don't want to have to look up the current experiment number and reconstruct the method name to add it. I just want `experiment=True`.  Maybe this doesn't require a code change at all, and we just always add the experiment folder as just experiment.  So in `hutch_python`:
```python
cur_exp = get_current_experiment()
setup_preset_paths(experiment="presets/{}.format(cur_exp)", beamline="presets")
```

pcdsdevices/mv_interface.py
teddyrendahl: Good point. Might be overkill. You can use it as a context manager and I assume since it is in standard lib it should handle most cases. Your are correct it might be overkill.

ZLLentz:
1. I'll check if there's a nice way to do the history in a yaml file with 2 elements at each time point, but I expect I'll just concatenate into a string as you have done. I was intending the comments to be why the state itself exists, but I see the usefulness of using them as a change log.
2. Thank you for bringing up the programmatic need here. I'll have a response to this in code. I agree that my PR description with the name of the experiment in the method call is a bad idea.

Also, I need to consider whether to add one more test. This wouldn't have worked on real devices for some inheritance reasons but it didn't come up until I was putting presets on every device always.

pcdsdevices/mv_interface.py
ZLLentz: I'll do this to protect ourselves when we decide to assign presets programatically as you described below

teddyrendahl:
Keeping the comments separate wouldn't be the end of the world, a little less readable in they YAML file but 🤷‍♂️ :
```yaml
  historic_values:
    26 Mar 2018 17:33:08: 1.023
    26 Mar 2018 17:33:11: 1
  historic_comments:
    26 Mar 2018 17:33:08: Aligned with ref. laser
    26 Mar 2018 17:33:11: Updated after mirror pointing
```

pcdsdevices/mv_interface.py
teddyrendahl: Yeah used in the `happi` JSON interface and no issues yet.

https://github.com/pcdshub/happi/blob/a1443cd76140aa562c40f86a4e13b301ffd16755/happi/backends/json_db.py#L91

ZLLentz:
FYI the issue that broke everything is that `name` isn't ready until `OphydObject` is done with it's `__init__`. `OphydObject` is written in such a way that it should always be the second-to-last thing in the `mro` (before `object`), but this didn't matter for us because we didn't care about anything it did for the purposes of the wonky interface stuff. Now we care because we need to locate the presets file.

pcdsdevices/mv_interface.py
ZLLentz: This ended up being very difficult to implement correctly.

pcdsdevices/mv_interface.py
```diff
@@ -202,16 +205,48 @@ def _read(self, preset_type):
...
+
+    @contextmanager
+    def _file_open_rlock(self, preset_type):
+        """
+        File locking context manager for this object.
+
+        Works like threading.Rlock in that you can acquire it multiple times
+        safely.
+
+        Raises an OSError if we cannot acquire the file lock.
```
teddyrendahl: Use the special `Raises` section?
```
Raises
-------
OSError:
      If we can not acquire the filelock

File: pcdsdevices/mv_interface.py [hunk 6]
```diff
@@ -252,14 +288,23 @@ def sync(self):
...
+                except OSError:
+                    self._log_flock_error()
+            else:
                 logger.debug('No %s preset file for %s',
                              preset_type, self._device.name)
         self._create_methods()
 
+    def _log_flock_error(self):
+        logger.error('Cannot acquire file lock for %s', self.name)
+        logger.debug('', exc_info=True)
```
teddyrendahl: Might encourage a more novice friendly error message. `"File is currently being edited, can not acquire filelock"`

ZLLentz: Your suggestion isn't necessarily true. Being unable to acquire a file lock can also be because something has gone very wrong.

teddyrendahl: `"Unable to acquire filelock. File may be being edited by another user"`

ZLLentz:
Solutions to point 2 are all that's left, I think. I can provide something like:
```
motor.presets.add('name', value, comment='hello', kind='experiment')
```
Or we can stick with what already exists:
```
motor.presets.add_experiment('name', value, comment='hello')
```
But I do not want to do boolean args because you can't turn them all on at once. e.g. the following is nonsensical:
```
motor.presets.add('name', value, comment='hello', experiment=True, beamline=True)
```

teddyrendahl:
I see the point about booleans kwargs. I think we can stick with what already exists as long as we use the `experiment` or `exp` for short notation, no proposal named functions. I do not like the `kind` kwarg as the string won't be tab completable. 

Just thinking out loud. There are certain devices we know that should probably never have beamline `presets`, like those created from the Questionnaire, should we not attach the `beamline` kwargs there? Probably overkill, just wanted to preserve the idea somewhere.

pcdsdevices/mv_interface.py
```diff
@@ -266,22 +266,22 @@ def _update(self, preset_type, name, value=None, comment=None,
...
                     data[name]['history'] = history
-                if comment is not None:
-                    data[name]['comment'] = comment
                 if active:
                     data[name]['active'] = True
                 else:
                     data[name]['active'] = False
                 self._write(preset_type, data)
-        except OSError:
+        except BlockingIOError:
```
teddyrendahl: Whoah. Are the Python docs wrong? They explicitly say `OSError`

PR #188, opened 2018-03-29
FIX: Wait a second for file lock

ZLLentz:
## Description

Wait for the file lock with a timeout instead of immediately giving up.

## Motivation and Context

The file lock error prevents presets from loading, so if it's a manner of a short wait we shouldn't go straight to throwing an error.

## How Has This Been Tested?

Tests still pass, they just take a little longer because we're waiting more in the file lock test.


teddyrendahl:
Not familiar enough with `signal` library to fully critique this, but I understand the basics and think that  this is a better solution.

PR #192, opened 2018-04-04
ENH: Add disable puts field (.DISP) to PCDSMotorBase

yashrao99:
## Description

This PR adds the disabled attribute (.DISP) to the epics_motor class. This will allow users to enable/disable the motor when they see fit.

## Motivation and Context

This was added to disable the motors if needed when changing the CXI beamline configuration (moving the stands). If one of the motors moves to the wrong position, all the motors will stop moving and be disabled. This is also a safety check prior to moving the motors.

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -145,16 +145,15 @@ class PCDSMotorBase(FltMvInterface, EpicsMotor):
     def check_value(self, value):
         """
         Check if the motor is disabled
-
-        Raises
-        ------
-        Exception
-
         Check if the value is within the soft limits of the motor.
 
         Raises
         ------
+        Exception
+            If the motor is passed any motion command when disabled
         ValueError
+            When the provided value is outside the range of the low
+            and high limits
         """
         # First check that the user has returned a valid EPICS value. It will
         # not consult the limits of the PV itself because limits=False
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -52,7 +52,7 @@ class PCDSMotorBase(FltMvInterface, EpicsMotor):
     # This attribute is tracked by the _pos_changed callback
     direction_of_travel = Component(Signal)
     # This attribute will show the current status of motor
-    current_motor_status = Component(EpicsSignal, ".DISP")
+    disabled = Component(EpicsSignal, ".DISP")
 
     @property
     def low_limit(self):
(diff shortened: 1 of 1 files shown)


pcdsdevices/epics_motor.py
```diff
@@ -51,6 +51,8 @@ class PCDSMotorBase(FltMvInterface, EpicsMotor):
     # Disable missing field that our EPICS motor record lacks
     # This attribute is tracked by the _pos_changed callback
     direction_of_travel = Component(Signal)
+    # This attribute will show the current status of motor
+    current_motor_status = Component(EpicsSignal, ".DISP")
```
teddyrendahl: `current_motor_status` is a little misleading. I think the field is just here for disabling the motor. I would be tempted to call it what it is -> `disabled`.

pcdsdevices/epics_motor.py
```diff
@@ -125,6 +127,31 @@ def limits(self, limits):
         self.low_limit = limits[0]
         self.high_limit = limits[1]
 
+    @property
+    def motor_status(self):
```
teddyrendahl: This is a little redundant. If you make the signal we can just call -> `disabled.get()` or `disabled.set()`. I would scratch these these two properties and just make a function that looks like:

```python
def enable(self, enable):
    return self.disabled.set(not enable)
```

pcdsdevices/epics_motor.py
```diff
@@ -125,7 +127,38 @@ def limits(self, limits):
...
+        """
+        # First check that the user has returned a valid EPICS value.
+        super().check_value(value)
+
+        # Find the value for the disabled attribute
+        is_disabled = self.disabled.value
+        if is_disabled == 1:
+            raise Exception("Motor is not enabled. Motion requests "
+                            "ignored")
+
```
teddyrendahl: Almost! You made a function with two docstrings that calls `super().check_value` twice. Make this a single function with a single docstring. 

Also condense the if statement by just comparing against value specifically. No need to make the `is_disabled` variable.

yashrao99: @teddyrendahl I merged the doctrings - I wasn't sure if I just keep it in order of what was called, or if I keep the Raises sections together and the descriptions together

yashrao99:
@teddyrendahl hopefully that does it

pcdsdevices/epics_motor.py
```diff
@@ -125,17 +127,43 @@ def limits(self, limits):
         self.low_limit = limits[0]
         self.high_limit = limits[1]
 
+    def enable(self, enable):
```
ZLLentz: I expect a user will see a method named `enable` and immediately expect that calling `motor.enable()` will enable a motor. Perhaps we can be kind to the user and default this to `enable=True`. If you don't expect most users to try to call this you can ignore this comment.

pcdsdevices/epics_motor.py
```diff
@@ -125,17 +127,43 @@ def limits(self, limits):
...
+        """
+        Sets the state for the motor
```
ZLLentz: This docstring is not helpful. What does setting the motor state mean?

pcdsdevices/epics_motor.py
```diff
@@ -125,17 +127,43 @@ def limits(self, limits):
...
+            and high limits
         """
         # First check that the user has returned a valid EPICS value. It will
         # not consult the limits of the PV itself because limits=False
         super().check_value(value)
+
+        # Find the value for the disabled attribute
+        if self.disabled.value == 1:
+            raise Exception("Motor is not enabled. Motion requests "
+                            "ignored")
```
ZLLentz: This should be more specific than `Exception`, since every error is an `Exception`, but I'm struggling to pick a good replacement. I'm tempted to say this should be a `RuntimeError`,  but that's not much better. Any suggestions @teddyrendahl ?

pcdsdevices/epics_motor.py
```diff
@@ -51,6 +51,8 @@ class PCDSMotorBase(FltMvInterface, EpicsMotor):
     # Disable missing field that our EPICS motor record lacks
     # This attribute is tracked by the _pos_changed callback
     direction_of_travel = Component(Signal)
+    # This attribute will show the current status of motor
```
ZLLentz: This comment doesn't make any sense. The "disable puts" field has little to do with the motor's status.

ZLLentz:
I think `enable()` and `disable()` feels good for a user, because it's easier to tab-complete than `enable(True)` and `enable(False)`. It also makes it clearer that disabling is a thing you can do, since not everyone is going to read all the docstrings, but everyone is going to press tab. I think that is a good solution, just for usability. We can have both syntaxes available.

yashrao99:
@ZLLentz That should do it for `enable()` and `disable()`. I'm not sure what the alternative for the raised `Exception` would be, so I can change that as well if you guys decide on something.

ZLLentz:
My intuition is that we should define a new exception type in case anyone wants to wrap in a try/except block and needs to differentiate between, say, a `DisabledError` and a `ValueError`. I don't think this is super critical, I like this pull request as-is and we can add features if a need arises rather than trying to anticipate them.

PR #196, opened 2018-04-10
EventSequencer

### LINKED ISSUE #195: EventSequencer
## Expected Behavior

Create a class that can start and stop the EventSequencer in the context of a `bluesky` plan

teddyrendahl:
## Description

A basic `EventSequencer` object. 

This handles **only** starting and stopping the `EventSequencer` in the various modes of operation. Actually configuring what the sequence loaded is should be discussed outside of this PR. 

I suspect the main controversy will be the fact that this is a `bluesky` flyer. I decided to go this route because the `set` method does not make much sense. In reality most of the time you want to start and stop the Event Sequencer as two entirely separate actions.  This works especially nicely if you just want to run the sequencer the entire time you are scanning:

```python
fly_during_wrapper(my_scan, [sequencer])
```

I do agree that when you want to keep starting and stopping the sequencer during a scan it is a little more awkward, but I don't think the `set` and `wait` syntax really makes much more sense. I like the granularity of this. It is clear exactly when I want to start the sequence, and exactly when I want to wait for my sequence to complete.
```python
def step_then_sequence(detectors, step, pos_cache):
   yield from one_nd_step(detectors, step, pos_cache)
   yield from kickoff(sequencer)
   yield from complete(sequencer, wait=True)

scan([det], motor, ..., per_step=step_then_sequence)
```
## Motivation and Context

Closes #195 

## How Has This Been Tested?

* Added to test suite a series of methods that run the `EventSequencer` in a `bluesky` plan as well as some more in-depth tests on specific methods
* Tested with the CXI EventSequencer `R3.8.0`

Added to `device_types` table. Included some sample use cases

### DIFF
--- a/pcdsdevices/device_types.py
+++ b/pcdsdevices/device_types.py
@@ -10,3 +10,4 @@ from .pim import PIM
 from .pulsepicker import PulsePicker
 from .slits import Slits
 from .valve import Stopper, GateValve
+from .sequencer import EventSequencer
--- a/pcdsdevices/device_types.py
+++ b/pcdsdevices/device_types.py
@@ -8,6 +8,6 @@ from .mirror import OffsetMirror, PointingMirror
 from .movablestand import MovableStand
 from .pim import PIM
 from .pulsepicker import PulsePicker
+from .sequencer import EventSequencer
 from .slits import Slits
 from .valve import Stopper, GateValve
-from .sequencer import EventSequencer
--- a/pcdsdevices/sequencer.py
+++ b/pcdsdevices/sequencer.py
@@ -1,6 +1,4 @@
-import time
 import logging
-from functools import partial
 
 from ophyd import Device, EpicsSignal, EpicsSignalRO, Component as Cpt
 from ophyd.status import DeviceStatus, SubscriptionStatus
@@ -58,8 +56,7 @@ class EventSequencer(Device, MonitorFlyerMixin, FlyerInterface):
     _default_read_attrs = ['play_status']
     _default_configuration_attrs = ['play_mode', 'sequence_length']
 
-    def __init__(self, prefix, *, name=None,
-                  monitor_attrs=None, **kwargs):
+    def __init__(self, prefix, *, name=None, monitor_attrs=None, **kwargs):
         monitor_attrs = monitor_attrs or ['current_step', 'play_count']
         # Device initialization
         super().__init__(prefix, name=name,
--- a/pcdsdevices/sequencer.py
+++ b/pcdsdevices/sequencer.py
@@ -5,12 +5,12 @@ from functools import partial
 from ophyd import Device, EpicsSignal, EpicsSignalRO, Component as Cpt
 from ophyd.status import DeviceStatus, SubscriptionStatus
 from ophyd.utils.epics_pvs import raise_if_disconnected
-
+from ophyd.flyers import FlyerInterface, MonitorFlyerMixin
 
 logger = logging.getLogger(__name__)
 
 
-class EventSequencer(Device):
+class EventSequencer(Device, MonitorFlyerMixin, FlyerInterface):
     """
     Event Sequencer
 
@@ -58,12 +58,12 @@ class EventSequencer(Device):
     _default_read_attrs = ['play_status']
     _default_configuration_attrs = ['play_mode', 'sequence_length']
 
-    def __init__(self, prefix, *, name=None,  **kwargs):
+    def __init__(self, prefix, *, name=None,
+                  monitor_attrs=None, **kwargs):
+        monitor_attrs = monitor_attrs or ['current_step', 'play_count']
         # Device initialization
-        super().__init__(prefix, name=name, **kwargs)
-        # Data caches
-        self._steps = list()
-        self._iterations = list()
+        super().__init__(prefix, name=name,
+                         monitor_attrs=monitor_attrs, **kwargs)
 
     @raise_if_disconnected
     def kickoff(self):
@@ -75,23 +75,16 @@ class EventSequencer(Device):
         status : SubscriptionStatus
             Status indicating whether or not the EventSequencer has started
         """
-        # Clear cached data from prior runs
-        self._steps.clear()
-        self._iterations.clear()
         # Start the sequencer
         logger.debug("Starting EventSequencer ...")
         self.play_control.set(1)
-
-        # Subscribe to changes in the play step and iteration
-        def add_event(cache, timestamp=None, value=None, **kwargs):
-            cache.append((timestamp, value))
-
-        self.current_step.subscribe(partial(add_event, self._steps))
-        self.play_count.subscribe(partial(add_event, self._iterations))
+        # Start monitor signals
+        super().kickoff()
 
         # Create our status
         def done(*args, value=None, old_value=None, **kwargs):
             return value == 2 and old_value == 0
+
         # Create our status object
         return SubscriptionStatus(self.play_status, done, run=True)
 
@@ -118,17 +111,13 @@ class EventSequencer(Device):
         configuration use the :meth:`.stop` command.
 
         """
-        # Clear subscriptions
-        def clear_subs(**kwargs):
-            self.current_step.unsubscribe_all()
-            self.play_count.unsubscribe_all()
-
+        # Clear the monitor subscriptions
+        super().complete()
         # If we are running forever we can stop whenever
         if self.play_mode.get() == 2:
             logger.debug("EventSequencer is set to run forever, "
                          "stopping immediately")
             self.stop()
-            clear_subs()
             return DeviceStatus(self, done=True, success=True)
 
         # Otherwise we should wait for the sequencer to end
@@ -139,36 +128,7 @@ class EventSequencer(Device):
         logger.debug("EventSequencer has a determined stopping point, "
                      " waiting for sequence to complete")
         st = SubscriptionStatus(self.play_status, done, run=True)
-        st.add_callback(clear_subs)
         return st
-
-    def collect(self):
-        """
-        Gather information about the state of the Sequencer
-
-        During :meth:`.kickoff`, a monitor is started on the ``current_step``
-        and ``play_count`` signals. The timestamps and values of these signals
-        are stored internally and can be added back into the data stream
-        asynchronously by calling this method.
-        """
-        # Create partial event document
-        logger.debug("Interpreting stored EventSequencer information")
-        event = {'time': time.time(), 'timestamps': dict(), 'data': dict()}
-        for sig, cache in zip((self.current_step, self.play_count),
-                              (self._steps, self._iterations)):
-            event['timestamps'][sig.name] = [evt[0] for evt in cache]
-            event['data'][sig.name] = [evt[1] for evt in cache]
-        # Clear caches
-        self._steps.clear()
-        self._iterations.clear()
-        yield event
-
-    def describe_collect(self):
-        """Describe the collection information"""
-        dd = dict()
-        dd.update(self.play_count.describe())
-        dd.update(self.current_step.describe())
-        return {'sequence': dd}
 
     def stop(self):
         """Stop the EventSequencer"""
--- a/pcdsdevices/sequencer.py
+++ b/pcdsdevices/sequencer.py
@@ -75,6 +75,7 @@ class EventSequencer(Device, MonitorFlyerMixin, FlyerInterface):
         self.start()
         # Start monitor signals
         super().kickoff()
+
         # Create our status
         def done(*args, value=None, old_value=None, **kwargs):
             return value == 2 and old_value == 0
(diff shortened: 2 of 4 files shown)


pcdsdevices/sequencer.py
```diff
@@ -0,0 +1,169 @@
...
+        fly_during_wrapper(scan([det], motor, ...), [sequencer])
+
+    Run the EventSequencer at each step in my scan, after the completing the
+    motor move.
+
+    .. code::
+
+        def step_then_sequence(detectors, step, pos_cache)
+            yield from one_nd_step(detectors, step, pos_cache)
+            yield from kickoff(sequencer)
```
ZLLentz: This will move, trigger, read, save, and THEN start the sequencer. Is that really what we want? I think what we really should be doing is move/wait, kickoff, trigger/read/save, complete (despite that being more annoying to write out)

pcdsdevices/sequencer.py
```diff
@@ -0,0 +1,169 @@
...
+
+from ophyd import Device, EpicsSignal, EpicsSignalRO, Component as Cpt
+from ophyd.status import DeviceStatus, SubscriptionStatus
+from ophyd.utils.epics_pvs import raise_if_disconnected
+
+
+logger = logging.getLogger(__name__)
+
+
+class EventSequencer(Device):
```
ZLLentz: Maybe also subclass `ophyd.flyers.FlyerInterface` in case we want to do `isinstance(obj, FlyerInterface)`

pcdsdevices/sequencer.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def __init__(self, prefix, *, name=None,  **kwargs):
+        # Device initialization
+        super().__init__(prefix, name=name, **kwargs)
+        # Data caches
+        self._steps = list()
+        self._iterations = list()
+
+    @raise_if_disconnected
+    def kickoff(self):
```
ZLLentz: It seems like we're committing to having the `bluesky` flyer interface be the only way to start the sequencer. Just making sure there was a conscious decision to omit a `start` method similar to the `stop` method. I think this is ok but it needs to be stated. Would anyone ever be calling `seq.kickoff()` outside of a plan? I guess it's ok to force the `bluesky` API over another arbitrary API. End ramble.

pcdsdevices/sequencer.py
```diff
@@ -0,0 +1,169 @@
...
+    def describe_collect(self):
+        """Describe the collection information"""
+        dd = dict()
+        dd.update(self.play_count.describe())
+        dd.update(self.current_step.describe())
+        return {'sequence': dd}
+
+    def stop(self):
+        """Stop the EventSequencer"""
+        self.play_control.put(0)
```
ZLLentz: Do you plan to implement `pause` and `resume`? http://nsls-ii.github.io/bluesky/hardware.html#flyer-interface

pcdsdevices/device_types.py
```diff
@@ -10,3 +10,4 @@
 from .pulsepicker import PulsePicker
 from .slits import Slits
 from .valve import Stopper, GateValve
+from .sequencer import EventSequencer
```
ZLLentz: Can you sort this into the alphabetized spot? I feel like having a set way to order imports is very useful when there are tons of imports in a module.

pcdsdevices/sequencer.py
teddyrendahl: I agree. Do you think this level of granularity makes the example more or less confusing?

ZLLentz: I think, as stated, it doesn't match the description of the example. It'll make the example more confusing if we add granularity. Perhaps we just change the description to make it clear what it does.

pcdsdevices/sequencer.py
teddyrendahl: They are a little different. `stop` is used by `bluesky` to make sure that we don't leave the EventSequencer running if the scan exits in an ugly state.  `start` is a wrapper for user API. I am voting for forcing `bluesky` API

pcdsdevices/sequencer.py
teddyrendahl: Thank you for making me look at this file closely. It looks like I reimplemented part of `MonitorFlyerMixin` I will subclass this and `FlyerInterface`

pcdsdevices/sequencer.py
teddyrendahl: Hmmmmmm.  I am not sure about this. If we do `start` and `stop` we restart the sequence. This could have unintended consequences. I could see it being useful in certain cases though. Requires thought ...

pcdsdevices/sequencer.py
ZLLentz: I didn't even notice `MonitorFlyerMixin`. Good catch.

pcdsdevices/sequencer.py
ZLLentz: You should at least make sure it does the right thing if we ctrl+c during a plan e.g. does the sequence restart when the RE resumes. I don't recall if it calls `kickoff` again or if it just checks for `resume`.

teddyrendahl: `kickoff` is not called again.  What I am suggesting that the sequence will not stop... Doesn't seem right either.

pcdsdevices/sequencer.py
ZLLentz: I support your decision, but will direct all questions of "why is there a `stop` but no `start`" to you.

pcdsdevices/sequencer.py
ZLLentz: I expect the sequence does stop because you have defined `stop`

teddyrendahl: Disagree. `stop_movable_objects` called on `pause` only stops objects that received a `set` command.

https://github.com/NSLS-II/bluesky/blob/a64a205f1b7f090c99ab74074e913ce444c56841/bluesky/run_engine.py#L985

ZLLentz: I stand corrected. I still think it's weird to keep the sequence running when the scan is paused.

pcdsdevices/sequencer.py
ZLLentz: Seems fine, just want no surprises.

ZLLentz:
One more question: is this just for things like the laser triggers? There already was a sync sequence with DAQ button in the DAQ interface.

pcdsdevices/sequencer.py
```diff
@@ -72,19 +72,37 @@ def kickoff(self):
...
+        # Start the sequencer
+        logger.debug("Starting EventSequencer ...")
+        self.play_control.set(1)
+
+    def pause(self):
+        """Stop the event sequencer and stop monitoring events"""
+        # Order a stop
+        self.stop()
+        # Pause monitoring
+        super().pause
```
ZLLentz: typo

pcdsdevices/sequencer.py
```diff
@@ -0,0 +1,160 @@
...
+
+from ophyd import Device, EpicsSignal, EpicsSignalRO, Component as Cpt
+from ophyd.status import DeviceStatus, SubscriptionStatus
+from ophyd.utils.epics_pvs import raise_if_disconnected
+from ophyd.flyers import FlyerInterface, MonitorFlyerMixin
+
+logger = logging.getLogger(__name__)
+
+
+class EventSequencer(Device, MonitorFlyerMixin, FlyerInterface):
```
n-wbrown: I'm comparing this with the sequencer we made for sxr.

I notice that there's no way to access the beam owner name or ID. Is this functionality put somewhere else?

There's no :CURSTP pv which tells the current step in the sequencer. Is this intentional?

There's no components for the PVs in the right hand column of the EDM screen for sequencers. These are the synchronization settings.

There's no interface for getting beam rate. Has this been moved to a better place?

teddyrendahl:
@ZLLentz I think it is two ways to solve the same problem. It sounds like from talking to Silke that this is the common mode of operation in XPP. I can tell you from experience that isn't what is typically done in CXI/MFX. It also appears like a need as SXR wrote an EventSequencer class for their own scans.

Also, since the discussion is now hidden, I have implemented `pause`, `resume` and `start` for the EventSequencer. This way the EventSequencer starts and stops when a user interrupts their scan. The one thing I'm worried about is that this will restart what ever sequence was programmed. If you are repeating ad infinitum this is probably not a huge deal. However, if you are running some large single looped sequence it might affect you. I placed the following note in the docstring to remind users:

```
    It is ambiguous what the correct behavior for the EventSequencer is when we
    pause and resume during a scan. The current implementation will stop the
    EventSequencer and restart the sequence from the beginning. This may impact
    applications which depend on a long single looped sequence running through
    out the scan
```

pcdsdevices/sequencer.py
ZLLentz: Stepping in to note that `CURSTP` is included and that beam rate is actually a different IOC, though admittedly we haven't made an object for it.

teddyrendahl: Thanks for comparing! I do have `CURSTP` but I am missing the rest. I'll add all these in.

teddyrendahl: Argh. Ownership PVs have a different prefix. Adding sync settings then calling it good.

ZLLentz: I'm surprised that the ownership PVs have a different prefix. Are they tied to the beam rate ioc or something completely different?

teddyrendahl: I think there is some base `EventSequencer` that is referenced for these `ECS:SYS:0`. This is where the beam rate and ownership come from. Other EventSequencers have `ECS:SYS:{HUTCH_NO}`. I guess if they are constant for all the sequencers I could just put them in as FormattedComponents and tell it not to add prefixes?

ZLLentz: I decided to check, because I also have a computer. This is True for the beam owner, but not for the hutch owner. I think we should include `HUTCH_ID` for sure, but I'm 50/50 on doing the `ECS:SYS:0` PVs.
```
$ caget ECS:SYS0:{0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16}:HUTCH_ID
Channel connect timed out: some PV(s) not found.
ECS:SYS0:0:HUTCH_ID            *** not connected
ECS:SYS0:1:HUTCH_ID            1
ECS:SYS0:2:HUTCH_ID            2
ECS:SYS0:3:HUTCH_ID            3
ECS:SYS0:4:HUTCH_ID            4
ECS:SYS0:5:HUTCH_ID            5
ECS:SYS0:6:HUTCH_ID            6
ECS:SYS0:7:HUTCH_ID            7
ECS:SYS0:8:HUTCH_ID            3
ECS:SYS0:9:HUTCH_ID            7
ECS:SYS0:10:HUTCH_ID           2
ECS:SYS0:11:HUTCH_ID           3
ECS:SYS0:12:HUTCH_ID           4
ECS:SYS0:13:HUTCH_ID           0
ECS:SYS0:14:HUTCH_ID           0
ECS:SYS0:15:HUTCH_ID           0
ECS:SYS0:16:HUTCH_ID           5

$ caget ECS:SYS0:{0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16}:BEAM_OWNER_ID
Channel connect timed out: some PV(s) not found.
ECS:SYS0:0:BEAM_OWNER_ID       2
ECS:SYS0:1:BEAM_OWNER_ID       *** not connected
ECS:SYS0:2:BEAM_OWNER_ID       *** not connected
ECS:SYS0:3:BEAM_OWNER_ID       *** not connected
ECS:SYS0:4:BEAM_OWNER_ID       *** not connected
ECS:SYS0:5:BEAM_OWNER_ID       *** not connected
ECS:SYS0:6:BEAM_OWNER_ID       *** not connected
ECS:SYS0:7:BEAM_OWNER_ID       *** not connected
ECS:SYS0:8:BEAM_OWNER_ID       *** not connected
ECS:SYS0:9:BEAM_OWNER_ID       *** not connected
ECS:SYS0:10:BEAM_OWNER_ID      *** not connected
ECS:SYS0:11:BEAM_OWNER_ID      *** not connected
ECS:SYS0:12:BEAM_OWNER_ID      *** not connected
ECS:SYS0:13:BEAM_OWNER_ID      *** not connected
ECS:SYS0:14:BEAM_OWNER_ID      *** not connected
ECS:SYS0:15:BEAM_OWNER_ID      *** not connected
ECS:SYS0:16:BEAM_OWNER_ID      *** not connected
```

ZLLentz: Maybe it's a good candidate for the beam stats class, beam owner seems closely tied to beam stats. The beam stats class will have tons of ugly PVs from different sources anyway.

ZLLentz:
@teddyrendahl congrats on finding another fake epics signal testing race condition

teddyrendahl:
Race conditions cleaned up. Ran this consecutively for about half hour with no issues 🤷‍♂️

PR #204, opened 2018-04-13
WIP: Dynamic States

ZLLentz:
## Description

Allow setting a `_dynamic_states` boolean to `True` to grab enum states information from the `states` signal. Everything else is a consequence of that change, while continuing to meet the design constraint of not using our signal connections during `__init__`.

## Motivation and Context

IPMS and XFLS have a varied array of states names. We can't cover them all with different classes.

## How Has This Been Tested?

n/a

n/a

This is a work in progress, I haven't reviewed the code myself. I'm stopping here due to splitting headache. This needs to be thought over, completed, tested, documented, etc.

### DIFF
--- a/pcdsdevices/inout.py
+++ b/pcdsdevices/inout.py
@@ -21,13 +21,13 @@ class InOutPositioner(StatePositioner):
 %s
     Attributes
     ----------
-    in_states: ``list` of ``str``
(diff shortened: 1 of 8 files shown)


ZLLentz:
#205 is a better approach

PR #205, opened 2018-04-16
ENH: States Updates for IPM and XFLS

### LINKED ISSUE #150: IPM State Names
## Expected Behavior

The `IPM` class does not work with the default IOC configuration. The class assumes that the state names are `T*` while the PVs are `TARGET*`.

## Current Behavior

Assumes PV names are `T*`

## Possible Solution

1. Have a `kwarg` to pass in the state prefix i.e `TARGET*`, `T*`
2. Set it to the default `TARGET` and have XPP have a special class

### LINKED ISSUE #203: IPM factory function
## Expected Behavior

The `IPM` IOC does not have any specific names for the targets therefore there is a large variance from hutch to hutch on naming conventions.

## Current Behavior

Must subclass `IPM` to change state names

## Possible Solution

Split up `IPM` and `IPMBase` where `IPMBase` is almost identical to its' current from with the exception of the `states_list` and `target_names`. `IPM` is then a factory function that creates an a class of type `IPMBase`  with these states specified. By setting the default states in this function to match the current `IPM` class we maintain back-compat ⛪️ 

### Note
All this applies to the `XFLS` class as well.

ZLLentz:
## Description

- Add EPICS enums as primary aliases
- Put integer instead of str for moves
- Map transmissions with integers
- Adjust IPMs to work with these changes
- Fix bug where readback was broken for all IPM
- Remove redundant docstrings

## Motivation and Context

Force IOCs to have the same PVs for configuring the same devices. Take the EPICS enum so that users can create aliases in EPICS.
closes #203 
closes #150 

## How Has This Been Tested?

Updated tests and it works with the real IPM objects

### DIFF
--- a/pcdsdevices/pulsepicker.py
+++ b/pcdsdevices/pulsepicker.py
@@ -3,7 +3,7 @@ Module for the LCLS1 `PulsePicker`
 """
 import logging
 
-from ophyd.device import Component as Cmp, FormattedComponent as FCmp
+from ophyd.device import Component as Cpt, FormattedComponent as FCpt
 from ophyd.signal import EpicsSignal, EpicsSignalRO
 from ophyd.status import SubscriptionStatus, wait as status_wait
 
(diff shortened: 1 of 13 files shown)


ZLLentz:
@teddyrendahl I fixed the issues, some of it was testing related and some of it was real bugs.

pcdsdevices/pulsepicker.py
```diff
@@ -173,14 +173,14 @@ class PulsePickerInOut(PulsePicker):
...
+    inout = FCpt(InOutRecordPositioner, '{self._inout}')
 
     states_list = ['OUT', 'OPEN', 'CLOSED']
     out_states = ['OUT', 'OPEN']
-    _state_logic = {'inout.state': {1: 'OUT',
-                                    2: 'defer',
-                                    'OUT': 'OUT',
-                                    'IN': 'defer'},
+    _state_logic = {'inout.state': {1: 'defer',
+                                    2: 'OUT',
```
teddyrendahl: 😨

PR #207, opened 2018-04-17
FIX: flock test on NFS

### LINKED ISSUE #206: Preset Tests Fail on NFS
Mildly alarming, these tests do not pass on NFS, e.g. `psbuild-rhel7` and `psbuild-rhel6`. The tests fail right near the end, so most of the cases pass, so this isn't a five-alarm fire, but it's concerning.

The tests work on my laptop, on my desktop, and on travis.

ZLLentz:
closes #206 

Nearest I can tell, `flock` on NFS doesn't work unless the file is opened in write or append or some equivalent mode... We don't "need" to write to the file to run the test here, but empirically it seems like we do need it to avoid an exception on NFS.

The code bit here is simply starting a second process and locking the file, and checking that the presets behave as expected in this situation. The `mp.Lock` is used to tell the process to release the file lock.

### DIFF
--- a/tests/test_mv_interface.py
+++ b/tests/test_mv_interface.py
@@ -142,7 +142,7 @@ def test_presets(presets, motor):
     assert len(motor.presets.positions.sample.history) == 2
 
     def block_file(path, lock):
-        with open(path, 'r') as f:
+        with open(path, 'r+') as f:
             fcntl.flock(f, fcntl.LOCK_EX)
             lock.acquire()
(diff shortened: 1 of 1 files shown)


teddyrendahl:
Fascinating. Good news is this shouldn't affect actual `Presets` in the field then since they should always be opened in write mode.

https://github.com/pcdshub/pcdsdevices/blob/b047577afeb9ae70e0e96ec7bb5fe74748cfd6f0/pcdsdevices/mv_interface.py#L248

PR #208, opened 2018-04-19
FIX: IMS velocity has limits

### LINKED ISSUE #209: Velocity in IMS record does not check limits
## Description

The `IMS` record has limits on the velocity field. By default, signals do not look at limits when setting. In the current software, an incorrect setting is not noticed by Python until after the device moves and an alarm is generated. 

Issue noticed by @koglin @yashrao99

ZLLentz:
closes #209

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -213,6 +213,9 @@ class IMS(PCDSMotorBase):
     error_severity = Component(EpicsSignal, '.SEVR')
     part_number = Component(EpicsSignalRO, '.PN')
 
+    # IMS velocity has limits
+    velocity = Component(EpicsSignal, '.VELO', limits=True)
+
     def stage(self):
         """
         State the IMS motor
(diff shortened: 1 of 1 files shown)


ZLLentz:
Open question: how to handle pyepics caching the limits

teddyrendahl:
@ZLLentz I agree it is an issue but not one that should hold up this Pull Request. Where should the relevant issue go? It sounds like we are confident this a `pyepics` issue....

ZLLentz:
@teddyrendahl this might actually be OK.

in `EpicsSignalBase`:

```
    @property
    @raise_if_disconnected
    def limits(self):
        '''The read PV limits'''


        # This overrides the base limits
        with self._lock:
            pv = self._read_pv
            pv.get_ctrlvars()
            return (pv.lower_ctrl_limit, pv.upper_ctrl_limit)
```

In `pyepics`

```
    def get_ctrlvars(self, timeout=5, warn=True):
        "get control values for variable"
        if not self.wait_for_connection():
            return None
        kwds = ca.get_ctrlvars(self.chid, timeout=timeout, warn=warn)
        self._args.update(kwds)
        self.force_read_access_rights()
        return kwds
```

Issue #214, opened 2018-04-20
PointingMirror states are broken


teddyrendahl:
## Current Behavior
```
ValueError: 1 is not a valid state for xrt_m1h. Valid state names are: ['Unknown', 'IN'], and their corresponding values are [0, 2].
```
There is also the issue that these PVs exist but do not function. I will make these state-less mirrors for now in the `happi` config but this should be solved.

<details closed>
<summary> Full Traceback </summary>
<pre>
In [3]: xrt_m1h.inserted
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/reg/neh/home4/trendahl/sandboxes/pcds/cxi/dev/devpath/pcdsdevices/state.py in get_state(self, value)
    211         try:
--> 212             return self.states_enum[value]
    213         except KeyError:

/reg/g/pcds/pyps/conda/py36/envs/pcds-1.0.0/lib/python3.6/enum.py in __getitem__(cls, name)
    326     def __getitem__(cls, name):
--> 327         return cls._member_map_[name]
    328 

KeyError: 1

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
/reg/neh/home4/trendahl/sandboxes/pcds/cxi/dev/devpath/pcdsdevices/state.py in get_state(self, value)
    214             try:
--> 215                 return self.states_enum(value)
    216             except ValueError:

/reg/g/pcds/pyps/conda/py36/envs/pcds-1.0.0/lib/python3.6/enum.py in __call__(cls, value, names, module, qualname, type, start)
    290         if names is None:  # simple value lookup
--> 291             return cls.__new__(cls, value)
    292         # otherwise, functional API: we're creating a new Enum type

/reg/g/pcds/pyps/conda/py36/envs/pcds-1.0.0/lib/python3.6/enum.py in __new__(cls, value)
    532         # still not found -- try _missing_ hook
--> 533         return cls._missing_(value)
    534 

/reg/g/pcds/pyps/conda/py36/envs/pcds-1.0.0/lib/python3.6/enum.py in _missing_(cls, value)
    545     def _missing_(cls, value):
--> 546         raise ValueError("%r is not a valid %s" % (value, cls.__name__))
    547 

ValueError: 1 is not a valid PointingMirrorStates

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
[...]

PR #216, opened 2018-04-20
ENH: fail in an obvious way if state is bad

ZLLentz:
## Description

- Raise a `ValueError` with a helpful string if something has gone wrong in creating the states enum

This is related to #214, to make it more obvious what has happened in these instances.

- Fix a bug where `None` could make it into the `_valid_states` list if used as a skip state
- Fix a bug where transmission would be malformed if `None` used as a skip state

In practice we aren't using this incidental `None` skip state feature, but I wanted to maintain it just in case. I noticed these bugs because other tests started failing after fixing the original bug.

### DIFF
--- a/pcdsdevices/state.py
+++ b/pcdsdevices/state.py
@@ -69,13 +69,13 @@ class StatePositioner(Device, PositionerBase, MvInterface):
     def __init__(self, prefix, *, name, **kwargs):
         super().__init__(prefix, name=name, **kwargs)
         self._valid_states = [state for state in self.states_list
-                              if state not in self._invalid_states]
+                              if state not in self._invalid_states
+                              and state is not None]
         if self._unknown:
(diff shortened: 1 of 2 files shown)


pcdsdevices/state.py
```diff
@@ -255,7 +257,12 @@ def _create_states_enum(self):
                 for alias in aliases:
                     state_def[alias] = i
         enum_name = self.__class__.__name__ + 'States'
-        return Enum(enum_name, state_def, start=0)
+        enum = Enum(enum_name, state_def, start=0)
+        if len(enum) != state_count:
+            raise ValueError(('Bad states definition! Inconsistency in '
+                              'states_list {} or _states_alias {}'
+                              ''.format(self.states_list, self._states_alias)))
```
teddyrendahl: Nice!

Issue #220, opened 2018-04-26
AreaDetector Setup is Confusing and Unintuitive


ZLLentz:
From a discussion I had with @koglin.
The way that we're doing area detector here is very confusing and hard to follow. We have specific `DetectorBase` subclasses that do the same thing (adapt to our PV conventions) and what we use in the `PIM` object is a holdover from skywalker where very specific things are emphasized over everything else.

I suggest the following fixes:
- areadetector/detectors.py should have an `AreaDetector` class instead of `PulnixDetector` and `FeeOpalDetector` and a `DefaultAreaDetector` class with sane default plugins
- areadetector/cam.py should be cleaned up and looked over, it is tough to follow. This isn't being used anywhere, but you could imagine that these sorts of classes would be useful.
- areadetector/plugins.py should be cleaned up and looked over, we should see how many of our overrides made it into the `ophyd` library
- `PIM` should switch to using `DefaultAreaDetector` instead of having a special `PIMPulnixDetector`.

teddyrendahl:
Grabbing this one!

- [ ] Support for `AreaDetector Stream`
- [ ] Remove `areadetector` support from `PIM`
- [ ] Deprecate existing plugins

Issue #120, opened 2018-01-11
PCDS Areadetector Setup


teddyrendahl:
## Expected Behavior

We should have a base AreaDetector implementation for the standard PCDS setup. This will replace all GigECam, edtCam setup scripts everywhere. 

* `IMAGE1` -> Full Rate
* `IMAGE2` -> Low rate

ZLLentz:
closed by #221

PR #221, opened 2018-04-26
MNT: Clean up ad detector classes

ZLLentz:
WIP: See #220

### DIFF
--- a/pcdsdevices/areadetector/detectors.py
+++ b/pcdsdevices/areadetector/detectors.py
@@ -16,18 +16,18 @@ from .plugins import ImagePlugin, StatsPlugin
 logger = logging.getLogger(__name__)
 
 
-__all__ = ['AreaDetector',
-           'DefaultAreaDetector']
+__all__ = ['PCDSDetectorBase',
+           'PCDSDetector']
 
 
-class AreaDetector(DetectorBase):
+class PCDSDetectorBase(DetectorBase):
     """
     Standard area detector with no plugins.
     """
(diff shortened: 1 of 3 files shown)


teddyrendahl:
Also addresses #120

teddyrendahl:
Few comments:

1) I don't think `Stats` should be enabled by default. I see the benefit of having it declared as a `Component` but most AreaDetector set ups won't want this enabled. 
2) Don't we want `Image1` and `Image2`? Shouldn't we be setting at least one of these to a specific rate for the viewer?
3) I'm not sure I like the name `DefaultAreaDetector` I think I would prefer `PCDSAreaDetector` but can't justify this with any logic.

ZLLentz:
1. Agree on not enabling `Stats` by default. I started from a long list of stage sigs and wasn't sure how far to cut it down.
2. Assumed we only care about the low-rate or ROI image in Python 99% of the time
3. No preference on the name, happy to change it

ZLLentz:
Actually I'm backtracking on point 1 a bit. When do we stage an areadetector but not want the stats?

teddyrendahl:
I think we might not be imagining the same endpoint for this device.

ZLLentz:
@teddyrendahl I'd at least like to finalize class names here so I can work on other issues that need a detector subclass.

ZLLentz:
I will change the names and merge the PR, then we need to have a larger discussion about cleaning up this file and the other two in our areadetector dir

PR #225, opened 2018-05-05
FIX: Misc areadetector fixes

ZLLentz:
## Description

- switch off of ADComponent because lazy=True
- fix typo in IMAGE2
- remove stage_sigs covered in base classes
- express remaining stage_sigs as strings to avoid signal getattrs at init
- patch around signal gets in ophyd's plugin base initialization

## Motivation and Context

- lazy=True is bad
- redundancy is bad
- Our AD classes blow up if certain PVs are missing, which means they access PVs at init... unless we do these things, or some equivalent

### DIFF
--- a/pcdsdevices/areadetector/detectors.py
+++ b/pcdsdevices/areadetector/detectors.py
@@ -7,7 +7,6 @@ functions needed by all instances of a detector are added here.
 import logging
 
 from ophyd.areadetector import cam
-from ophyd.areadetector.base import ADComponent
 from ophyd.areadetector.detectors import DetectorBase
 from ophyd.device import Component as Cpt
 
@@ -24,7 +23,7 @@ class PCDSDetectorBase(DetectorBase):
     """
     Standard area detector with no plugins.
     """
-    cam = ADComponent(cam.CamBase, ":")
+    cam = Cpt(cam.CamBase, ":")
 
 
 class PCDSDetector(PCDSDetectorBase):
@@ -36,7 +35,7 @@ class PCDSDetector(PCDSDetectorBase):
     IMAGE2: reduced rate image
     Stats2: reduced rate stats
     """
-    image = Cpt(ImagePlugin, ':IMAGE1:', read_attrs=['array_data'])
+    image = Cpt(ImagePlugin, ':IMAGE2:', read_attrs=['array_data'])
     stats = Cpt(StatsPlugin, ':Stats2:', read_attrs=['centroid',
                                                      'mean_value',
                                                      'sigma_x',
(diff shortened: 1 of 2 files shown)


pcdsdevices/areadetector/detectors.py
```diff
@@ -36,15 +35,13 @@ class PCDSDetector(PCDSDetectorBase):
...
+    image = Cpt(ImagePlugin, ':IMAGE2:', read_attrs=['array_data'])
     stats = Cpt(StatsPlugin, ':Stats2:', read_attrs=['centroid',
                                                      'mean_value',
                                                      'sigma_x',
                                                      'sigma_y'])
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.image.stage_sigs[self.image.enable] = 1
-        self.stats.stage_sigs[self.stats.enable] = 1
```
teddyrendahl: Can you explain why these are redundant?

pcdsdevices/areadetector/plugins.py
```diff
@@ -21,6 +21,14 @@ class PluginBase(ophyd.plugins.PluginBase, ADBase):
     """
     enable = C(EpicsSignal, 'EnableCallbacks_RBV.RVAL', write_pv="EnableCallbacks", string=False)
 
+    def __init__(self, *args, **kwargs):
+        # Avoid ophyd PluginBase's init
+        ADBase.__init__(self, *args, **kwargs)
```
teddyrendahl: Dislike this. Divorcing ourselves from the constructor emit seems drastic to avoid one `caget`. I think this is really a problem with the fact that this kind of behavior doesn't belong ing `PluginBase.__init__` and it should be fixed on the `ophyd` end not patched here.

pcdsdevices/areadetector/detectors.py
ZLLentz: https://github.com/NSLS-II/ophyd/blob/2cbfccdc54d749e17e87cda0d5d47c129749c2ee/ophyd/areadetector/plugins.py#L70-L71

pcdsdevices/areadetector/plugins.py
ZLLentz: It avoids two cagets btw

ZLLentz: I'm willing to submit the less obvious one as a PR:
`'cam' in self.parent.component_names` vs `hasattr(self.parent, 'cam')`

But it's hard for me to find words to justify not checking the plugin type

teddyrendahl: I mean.  I think the main problem is that the `plugin_type` not matching is an issue in the class construction. Why are we checking this on every `__init__`?

ZLLentz: it's not the getting that's bad, it's the `wait_for_connection` that gunks up the whole initialization

ZLLentz:
I'm going to trash this PR and make a new one that just changes `IMAGE1` to `IMAGE2`. At some point we need to revisit AD devices but we'll wait and see if there are problems with the real devices rather than trying to anticipate. The current organization with the base device is at least useful and isn't "wrong".

PR #227, opened 2018-05-10
ENH: Add AvgSignal, add mj_avg to BeamStats

ZLLentz:
## Description

- Port and update `AvgSignal` from `pswalker`
- add `mj_avg` field to `BeamStats`

## Motivation and Context

- `AvgSignal` was useful for suspenders, and it will be useful in `nabs` for the suspend-and-drop construct I'm working on
- The beam mj level is a common thing to be averaged, since an individual reading going bad does not necessarily mean the beam is down.

## How Has This Been Tested?

Tests added, and did "feel" tests on psbuild-rhel7

- `AvgSignal` automatically included in API docs
- `BeamStats` added to API docs

### DIFF
--- a/pcdsdevices/signal.py
+++ b/pcdsdevices/signal.py
@@ -125,14 +125,14 @@ class AvgSignal(Signal):
 
     def get(self, *args, **kwargs):
         self._ensure_sub()
-        super().get(*args, **kwargs)
+        return super().get(*args, **kwargs)
 
     def put(self, *args, **kwargs):
         raise ReadOnlyError()
 
     def subscribe(self, *args, **kwargs):
         self._ensure_sub()
-        super().subsribe(*args, **kwargs)
+        return super().subsribe(*args, **kwargs)
 
     @property
     def averages(self):
--- a/pcdsdevices/signal.py
+++ b/pcdsdevices/signal.py
@@ -3,7 +3,9 @@ Module to define ophyd Signal subclass utilities.
 """
 import logging
 from threading import RLock
+
 from ophyd.signal import Signal
+from ophyd.utils.errors import ReadOnlyError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


pcdsdevices/signal.py
```diff
@@ -100,3 +102,70 @@ def _run_sub_value(self, *args, **kwargs):
...
+
+    def _update_avg(self, *args, value, **kwargs):
+        """
+        Add new value to the buffer, overriding old values if needed.
+        """
+        with self._lock:
+            self.values[self.index] = value
+            self.index += 1
+            if self.index == len(self.values):
+                self.index = 0
```
teddyrendahl: Instead of keeping track of a `self.index` you could always use `np.roll` and replace the first value.

```python
x = array([1, 2, 3, 4])
x = np.roll(x, shift=1)
x
array([4, 1, 2, 3])
```

pcdsdevices/signal.py
```diff
@@ -100,3 +102,70 @@ def _run_sub_value(self, *args, **kwargs):
...
+            # This takes a mean, skipping nan values.
+            self.put(np.nanmean(self.values))
```
teddyrendahl: Do we want to report a value if the buffer isn't filled?

ZLLentz: It makes the code more complicated to hold back values until the buffer is ready. We have no guarantee on how long it will take the buffer to fill, so I opted for this approach. Other approaches would be to have `get` return a `None` or `NaN`, or have `get` try a `wait_for_connection` and keep `connected=False` until the buffer is full. But again, this can take indeterminate time to occur, though it will be finite time as long as the PV is updating. Let me know what you think.

pcdsdevices/signal.py
ZLLentz: I didn't know about np.roll.
```
x = np.array([np.nan, np.nan, np.nan, np.nan])
for i in range(5):
    x = np.roll(x, shift=1)
    x[0] = i
    print(x)
[ 0. nan nan nan]
[ 1.  0. nan nan]
[ 2.  1.  0. nan]
[3. 2. 1. 0.]
[4. 3. 2. 1.]
```
Seems ok, but I read the source code and this allocates a new array every time. I'd much rather keep modifying the array in place because this callback might be run at 120Hz.

pcdsdevices/signal.py
teddyrendahl: I think I agree. Maybe just put a note in the docstring so the behavior does not surprise anyone.

pcdsdevices/signal.py
teddyrendahl: Good point. The idea was that it would be nice if you could always tell the order of events but this is probably unnecessary especially at the cost of performance. You checked this against a 120Hz PV at all?

pcdsdevices/signal.py
ZLLentz: I agree, I'll extend the docstring so the behavior is not transparent.

pcdsdevices/signal.py
ZLLentz: I checked it in the `BeamStats` object and it seems ok, though python prints at 120Hz are surprisingly expensive. The stuttering is clearly from the gateway issue.

PR #228, opened 2018-05-17
ENH: CCM, Conventions for Combined Axes and Calc Motors

### LINKED ISSUE #222: Add CCM Class
See title

ZLLentz:
## Description

- Add `CCM`, which includes two coordinated motor assemblies and an `energy` pseudomotor
- Add `SyncAxes` class factory

Unfortunately, the CCM currently uses a bunch of magic constants for the calcs and kwargs for the in/out positions.

closes #222 

## Motivation and Context

- Need a CCM for XPP (eventually XCS will use theirs again maybe), and for downstream hutches to verify CCM is out
- Need conventions for synchronized axes and calculation motors

## How Has This Been Tested?

Added tests

Added to device docs

### DIFF
--- a/pcdsdevices/ccm.py
+++ b/pcdsdevices/ccm.py
@@ -224,6 +224,6 @@ class CCM(InOutPositioner):
     @_state.setter
     def _state(self, value):
         if value == 1:
-            self.x.move(self._inpos)
+            self.x.move(self._inpos, wait=False)
         elif value == 2:
-            self.x.move(self._outpos)
+            self.x.move(self._outpos, wait=False)
--- a/pcdsdevices/ccm.py
+++ b/pcdsdevices/ccm.py
@@ -1,3 +1,5 @@
+import time
+
 import numpy as np
 from ophyd.device import Component as Cpt, FormattedComponent as FCpt
 from ophyd.pseudopos import PseudoPositioner, PseudoSingle
@@ -30,8 +32,20 @@ def alio_to_theta(alio, gTheta0, gR, gD):
     """
     Converts alio position (mm) to theta angle (rad)
     """
-    theta = 2*np.arctan((np.sqrt(alio**2+gD**2+2*gR*alio)-gD)/(2*gR+alio))
-    return gTheta0 + theta
+    low = -1.0
+    high = 1.0
+    timeout = 1.0
+    start = time.time()
+    while time.time() - start < timeout:
+        theta_guess = (low+high)/2
+        alio_calc = theta_to_alio(theta_guess, gTheta0, gR, gD)
+        if np.isclose(alio, alio_calc):
+            break
+        elif alio_calc > alio:
+            high = theta_guess
+        else:
+            low = theta_guess
+    return theta_guess
 
 
 def wavelength_to_theta(wavelength, gdspacing):
--- a/pcdsdevices/ccm.py
+++ b/pcdsdevices/ccm.py
@@ -1,3 +1,4 @@
+import logging
 import time
 
 import numpy as np
@@ -10,6 +11,7 @@ from .epics_motor import IMS
 from .inout import InOutPositioner
 from .pseudopos import SyncAxesBase
 
+logger = logging.getLogger(__name__)
 
 # Constants
 si_111_dspacing = 3.1356011499587773
@@ -114,7 +116,7 @@ class CCMCalc(PseudoPositioner):
 
     def __init__(self, *args, theta0=default_theta0, dspacing=default_dspacing,
                  gr=default_gr, gd=default_gd, **kwargs):
-        super().__init__(*args, **kwargs)
+        super().__init__(*args, auto_target=False, **kwargs)
         self.theta0 = theta0
         self.dspacing = dspacing
         self.gr = gr
@@ -135,6 +137,7 @@ class CCMCalc(PseudoPositioner):
             theta = pseudo_pos.theta*np.pi/180
         else:
             alio = self.alio.position
+        logger.debug((energy, wavelength, theta))
         if energy is not None:
             wavelength = energy_to_wavelength(energy)
         if wavelength is not None:
--- a/pcdsdevices/ccm.py
+++ b/pcdsdevices/ccm.py
@@ -32,7 +32,7 @@ class CCMMotor(PVPositionerPC):
     loading, and make the wait for done just compare the values.
     """
     setpoint = Cpt(EpicsSignal, ":POSITIONSET")
-    readback = Cpt(EpicsSignalRO, ":POSITIONGET")
+    readback = Cpt(EpicsSignalRO, ":POSITIONGET", kind='hinted')
 
     limits = None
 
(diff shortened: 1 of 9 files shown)


pcdsdevices/ccm.py
```diff
@@ -0,0 +1,185 @@
...
+    start = time.time()
+    while time.time() - start < timeout:
+        theta_guess = (low+high)/2
+        alio_calc = theta_to_alio(theta_guess, gTheta0, gR, gD)
+        if np.isclose(alio, alio_calc):
+            break
+        elif alio_calc > alio:
+            high = theta_guess
+        else:
+            low = theta_guess
```
teddyrendahl: Oh boy. This is unclean. If you tell me its necessary than its neccesary I suppose. Should `np.isclose` have a set-able tolerance?

pcdsdevices/pseudopos.py
```diff
@@ -0,0 +1,59 @@
...
+    its position as the average value of all axes.
+
+    Attributes
+    ----------
+    axes: ``list of str``
+        Names of the components that correspond to the real axes.
+    """
+    pseudo = Cpt(PseudoSingle)
+
+    def forward(self, pseudo_pos):
```
teddyrendahl: This is a little oversimplified than what I think is needed for other applications. I think that is fine for now, but should be addressed later. For instance this will not preserve the difference between all the motors it will just set them to the same value. This will certainly cause issues with the stands but maybe the CCM has nice enough offsets that this doesn't matter

pcdsdevices/ccm.py
ZLLentz: You are free to find the analytic inversion of the other method if you'd like. I think this binary search inverter should be a generic function in a utils package. There is no need to have a settable tolerance because nobody will ever use the knob. I might tweak the default tolerance to get a tighter fit.

pcdsdevices/pseudopos.py
ZLLentz: CCM offsets are all zero. I didn't think we would be ok with having level stands with nonzero offsets.

ZLLentz: After some thought, probably best to record axis offsets before the first move and whenever requested and maintain those offsets with absolute moves

ZLLentz:
@teddyrendahl this is mostly ready to go now. This was a struggle because I was learning the nuances of `PseudoPositioner`. I think the result is fairly clean.

Further improvement would be to mandate a state in/out record (none exist, so the python takes care of it...) and then remove the top-level CCM device in favor of a happi namespace like we were talking about.

I guess I could also just move the in/out handling down to the relevent coordinated axis... and remove the top-level class. Yeah ok maybe tomorrow.

pcdsdevices/pseudopos.py
```diff
@@ -0,0 +1,83 @@
...
+        If not done earlier, this will be automatically run before it is first
+        needed (generally, right before the first move).
+
+        Note that this method does not protect you from getting your offsets
+        into an inconsistent state! Pass arguments into the `offsets` param at
+        your own risk.
+
+        Parameters
+        ----------
+        **offsets: ``int`` or ``float``, optional
```
teddyrendahl: Tempted to say this is too much power. Maybe they just adjust the motors and `save_offsets`...

pcdsdevices/pseudopos.py
```diff
@@ -0,0 +1,83 @@
...
+            Mapping of axis name to offset. If omitted, we'll use the current
+            positions. If a partial dictionary is provided, the omitted values
+            are assumed to be either 0 (if no value is available) or unchanged
+            (if previous saved values are available).
+        """
```
teddyrendahl: Maybe a `logger.info` message in here. The operator should be informed when we update the `offsets`

pcdsdevices/pseudopos.py
```diff
@@ -0,0 +1,83 @@
...
+    ----------
+    position_mode: ``func``, optional
+        The function to apply to the list of current positions to determine
+        the combined position. By default, this is the minimum of all axis
+        positions.
+
+    Attributes
+    ----------
+    axes: ``list of str``
+        Names of the components that correspond to the real axes.
```
teddyrendahl: You mind just adding a simple example here? I understand what you meant but I think seeing it might help newer folks:

```python
class Parallel(SyncAxesBase):
     left = Cpt(EpicsMotor, ':01')
     right = Cpt(EpicsMotor(':02')
```
?

pcdsdevices/pseudopos.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    pseudo = Cpt(PseudoSingle)
+
+    _default_mode = min
```
teddyrendahl: Any reason `min` would be the default? `mean` seems like the most intuitive to me but obviously an opinion.

pcdsdevices/pseudopos.py
ZLLentz: I'm tempted to agree, it's not hard to remove

pcdsdevices/pseudopos.py
ZLLentz: Sure, sounds reasonable.

pcdsdevices/pseudopos.py
ZLLentz: I agree that any of these `Base` classes should have an example, I'll add it in.

The docstring in this code snippet is wrong btw... from an old version of the file.

pcdsdevices/pseudopos.py
ZLLentz: I originally had `mean` as the default but decided that it was conceptually easier to think of one motor being the "anchor" of the position and defining everything else relative to the motor, like the Beckhoff coupled axes with the master and slave axis. The `mean` is more nebulous as the position returned may not match any of the composite motors.

`min` vs `max` vs picking a random motor in the middle was pretty arbitrary.

pcdsdevices/pseudopos.py
ZLLentz: On further consideration, is this needed? It will happen every first time you move a SyncAxes... e.g. `ccm.insert()` will start with `INFO: Offsets saved`

teddyrendahl: hmmm. well a `debug` at the least.

pcdsdevices/pseudopos.py
teddyrendahl: I see where you are coming from but I think there is a clear difference for this case. You are not explicitly labeling a leader and a follower. You'll read from one or the other based on position essentially swapping the "leader". I think unless you actually label one the follower `mean` is more intuitive.

ZLLentz: Strong disagree. I'd rather refactor so that motor #1 is the anchor value than swap from `min` to `mean`.

ZLLentz: e.g. I'd rather set it as

```
_default_mode = lambda pos: pos.__getitem__(0)
```

teddyrendahl: That makes more sense to me than `min` ⬆️

ZLLentz: After thinking it over, I share you concerns with the arbitrary nature of `min`. I'll swap "first" to be the default behavior and adjust the tests accordingly.

I think `mean` makes sense for the detector tables fwiw, it's just the general case that has me worried.

teddyrendahl: :+1: The important part is the user can pick with a line of Python code 😄

ZLLentz: Something "fun": built-ins attached to classes do not get passed "self", but all other functions do. So when I swapped `min` to `first` I got sad.

This will now be a subclass thing instead of a passable/class def override so that people can't make the mistake I just did.

ZLLentz:
@teddyrendahl this is also ready for the final go-ahead

pcdsdevices/pseudopos.py
```diff
@@ -24,24 +29,31 @@ class Parallel(SyncAxesBase):
...
 
+    def calc_combined(self, real_position):
+        """
+        Calculate the combined pseudo position.
+
+        By default, this is just the position of our first axis.
+
+        Parameters
+        ----------
+        real_position: `namedtuple`
```
teddyrendahl: Nice! Like the `namedtuple`

ZLLentz: This is inherited from the PseudoMotor schema

PR #229, opened 2018-05-18
ENH: Add make_fake_class factory function

ZLLentz:
## Description

`make_fake_class(cls)` creates a fake version of `cls`. All `EpicsSignal` and `EpicsSignalRO` will be replaced. They will start at value `None` and stay at whatever value you set. There are minimal hooks for recreating functionality of the real things.

## Motivation and Context

We need a dead-simple way to test classes that doesn't have update loops or threading with the potential for race conditions.

## How Has This Been Tested?

Added tests

I will add to the docs once we agree on the filename it should go in

### DIFF
--- a/pcdsdevices/utils.py
+++ b/pcdsdevices/utils.py
@@ -1,0 +1,148 @@
+import copy
+import logging
+
+from ophyd.device import (Device, Component as Cpt,
+                          DynamicDeviceComponent as DDCpt)
+from ophyd.signal import Signal, EpicsSignal, EpicsSignalRO
+from ophyd.utils import ReadOnlyError, LimitError
+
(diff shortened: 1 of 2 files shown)


ZLLentz:
I plan to switch to the `SynSignal`

tacaswell:
Can you move this to ohpyd?

ZLLentz:
@tacaswell I'm happy to clean this up and move it to ophyd, but it's not obvious to me where in the module it belongs. `ophyd.sim`? `ophyd.device`? `ophyd.utils`?

ZLLentz:
I'll slip it into `ophyd.sim`

ZLLentz:
Closing because this made it into ophyd

PR #230, opened 2018-05-18
Beam stats additions

n-wbrown:
This PR adds new static components to BeamStats object.

## Description
The BeamStats objects now has two additional PVs. `accel_ev` watches the accelerator-reported ev levels and the other, `sxr_gmd`, watches the SXR GMD. The accelerator-reported ev levels, `accel_ev`, update much more quickly than the `ev` component that already exists in the object.

## Motivation and Context
These additions provide the necessary PVs for basic tests of the [auto_monochromator](https://github.com/pcdshub/auto_monochromator). The module uses a tiny event builder and so requires high time-resolution information in order to better generate complete event-built sets with which to generate histograms.

## How Has This Been Tested?
The object has been instantiated in an iPython session and the new components are able to get values and subscribe callbacks.

An in-line comment has been added to differentiate the two pv's reporting accelerator packet energy.

### DIFF
--- a/pcdsdevices/beam_stats.py
+++ b/pcdsdevices/beam_stats.py
@@ -6,13 +6,14 @@ from .signal import AvgSignal
 
 class BeamStats(Device):
     mj = Cpt(EpicsSignalRO, 'GDET:FEE1:241:ENRC')
-    ev = Cpt(EpicsSignalRO, 'SIOC:SYS0:ML00:AO541')
+    ev = Cpt(EpicsSignalRO, 'BLD:SYS0:500:PHOTONENERGY')
     rate = Cpt(EpicsSignalRO, 'EVNT:SYS0:1:LCLSBEAMRATE')
     owner = Cpt(EpicsSignalRO, 'ECS:SYS0:0:BEAM_OWNER_ID')
-    # This one is much faster than ev
-    accel_ev = Cpt(EpicsSignalRO, 'BLD:SYS0:500:PHOTONENERGY')
-    sxr_gmd = Cpt(EpicsSignalRO, 'SXR:GMD:BLD:milliJoulesPerPulse')
-
+    
+    #########
+    xpp_ipm2 = Cpt(EpicsSignalRO, 'XPP:SB2:IPM:01:SUM')
+    xpp_ipm3 = Cpt(EpicsSignalRO, 'XPP:SB3:IPM:01:SUM')
+    #########
     mj_avg = Cpt(AvgSignal, 'mj', averages=120)
     mj_buffersize = Cpt(AttributeSignal, 'mj_avg.averages')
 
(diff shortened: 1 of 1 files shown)


teddyrendahl:
👍  to accelerator eV but I would be curious to see how that value is actually calculated and why it is fundamentally different than the `eV` signal itself. Any ideas?

I don't think the SXR specific signal belongs here. The `BeamStats` object is meant to be general and hutch specific PVs should remain separate.  You can either instantiate this as need as  an`EpicsSignalRO` or something like the `GMD` will probably deserve it's own `Device` class someday.

pcdsdevices/beam_stats.py
```diff
@@ -9,6 +9,9 @@ class BeamStats(Device):
     ev = Cpt(EpicsSignalRO, 'SIOC:SYS0:ML00:AO541')
     rate = Cpt(EpicsSignalRO, 'EVNT:SYS0:1:LCLSBEAMRATE')
     owner = Cpt(EpicsSignalRO, 'ECS:SYS0:0:BEAM_OWNER_ID')
+    # This one is much faster than ev
+    accel_ev = Cpt(EpicsSignalRO, 'BLD:SYS0:500:PHOTONENERGY')
```
ZLLentz: This should have a more descriptive name, since both `ev` PVs come from the accelerator. Perhaps something like `ev_fast` or `fast_ev` or `ev_120`, etc.

I was waffling on whether or not we should have both PVs, it might make sense to just replace the other `ev` measurement with this faster one. The old one is like 2/3Hz, and this one is 120Hz.

@teddyrendahl any opinion about this?

pcdsdevices/beam_stats.py
```diff
@@ -9,6 +9,9 @@ class BeamStats(Device):
...
+    sxr_gmd = Cpt(EpicsSignalRO, 'SXR:GMD:BLD:milliJoulesPerPulse')
```
ZLLentz: This is good to have. In the future we might want a full `SxrGmd` Device with all of the sxr statistics. I'll leave it up to you if you want to start that now or wait to see if it's needed.

I'm bringing this up because `sxr_gmd` seems like the name of the full device, rather than just the mJ measurement.

pcdsdevices/beam_stats.py
teddyrendahl: Are they calculated differently? Is there any reason to compare the two? I thought `SIOC` was just an estimate maybe the `BLD` is more of a systematic measurement. The fact that it is fact makes it sound to me like it is being triggered by something?

n-wbrown:
- moved  the SXR GMD into its own object
 - deleted the slow `ev` measurement from BeamStats
 - The faster measurement has been renamed from `accel_ev` to simply `ev`

teddyrendahl:
I would just remove the `SXRGmD` for now unless you plan to expand it soon. One signal just isn't enough to really require a whole device.

ZLLentz:
I agree that it looks silly to have a one-signal device, but I disagree to removing it. There are like 10+ other PVs associated with the SXR GMD that could be brought in when needed.

teddyrendahl:
Alright. Fine with it as a placeholder I suppose.

pcdsdevices/beam_stats.py
```diff
@@ -22,3 +22,14 @@ def __init__(self, prefix='', name='beam_stats', **kwargs):
     @property
     def hints(self):
         return {'fields': [self.mj.name]}
+
+
+class SxrGmd(Device):
+    mj = Cpt(EpicsSignalRO, 'SXR:GMD:BLD:milliJoulesPerPulse')
+
+    def __init__(self, prefix='', name='SxrGmd', **kwargs):
```
ZLLentz: I recommend following the `lowercase_underscore` name convention here instead of `CapitalizedCamelCase` because `ophyd` extends the names with underscores. So right now your inner signal's name is `SxrGmd_mj` which is a mix of the styles, rather than `sxr_gmd_mj`. This of course only matters if you're doing `bluesky` scans.

PR #233, opened 2018-05-30
DOC: Add release history page

ZLLentz:
I went as far back as the release notes were actually good. The idea is that all future PRs add a summary of their changes to this release notes page, so when we make a new release we just have to copy/paste a text blob and maybe change formatting.

I'm going to go through and do this on a bunch of our repos.

### DIFF
--- a/docs/source/releases.rst
+++ b/docs/source/releases.rst
@@ -1,0 +1,64 @@
+Release History
+###############
+
+Next Release
+============
+
+Features
+--------
+- Added `AvgSignal` class. This can be used when you want to
+  run a callback on a rolling average of a ``Signal``. (#227)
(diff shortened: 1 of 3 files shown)


docs/source/releases.rst
```diff
@@ -0,0 +1,64 @@
+Release History
+###############
+
+Next Release
+============
+
+Features
+--------
+- Added `AvgSignal` class. This can be used when you want to
+  run a callback on a rolling average of a ``Signal``. (#227)
```
teddyrendahl: Do these references to PRs play nice in Sphinx? Do we want these still?

ZLLentz: they end up as plain text and then become links when copied to the releases page

PR #236, opened 2018-06-02
PCDSMotorBase .SPG Implementation

rajanplumley:
Implemented motor SPG field to PCDSMotorBase
## Description
Calling <motor>.stop(), <motor>.pause(), and <motor>.go() will put to the EPICS .SPG field and stop pause or resume motion.  <motor>.check_value also considers the .SPG state 

## Motivation and Context
The .SPG field is used by operators all the time from the expert screen.  Having it in Python will help abort moves or contrain motors easily. 

## How Has This Been Tested?
Tested in pcdsdevices test suite (run_tests.py) with a test that checks .SPG puts in test_epics_motor.py

I think Sphinx does this, right?

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -56,6 +56,8 @@ class PCDSMotorBase(FltMvInterface, EpicsMotor):
     direction_of_travel = Cpt(Signal)
     # This attribute will show if the motor is disabled or not
     disabled = Cpt(EpicsSignal, ".DISP")
+    # This attribute changes if the motor is stopped and unable to move 'Stop', paused and ready to resume on Go 'Paused', and able to move 'Go' or resume a move.
+    motor_spg = Cpt(EpicsSignal, ".SPG")
 
     @property
     def low_limit(self):
@@ -152,6 +154,50 @@ class PCDSMotorBase(FltMvInterface, EpicsMotor):
         """
         return self.disabled.set(value=1)
 
+    def stop(self):
+        """
+        Stops the motor.
+
+        Returns
+        -------
+        status: Status Object
+            Status object of the set
+        """
+        return self.motor_spg.put(value='Stop')
+
+    def pause(self):
+        """
+        Pauses a move.
+
+        Returns
+        -------
+        status: Status Object
+            Status object of the set
+        """
+        return self.motor_spg.put(value='Pause')
+
+    def resume(self):
+        """
+        Sets motor ready to move or resumes a paused move (same as <motor>.go())
+
+        Returns
+        -------
+        status: Status Object
+            Status object of the set
+        """
+        return self.motor_spg.put(value='Go')
+
+    def go(self):
+        """
+        Sets motor ready to move or resumes a paused move.
+
+        Returns
+        -------
+        status: Status Object
+            Status object of the set
+        """
+        return self.motor_spg.put(value='Go')
+
     def check_value(self, value):
         """
         Check if the motor is disabled
@@ -173,6 +219,14 @@ class PCDSMotorBase(FltMvInterface, EpicsMotor):
         if self.disabled.value == 1:
             raise Exception("Motor is not enabled. Motion requests "
                             "ignored")
+
+        if self.motor_spg.value == 0:
+            raise Exception("Motor is stopped.  Motion requests "
+                            "ignored until motor is set to 'Go')")
+
+        if self.motor_spg.value == 1:
+            raise Exception("Motor is paused.  Motion requests "
+                            "ignored until motor is set to 'Go')")
 
         # Find the soft limit values from EPICS records and check that this
         # command will be accepted by the motor
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -218,8 +218,8 @@ class PCDSMotorBase(FltMvInterface, EpicsMotor):
                             "ignored until motor is set to 'Go'")
 
         if self.motor_spg.value in [1, 'Pause']:
-            raise Exception("Motor is paused.  If a move is set, motion
-            will resume when motor is set to 'Go'")
+            raise Exception("Motor is paused.  If a move is set, motion "
+                            "will resume when motor is set to 'Go'")
 
         # Find the soft limit values from EPICS records and check that this
         # command will be accepted by the motor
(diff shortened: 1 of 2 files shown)


pcdsdevices/epics_motor.py
```diff
@@ -56,6 +56,8 @@ class PCDSMotorBase(FltMvInterface, EpicsMotor):
     direction_of_travel = Cpt(Signal)
     # This attribute will show if the motor is disabled or not
     disabled = Cpt(EpicsSignal, ".DISP")
+    # This attribute changes if the motor is stopped and unable to move 'Stop', paused and ready to resume on Go 'Paused', and able to move 'Go' or resume a move.
```
ZLLentz: `travis` is upset because this line is too long (79 characters is the max). This isn't a functional thing, it's just a python coding style thing.

pcdsdevices/epics_motor.py
```diff
@@ -152,6 +154,50 @@ def disable(self):
...
+        return self.motor_spg.put(value='Stop')
+
+    def pause(self):
+        """
+        Pauses a move.
+
+        Returns
+        -------
+        status: Status Object
+            Status object of the set
```
ZLLentz: This doc string (and the others) are incorrect. `put` does not return a `Status` object, it returns `None`.

tests/test_epics_motor.py
```diff
@@ -103,3 +104,25 @@ def plan():
         yield from close_run()
 
     RE(plan())
+
+@using_fake_epics_pv
+def test_resume_pause_stop():
+    m = fake_motor()
+    m.stop()
+    assert m.motor_spg.get(as_string = True) == 'Stop'
```
ZLLentz: I feel like the most pedantic person right now, but the convention is to not put the spaces around the `=` sign when they are function arguments, e.g.
`m.motor_spg.get(as_string=True)`

pcdsdevices/epics_motor.py
```diff
@@ -152,6 +154,50 @@ def disable(self):
...
+        """
+        return self.motor_spg.put(value='Pause')
+
+    def resume(self):
+        """
+        Sets motor ready to move or resumes a paused move (same as <motor>.go())
```
ZLLentz: `travis` is upset about this line too because it's longer than 79 characters

tests/test_epics_motor.py
```diff
@@ -103,3 +104,25 @@ def plan():
...
+    m.pause()
+    assert m.motor_spg.get(as_string = True) == 'Pause'
+    with pytest.raises(Exception):
+        m.move(10, wait=False)
+    m.go()
+    assert m.motor_spg.get(as_string = True) == 'Go'
+    m.check_value(10)
+    m.resume()
+    assert m.motor_spg.get(as_string = True) == 'Go'
+    m.check_value(10)
```
teddyrendahl: Nice work on the tests.

pcdsdevices/epics_motor.py
```diff
@@ -174,6 +220,14 @@ def check_value(self, value):
...
                             "ignored")
 
+        if self.motor_spg.value in [0, 'Stop']:
+            raise Exception("Motor is stopped.  Motion requests "
+                            "ignored until motor is set to 'Go'")
+
+        if self.motor_spg.value in [1, 'Pause']:
+            raise Exception("Motor is paused.  Motion requests "
+                            "ignored until motor is set to 'Go'")
+
```
teddyrendahl: This is properly implemented just thinking out loud. It might be getting close to time to have a custom `Exception` for this kind of "Motor Configuration" Error. We now check quite a few things at the start of a move, and more descriptive errors are helpful for operators.

pcdsdevices/epics_motor.py
```diff
@@ -152,6 +154,50 @@ def disable(self):
...
+    def go(self):
+        """
+        Sets motor ready to move or resumes a paused move.
+
+        Returns
+        -------
+        status: Status Object
+            Status object of the set
+        """
+        return self.motor_spg.put(value='Go')
```
teddyrendahl: I am uncomfortable with a `method` called `go` that doesn't neccesarily `move` something. I'd leave it at just `resume`. Just a thought.

teddyrendahl:
Oh also one more generic thing. We like to include [these](https://pcdshub.github.io/development.html#commit-guidelines) tags at the start of our commits. This makes it easy for us to sift through the `git` history. Don't worry about retroactively adding these, but think about including them in the future.

pcdsdevices/epics_motor.py
ZLLentz: I had the same thought

pcdsdevices/epics_motor.py
rajanplumley: ```go```  was added in because it matches what staff are used to on the motor expert screen.  It really does make a motor *go* provided a move has been set and paused.  Maybe we can keep it around and see if people don't like it?

About setting the motor SPG during Bluesky staging, I'm not sure if we want to do this if the motor is stopped.  Staff often stop motors as a way of "soft-disabling" them because they haven't been completely set up yet or they don't want them moved for whatever reason.  I think its better to display a transparent error message about a motor being stopped and let them set it themselves than to assume its OK to move and go ahead.

rajanplumley: However I don't see anything wrong with calling ```go``` during a stage if the motor is paused.

ZLLentz:
Added branch protection to prevent people from accidentally pushing to master in the future

pcdsdevices/epics_motor.py
ZLLentz: I think this docstring is the only complaint I have left

teddyrendahl:
@rajanplumley so close! Just edit that docstring and this will head in

ZLLentz:
Don't worry about the travis failure here, it's related to my "fix the tests" PR

teddyrendahl:
Okay I lied. Two more things. I apologize for being so pedantic, I hope this doesn't discourage you in the future 😢 

1. The docstrings should have a one line description, then a paragraph description. This helps Sphinx. http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html. 

```python
def stop(self):
    """ 
    Stops the motor. 

    The motor  must be set back to 'go' via <motor>.go() in order to move again.
```
2. We should repeat ourselves. `resume` should just call `go`. That way if we ever do anything to `go` it automatically will be updated in the other function.

```python
def resume(self):
       """
       ....
       """
       return self.go()
```

PR #241, opened 2018-06-28
added mv_ginput, a way for the user to click on a plot to change the motor's position.

MattHooverSLAC:
## Description
mv_ginput gives the user a plot based on the motor's limits that can be clicked on to change the motor's x value.  If there are no limits provided, it goes from 0 to 100 x, and if there is already a plot created, it will detect and is it instead.  Also imported the necessary packages.
## Motivation and Context
This was a feature in the old motor file, and has been ported over to the new system.

### DIFF
--- a/pcdsdevices/mv_interface.py
+++ b/pcdsdevices/mv_interface.py
@@ -6,6 +6,8 @@ import fcntl
 import logging
 import numbers
 import signal
+import pylab
+from pylab import plot, ginput, show, axis
 from contextlib import contextmanager
 from pathlib import Path
 from types import SimpleNamespace, MethodType
@@ -145,6 +147,24 @@ class FltMvInterface(MvInterface):
         """
         self.umv(delta + self.wm(), timeout=timeout)
 
+    def mv_ginput(self, timeout=None):
+        """
+        Moves to a location the user clicks on.
+        """
+        print("Select new motor x-position in current plot by mouseclick")
+        if not pylab.get_fignums():
+            upperLimit = 0
+            lowerLimit = self.limits[0]
+            if self.limits[1] == 0:
+                 upperLimit = 100
+            else:
+                 uppderLimit = self.limits[1]
+            limitPlotter = []
+            for x in range(lowerLimit,upperLimit):
+                limitPlotter.append(x)
+            pylab.plot(limitPlotter)
+        pos = ginput(1)[0][0]
+        self.move(pos, timeout=timeout, wait=True)
 
 def setup_preset_paths(**paths):
     """


pcdsdevices/mv_interface.py
```diff
@@ -145,6 +147,24 @@ def umvr(self, delta, timeout=None):
...
+        Moves to a location the user clicks on.
+        """
+        print("Select new motor x-position in current plot by mouseclick")
+        if not pylab.get_fignums():
+            upperLimit = 0
+            lowerLimit = self.limits[0]
+            if self.limits[1] == 0:
+                 upperLimit = 100
+            else:
+                 uppderLimit = self.limits[1]
```
ZLLentz: typo right here `uppderLimit` instead of `upperLimit`

pcdsdevices/mv_interface.py
```diff
@@ -145,6 +147,24 @@ def umvr(self, delta, timeout=None):
         """
         self.umv(delta + self.wm(), timeout=timeout)
 
+    def mv_ginput(self, timeout=None):
```
ZLLentz: I feel like there should be a `wait` keyword argument here to compliment the `move` call

pcdsdevices/mv_interface.py
```diff
@@ -145,6 +147,24 @@ def umvr(self, delta, timeout=None):
...
+            if self.limits[1] == 0:
+                 upperLimit = 100
+            else:
+                 uppderLimit = self.limits[1]
+            limitPlotter = []
+            for x in range(lowerLimit,upperLimit):
+                limitPlotter.append(x)
+            pylab.plot(limitPlotter)
+        pos = ginput(1)[0][0]
+        self.move(pos, timeout=timeout, wait=True)
```
ZLLentz: I wonder if we can add this method to the tests using the magic clicky stuff in `pytest-qt` @teddyrendahl is this something we can do? Or is simulating a click at a specific coordinate outside the range of capabilities?

ZLLentz:
I wonder if we can add this method to the tests using the magic clicky stuff in `pytest-qt` @teddyrendahl is this something we can do? Or is simulating a click at a specific coordinate outside the range of capabilities?

pcdsdevices/mv_interface.py
```diff
@@ -6,6 +6,8 @@
 import logging
 import numbers
 import signal
+import pylab
+from pylab import plot, ginput, show, axis
```
ZLLentz: You should pick one of these schemas and stick to it to avoid a messy import block. We don't need to import `plot` from `pylab` and then later call `pylab.plot` instead of just `plot`. One or the other would make the code more readable for all the `pylab` functions.

ZLLentz:
Here is the `flake8` style error output from the automated tests:
```
pcdsdevices/mv_interface.py:10:1: F401 'pylab.plot' imported but unused
pcdsdevices/mv_interface.py:10:1: F401 'pylab.show' imported but unused
pcdsdevices/mv_interface.py:10:1: F401 'pylab.axis' imported but unused
pcdsdevices/mv_interface.py:159:18: E111 indentation is not a multiple of four
pcdsdevices/mv_interface.py:161:18: E111 indentation is not a multiple of four
pcdsdevices/mv_interface.py:161:18: F841 local variable 'uppderLimit' is assigned to but never used
pcdsdevices/mv_interface.py:163:38: E231 missing whitespace after ','
pcdsdevices/mv_interface.py:169:1: E302 expected 2 blank lines, found 1
```

teddyrendahl:
I believe the `mouseClick` button takes a positional argument. 

https://pytest-qt.readthedocs.io/en/latest/tutorial.html

ZLLentz:
Also, I hate to be so picky, but in `Python` we usually have a convention where variable names are `separated_by_underscores` instead of `camelCase`

hhslepicka:
You can even just create the Event without pytest-qt and invoke the method
sending your custom event with the coordinates. The pytest-qt click and
mouse was not working nice for us so we did the click tests ourselves at
pydm.

pcdsdevices/mv_interface.py
MattHooverSLAC: fixed

ZLLentz:
closing and moving to #258

Issue #183, opened 2018-03-20
wm_update


ZLLentz:
Port the old motor `wm_update` feature, functions as a `camonitor`. I've seen the scientists use this from time to time, so I know they will ask for it.

ZLLentz:
I'd like this to hook into the progress bars

ZLLentz:
closing by #242, Yesenia implemented the `camonitor` method. This is done via a loop because not every positioner has a signal dictating their position. We may consider in the future putting something like this on all signals, and we may consider doing this via subscriptions for classes that get their position from a signal.

PR #242, opened 2018-06-28
Camonitor Implementation 

ygarnica:
## Description
Implemented the camonitor such that it displays the current position of the motor until Ctrl+C is used.

## Motivation and Context
- Transfering old wm_update into python 3 code.
- addresses issue:  https://github.com/pcdshub/pcdsdevices/issues/183

## How Has This Been Tested?
- used a disconnected moving motor to test current moving position 

- Function was implemented in IMS(PCDSMotorBase)

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -375,14 +375,6 @@ class IMS(PCDSMotorBase):
             status_wait(st, timeout=timeout)
         return st
 
-    def camonitor(self):
-        try:
-            while True:
-                time.sleep(0.1)
-                print("\r {0:4f}".format(self.user_readback.value), end=" ")
-        except KeyboardInterrupt:
-            pass
-
 
 class Newport(PCDSMotorBase):
     """
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -375,6 +375,25 @@ class IMS(PCDSMotorBase):
         if wait:
             status_wait(st, timeout=timeout)
         return st
+    def cb(self,moving,timeout):
+    """Display motor position second by second"""
+        import time
+        if timeout<=0:
+           timeout=float("inf")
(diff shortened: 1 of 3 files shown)


pcdsdevices/epics_motor.py
```diff
@@ -376,6 +376,13 @@ def flag_is_cleared(value=None, **kwargs):
             status_wait(st, timeout=timeout)
         return st
 
+    def camonitor(self):
+        try:
+            while True:
+                print("\r {0:4f}".format(self.user_readback.value), end=" ")
```
ZLLentz: This is why the cursor is flashing across the screen so fast. Python is racing through this `while` loop as fast as it can and repeatedly printing to the screen at max speed. We should either slow it down with a `time.sleep` call or implement this as a subscription callback e.g.
```
def cb(value=None, **kwargs):
    print("\r {0:4f}".format(value), end=" ")
cbid = self.user_readback.subscribe(cb)
```
And then later at the end, probably in a `finally` block:
`self.user_readback.unsubscribe(cbid)`

pcdsdevices/epics_motor.py
```diff
@@ -376,6 +376,13 @@ def flag_is_cleared(value=None, **kwargs):
             status_wait(st, timeout=timeout)
         return st
 
+    def camonitor(self):
```
ZLLentz: I wonder if this should become a standard method on every `Signal` object...

ZLLentz: @teddyrendahl are you opposed to awful monkeypatching? E.g. if we defined a `camonitor` method standalone and did:
```
from ophyd.signal import Signal

def camonitor(self):
    <code>

Signal.camonitor = camonitor
```
Then we could use this in literally every class as desired. I'm not sure what Dan and Tom would think about putting this as part of `ophyd`, but it seems like a reasonable interactive debugging tool.

PR #243, opened 2018-07-02
ENH: Add utils with keypress getter

ZLLentz:
## Description

Port of old hutch-python `keypress` module, but only the useful functions

## Motivation and Context

We need this for `tweak`. This should be ported to a utilities module at some point, but I didn't feel like making a module with one file.

## How Has This Been Tested?

I have not written tests, but I plan to do this later at the same time we do the tweak tests.

This has really good comments, but I'd prefer to leave it out of the sphinx docs until it's in the correct module.

### DIFF
--- a/pcdsdevices/utils.py
+++ b/pcdsdevices/utils.py
@@ -1,22 +1,22 @@
-import sys
 import select
-import tty
+import sys
 import termios
 import time
+import tty
 
-arrow_up   = '\x1b[A'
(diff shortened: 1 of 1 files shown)


ZLLentz:
@teddyrendahl I think this is the point of the special cbreak mode. I did a lot of reading up on this before committing to using the old code. I was also skeptical, but it seems to work.
https://en.wikipedia.org/wiki/Terminal_mode

PR #244, opened 2018-07-03
ENH: Beckhoff Axis Class (and some refactor)

ZLLentz:
## Description

- Add the `Beckhoff` class for controlling the motor record from ESS.
- Add specific error classes for things like SPG and DISP
- Pare down unnecessarily wordy docstrings (when your function just sets the property, you don't have to copy the whole thing again...)
- Refactor `epics_motor` module to split the things we voluntarily add to the base `EpicsMotor` class from the things we are forced to add because our "motor-record-like" IOCs aren't quite compliant.
- Refactor the tests to be more thorough

## Motivation and Context

- We needed a `Beckhoff` class that could at least reset the errors. We may need to add more later, but it's unclear.
- The refactor was the cleanest way to include the `Beckhoff` class.
- I felt like the tests were weak here, we didn't even instantiate most of the classes. I parametrized the tests so that we can run each set of tests on all subclasses that should have the functionality. I also added the missing `DISP` test.

## How Has This Been Tested?

We're all green captain

Added to the `device_types` page

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -19,7 +19,7 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
     """
     The standard EpicsMotor class, but with our interface attached.
 
-    This includes some pcds preferences, but not any IOC differences.
+    This includes some PCDS preferences, but not any IOC differences.
 
     Notes
     -----
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -99,14 +99,18 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
 
     def check_value(self, value):
         """
-        Check if the motor is disabled
-        Check if the value is within the soft limits of the motor.
+        Raise an exception if the motor cannot move to value.
+
+        The full list of checks done in this method:
+
+            - Check if the value is within the soft limits of the motor.
+            - Check if the motor is disabled (``.DISP`` field)
 
         Raises
         ------
-        MotorDisabledError
+        `MotorDisabledError`
             If the motor is passed any motion command when disabled
-        LimitError(ValueError)
+        ``LimitError(ValueError)``
             When the provided value is outside the range of the low
             and high limits
         """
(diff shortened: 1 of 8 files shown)


pcdsdevices/epics_motor.py
```diff
@@ -15,38 +15,27 @@
...
     """
-    EpicsMotor for PCDS
+    The standard EpicsMotor class, but with our interface attached.
 
-    This incapsulates all motor record implementations standard for PCDS,
-    including but not exclusive to Pico, Piezo, IMS and Newport motors. While
-    these types of motors may have additional records and configuration
-    requirements, this class is meant to handle the shared records between
-    them.
+    This includes some pcds preferences, but not any IOC differences.
```
teddyrendahl: Capital PCDS?

pcdsdevices/epics_motor.py
```diff
@@ -55,109 +44,123 @@ class PCDSMotorBase(FltMvInterface, EpicsMotor):
...
-        status: Status Object
-            Status object of the set
+        When disabled, all EPICS puts to the base record will be dropped.
+        """
+        return self.disabled.put(value=1)
+
+    def check_value(self, value):
+        """
+        Check if the motor is disabled
+        Check if the value is within the soft limits of the motor.
```
teddyrendahl: Fix this docstring?

pcdsdevices/epics_motor.py
```diff
@@ -200,39 +203,28 @@ def check_value(self, value):
...
         Raises
         ------
-        Exception
+        MotorDisabledError(MoveForbiddenError)
             If the motor is passed any motion command when disabled
-        ValueError
+        LimitError(ValueError)
             When the provided value is outside the range of the low
             and high limits
+        MotorSPGError(MoveForbiddenError)
```
teddyrendahl: Not sure we need this much resolution. I'm kind of fine with a `MotorDisabledError` for both cases

pcdsdevices/epics_motor.py
```diff
@@ -412,3 +404,66 @@ class PMC100(PCDSMotorBase):
 
     def home(self, *args, **kwargs):
         raise NotImplementedError("PMC100 motors have no homing procedure")
+
+
+class Beckhoff(EpicsMotorInterface):
```
teddyrendahl: Not in love with just `Beckhoff` as the name but it does follow precedent...

ZLLentz: The part of precedent that is broken here is that we use the `Beckhoff` rails for vacuum too, so there is ambiguity. We have no such issues with `IMS` and `Newport`.

Some suggested renames:
`BeckhoffMotor`
`BeckhoffAxis`

pcdsdevices/epics_motor.py
ZLLentz: I opted for high resolution so that other code could decide how to enable the motor (call `go()` or `enable()`), but perhaps this is unnecessary because you can just call both with no ill effects.

As an aside, having an `enable` that doesn't set `go` makes me slightly uncomfortable, but I'm not sure what to do better here.

pcdsdevices/epics_motor.py
ZLLentz: I think you mean that it doesn't follow the one-line-summary convention, since the actual text here is accurate. I'll fix it.

pcdsdevices/epics_motor.py
ZLLentz: .

pcdsdevices/epics_motor.py
ZLLentz: What I want to do here is rename all of them to `<brand>Motor` and leave `<brand>` as an alias.

ZLLentz: Though this may be confusing and it will lead to verbose repetition in the autodoc

pcdsdevices/epics_motor.py
```diff
@@ -99,14 +99,18 @@ def disable(self):
...
+
+        The full list of checks done in this method:
+
+            - Check if the value is within the soft limits of the motor.
+            - Check if the motor is disabled (``.DISP`` field)
 
         Raises
         ------
-        MotorDisabledError
+        `MotorDisabledError`
```
teddyrendahl: Missed a set of `

pcdsdevices/epics_motor.py
teddyrendahl: I like `BeckhoffAxis`.

Also, so-so on the full rename. Seems unnecessarily verbose for most of them.

pcdsdevices/epics_motor.py
ZLLentz: Using one set instead of two creates an auto link to the definition because `default_role = any` is set in the docs conf

ZLLentz:
@teddyrendahl I think I addressed all of your comments

PR #252, opened 2018-07-12
MAINT: Kind

### LINKED ISSUE #237: MAINT: Read, Configuration and Hints on All Devices
## Expected Behavior

All devices should have sensible read / configuration attrs and hints. These should be specified with the new `Kind` functionality.

teddyrendahl:
## Description

Added `Kind` specification to every Device in the library. 

I tried to keep a consistent methodology (mostly using `EpicsMotor` as a guideline), but happy to debate any specific signals that are in the wrong category. I'm taking a guess that most people won't care as almost all of them were in the wrong category before this PR 😄 

## Motivation and Context

Closes #237

### DIFF
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -31,7 +31,7 @@ class Filter(InOutPositioner):
     """
     state = Cmp(EpicsSignal, ':STATE', write_pv=':GO', kind='hinted')
     thickness = Cmp(EpicsSignal, ':THICK', kind='config')
-    material = Cmp(EpicsSignal, ':MATERIAL', kind='omitted')
+    material = Cmp(EpicsSignal, ':MATERIAL', kind='config')
     stuck = Cmp(EpicsSignal, ':IS_STUCK', kind='omitted')
 
 
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -29,10 +29,10 @@ class Filter(InOutPositioner):
     a ``Component`` in a subclass of `AttBase`. You can instantiate these
     classes via the `Attenuator` factory function.
     """
-    state = Cmp(EpicsSignal, ':STATE', write_pv=':GO')
-    thickness = Cmp(EpicsSignal, ':THICK')
-    material = Cmp(EpicsSignal, ':MATERIAL')
-    stuck = Cmp(EpicsSignal, ':IS_STUCK')
+    state = Cmp(EpicsSignal, ':STATE', write_pv=':GO', kind='hinted')
+    thickness = Cmp(EpicsSignal, ':THICK', kind='config')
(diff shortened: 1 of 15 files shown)


pcdsdevices/attenuator.py
```diff
@@ -29,10 +29,10 @@ class Filter(InOutPositioner):
     a ``Component`` in a subclass of `AttBase`. You can instantiate these
     classes via the `Attenuator` factory function.
     """
-    state = Cmp(EpicsSignal, ':STATE', write_pv=':GO')
-    thickness = Cmp(EpicsSignal, ':THICK')
-    material = Cmp(EpicsSignal, ':MATERIAL')
-    stuck = Cmp(EpicsSignal, ':IS_STUCK')
+    state = Cmp(EpicsSignal, ':STATE', write_pv=':GO', kind='hinted')
+    thickness = Cmp(EpicsSignal, ':THICK', kind='config')
+    material = Cmp(EpicsSignal, ':MATERIAL', kind='omitted')
```
ZLLentz: Both thickness and material should be the same interest level

pcdsdevices/movablestand.py
```diff
@@ -16,8 +16,8 @@ class MovableStand(InOutPVStatePositioner):
 
     name: ``str``, required keyword
     """
-    in_limit = Cmp(EpicsSignalRO, ':IN_DI')
-    out_limit = Cmp(EpicsSignalRO, ':OUT_DO')
+    in_limit = Cmp(EpicsSignalRO, ':IN_DI', kind='normal')
+    out_limit = Cmp(EpicsSignalRO, ':OUT_DO', kind='normal')
```
ZLLentz: Are we including this every time just so that we know someone has thought about it? This is the default value after all.

pcdsdevices/slits.py
```diff
@@ -114,23 +114,19 @@ class Slits(Device, MvInterface):
...
-    ycenter = C(SlitPositioner, '', slit_type="YCENTER")
-    ywidth = C(SlitPositioner, '', slit_type="YWIDTH")
-    blocked = C(EpicsSignalRO, ":BLOCKED")
-    open_cmd = C(EpicsSignal, ":OPEN")
-    close_cmd = C(EpicsSignal, ":CLOSE")
-    block_cmd = C(EpicsSignal, ":BLOCK")
+    xcenter = C(SlitPositioner, '', slit_type="XCENTER", kind='hinted')
+    xwidth = C(SlitPositioner, '', slit_type="XWIDTH", kind='config')
+    ycenter = C(SlitPositioner, '', slit_type="YCENTER", kind='hinted')
+    ywidth = C(SlitPositioner, '', slit_type="YWIDTH", kind='config')
```
ZLLentz: I feel like we want the widths to be `normal`, because you want to know if they have changed in the data

PR #254, opened 2018-07-13
API: Sim rework

### LINKED ISSUE #110: Use ophyd using_fake_epics_pv
## Expected Behavior

Deprecate our own version of `using_fake_epics_pv`. We should just use the ophyd version instead. This requires closing https://github.com/NSLS-II/ophyd/issues/477

### LINKED ISSUE #239: tests folder is installed as a full module
Uh... weird thing here. Since there is an `__init__.py` file in the tests folder, the `find_packages()` method used in `setup.py` thinks it's a standalone module, and it gets installed as the `tests` module!

Perhaps it's time to move this into the subfolder and include it in the full distribution.

ZLLentz:
## Description

- Remove `sim` module
- Make new `sim` module, populate with the simulated motors from `test_mv_interface` that were good for testing
- Remove `tests/__init__.py`

## Motivation and Context

- We aren't maintaining it, the bits used for pswalker, etc have already be copied over
- Can use new sim module for purely device-based sim if needed (not simulating data for plans, which should be the responsibility of a different module).
- closes #239 
- closes #110 

## How Has This Been Tested?

Still passes

The old module never was, perhaps we should document now... but it's after 7pm


pcdsdevices/sim/motor.py
```diff
@@ -0,0 +1,69 @@
...
+
+from ophyd.device import Device, Component as Cpt
+from ophyd.positioner import SoftPositioner
+from ophyd.signal import AttributeSignal
+from ophyd.sim import SynAxis
+
+from pcdsdevices.mv_interface import FltMvInterface
+
+
+class SynAxisInterface(FltMvInterface, SynAxis):
```
teddyrendahl: Too confusing just to refer to as `SynAxis`? `SynMotor`? Find this name kind of a drag ...

pcdsdevices/sim/motor.py
```diff
@@ -0,0 +1,69 @@
...
+    SynAxis that has our special interface on it.
+
+    This can be used to test when you need the readback_func feature or if you
+    just want your test motor to finish immediately.
+    """
+    def move(self, position, *args, **kwargs):
+        return super().set(position)
+
+
+class SimMotor(FltMvInterface, SoftPositioner, Device):
```
teddyrendahl: If we were doing everything from the start I'd say `Sim` is better, but might vote for keeping the `Syn-` prefix.

pcdsdevices/sim/motor.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    Simulated motor with our interface attached.
+
+    Unlike the SynAxis built into ophyd, this takes some time to reach the
+    destination. Use this when you need some sort of delay.
+    """
+    user_readback = Cpt(AttributeSignal, 'position')
```
teddyrendahl: You can't keep this the the same and just have this update `self.sim_state['position']`?

pcdsdevices/sim/motor.py
```diff
@@ -0,0 +1,69 @@
...
+
+class SimTwoAxis(Device):
+    """
+    Test assembly with two motors
+    """
+    x = Cpt(SimMotor)
+    y = Cpt(SimMotor)
+
+    # def tweak(self):
+    #     return tweak_2d(self.x, self.y)
```
teddyrendahl: Why is the comment preserved?

pcdsdevices/sim/pv.py
```diff
@@ -1,290 +0,0 @@
...
-
-import numpy as np
-import ophyd.control_layer as cl
-
-logger = logging.getLogger(__name__)
-
-_FAKE_PV_LIST = []
-
-
-class FakeEpicsPV(object):
```
teddyrendahl: DEATH TO FAKE EPICS PV!!

pcdsdevices/sim/motor.py
ZLLentz: I thought having a `SynMotor` and a `SimMotor` would be annoying. I thought that copying the `SynAxis` named would be annoying.

How about `SynMotor` and `SlowMotor`?

pcdsdevices/sim/motor.py
ZLLentz: I don't understand this comment, but this class was much easier to implement as a `SoftPositioner` than a `SynAxis` so that might cause some of the discrepancies here. Rewriting as a `SynAxis` would require a full override of the `move` method iirc.

pcdsdevices/sim/pv.py
ZLLentz: :fire:

pcdsdevices/sim/motor.py
ZLLentz: To remind me to put the tweak 2d function onto the fake 2d motor once Yesenia is done with it

pcdsdevices/sim/motor.py
ZLLentz: See above about potential renames. I also don't *really* like these names.

ZLLentz: This one should not be named `Syn` though so that nobody expects it to have all the features of a `SynAxis`.

pcdsdevices/sim/motor.py
teddyrendahl: I like `SynMotor` `SlowMotor`

pcdsdevices/sim/motor.py
teddyrendahl: Ah did this review too late at night. I thought this inherited from `SynAxis`. I think this could probably still be implemented as a `SynAxis` with a clever `readback_func`:

https://github.com/NSLS-II/ophyd/blob/f4fdae506ac33b1611471eb7f5a406c3b39e1f0c/ophyd/sim.py#L285

That is a nitpick though...

pcdsdevices/sim/motor.py
teddyrendahl: 👍

pcdsdevices/sim/motor.py
ZLLentz: Can `readback_func` be hacked to allow a slow update over time?

ZLLentz:
I think I've added adequate docs. I just realized that I've been neglecting the release history pages on all the repos, so I might take some time catching it up in a different PR.

Issue #174, opened 2018-03-19
Tweak, Tweak2D


ZLLentz:
## Expected Behavior

Old Python feature where you can use the arrow keys to manipulate motors

teddyrendahl:
I actually kind of like this one 😄

ZLLentz:
This is very easy to implement for the 1D tweak, it's less clear how to decide where and how to implement 2D tweak in a clean way

teddyrendahl:
⬆️ for scientist request!

ZLLentz:
description is:
1d tweak: `left` moves negative, `right` moves positive, `up` doubles step size, `down` halves step size
2d tweak: `left` moves x negatives, `right` moves x positive, `up` moves y positive, `down` moves y negative, and `shift+dir` changes the step size on that axis

ZLLentz:
closed by #255

PR #255, opened 2018-07-13
Add Tweak_base and Tweak functions

ygarnica:
## Description
Tweak base function created so Tweak function can be used to move one or two motors. 

## Motivation and Context
Issue https://github.com/pcdshub/pcdsdevices/issues/174 
## How Has This Been Tested?
Tested using the TwoMotor class in the tests folder.
<!--- Why is this change required? What problem does it solve?

### DIFF
--- a/pcdsdevices/mv_interface.py
+++ b/pcdsdevices/mv_interface.py
@@ -73,6 +73,9 @@ class MvInterface:
             self.mv(position, timeout=timeout, wait=wait)
 
     def camonitor(self):
+        """
+        Updates current position of the motor.
+        """
         try:
             while True:
                 time.sleep(0.1)
--- a/pcdsdevices/mv_interface.py
+++ b/pcdsdevices/mv_interface.py
@@ -86,7 +86,7 @@ class MvInterface:
             self._mov_ev.clear()
             while not self._mon_ev.is_set():
                 print("\r {0:4f}".format(self.position), end=" ")
-                self._mon_ev.wait(0.1)
+                self._mov_ev.wait(0.1)
         except KeyboardInterrupt:
             pass
         finally:
--- a/pcdsdevices/mv_interface.py
+++ b/pcdsdevices/mv_interface.py
@@ -4,6 +4,8 @@ Module for defining bell-and-whistles movement features
 import time
 import fcntl
 import logging
+from threading import Thread
+from .utils import *
 import numbers
 import signal
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


pcdsdevices/mv_interface.py
```diff
@@ -4,6 +4,9 @@
 import time
 import fcntl
 import logging
+import threading
+from threading import Thread
```
ZLLentz: This should be rewritten to either be:
`import threading`
or
`from threading import Thread, Event`

Instead of doing it both ways, for consistency's sake. You may have to change how this is used elsewhere in the file.

pcdsdevices/mv_interface.py
```diff
@@ -77,11 +84,86 @@ def camonitor(self):
...
+                self._mov_ev.wait(0.1)
         except KeyboardInterrupt:
             pass
+        finally:
+            self._mov_ev.clear()
+
+    def _stop_monitor(self):
+        self._mov_ev.set()
+
+    def Tweak_base(*args):
```
ZLLentz: another convention: method names are supposed to be lowercase

ZLLentz: actually one more thing, this shouldn't be attached to the `MvInterface` class, I think it should float near the bottom of this file as a standalone function so that we can do something like:
```
from mv_interface import tweak_base
```
in another file

pcdsdevices/mv_interface.py
```diff
@@ -77,11 +84,86 @@ def camonitor(self):
...
+        """
+        Base function to tweak motors
+        """
+        up = '\x1b[A'
+        down = '\x1b[B'
+        left = '\x1b[D'
+        right = '\x1b[C'
+        shift_up = '\x1b[1;2A'
+        shift_down = '\x1b[1;2B'
```
ZLLentz: don't have to change it now, but maybe in the future try to refer to `utils.up`, `utils.down` to avoid potential copy paste errors

pcdsdevices/mv_interface.py
```diff
@@ -156,6 +238,12 @@ def umvr(self, delta, timeout=None):
         """
         self.umv(delta + self.wm(), timeout=timeout)
 
+    def Tweak(*args):
+        if len(args) > 1:
+            return args[0].Tweak_base(args[1])
+        else:
+            return args[0].Tweak_base()
```
ZLLentz: when tweak_base moves to the module level you'll have to edit this a bit

ZLLentz:
Looks good, I'm merging this in. Still on the to-do is thorough tests, but let's keep moving forward.

PR #257, opened 2018-07-14
MAINT: PVStateSignal Description

teddyrendahl:
## Description

`PVStateSignal` now gives more information about itself in the `describe` dictionary for downstream consumers. This includes the `enum_strs` and which signals it is summarizing. This was largely for compatibility with `Typhon`, but I think it could be a helpful debugging tool.

#### Example
```python
{'mfx_dg1_vgc_01_state': {'dtype': 'string',
  'enum_strs': ['Unknown', 'IN', 'OUT'],
  'shape': [],
  'source': 'SUM:mfx_dg1_vgc_01_open_limit,mfx_dg1_vgc_01_closed_limit'}}
```
## Tests
Added a test for the contents of the `describe` dictionary

### DIFF
--- a/pcdsdevices/state.py
+++ b/pcdsdevices/state.py
@@ -281,7 +281,8 @@ class PVStateSignal(AggregateSignal):
         desc = {'source': 'SUM:{}'.format(','.join(sub_sigs)),
                 'dtype': 'string',
                 'shape': [],
-                'enum_strs': [state.name for state in self.parent.states_enum]}
+                'enum_strs': tuple(state.name
+                                   for state in self.parent.states_enum)}
         return {self.name: desc}
 
     def _calc_readback(self):
--- a/pcdsdevices/state.py
+++ b/pcdsdevices/state.py
@@ -275,6 +275,15 @@ class PVStateSignal(AggregateSignal):
             self._sub_signals.append(sig)
             self._sub_map[signal_name] = sig
 
+    def describe(self):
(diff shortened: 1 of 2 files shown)


pcdsdevices/state.py
```diff
@@ -275,6 +275,15 @@ def __init__(self, *, name, **kwargs):
             self._sub_signals.append(sig)
             self._sub_map[signal_name] = sig
 
+    def describe(self):
+        # Base description information
+        sub_sigs = [sig.name for sig in self._sub_signals]
+        desc = {'source': 'SUM:{}'.format(','.join(sub_sigs)),
```
ZLLentz: Not sure why this is called `SUM` but I don't have any reasonable suggestions for it.

teddyrendahl: Good question. Meant it as short for summary.  Open to suggestions.

teddyrendahl:
I can add it if you want it, though I don't think a bad `describe` will break anything (maybe callbacks that use it). Prior iterations would have had the signal listed as a `"number"`!

ZLLentz:
`describe` will definitely break the `RE` if it raises an exception, so maybe I'm internally paranoid. I'm happy to merge this as-is.

ZLLentz:
Happy to merge this if no objections

teddyrendahl:
Realized that this should be a tuple not a list. Ready for merge

PR #265, opened 2018-07-19
FIX: Warn user about subclassing base classes

ZLLentz:
## Description

Raise a concise, helpful error when a user tries to create a base-class-only device.

## Motivation and Context

This used to give a wide range of unintelligible errors depending on which class you picked.

## How Has This Been Tested?

Added tests

Some docstrings updated to indicate that subclassing is needed


ZLLentz:
hmm, that's definitely a thing to do, but I'm not motivated today. I'll open an issue instead so we don't forget

PR #270, opened 2018-07-19
MAINT: Interlocks belong in check_value

### LINKED ISSUE #259: PPSStopper has move options
The `lightpath` looks to see if objects have `remove` and `insert` methods to see if they should have corresponding widgets. Currently the `PPSStopper` has the both of these and they just raise an Exception.

Can we remove these methods while still inheriting? Can we do something intelligent if `state` is a read-only signal?

teddyrendahl:
## Description

External applications may want to check whether a move is possible without requesting one. Because `PPSStopper` and `GateValve` had interlock logic hidden in `_do_move` this was not possible. This places that functionality within `check_value`. 

We may still want to make this process more explicit, but this is the first step.

## Motivation and Context

Related but might not fully close #259 

## How Has This Been Tested?

Existing interlock tests still pass

### DIFF
--- a/pcdsdevices/valve.py
+++ b/pcdsdevices/valve.py
@@ -95,7 +95,6 @@ class GateValve(Stopper):
     command = Cpt(EpicsSignal,   ':OPN_SW', kind='omitted')
     interlock = Cpt(EpicsSignalRO, ':OPN_OK', kind='normal')
 
-
     def check_value(self, value):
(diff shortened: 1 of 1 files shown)


teddyrendahl:
`super().check_value` will always give us back the enum I believe

ZLLentz:
Oh right! I forgot we did that.

PR #271, opened 2018-07-19
ENH: Lens stack starting point

ZLLentz:
## Description

Add a basic XYZ stack for the be lens alignment/calibration classes

## Motivation and Context

This had to be done eventually, and is a good starting point for Matt and Yesenia to implement the xy calibration and spot size calculations respectively

## How Has This Been Tested?

This breaks flake8 but I want to put it in here anyway


teddyrendahl:
Maybe add `kind`?

ZLLentz:
@teddyrendahl kind=normal and then beam size will be hinted I think

ZLLentz:
I'll go through and fill them in when I do the great calculation merge

PR #273, opened 2018-07-26
ENH: expand the sim classes

ZLLentz:
Some more testing classes, including a fake lens stack

### DIFF
--- a/pcdsdevices/sim.py
+++ b/pcdsdevices/sim.py
@@ -6,7 +6,8 @@ from ophyd.positioner import SoftPositioner
 from ophyd.signal import AttributeSignal
 from ophyd.sim import SynAxis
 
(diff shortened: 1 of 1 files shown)


ZLLentz:
Actually this appears to work 100%, so I'm retracting the WIP

pcdsdevices/sim.py
```diff
@@ -60,10 +69,19 @@ def stop(self):
...
     x = Cpt(SlowMotor)
     y = Cpt(SlowMotor)
 
-    # def tweak(self):
-    #     return tweak_2d(self.x, self.y)
+    def tweak(self):
+        return tweak_base(self.x, self.y)
+
+
+class SimLensStack(LensStack):
```
teddyrendahl: Does this belong here instead of the `tests` folder?

PR #274, opened 2018-07-26
align function and align_z motor added to LensStack

MattHooverSLAC:
Align and align_z are added to LensStack class, to help users create a virtual line of calibrated positions for the beam in a LensStack.  The align class will use tweak on the x and y motors at the upper and lower limits of the z motor, and using that information to create a useful pseudo positioner that, when moved, will move the z axis to the position and update the x and y motors to a calibrated position.
 
There's also some changes in the sim class, that's from Zach.  Includes a tester for the LensStack, updates to the synthetic motors with the "FastMotor" that is the old SlowMotor but with instant movement.  This can be more helpful than the SynMotor when a SoftPositioner is needed.


pcdsdevices/lens.py
```diff
@@ -29,15 +29,68 @@ def __init__(self, prefix, *, name, **kwargs):
...
+               self.z.presets.positions.exit.pos]
+        x_pos = ((pos[0]-pos[3])/(pos[2]-pos[5]))*(z_pos-pos[2])+pos[0]
+        y_pos = ((pos[1]-pos[4])/(pos[2]-pos[5]))*(z_pos-pos[2])+pos[1]
+        return self.RealPosition(x = x_pos, y = y_pos, z = z_pos)
+
+    @real_position_argument
+    def inverse(self, real_pos):
+        return self.PseudoPosition(calib_z = self.z.position)
+
+    def align(self,z_position=None):
```
ZLLentz: I was about to check what this used to be called, but the servers are down so I can't. We'll sync the name with whatever the old code defined.

pcdsdevices/lens.py
```diff
@@ -29,15 +29,68 @@ def __init__(self, prefix, *, name, **kwargs):
...
+        print()
+        self.tweak()
+        pos.extend([self.x.position,self.y.position,self.z.position])
+        self.x.presets.add_hutch(value=pos[0],name="entry")
+        self.x.presets.add_hutch(value=pos[3],name="exit")
+        self.y.presets.add_hutch(value=pos[1],name="entry")
+        self.y.presets.add_hutch(value=pos[4],name="exit")
+        self.z.presets.add_hutch(value=pos[2],name="entry")
+        self.z.presets.add_hutch(value=pos[5],name="exit")
+        self.calib_z.move(z_position)
```
ZLLentz: You need to skip this call if `z_position` is `None`, the default value, or the real motors will throw errors

pcdsdevices/lens.py
```diff
@@ -29,15 +29,68 @@ def __init__(self, prefix, *, name, **kwargs):
...
+        Generates equations for aligning the beam based on user input.
+
+        This program uses two points, one made on the lower limit
+        and the other made on the upper limit, after the user uses tweak function 
+        to put the beam into alignment, and uses those two points
+        to make two equations to determine a y- and x-position
+        for any z-value the user wants that will keep the beam focused.
+        The beam line will be saved in a file in the presets folder,
+        and can be used with the pseudo positioner on the z axis.
+        """
```
ZLLentz: This is a good docstring, but it also needs to explain what the optional z_position argument does.

pcdsdevices/lens.py
```diff
@@ -29,15 +29,68 @@ def __init__(self, prefix, *, name, **kwargs):
...
+        setup_preset_paths(hutch='presets',exp='presets')
```
ZLLentz: We can't do this here, it needs to be done outside of the class.

pcdsdevices/lens.py
```diff
@@ -29,15 +29,68 @@ def __init__(self, prefix, *, name, **kwargs):
...
+        self.z.move(self.z.limits[1])
+        print()
+        self.tweak()
+        pos.extend([self.x.position,self.y.position,self.z.position])
+        self.x.presets.add_hutch(value=pos[0],name="entry")
+        self.x.presets.add_hutch(value=pos[3],name="exit")
+        self.y.presets.add_hutch(value=pos[1],name="entry")
+        self.y.presets.add_hutch(value=pos[4],name="exit")
+        self.z.presets.add_hutch(value=pos[2],name="entry")
+        self.z.presets.add_hutch(value=pos[5],name="exit")
```
ZLLentz: I have to wonder if we should use more descriptive names than "entry" and "exit" to guarantee we don't conflict with a user's preset positions.

We should definitely check if `add_hutch` exists, and if not, ask the user to set up their preset paths correctly before doing this.

pcdsdevices/lens.py
```diff
@@ -29,15 +29,68 @@ def __init__(self, prefix, *, name, **kwargs):
...
         self.y_prefix = y_prefix
         self.z_prefix = z_prefix
         super().__init__(x_prefix, *args, **kwargs)
+
+    def tweak(self):
+        """
+        Calls the tweak function from mv_interface
+        with the x and y motors.
+        """
+        tweak_base(self.x,self.y)
```
ZLLentz: This will upset the style checker, it should be:
`tweak_base(self.x, self.y)`

pcdsdevices/lens.py
```diff
@@ -29,15 +29,68 @@ def __init__(self, prefix, *, name, **kwargs):
...
+        setup_preset_paths(hutch='presets',exp='presets')
+        pos = [self.x.presets.positions.entry.pos,
+               self.y.presets.positions.entry.pos,
+               self.z.presets.positions.entry.pos,
+               self.x.presets.positions.exit.pos,
+               self.y.presets.positions.exit.pos,
+               self.z.presets.positions.exit.pos]
+        x_pos = ((pos[0]-pos[3])/(pos[2]-pos[5]))*(z_pos-pos[2])+pos[0]
+        y_pos = ((pos[1]-pos[4])/(pos[2]-pos[5]))*(z_pos-pos[2])+pos[1]
+        return self.RealPosition(x = x_pos, y = y_pos, z = z_pos)
```
ZLLentz: this will upset the style checker, it should be:
`return self.RealPosition(x=x_pos, y=y_pos, z=z_pos)`

pcdsdevices/lens.py
```diff
@@ -29,15 +29,68 @@ def __init__(self, prefix, *, name, **kwargs):
...
+
+    @real_position_argument
+    def inverse(self, real_pos):
+        return self.PseudoPosition(calib_z = self.z.position)
```
ZLLentz: this will upset the style checker, it should be:
`return self.PseudoPosition(calib_z=self.z.position)`

pcdsdevices/lens.py
ZLLentz: Also, the style checker will throw a fit here without a space after the comma:
`def align(self, z_position=None):`

pcdsdevices/lens.py
```diff
@@ -29,15 +29,68 @@ def __init__(self, prefix, *, name, **kwargs):
...
+        to put the beam into alignment, and uses those two points
+        to make two equations to determine a y- and x-position
+        for any z-value the user wants that will keep the beam focused.
+        The beam line will be saved in a file in the presets folder,
+        and can be used with the pseudo positioner on the z axis.
+        """
+        setup_preset_paths(hutch='presets',exp='presets')
+        self.z.move(self.z.limits[0])
+        self.tweak()
+        pos = [self.x.position,self.y.position,self.z.position]
```
ZLLentz: style checker wants a space after each comma

pcdsdevices/lens.py
```diff
@@ -29,15 +29,68 @@ def __init__(self, prefix, *, name, **kwargs):
...
+        self.z.move(self.z.limits[1])
+        print()
+        self.tweak()
+        pos.extend([self.x.position,self.y.position,self.z.position])
```
ZLLentz: style checker wants a space after each comma

pcdsdevices/lens.py
ZLLentz: style checker wants a space after each comma

ZLLentz:
Thanks Matt! I'm merging this now.

PR #275, opened 2018-08-03
TST: cover the keypress utils in tests

ZLLentz:
I figured out how to simulate `stdin` in a test when the code is relying on the `stdin` object to be a real file descriptor for methods like `select.select` and `tty.setcbreak`

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,62 @@
+import logging
+import pty
+import sys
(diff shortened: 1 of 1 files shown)


ZLLentz:
I'm merging this because Teddy is on vacation and it is innocuous, as it is just tests.

teddyrendahl:
👍

PR #277, opened 2018-08-08
MNT: Style cleanup

ZLLentz:
## Description

Some code cleanup, some tests added, some bugs fixed. Not much changed in the way of functionality. I'm going to self-merge because Teddy is out, but we can revisit when he returns.

## Motivation and Context

First step towards the next release. I don't want to release untested code now that I know how to properly test thinks like `mv_ginput`, `tweak`, etc.

I'm doing this myself rather than assigning it to Matt/Yesenia because it's my job to deploy this stuff. It isn't very instructive to levy out manifestos of style preferences.

## How Has This Been Tested?

Tests added

Some docstrings have been updated too

### DIFF
--- a/pcdsdevices/lens.py
+++ b/pcdsdevices/lens.py
@@ -9,6 +9,7 @@ from ophyd.pseudopos import (PseudoPositioner, PseudoSingle,
 from .doc_stubs import basic_positioner_init
 from .epics_motor import IMS
 from .inout import InOutRecordPositioner
+from .sim import FastMotor
 
 
 class XFLS(InOutRecordPositioner):
@@ -41,3 +42,12 @@ class LensStack(Device):
         self.y_prefix = y_prefix
         self.z_prefix = z_prefix
         super().__init__(x_prefix, *args, **kwargs)
+
+
+class SimLensStack(LensStack):
+    """
+    Test version of the lens stack for testing the Be lens class.
+    """
+    x = Cpt(FastMotor, limits=(-10, 10))
+    y = Cpt(FastMotor, limits=(-10, 10))
+    z = Cpt(FastMotor, limits=(-100, 100))
--- a/pcdsdevices/mv_interface.py
+++ b/pcdsdevices/mv_interface.py
@@ -204,11 +204,15 @@ class FltMvInterface(MvInterface):
         pos = pylab.ginput(1)[0][0]
         self.move(pos, timeout=timeout, wait=wait)
 
-    def tweak(*args):
-        if len(args) > 1:
-            return tweak_base(args[0], args[1])
-        else:
-            return tweak_base(args[0])
+    def tweak(self):
+        """
+        Control this motor using the arrow keys.
+
+        Use left arrow to step negative and right arrow to step positive.
+        Use up arrow to increase step size and down arrow to decrease step
+        size. Press q or ctrl+c to quit.
+        """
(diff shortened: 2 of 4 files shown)


pcdsdevices/lens.py
```diff
@@ -41,3 +42,12 @@ def __init__(self, x_prefix, y_prefix, z_prefix, *args, **kwargs):
         self.y_prefix = y_prefix
         self.z_prefix = z_prefix
         super().__init__(x_prefix, *args, **kwargs)
+
+
+class SimLensStack(LensStack):
```
teddyrendahl: Still thinking about where these all belong. It would be nice if all these live under the "`.sim`" module but would like to avoid repeating the file structure. Maybe we can think of something clever.

pcdsdevices/mv_interface.py
```diff
@@ -17,7 +17,7 @@
 from bluesky.utils import ProgressBar
 from ophyd.status import wait as status_wait
 
-from .utils import get_input
+from . import utils as util
```
teddyrendahl: Just curious. Why?

pcdsdevices/mv_interface.py
```diff
@@ -173,8 +187,13 @@ def umvr(self, delta, timeout=None):
     def mv_ginput(self, timeout=None, wait=True):
         """
         Moves to a location the user clicks on.
+
+        If there are existing plots, this will be the position on the most
+        recently active plot. If there are no existing plots, an empty plot
+        will be created with the motor's limits as the range.
         """
-        print("Select new motor x-position in current plot by mouseclick")
+        logger.info(("Select new motor x-position in current plot "
```
teddyrendahl: Duplicated parentheses

pcdsdevices/mv_interface.py
```diff
@@ -707,29 +739,35 @@ def movement(scale, direction):
...
+                else:
+                    print(" Left: move x motor left")
+                    print(" Right: move x motor right")
+                    print(" Down: move y motor down")
+                    print(" Up: move y motor up")
+                    print(" Shift_Up: scale*2")
+                    print(" Shift_Down: scale/2")
+                print(" Press q to quit."
                       " Press any other key to display this message.")
+                print()  # Newline
```
teddyrendahl: `'Press any other key ... .\n'`?

pcdsdevices/lens.py
ZLLentz: I keep flip-flopping. I started to put them into `sim`, but then I thought that everything will be easier if `sim.py` simply provided utilities for creating simulated devices, and then in the device-specific files we put the actual simulated devices. Then the `SimLensStack` is bundled with the `LensStack` and it's really easy to find it.

pcdsdevices/mv_interface.py
ZLLentz: technically the "correct" way to do this is actually `print(<msg>, end='\n\n')`

pcdsdevices/mv_interface.py
ZLLentz: to refer to `util.arrow_up`, etc.

pcdsdevices/mv_interface.py
ZLLentz: Somehow I thought that you needed the inner parentheses to concat the multiline string, but I am mistaken.

PR #278, opened 2018-08-08
ENH: Delay stage

ZLLentz:
## Description

Delay stage, fully generalized with real motor and delay motor units.

## Motivation and Context

Laser delay etc etc

## How Has This Been Tested?

A single sanity test

Added to docs

### DIFF
--- a/pcdsdevices/pseudopos.py
+++ b/pcdsdevices/pseudopos.py
@@ -1,8 +1,10 @@
 import logging
 
-from ophyd.device import Component as Cpt
+from cf_units import Unit
+from ophyd.device import Component as Cpt, FormattedComponent as FCpt
 from ophyd.pseudopos import (PseudoPositioner, PseudoSingle,
                              real_position_argument, pseudo_position_argument)
+from scipy.constants import speed_of_light
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


pcdsdevices/utils.py
```diff
@@ -69,3 +71,27 @@ def get_input():
...
+
+    new_unit: ``str``
+        The desired unit for the conversion
+
+    Returns
+    -------
+    new_value: ``float``
+        The starting value, but converted to the new unit.
+    """
+    start_unit = Unit(unit)
```
teddyrendahl: What exception gets raised on a bad unit? A sensical one?

pcdsdevices/pseudopos.py
```diff
@@ -89,5 +92,75 @@ def inverse(self, real_pos):
...
+        meters = seconds * speed_of_light / self.n_bounces
+        motor_value = convert_unit(meters, 'meters', self.motor.egu)
+        return self.RealPosition(motor=motor_value)
+
+    @real_position_argument
+    def inverse(self, real_pos):
+        """
+        Convert motor unit to delay unit
+        """
+        meters = convert_unit(real_pos.motor, self.motor.egu, 'meters')
```
teddyrendahl: Clever

pcdsdevices/utils.py
ZLLentz: fyi this is also throwing `/u1/zlentz/conda/envs/dev/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88` on import because `numpy` had a regression in the warning ignores block. I've looked into it and found that it isn't dangerous, but I might go back and suppress the warning explicitly.

pcdsdevices/pseudopos.py
ZLLentz: The approach here was to protect against people using motors with random scales

ZLLentz:
I'm merging this if the travis passes again. Very overdue.

PR #282, opened 2018-08-17
FIX: Lens calibration edge cases

ZLLentz:
- lens stack can't move to the exact limits in the real hardware
- tweak should catch and report motion errors without stopping
- tweak should announce that it is tweaking

I want to edit this some more but I'm putting it as a PR so I don't forget about it. These were the changes that Matt and I made to the Be lens axis calibration that made it work.


ZLLentz:
@teddyrendahl I've completely forgotten what else I wanted to do here, so I'm going to merge it because it fixed some issues. Can I get a quick review?

PR #287, opened 2018-09-11
MAINT: Add icons for lightpath

teddyrendahl:
## Description

A first pass at adding all symbols for lightpath devices. Happy to show these to anyone who wants to help me pick these. The good news is we can always adjust these at a later date.

## Motivation and Context

Used to render symbols for the PCDS Lightpath after https://github.com/pcdshub/lightpath/pull/85

### DIFF
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -78,7 +78,7 @@ class AttBase(FltMvInterface, PVPositioner):
     done_value = 0
 
     # QIcon for UX
-    _icon = 'ei.lines'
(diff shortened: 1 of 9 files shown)


teddyrendahl:
Here is a screenshot if it helps evaluate:
screen shot 2018-09-11 at 2 08 30 pm

teddyrendahl:
Who ever cares enough to suggest an alternative wins!

ZLLentz:
My main issue with these is that they seem to be oriented incorrectly

PR #288, opened 2018-09-11
MAINT: Subscription States

teddyrendahl:
## Description

Both the `OffsetMirror` and the `Attenuator` classes didn't have `SUB_STATE` subscription types. This confused the `lightpath`:

### Attenuator
A call to subscribe to the  `SUB_STATE`  method subscribes to all of the filter states. Any change in state of a child filter triggers a call. This is better than the alternative of subscribing to the actual transmission value which jitters with beam energy.

### OffsetMirror
The `OffsetMirror` never changes states. It is always considered inserted. The `SUB_STATE` is never called. This is slightly hacky and I can be convinced to remove it, but it solves my problem for now 😄 

## Motivation and Context

Related to https://github.com/pcdshub/lightpath/issues/75

## How Has This Been Tested?

Added a test in the attenuator class for the `SUB_STATE` subscription call

### DIFF
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -214,8 +214,7 @@ class AttBase(FltMvInterface, PVPositioner):
         if event_type is None:
             event_type = self._default_sub
         if event_type == self.SUB_STATE and not self._has_subscribed_state:
-            for filt in self.filters:
-                filt.subscribe(self._run_filt_state, run=False)
+            self.done.subscribe(self._run_filt_state, run=False)
             self._has_subscribed_state = True
         return cid
 
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -79,10 +79,13 @@ class AttBase(FltMvInterface, PVPositioner):
 
     # QIcon for UX
     _icon = 'fa.barcode'
+    # Subscription Types
+    SUB_STATE = 'state'
 
     def __init__(self, prefix, *, name, **kwargs):
         super().__init__(prefix, name=name, limits=(0, 1), **kwargs)
         self.filters = []
+        self._has_subscribed_state = False
         for i in range(1, MAX_FILTERS + 1):
             try:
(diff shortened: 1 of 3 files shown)


pcdsdevices/attenuator.py
```diff
@@ -203,6 +207,21 @@ def _setup_move(self, position):
             self._move_changed(value=moving_val)
             self._move_changed(value=self.done_value)
 
+    def subscribe(self, cb, event_type=None, run=True):
+        cid = super().subscribe(cb, event_type=event_type, run=run)
+        if event_type is None:
+            event_type = self._default_sub
+        if event_type == self.SUB_STATE and not self._has_subscribed_state:
+            for filt in self.filters:
+                filt.subscribe(self._run_filt_state, run=False)
```
ZLLentz: It's unintuitive, but it might be better to subscribe to the single "done" signal instead of to all the filter states. The "done" signal is the attenuator status readback. I'm imagining the callback being run up to 10 times after an attenuation change.

pcdsdevices/mirror.py
```diff
@@ -187,6 +187,8 @@ class OffsetMirror(Device):
                    kind='config')
     # Transmission for Lightpath Interface
     transmission = 1.0
+    # Subscription types
+    SUB_STATE = 'state'
```
ZLLentz: Was this really all that had to be added here? I'm surprised.

teddyrendahl: 😄

teddyrendahl: A subscription that is never run can still be subscribed to nonetheless.

pcdsdevices/attenuator.py
teddyrendahl: Good point. Will modify

teddyrendahl:
Swapped to subscribe to `done` signal to avoid repetitive subscription calls

teddyrendahl:
Rebased to resolve conflicts with #287

teddyrendahl:
@ZLLentz Ready for follow-up review

PR #289, opened 2018-09-12
ENH: Reflaser

teddyrendahl:
## Description

* Add a basic `Reflaser` class with an icon
* Change `TTReflaser` to inherit from the new base `Reflaser`

Also added a bit to `setup.cfg` so `flake8` wouldn't search in the `docs` or `pcdsdevices/areadetector` folders. Maybe something we should consider adding to standard packaging

## Motivation and Context

`Reflaser` wants an icon for the `lightpath` but we were just using `InOutRecordPositioner` before. This gives us a way to have a unique icon and also gives a nice blueprint if anyone wants to expand on the class.

Added to `device_types` table

### DIFF
--- a/pcdsdevices/inout.py
+++ b/pcdsdevices/inout.py
@@ -126,12 +126,16 @@ class InOutRecordPositioner(StateRecordPositioner, InOutPositioner):
     __doc__ += basic_positioner_init
 
 
-class TTReflaser(InOutRecordPositioner):
+class Reflaser(InOutRecordPositioner):
+    """Simple ReferenceLaser with In/Out States"""
+    _icon = 'fa.empire'
+    __doc__ += basic_positioner_init
+
+
+class TTReflaser(Reflaser):
     """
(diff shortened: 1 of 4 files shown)


pcdsdevices/inout.py
```diff
@@ -126,12 +126,16 @@ class InOutRecordPositioner(StateRecordPositioner, InOutPositioner):
...
+    _icon = 'fa.empire'
+    __doc__ += basic_positioner_init
+
+
+class TTReflaser(Reflaser):
     """
     Motor stack that includes both a timetool and a reflaser.
     """
-    __doc__ += basic_positioner_init
-
```
ZLLentz: Did you mean to remove these lines?

teddyrendahl: It was moved to the base `Reflaser` class as it pertains to both.

ZLLentz: But it is immediately overriden by the new subclass

PR #291, opened 2018-09-13
WIP: LODCM Branching Specification

teddyrendahl:
## Description

The `lightpath` requires that an optic report all possible beamlines that may branch off the line it is on. Unfortunately, since both `LODCM` are on the `HXD` branch which has **all** the HXR beamlines it must report them. The `main_line` argument has been changed to accept a `list`. For consistency sake I changed the `mono_line` to use a list as well. This last choice is debatable. 

## Motivation and Context

Have the lightpath report accurately

## How Has This Been Tested?

* I tried this with the `lightpath` and it correctly identified the LODCM as a blocking device for the `HXR` hutches. 
* Existing tests pass unmodified

Updated docstrings

### DIFF
--- a/pcdsdevices/lodcm.py
+++ b/pcdsdevices/lodcm.py
@@ -8,6 +8,7 @@ downstream hutches.
 It will not be possible to align the device with the class, and there will be
 no readback into the device's alignment. This is intended for a future update.
 """
+import itertools
 import logging
 import functools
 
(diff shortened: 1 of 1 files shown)


ZLLentz:
Are you saying that the `lightpath` needs each `LODCM` to report all possible destinations, even if the `LODCM` doesn't directly feed into them? Is this true for all devices? Seems like a strict requirement, maybe I misunderstand.

teddyrendahl:
I wrote a long response to why, but after thinking about it more I might just patch lightpath. Posting below for posterity.

teddyrendahl:
We changed how the `lightpath` worked since `skywalker`. The logic is as following:

1. Traverse the beamline, remember any optics we see
2. Stop when we find a device with a different beamline then the one we are on.
3. Check that all optics we have seen up to this point are configured to deliver to this new beamline. Otherwise, mark them as blocking.
4. Clear our cache of seen optics and repeat

This was done to deal with some complicated situations without giving the beamlines between every optic a unique name. This is a special pain for the few devices in between XRT M1H / XRT M2H.

Now, the way we define our beamlines is that the HXD line starts in the FEE and travels all the way to the XCS LODCM. That means when we split off of the main line all the way down at CXI, we still have both LODCMs in our "cache" of seen optics and want to see they point to CXI.

PR #295, opened 2018-09-19
MAINT: Slit Interface

teddyrendahl:
## Description
These are some changes to make the Slits nicer to work with in the `lightpath`. Largely reducing complexity and making their behavior easier to understand.

1. No more complex `transmission` calculation based on their `nominal_aperture`. If either dimension of the slit is closed past the `nominal_aperture` they are inserted, otherwise they removed.
2. `nominal_aperture` is now an `AttributeSignal`. This field will be populated by the `happi` database and now contained in the `Typhon` screen 😄
3. I added a few lines in the `Slits.__init__` to change the `Kind` of its children. I think this is the proper way to handle this situation in this case although it looks a bit strange. If you load the `SlitPositioner` by itself the `readback` should be a `hint` and the `setpoint` should be `normal`. If you load them in the context of the entire `Slits` object the `hints` should be the readback of the `width` and the `height` (not all four). Thoughts? 

## How Has This Been Tested?

* Remove tests for `transmission` calculations
* Added tests for subscription callbacks

### DIFF
--- a/pcdsdevices/slits.py
+++ b/pcdsdevices/slits.py
@@ -124,7 +124,6 @@ class Slits(Device, MvInterface):
 
     def __init__(self, *args, nominal_aperture=5.0, **kwargs):
         self._has_subscribed = False
-        self._nom = nominal_aperture
         super().__init__(*args, **kwargs)
         # Initialize nominal_aperture behind the scenes
         self.nominal_aperture._readback = nominal_aperture
--- a/pcdsdevices/slits.py
+++ b/pcdsdevices/slits.py
@@ -18,7 +18,7 @@ import time
 from ophyd.status import wait as status_wait
 from ophyd.pv_positioner import PVPositioner
 from ophyd import (Device, EpicsSignal, EpicsSignalRO, Component as Cpt,
-                   FormattedComponent as FCpt, Kind)
+                   FormattedComponent as FCpt)
 from ophyd.signal import AttributeSignal
 
 from .mv_interface import MvInterface, FltMvInterface
@@ -128,8 +128,8 @@ class Slits(Device, MvInterface):
         self._nom = nominal_aperture
         super().__init__(*args, **kwargs)
         # Modify Kind of center readbacks
-        self.xcenter.readback.kind = Kind.normal
-        self.ycenter.readback.kind = Kind.normal
+        self.xcenter.readback.kind = 'normal'
+        self.ycenter.readback.kind = 'normal'
 
     def move(self, size, wait=False, moved_cb=None, timeout=None):
         """
--- a/pcdsdevices/slits.py
+++ b/pcdsdevices/slits.py
@@ -57,10 +57,10 @@ class SlitPositioner(FltMvInterface, PVPositioner, Device):
     ``ophyd.PVPositioner``
         ``SlitPositioner`` inherits directly from ``PVPositioner``.
     """
-    setpoint = FCpt(EpicsSignal, "{self.prefix}:{self._dirshort}_REQ",
-                    kind='normal')
     readback = FCpt(EpicsSignalRO, "{self.prefix}:ACTUAL_{self._dirlong}",
                     kind='hinted')
+    setpoint = FCpt(EpicsSignal, "{self.prefix}:{self._dirshort}_REQ",
+                    kind='normal')
     done = Cpt(EpicsSignalRO, ":DMOV", kind='omitted')
 
     def __init__(self, prefix, *, slit_type="", name=None,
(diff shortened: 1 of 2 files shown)


pcdsdevices/slits.py
```diff
@@ -99,40 +100,36 @@ class Slits(Device, MvInterface):
...
+    blocking the beamline. Both the ``width`` and the ``height`` need to exceed
+    this ``nominal_aperture`` for the ``Slits`` to be considered removed.
     """
-    xcenter = Cpt(SlitPositioner, '', slit_type="XCENTER", kind='hinted')
-    xwidth = Cpt(SlitPositioner, '', slit_type="XWIDTH", kind='normal')
-    ycenter = Cpt(SlitPositioner, '', slit_type="YCENTER", kind='hinted')
-    ywidth = Cpt(SlitPositioner, '', slit_type="YWIDTH", kind='normal')
+    xwidth = Cpt(SlitPositioner, '', slit_type="XWIDTH", kind='hinted')
+    ywidth = Cpt(SlitPositioner, '', slit_type="YWIDTH", kind='hinted')
+    nominal_aperture = Cpt(AttributeSignal, attr='_nominal', kind='normal')
```
ZLLentz: Why `AttributeSignal` and not just `Signal` with something subscribed to it?

pcdsdevices/slits.py
```diff
@@ -99,40 +100,36 @@ class Slits(Device, MvInterface):
...
 
-    def __init__(self, *args, nominal_aperture=(5.0, 5.0), **kwargs):
+    def __init__(self, *args, nominal_aperture=5.0, **kwargs):
         self._has_subscribed = False
-        self.nominal_aperture = nominal_aperture
+        self._nom = nominal_aperture
         super().__init__(*args, **kwargs)
+        # Modify Kind of center readbacks
+        self.xcenter.readback.kind = Kind.normal
+        self.ycenter.readback.kind = Kind.normal
```
ZLLentz: Are you required to import `Kind` to do these? Kind of annoying they don't just accept `'normal'`

pcdsdevices/slits.py
```diff
@@ -55,10 +56,10 @@ class SlitPositioner(FltMvInterface, PVPositioner, Device):
     ``ophyd.PVPositioner``
         ``SlitPositioner`` inherits directly from ``PVPositioner``.
     """
-    setpoint = FCpt(EpicsSignal, "{self.prefix}:{self._dirshort}_REQ",
-                    kind='normal')
     readback = FCpt(EpicsSignalRO, "{self.prefix}:ACTUAL_{self._dirlong}",
                     kind='hinted')
+    setpoint = FCpt(EpicsSignal, "{self.prefix}:{self._dirshort}_REQ",
+                    kind='normal')
```
ZLLentz: I appreciate small things like rearranging the signals to the correct order

ZLLentz:
I'd like to note here as we talked about I think base `ophyd` should be in charge of masking subcomponent hints when the parent component doesn't have something hinted. Setting over them in the `__init__` is a harmless workaround, but it will be tiring to do this everywhere.

pcdsdevices/slits.py
teddyrendahl: This actually matters for `typhon`

ZLLentz: Didn't realize, that makes sense though

pcdsdevices/slits.py
teddyrendahl: Haven't tried this.

teddyrendahl: It worked

teddyrendahl:
My one consideration of keeping it an `AttributeSignal` is we may want the `setattr` to write back to the `happi` database. On second thought I might make this a `SignalRO` and say that all `nominal aperture` changes need to be done in the database itself.

ZLLentz:
I don't mind you keeping it an attribute signal, it just seemed like overkill for the current use

teddyrendahl:
I swapped it out for `SignalRO`. We don't want operators writing to the database and we don't want people changing the `nominal_aperture` temporarily then that information being wiped when we leave the `lightpath` session. 

This removes the need to fire any subscriptions as we assume this value is static through out operations.

ZLLentz:
Fair enough, keeping it simple is good. I have a feeling that `SignalRO` is located in the wrong module on the `ophyd` side. This means we should be passing in a `nominal_aperture` kwarg from `happi`, right?

pcdsdevices/slits.py
```diff
@@ -99,40 +99,38 @@ class Slits(Device, MvInterface):
...
     SUB_STATE = 'sub_state_changed'
     _default_sub = SUB_STATE
     # QIcon for UX
     _icon = 'fa.th-large'
 
-    def __init__(self, *args, nominal_aperture=(5.0, 5.0), **kwargs):
+    def __init__(self, *args, nominal_aperture=5.0, **kwargs):
         self._has_subscribed = False
-        self.nominal_aperture = nominal_aperture
+        self._nom = nominal_aperture
```
ZLLentz: it seems like this attribute is no longer used

teddyrendahl:
Correct about the `SignalRO`. I think they are waiting to address this in https://github.com/NSLS-II/ophyd/issues/571.

Also correct about the `happi` configuration.

ZLLentz:
:+1: merge at will if you're done with this

PR #299, opened 2018-09-24
ENH: EventSequencer Trigger

teddyrendahl:
## Description

The EventSequencer now has a `trigger` method that institutes the following behavior:
* If set to run forever, the returns a Status thats is done right away
* If set to a finite run time, wait for the Sequencer to complete its run

## Motivation and Context

Allows for use in of both the DAQ and the EventSequencer together with a simple:
`trigger_and_read` call.

## How Has This Been Tested?

Added a test for new behavior

Docstrings updated. A thorough explanation of how to utilize this object in conjunction with the DAQ is probably owed, but maybe not in this repository

### DIFF
--- a/pcdsdevices/sequencer.py
+++ b/pcdsdevices/sequencer.py
@@ -39,13 +39,7 @@ class EventSequencer(Device, MonitorFlyerMixin, FlyerInterface):
 
     .. code::
 
-        def step_then_sequence(detectors, step, pos_cache)
-            yield from one_nd_step(detectors, step, pos_cache)
-            yield from kickoff(sequencer)
-            yield from complete(sequencer)  # Waits for sequence to complete if
-                                            # not in "Run Forever" mode
-
-        scan([det], motor, ..., per_step=step_then_sequence)
+        scan([sequencer], motor, ....)
 
     Note
     ----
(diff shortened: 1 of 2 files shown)


ZLLentz:
This is missing a call to stop the infinite sequence (like how I did with the daq), which may be desirable if you're putting it at every scan step. I imagine that if your sequence is long you may want to ensure it starts from the beginning at each scan step. Otherwise you probably just run the sequencer independently of the scan, from python or gui.

teddyrendahl:
So when you call `trigger`, stop the `EventSequencer` then restart? Makes sense

ZLLentz:
Yes, either stop/restart at trigger or stop at read.

teddyrendahl:
Ready for a full review. One thought. I'd like to move the `DEFAULT_SLEEP` to the DAQ side. In reality the sleep is necessary because when the DAQ says it is ready it actually isn't so we need a little "settle time". Since this behavior is really intertwined with the DAQ I think it makes more sense to live there and not on the EventSequencer object.

Thoughts?

ZLLentz:
RE: default sleep: you're saying this should go into the built-in wait for `daq.trigger`?

This is fine, especially if you think other things will need to wait for the `daq`. It should be wrapped into the raw begin call as a configurable extra wait. The goal is for trigger to not return until acquisition has started.

teddyrendahl:
@ZLLentz I agree. I'll remove those commits from this PR and put an issue on `pcdsdaq`

ZLLentz:
I'm merging this soon if there are no objections

PR #300, opened 2018-09-27
MAINT: SPG in a scan

### LINKED ISSUE #276: Potential Issues with IMS/Newport Pause
I quit out of a `bluesky` scan today during a scan motion step via ctrl+c spam, and the motors got left in a `pause` state. This means the next scan didn't work.

Perhaps we need to either rethink SPG or set it to "stop" and then to "go" on stage.

teddyrendahl:
## Description

Setting the SPG field on `pause` proved very cumbersome in deployment. Largely because the SPG field is buried on most of the high level motor screens. It is also not clear if there is much added benefit. This PR:

* Renames `pause` and `go` to `spg_pause` and `spg_go`. This keeps the functionality but makes sure that they are not called via `bluesky`
* Removes the `resume` method. This was only necessary for the `bluesky` interface
* I left `stop`. I think this is actually the correct behavior for this function. 
* I put the `motor_spg`as a `stage_sig`. This means that at the end of the scan when we `unstage` we will always set the `SPG` field back to what it was at the start of the scan.

## Motivation and Context

Closes #276 

## How Has This Been Tested?

Modified and added tests

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -170,6 +170,10 @@ class PCDSMotorBase(EpicsMotorInterface):
     # paused and ready to resume on Go 'Paused', and to resume a move 'Go'.
     motor_spg = Cpt(EpicsSignal, ".SPG", kind='omitted')
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.stage_sigs[self.motor_spg] = 2
+
     def stop(self):
         """
         Stops the motor.
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -174,36 +174,22 @@ class PCDSMotorBase(EpicsMotorInterface):
         """
         Stops the motor.
 
-        After which the motor
-        must be set back to 'go' via <motor>.go()
-        in order to move again.
+        After which the motor must be set back to 'go' via <motor>.spg_go() in
+        order to move again.
         """
         return self.motor_spg.put(value='Stop')
 
-    def pause(self):
+    def spg_pause(self):
         """
         Pauses a move.
 
-        Move will resume if <motor>.resume()
-        or <motor>.go() are called.
+        Move will resume if <motor>.go() is called.
         """
         return self.motor_spg.put(value='Pause')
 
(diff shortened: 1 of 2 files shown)


teddyrendahl:
Silently wondering if just the `stage_sig` is enough ....

rajanplumley:
If this is mostly renaming, how exactly does this change behavior?  Was there crosstalk with bluesky?

ZLLentz:
Bluesky was calling `pause` when you spammed ctrl+c out of a scan, leaving the system in an unintentionally changed state. We should talk about what `stop` should do.

ZLLentz:
Note that with this change, we'll have a similar issue: `stop` will be called when we quit the scan, leaving the motors with spg set to stop. This is why I think we should rename that one as well so that it isn't called by `bluesky`, instead the `ophyd` `stop` can be used which maintains the state of `spg` but brings the motor to a stop.

https://github.com/NSLS-II/bluesky/blob/df6b93641c93de157165ebe1559c68e85b0f29b2/bluesky/run_engine.py#L611-L618

teddyrendahl:
I think this warrants an in-depth exploration of how this works with staging, e.t.c. I think I'm fine with:
1. Scan is started, Motor is staged, starting point of `SPG` is stored
2. Operator requests pause
3. Motor is paused
4. Operator aborts scan
5. Motor is unstaged back to setting prior setting, usually "Go"

The only thing that breaks this flow is if the scan errors and we never reach "unstage". I don't recall how robust the `RunEngine` is at catching exceptions and making sure things are `unstaged`

ZLLentz:
The `RunEngine` will do `unstage` if you are patient and don't spam ctrl+c to death iirc

ZLLentz:
My perspective is that `RunEngine` wants control over the motor state, so I'd rather give the `RunEngine` full control than let the IOC keep that state.

teddyrendahl:
Renamed to `spg_stop`. Really think this should be `halt` or something, but it'll make more sense this way for anyone with `EpicsMotor` experience. 

Leaving `stage_sigs` in. If you put a motor in a scan that is stopped, we'll put it to start then put it back when we are done. I kind of like this behavior.

Issue #198, opened 2018-04-11
Event Sequencer Setup


ZLLentz:
- Provide an intuitive interface for setting up an event sequencer sequence programatically.
- Provide a way to save/restore sequences

PR #302, opened 2018-10-10
ENH: Event Sequence Read/Write

slactjohnson:
Add in ability to read/write sequences to the event sequencer. 

<sequencer_object>.sequence.read() to read the current sequence,  <sequencer_object>.sequence.write(sequence) to write a new sequence.

Should not require any adjustment to the instantiation of the event sequencer  object. The sequence class initializes itself based on the prefix for the event sequencer. 

## Motivation and Context

Needed for MEC to run the laser in BlueSky/Ophyd hutch python. Should be useful for other hutches as well.

Appears to address issue #198.

## How Has This Been Tested?

Tested it on  MEC event sequencer with my local copy of mecpython and pcdsdevices.

I put in docstrings as I thought appropriate. I'm sure these could be improved.

### DIFF
--- a/pcdsdevices/sequencer.py
+++ b/pcdsdevices/sequencer.py
@@ -1,12 +1,144 @@
 import logging
 
 from ophyd import Device, EpicsSignal, EpicsSignalRO, Component as Cpt
-from ophyd import FormattedComponent as FCpt
 from ophyd.status import DeviceStatus, SubscriptionStatus
 from ophyd.utils.epics_pvs import raise_if_disconnected
 from ophyd.flyers import FlyerInterface, MonitorFlyerMixin
 
 logger = logging.getLogger(__name__)
+
+
+class EventSequence(Device):
+    """Class for the event sequence of the event sequencer."""
+
+    ec_array = Cpt(EpicsSignal, ':SEQ.A')
+    bd_array = Cpt(EpicsSignal, ':SEQ.B')
+    fd_array = Cpt(EpicsSignal, ':SEQ.C')
+    bc_array = Cpt(EpicsSignal, ':SEQ.D')
+
+    def get_seq(self, current_length=True):
+        """Retrieve the current event sequence. Returns a list of 4 lists, with
+        the form [[Event Code], [Beam Delta], [Fiducial Delta], [Burst Count]].
+        Returns the current sequence up to the current play length, the
+        {prefix}:LEN PV, unless the current_length option is set to False. If
+        set to false, the whole sequence will be returned.
+
+        Parameters
+        ----------
+        current_length : bool
+            Option to retrieve the sequence up to the current length. Defaults
+            to True.
+
+        Examples
+        --------
+
+        EventSequence.get_seq()
+
+        EventSequence.get_seq(current_length=False) # Get whole sequence
+
+        """
+
+        if self.parent and current_length is True:
+            seq_length = self.parent.sequence_length.get()
+        else:
+            seq_length = 2048  # Whole thing
+
+        sequence = [[], [], [], []]
+        sequence[0] = self.ec_array.get()[0:seq_length]
+        sequence[1] = self.bd_array.get()[0:seq_length]
+        sequence[2] = self.fd_array.get()[0:seq_length]
+        sequence[3] = self.bc_array.get()[0:seq_length]
+
+        return sequence
+
+    def put_seq(self, sequence, update_length=True):
+        """Write a sequence to the event sequencer. Takes a list of lists,
+        with each sub-list representing one part of the event sequence (event
+        codes, beam deltas, etc). The written sequence will overwrite the
+        current sequence in order, up to the specified length. The play length
+        of the sequencer will automatically be updated, unless the
+        update_length flag is set to False.
+
+        Parameters
+        ----------
+        sequence: list
+            List of lists describing the event sequence. The list takes the
+            form [[Event Code], [Beam Delta], [Fiducial Delta], [Burst Count]].
+
+        update_length: bool
+            Option to automatically update the play length, the
+            {prefix}:LEN PV, to the length of the written sequence. Defaults
+            to True.
+
+        Examples
+        --------
+        seq = [[167, 168, 182, 176, ... ], # Event Codes
+               [ 19,   4,   0,   0, ... ], # Beam Deltas
+               [  0,   0,   0,   0, ... ], # Fiducial Deltas
+               [  0,   0,   0,   0, ... ]] # Burst Counts
+
+        EventSequence.put_seq(seq)
+
+        EventSequence.put_seq(seq, update_length=False) # Don't update length
+
+        """
+
+        curr_seq = self.get_seq(current_length=False)
+        new_seq = curr_seq.copy()
+
+        # Modify just the requested lines of the current sequence
+        for i in range(len(curr_seq)):
+            for j in range(len(sequence[i])):
+                new_seq[i][j] = sequence[i][j]
+
+        # Update the length of the sequence if update_length == True and
+        # the event sequence is a child of the EventSequencer
+        if self.parent and update_length is True:
+            # Sequence length from number of event codes
+            new_len = len(sequence[0])
+            self.parent.sequence_length.put(new_len)
+
+        self.ec_array.put(new_seq[0])
+        self.bd_array.put(new_seq[1])
+        self.fd_array.put(new_seq[2])
+        self.bc_array.put(new_seq[3])
+
+    def show(self, num_lines=None):
+        """Print a human readable copy of the current event sequence. Shows
+        the current sequence up to the length of the sequencer play length,
+        unless otherwise specified by the num_lines parameter.
+
+        Parameters
+        ----------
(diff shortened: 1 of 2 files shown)


teddyrendahl:
Hell yeah! This needed to get done! 

@slactjohnson do you have time to meet this week and discuss? It looks like we just need to get the tests to pass here and a few minor tweaks on the `Device` side. Should be quick?

slactjohnson:
@teddyrendahl Sure, I am available today, but will be out tomorrow and Friday. I will be back Monday.

pcdsdevices/sequencer.py
```diff
@@ -64,6 +67,13 @@ class EventSequencer(Device, MonitorFlyerMixin, FlyerInterface):
 
     def __init__(self, prefix, *, name=None, monitor_attrs=None, **kwargs):
         monitor_attrs = monitor_attrs or ['current_step', 'play_count']
+
+        # Setup Event Sequence
+        hutch_map = {1:'AMO', 2:'SXR', 3:'XPP', 4:'XCS', 5:'CXI', 6:'MEC', 7:'MFX'}
+        hutch = hutch_map[int(prefix[-1])]
```
ZLLentz: This isn't sufficient to cover spare sequences:
```
$ caget ECS:SYS0:{1..16}:HUTCH_NAME
ECS:SYS0:1:HUTCH_NAME          AMO
ECS:SYS0:2:HUTCH_NAME          SXR
ECS:SYS0:3:HUTCH_NAME          XPP
ECS:SYS0:4:HUTCH_NAME          XCS
ECS:SYS0:5:HUTCH_NAME          CXI
ECS:SYS0:6:HUTCH_NAME          MEC
ECS:SYS0:7:HUTCH_NAME          MFX
ECS:SYS0:8:HUTCH_NAME          XPP
ECS:SYS0:9:HUTCH_NAME          MFX
ECS:SYS0:10:HUTCH_NAME         SXR
ECS:SYS0:11:HUTCH_NAME         XPP
ECS:SYS0:12:HUTCH_NAME         XCS
ECS:SYS0:13:HUTCH_NAME         None
ECS:SYS0:14:HUTCH_NAME         None
ECS:SYS0:15:HUTCH_NAME         None
ECS:SYS0:16:HUTCH_NAME         CXI
```

pcdsdevices/sequencer.py
```diff
@@ -189,3 +199,131 @@ def stop(self):
...
+               [182,  1, 0, 0],
+               [176,  0, 0, 0],
+               [169,  0, 0, 0]]
+
+        EventSequence.write(seq)
+        
+        """
+
+        if len(sequence) > 20:
+            raise ValueError("The sequence length cannot be longer than 20!")
```
ZLLentz: Despite the GUI having only 20 lines, the max sequence length is actually 2048. You just need to set the `stop at step` field  (`LEN` pv) to a higher value

pcdsdevices/sequencer.py
```diff
@@ -64,6 +67,13 @@ class EventSequencer(Device, MonitorFlyerMixin, FlyerInterface):
 
     def __init__(self, prefix, *, name=None, monitor_attrs=None, **kwargs):
         monitor_attrs = monitor_attrs or ['current_step', 'play_count']
+
+        # Setup Event Sequence
+        hutch_map = {1:'AMO', 2:'SXR', 3:'XPP', 4:'XCS', 5:'CXI', 6:'MEC', 7:'MFX'}
+        hutch = hutch_map[int(prefix[-1])]
+    
+        self.sequence = EventSequence('{}:ECS:IOC:01'.format(hutch), hutch_num=prefix[-1], name='{}_sequence'.format(hutch))
```
ZLLentz: Setting up `ophyd` subdevices like this is not recommended because it confuses some of the `bluesky` machinery. Every subdevice should be included as a `Component` in the class definition. However, there is no reason that this `EventSequence` object needs to be a `Device` subclass, so maybe we just nix that.

pcdsdevices/sequencer.py
```diff
@@ -189,3 +199,131 @@ def stop(self):
...
+    bc = FCpt(EpicsSignal, '{self.prefix}:BC_{self._ID}:{self._line}')
+
+    def __init__(self, prefix, hutch_id=None, line=None, **kwargs):
+
+        self._ID = hutch_id
+        self._line = line
+
+        super().__init__(prefix, **kwargs)
+
+    def read(self):
```
ZLLentz: This breaks the `bluesky` API by redefining `read`. I don't think this device will ever be read by `bluesky`, but if it does we should be careful here. Even if we don't, I suspect this could get confusing.

ZLLentz:
I'm a fan of your implementation in general, assembling a mega list that looks like edm screen table is more intuitive than repeatedly calling `setstep` as in the old python. We can do everything we could do in the old python in this one, all of the `setstep` calls will be replaced by assembling the list and then making one call to the object. I commented on some `bluesky`/`ophyd`-specific things, as well as a few things I learned from poking around the edges of the sequencer a bit. `bluesky`/`ophyd` fixes (as well as beating the tests into submission) are more important than the other things. The old sequencer class also had some errors in handling the spare sequences.

pcdsdevices/sequencer.py
```diff
@@ -189,3 +199,131 @@ def stop(self):
...
+            [<event code>, <beam delta>, <fiducial delta>, <burst count>].
+
+        Examples
+        --------
+        SequenceLine.write([140, 12, 0, 0])
+
+        """
+
+        if len(line) != 4:
+            raise ValueError("The sequence line must be a 4 item list!")
```
ZLLentz: One more thought: it'd be cool to be able to use a 5th slot for the comment/DESC field!

slactjohnson:
@teddyrendahl @ZLLentz Thanks for your comments! I think my recent edits have all of these covered except for the variable number of lines of the event sequencer (I was unaware that this could go beyond 20). 

I am leaving on vacation for the next 2 days, but will take another swing at this when I return on Monday.

ZLLentz:
@slactjohnson Thanks for working on this Tyler! Remember that you do not need to address literally everything right now: having a good working base is most important. We can decide to not resolve everything and save it for a later pull request, adding relevant issues to the issues page. We'll talk next week.

slactjohnson:
@ZLLentz Under what circumstances are more than 20 event codes used? A quick check showed that nobody is using anything beyond <prefix>:19 (or, at least, a caget of the PV showed that PVs beyond 19 were disconnected). As long as I have been here MEC has used 20 or less, but I am not as plugged in to the activities of the other hutches.

teddyrendahl:
We use very long sequences in MFX quite often

ZLLentz:
Sometimes XPP uses a very long laser sequence with "random" on/off patterns

ZLLentz:
Of course, the long sequences can be a different PR, and Teddy or I can do it, especially if it's an XPP/MFX thing and not an MEC thing.

ZLLentz:
Here's a dirty trick to fix some testing oddities, put this near the top with the other fake device stuff:
```
import pcdsdevices.sequencer
pcdsdevices.sequencer.SequenceLine = make_fake_device(pcdsdevices.sequencer.SequenceLine)
```
The `make_fake_device` method assumes that we are not creating `ophyd` subdevices anywhere other than `Component` declarations. This trick replaces the module's `SequenceLine` with a proper fake one.

pcdsdevices/sequencer.py
```diff
@@ -189,3 +204,135 @@ def stop(self):
...
+    ds = FCpt(EpicsSignal, '{self.prefix}:EC_{self._ID}:{self._line}.DESC')
+
+    def __init__(self, prefix, hutch_id=None, line=None, **kwargs):
+
+        self._ID = hutch_id
+        self._line = line
+
+        super().__init__(prefix, **kwargs)
+
+    def get(self):
```
klauer: Believe it or not, the default implementation of `ophyd.Device.get` will do this for you - and (arguably) better. 

It gives you not just a tuple of values you have to refer to by index, but it's a `namedtuple` generated automatically of the form `namedtuple('SequenceLineTuple', 'ec bd fd bc ds')`. This means you can access the values either by index `seq.get()[0]` or `seq.get().ec`, for example.

teddyrendahl:
@slactjohnson This looks good to me! You want to add the longer sequences in this PR or a subsequent?

pcdsdevices/sequencer.py
teddyrendahl: I think what he is suggesting here is that if you delete this method it will do the same thing! It will just be a `collections.namedtuple` from the standard library not a `list`.

slactjohnson:
@teddyrendahl I would like to wait. I am working on a new version of the sequencer using the array PVs that will cover the full length of the sequencer. I should be able to push the updated version today.

ZLLentz:
I'm happy to merge this as-in and consider adding to it later, if there are no objections. `EventSequencer` having no setup options is a sore spot in the library.

klauer:
I think it's good to merge. Given that no one is using this just yet, significantly changing the API in the future shouldn't be a big deal.

PR #305, opened 2018-10-17
FIX: update requirements.txt according to conda recipe

klauer:


ZLLentz:
I always forget the requirements file exists since I focus on conda

PR #306, opened 2018-10-18
Motor Factory

teddyrendahl:
## Description

Instead of sorting out which kind of motor you have, we should be able to tell based on the prefix it self. To solve this I wrote `Motor`. This snoops the `prefix` finds the correct class and instantiates it

## Motivation and Context

Main motivation is ease of use for `happi`. Instead of having a one to one relationship between containers and classes, we should be building these factories that do the correct thing with the information provided. This simplifies the dependencies between the two libraries as we can repoint the objects by modifying the factories, not changing class specifications on both ends.

## How Has This Been Tested?

Added a test

Add to `device_types` table

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -456,3 +456,50 @@ class MotorDisabledError(Exception):
     Error that indicates that we are not allowed to move.
     """
     pass
+
+
+def Motor(prefix, **kwargs):
+    """
+    Load a PCDSMotor with the correct class based on prefix
+
+    The prefix is searched for one of the component keys in the table below. If
+    none of these are found, by default a ``ophyd.EpicsMotor`` will be used.
+
+    +---------------+------------------------+
+    | Component Key + Python Class           |
+    +===============+========================+
+    | MMS           | :class:`.IMS`          |
+    +---------------+------------------------+
+    | MMN           | :class:`.Newport`      |
+    +---------------+------------------------+
+    | MZM           | :class:`.PMC100`       |
+    +---------------+------------------------+
+    | MMB           | :class:`.BeckhoffAxis  |
+    +---------------+------------------------+
+    | PIC           | :class:`.PCDSMotorBase |
+    +---------------+------------------------+
+
+    Parameters
+    ----------
+    prefix: str
+        Prefix of motor
+
+    kwargs:
+        Passed to class constructor
+    """
+    # Available motor types
+    motor_types = (('MMS', IMS),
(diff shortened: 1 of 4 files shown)


klauer:
I know this isn't really the right place for such a question - but, uh, where did those abbreviations/acronyms come from for the motor types?

teddyrendahl:
https://confluence.slac.stanford.edu/display/PCDS/LCLS+Naming+Convention#LCLSNamingConvention-MotionComponents(CCC)

pcdsdevices/epics_motor.py
```diff
@@ -456,3 +456,50 @@ class MotorDisabledError(Exception):
...
+                   ('PIC', PCDSMotorBase))
+    # Search for component type in prefix
+    for cpt_abbrev, _type in motor_types:
+        if f':{cpt_abbrev}:' in prefix:
+            logger.debug("Found %r in prefix %r, loading %r",
+                         cpt_abbrev, prefix, _type)
+            return _type(prefix, **kwargs)
+    # Default to ophyd.EpicsMotor
+    logger.warning("Unable to find type of motor based on component. "
+                   "Using 'ophyd.EpicsMotor")
```
ZLLentz: typo, lingering apostrophe

pcdsdevices/epics_motor.py
```diff
@@ -456,3 +456,50 @@ class MotorDisabledError(Exception):
...
+    | MMS           | :class:`.IMS`          |
+    +---------------+------------------------+
+    | MMN           | :class:`.Newport`      |
+    +---------------+------------------------+
+    | MZM           | :class:`.PMC100`       |
+    +---------------+------------------------+
+    | MMB           | :class:`.BeckhoffAxis  |
+    +---------------+------------------------+
+    | PIC           | :class:`.PCDSMotorBase |
+    +---------------+------------------------+
```
ZLLentz: typos in this table, forgot to close your backquotes

pcdsdevices/epics_motor.py
```diff
@@ -456,3 +456,50 @@ class MotorDisabledError(Exception):
     Error that indicates that we are not allowed to move.
     """
     pass
+
+
+def Motor(prefix, **kwargs):
```
ZLLentz: At first I balked at this constructor, but upon closer inspection I see that there is only one positional arg to motors.

ZLLentz:
But please fix the docstring typos before merging so the docs look nice

teddyrendahl:
Will do! Thanks for looking closely

PR #307, opened 2018-10-25
FIX: The build was broken because Matplotlib/Conda

ZLLentz:
## Description

- Stop using `pylab` directly to simplify things
- Move imports of `pyplot` into methods to avoid selecting a backend early
- Pin `matplotlib` to under 3 because the conda builds are a mess

## Motivation and Context

- No reason to use `matplotlib` indirectly, made things harder to debug
- Backends why
- `3.0.0`+ conflicts with `pmgr` because something something `pyqt` must be `5.9`, but then it also breaks the travis here because it somehow installs `3.0.0` with pyqt `4.11`. I have no idea what is going on, I assume the recipes are broken.

### DIFF
--- a/pcdsdevices/mv_interface.py
+++ b/pcdsdevices/mv_interface.py
@@ -12,8 +12,8 @@ from threading import Thread, Event
 from types import SimpleNamespace, MethodType
 from weakref import WeakSet
 
-import pylab
 import yaml
+import matplotlib.pyplot as plt
 from bluesky.utils import ProgressBar
 from ophyd.status import wait as status_wait
 
@@ -194,7 +194,7 @@ class FltMvInterface(MvInterface):
         """
         logger.info(("Select new motor x-position in current plot "
                      "by mouseclick"))
-        if not pylab.get_fignums():
+        if not plt.get_fignums():
             upper_limit = 0
             lower_limit = self.limits[0]
             if self.limits[0] == self.limits[1]:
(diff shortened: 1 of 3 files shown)


pcdsdevices/mv_interface.py
```diff
@@ -192,9 +191,12 @@ def mv_ginput(self, timeout=None):
         recently active plot. If there are no existing plots, an empty plot
         will be created with the motor's limits as the range.
         """
+        # This is usually bad, but importing this at module level forces
+        # matplotlib to select a backend on import, breaking some things
+        import matplotlib.pyplot as plt  # NOQA
```
klauer: At times, importing like this is the best option you have. 👍 

(I'm not sure this warrants the 2-line disclaimer above, but that's just my opinion.)

tests/test_mv_interface.py
```diff
@@ -6,7 +6,6 @@
 import os
 import signal
 
-import pylab
```
klauer: I don't think it's such a big issue for the tests. Importing `pcdsdevices` and having it select a backend is the most annoying part.

pcdsdevices/mv_interface.py
ZLLentz: Sure, I'll make it a 1-line disclaimer. I just don't want someone to move it back because it looks wrong if you don't know.

tests/test_mv_interface.py
ZLLentz: You're probably correct, but I wanted to be thorough.

PR #313, opened 2018-11-01
ENH: add common plugins to PCDSDetector

klauer:
## Description

Adds all common plugins to PCDSDetector

## Motivation and Context

Plugins are available for most (all?) PCDS detectors and should be accessible to the user.

## How Has This Been Tested?

Interactively with CXI inline detectors. Default port chain information from `CXI:SC2:INLINE:` as it appears unmodified - but this may need revision.

See docstrings

## Backward incompatible changes
The leading `':'` has been stripped from all plugins. The logic is that the prefix should be what all PVs of a device have in common. This could be put back, but I wouldn't recommend it.

This means that beamline configurations need to be updated with the correct prefixes.

## TODO
PCDSDetector-related local changes should be brought back to ophyd. @teddyrendahl @ZLLentz and I should discuss what should remain for local customizations, if any.

### DIFF
--- a/pcdsdevices/areadetector/detectors.py
+++ b/pcdsdevices/areadetector/detectors.py
@@ -42,7 +42,37 @@ class PCDSDetector(PCDSDetectorBase):
 
     Default port chains
     -------------------
+    The default configuration of ports is as follows::
 
+        CAM -> CC1
+        CAM -> CC2
+        CAM -> HDF51
+        CAM -> IMAGE1:CC
+        CAM -> IMAGE1:Proc
+        CAM -> IMAGE1:ROI
+        CAM -> IMAGE2:CC
+        CAM -> IMAGE2:Over
+        CAM -> IMAGE2:Proc
+        CAM -> IMAGE2:ROI -> IMAGE2
+        CAM -> JPEG1
+        CAM -> NetCDF1
+        CAM -> Over1
+        CAM -> Proc1
+        CAM -> ROI1 -> IMAGE1
+        CAM -> ROI1 -> Stats1
+        CAM -> ROI2
+        CAM -> ROI3
+        CAM -> ROI4
+        CAM -> Stats2
+        CAM -> Stats3
+        CAM -> Stats4
+        CAM -> Stats5
+        CAM -> THUMBNAIL:CC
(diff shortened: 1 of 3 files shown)


ZLLentz:
Do you want to meet, go through the areadetector folder, and see how many things we can remove? My intuition is that a lot of the stuff we have in there is completely unnecessary.

klauer:
👍 @ZLLentz , let's meet and take a look.

slactjohnson:
If we are talking about removing plugin support, can I put in a request to keep one? 

Mike and I are hoping to use the area detector stats plugin to do our centroiding for us. It would be very useful (really, necessary) to have that available to us through python for the pointing/centering routines we want to do.

klauer:
@slactjohnson Plugin support will not go away. This PR improves the current situation by adding all the common, default plugins for all PCDS detectors that are deployed (to my knowledge, at least).

What we're referring to is removing unnecessary customizations in `pcdsdevices.areadetector` that already exist in the upstream `ophyd.areadetector` library.

pcdsdevices/areadetector/detectors.py
```diff
@@ -24,27 +25,127 @@ class PCDSDetectorBase(DetectorBase):
...
+        CAM -> Trans1
 
-    IMAGE2: reduced rate image
-    Stats2: reduced rate stats
+    Note
+    ----
+    Subclasses should replace 'cam' with that of the respective detector, such
+    as `ophyd.areadetector.cam.PilatusDetectorCam` for the Pilatus detector.
     """
-    image = Cpt(ImagePlugin, ':IMAGE2:', read_attrs=['array_data'])
```
teddyrendahl: Not at all concerned about back-compat with colons. This however might cause issues as I think we do have code out there that refers to `my_det.image` and `my_det.stats`. We should drop these names in the long run as they are too vague, but do you think you could alias these maybe we `DeprecationWarning` until then? Thoughts?

teddyrendahl: I like the numbers, this way the names match the screens as well. I was imagining a `property` based system.

klauer: See latest changes, noting that the default warning is `UserWarning`.

PR #320, opened 2018-12-07
BLD: Use Jinja magic to feed conda-recipe from requirements.txt

ZLLentz:
This isn't super important, but the idea snagged me today and I had to see it through. Curious what you think about this.

## Description

- Add `install_requires` to the `setup.py` file, feeding from `requirements.txt`
- Make `meta.yaml` read through the `install_requires` field and populate the run requirements
- Remove `matplotlib` pin, builds were fixed to avoid the issue

## Motivation and Context

I always forget to keep `meta.yaml` and `requirements.txt` in sync. Now it's easy. The requirements used by `pip` and the requirements used by `conda` should always be exactly the same.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -17,17 +17,16 @@ requirements:
 
   run:
     - python {{PY_VER}}*,>=3
+    # Inspect requirements.txt for the reqs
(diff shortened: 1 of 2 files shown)


klauer:
**This will remain unmerged until December 7th, 2019!**

PR #328, opened 2019-04-02
MNT: use mod to wrap index

tacaswell:
Marginally safer than an if statement.

### DIFF
--- a/pcdsdevices/signal.py
+++ b/pcdsdevices/signal.py
@@ -174,7 +174,6 @@ class AvgSignal(Signal):
         with self._lock:
             self.values[self.index] = value
             self.index += 1
-            if self.index == len(self.values):
-                self.index = 0
+            self.index = (self.index + 1) % len(self.values)            
             # This takes a mean, skipping nan values.
             self.put(np.nanmean(self.values))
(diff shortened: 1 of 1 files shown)


tacaswell:
Probably a bit of both.

The sketch I wrote up in the issues tried to be fixed time windows, this is always the last N points.  Not clear which one is better....which probable means we want both!

pcdsdevices/signal.py
```diff
@@ -174,7 +174,6 @@ def _update_avg(self, *args, value, **kwargs):
         with self._lock:
             self.values[self.index] = value
             self.index += 1
```
ZLLentz: We're incrementing the index twice now, so we should delete this line.
```suggestion
```

pcdsdevices/signal.py
```diff
@@ -174,7 +174,6 @@ def _update_avg(self, *args, value, **kwargs):
...
-            if self.index == len(self.values):
-                self.index = 0
+            self.index = (self.index + 1) % len(self.values)
```
ZLLentz: Trailing whitespace:
```suggestion
            self.index = (self.index + 1) % len(self.values)
```

teddyrendahl:
@tacaswell You mind accepting @ZLLentz suggestions? Both will be required to get this to pass on our CI

tacaswell:
fair enough.

That is what I get for trying to do the PR in the github editor

PR #329, opened 2019-05-06
Add more qs devs

jsheppard95:
## Description
Added classes for new devices that can be read by happi with questionnaire backend (https://github.com/pcdshub/happi/pull/94).

Imported `PCDSDetector` in `device_types` (already defined in `areadector.detectors`).

Created new file `analog_signals` containing the class `Acromag` and started to include some mesh functions - also imported in `device_types`

Created file for testing `analog_signals` and started constructing tests - still a WIP.

### DIFF
--- a/pcdsdevices/analog_signals.py
+++ b/pcdsdevices/analog_signals.py
@@ -149,7 +149,7 @@ class Mesh(Device):
         new_hv_sp = curr_hv_sp + delta_hv_sp
         self.set_mesh_voltage(new_hv_sp, wait=wait, do_print=do_print)
 
-    def tweak_mesh_voltage(self, delta_hv_sp):
+    def tweak_mesh_voltage(self, delta_hv_sp, test_flag=False):
         """
         Increase/decrease power supply setpoint by specified amount using
         the arrow keys
@@ -160,8 +160,13 @@ class Mesh(Device):
             Amount to change voltage from its current value at each step.
             After calling with specified step size, use arrow keys to keep
             changing. Use absolute value of increment size.
+
         ^C :
             exits tweak mode
+
+        test_flag : bool, opt
+            flag used in testing functions to only run `while True` loop
+            once - i.e single tweak mode
         """
         print('Use arrow keys (left, right) to step voltage (-, +)')
         while True:
@@ -174,3 +179,5 @@ class Mesh(Device):
             elif key == key_press.arrow_left:
                 self.set_rel_mesh_voltage(-delta_hv_sp, wait=False,
                                           do_print=False)
+            if test_flag:
+                return
--- a/pcdsdevices/analog_signals.py
+++ b/pcdsdevices/analog_signals.py
@@ -1,0 +1,167 @@
+import time
+import pcdsdevices.utils as key_press
+from ophyd import Device, Component as Cpt, EpicsSignal, EpicsSignalRO
+
+
+class Acromag(Device):
+    """
+    Class for Acromag analog input/ouput signals
+
+    Parameters:
+    -----------
+    prefix : str
+        The Epics base of the acromag
+
+    name : str
+        A name to prefer to the device
+    """
+    # Components for each channel
+    ao1_0 = Cpt(EpicsSignal, ":ao1:0", kind='normal')
+    ao1_1 = Cpt(EpicsSignal, ":ao1:1", kind='normal')
+    ao1_2 = Cpt(EpicsSignal, ":ao1:2", kind='normal')
+    ao1_3 = Cpt(EpicsSignal, ":ao1:3", kind='normal')
+    ao1_4 = Cpt(EpicsSignal, ":ao1:4", kind='normal')
+    ao1_5 = Cpt(EpicsSignal, ":ao1:5", kind='normal')
+    ao1_6 = Cpt(EpicsSignal, ":ao1:6", kind='normal')
+    ao1_7 = Cpt(EpicsSignal, ":ao1:7", kind='normal')
+    ao1_8 = Cpt(EpicsSignal, ":ao1:8", kind='normal')
+    ao1_9 = Cpt(EpicsSignal, ":ao1:9", kind='normal')
+    ao1_10 = Cpt(EpicsSignal, ":ao1:10", kind='normal')
+    ao1_11 = Cpt(EpicsSignal, ":ao1:11", kind='normal')
+    ao1_12 = Cpt(EpicsSignal, ":ao1:12", kind='normal')
+    ao1_13 = Cpt(EpicsSignal, ":ao1:13", kind='normal')
+    ao1_14 = Cpt(EpicsSignal, ":ao1:14", kind='normal')
+    ao1_15 = Cpt(EpicsSignal, ":ao1:15", kind='normal')
+
+    ai1_0 = Cpt(EpicsSignalRO, ":ai1:0", kind='normal')
+    ai1_1 = Cpt(EpicsSignalRO, ":ai1:1", kind='normal')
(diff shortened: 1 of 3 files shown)


pcdsdevices/analog_signals.py
```diff
@@ -0,0 +1,167 @@
...
+    Parameters:
+    -----------
+    prefix : str
+        The Epics base of the acromag
+
+    name : str
+        A name to prefer to the device
+    """
+    # Components for each channel
+    ao1_0 = Cpt(EpicsSignal, ":ao1:0", kind='normal')
```
klauer: This is a good use case for a `DynamicDeviceSignal`:

```python
from ophyd import DynamicDeviceComponent as DDC, Device

def _make_signals(channels):
    return {
        'signal{}'.format(channel): (EpicsSignal, ':ai1:{}'.format(channel),
                                     dict(kind='normal', write_pv=':ao1:{}'.format(channel)))
        for channel in channels

    }

class Acromag(Device):
    signals = DDC(_make_signals(range(0, 16)))
```

You could then access it by:

```python
In [1]: dev = Acromag('prefix', name='dev')

In [2]: dev.signals.signal0.pvname
Out[2]: 'prefix:ai1:0'

In [3]: dev.signals.signal0.setpoint_pvname
Out[3]: 'prefix:ao1:0'

In [4]: dev.signals.component_names
Out[4]:
('signal0',
 'signal1',
 'signal2',
 'signal3',
 'signal4',
 'signal5',
 'signal6',
 'signal7',
 'signal8',
 'signal9',
 'signal10',
 'signal11',
 'signal12',
 'signal13',
 'signal14',
 'signal15')
```

pcdsdevices/analog_signals.py
```diff
@@ -0,0 +1,167 @@
...
+
+        rb_ch : int
+            Read back Acromag channel to which high voltage readback is
+            connected. Range is 0 to 15
+
+        scale : float, optional
+            Gain for high voltage supply to be controlled by the Acromag
+        """
+        self.scale = scale
+        self.rb_pv = getattr(self, 'ai1_%s' % rb_ch).pvname
```
klauer: If you use the above methodology combining the `ao/ai` into a single `EpicsSignal`, you then only have to pick one of the signals for `mesh_raw` and you can do away with the `sp_pv` and `rb_pv`.

pcdsdevices/analog_signals.py
```diff
@@ -0,0 +1,167 @@
...
+    ai1_8 = Cpt(EpicsSignalRO, ":ai1:8", kind='normal')
+    ai1_9 = Cpt(EpicsSignalRO, ":ai1:9", kind='normal')
+    ai1_10 = Cpt(EpicsSignalRO, ":ai1:10", kind='normal')
+    ai1_11 = Cpt(EpicsSignalRO, ":ai1:11", kind='normal')
+    ai1_12 = Cpt(EpicsSignalRO, ":ai1:12", kind='normal')
+    ai1_13 = Cpt(EpicsSignalRO, ":ai1:13", kind='normal')
+    ai1_14 = Cpt(EpicsSignalRO, ":ai1:14", kind='normal')
+    ai1_15 = Cpt(EpicsSignalRO, ":ai1:15", kind='normal')
+
+    def initialize_mesh(self, sp_ch, rb_ch, scale=1000.0):
```
ZLLentz: I don't think these mesh methods belong on the `Acromag` class. I think these are more suitable for a subclass of `Acromag`. A majority of `Acromag` devices do not have the mesh power supply attached to them.

tests/test_analog_signals.py
```diff
@@ -0,0 +1,18 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope='function')
+def fake_acromag():
+    FakeAcromag = make_fake_device(Acromag)
+    acromag = FakeAcromag('Test:Acromag', name='test_acromag')
+    return acromag
+
```
ZLLentz: If we have methods that do things like set voltage, tweak voltage in this PR, they should also be tested in this PR.

pcdsdevices/analog_signals.py
jsheppard95: Oh that makes sense to me. I'll move the mesh functions into their own class that inherits from Acromag, that will clean up the way I was doing things as I would no longer need the `initialize_mesh` function. I'll also try to implement @klauer 's suggestion as I think that would clean things up.

tests/test_analog_signals.py
jsheppard95: Yes definitely, I'll implement tests for the mesh functions next and update with another commit

ZLLentz:
One more thought: is having the full `Acromag` as a device what we want? We typically use the `Acromag` as generic IO and reinterpret the channels between experiments, so what we think of as a "device" is typically a grouping of 1-4 of these PVs, not the whole set.

jsheppard95:
> One more thought: is having the full `Acromag` as a device what we want? We typically use the `Acromag` as generic IO and reinterpret the channels between experiments, so what we think of as a "device" is typically a grouping of 1-4 of these PVs, not the whole set.

@ZLLentz I see what you're saying here, and by that rationale it seems like `class Mesh` makes sense as it is a device made up of Acromag signals while the Acromag itself is generic IO. Really the reason I added `class Acromag` was so that I could import it in `pcdsdevices.device_types` which is necessary for happi's questionnaire backend to read these signals from the questionnaire. Maybe there is another way around that, but currently my approach has been to include `class Acromag` with just its component definitions while `class Mesh` is a standalone class that includes basic mesh functions.

pcdsdevices/analog_signals.py
```diff
@@ -50,45 +51,53 @@ class Acromag(Device):
...
+        Read back Acromag channel to which high voltage readback is
+        connected. Range is 0 to 15
+
+    scale : float, optional
+        Gain for high voltage supply to be controlled by the Acromag
+
+    """
+
+    write_cpt = FCpt(EpicsSignal, '{self.prefix}' + ':ao1:' + '{self.sp_ch}')
+    read_cpt = FCpt(EpicsSignalRO, '{self.prefix}' + ':ai1:' + '{self.rb_ch}')
```
ZLLentz: Super super small nitpick, but note that these become signals after instantiation, so calling them `cpt` here instead of `sig` or `ch` seems inaccurate

jsheppard95: That makes sense to me, just made that change.

jsheppard95:
The codecov check is failing due to not having wide enough coverage in my tests, but the only function I added left to test is `Mesh.tweak_mesh_voltage`. I'm currently unsure how to go about testing this as it requires user input, but I'm looking into this as part of Teddy's automated device commissioning/checkout effort so as I learn about it I will try to implement here as well.

pcdsdevices/analog_signals.py
teddyrendahl: I like this, but it isn't quite correct. The in and the out signals are not correlated (i.e they should not be in the same `EpicsSignal` as read and write pairs).

pcdsdevices/analog_signals.py
```diff
@@ -0,0 +1,176 @@
...
+        wait : bool, optional
+            Indicates whether or not the program should pause when writing
+            to a PV
+
+        do_print : bool, optional
+            Indicates whether or not the program should print it's
+            setpoint and readback values
+        """
+        if do_print:
+            print('Setting mesh voltage...')
```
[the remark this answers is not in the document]
jsheppard95: In 4a8618e, I changed each of the `print` statements to be `logger` calls which are shown if the users passes `verbose=True` when instantiating the `Mesh` class. This appears to work as expected, but I may be off with whether I should be using `logger.info` vs. `logger.debug`, etc.

PR #330, opened 2019-05-07
API: rename PCDSDetector to PCDSAreaDetector

ZLLentz:
## Description

Use sed to replace `PCDSDetector` with `PCDSAreaDetector`

## Motivation and Context

Requested by @silkenelson to avoid confusion with other pcds-group detectors that are not epics area detectors.

## How Has This Been Tested?

It might pass the tests

It will need to end up in the release notes under API breaks.

Note that this probably needs corresponding changes in some other libraries...


ZLLentz:
I have found one library that was using `PCDSDetector`: `hutch-python`, so maybe I'll go around and change that later.

PR #331, opened 2019-05-08
ENH: make states_list optional

ZLLentz:
## Description

If `states_list` isn't provided in a states device, use the metadata callback to get it.
Cleans up some device definitions to be simpler and more logically consistent.

## Motivation and Context

Silke asked for not having to make a separate timetool class for each hutch. I want to use this to clean up our device implementations and make it easier going forward.

## How Has This Been Tested?

One test was added

Docstrings and comments

### DIFF
--- a/pcdsdevices/state.py
+++ b/pcdsdevices/state.py
@@ -62,6 +62,7 @@ class StatePositioner(Device, PositionerBase, MvInterface):
 
     SUB_STATE = 'state'
     _default_sub = SUB_STATE
+    _state_meta_sub = EpicsSignal.SUB_VALUE
 
     egu = 'state'
 
@@ -76,7 +77,7 @@ class StatePositioner(Device, PositionerBase, MvInterface):
             self._state_init()
         else:
             cbid = self.state.subscribe(self._late_state_init,
-                                        event_type=EpicsSignal.SUB_VALUE,
+                                        event_type=self._state_meta_sub,
                                         run=False)
             self._state_init_cbid = cbid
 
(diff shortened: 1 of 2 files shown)


pcdsdevices/state.py
```diff
@@ -69,15 +71,40 @@ def __init__(self, prefix, *, name, **kwargs):
...
+        self._state_init_cbid = False
+        if self.states_list:
+            self._state_init()
+        else:
+            cbid = self.state.subscribe(self._late_state_init,
+                                        event_type=self._state_meta_sub,
+                                        run=False)
+            self._state_init_cbid = cbid
+
+    # @required_for_connection # This never made it into ophyd...
```
klauer: It's there - just not yet in a release: https://github.com/bluesky/ophyd/blob/be880e1bcc87c05668bf300845e32416537e647b/ophyd/device.py#L1549 ☹️

ZLLentz: Yeah, that's what I meant. I wrote this thing out assuming things like `required_for_connection` and `SUB_META` were available and then I had to redo it.

That being said, this PR will be much easier to review after the other PR goes in because the other PR accidentally fixed a problem that made this one possible.

ZLLentz: Almost a year later, I have uncommented this as it is in the upcoming 1.4.0 and we're not waiting for it to come out of rcs

ZLLentz:
I'm going to merge this if the tests pass because it's marginally useful and it doesn't seem to break anything. The idea needs more development though, stuff like automatically picking in/out states is one thing that should be implemented if possible and I'm sure there's more. There also should be a follow-up where we refactor some ugliness based on hard-coding states lists for things like the beryllium lens classes.

PR #332, opened 2019-05-13
Vacuum lcls i

silkenelson:
Adding classes for (ion) pumps and vacuum gauges
## Description
new files for (ion) pumps and vacuum gauges have been created w/ tests
shared doc strings have been added to doc__stubs

## Motivation and Context
the old hutch python had classes for these devices and I would like to replicate coverage
additionally, these classes show allow us to inspect the vacuum along a beam line from the command line.

## How Has This Been Tested?
I have checked more complicated functions on a few vacuum devices in the XCS line.

The classes have docstrings.

### DIFF
--- a/pcdsdevices/gauge.py
+++ b/pcdsdevices/gauge.py
@@ -1,0 +1,64 @@
+"""
+Standard classes for LCLS Gauges
+"""
+import logging
+from ophyd import EpicsSignal, EpicsSignalRO, Device, Component as Cpt
+
+logger = logging.getLogger(__name__)
+
+class BaseGauge(Device):
+    """
+    Vacuum gauge
+
+    A base class for a device with two limits switches controlled via an
+    external command PV. This fully encompasses the controls `Stopper`
+    installations as well as un-interlocked `GateValves`
+
+    Parameters
+    ----------
+    prefix : ``str``
+        Full Gauge base PV
+
+    name : ``str``
+        Alias for the gauge
+    """
+    # 
+    pressure = Cpt(EpicsSignalRO, ':PMON', kind='normal')
+    status = Cpt(EpicsSignalRO, ':STATUSMON', kind='normal')
+    pressure_status = Cpt(EpicsSignalRO, ':PSTATMON', kind='normal')
+    pressure_status_enable = Cpt(EpicsSignal, ':PSTATMSP', kind='normal')
+
+    def __repr__(self):
+        print('Gauge is %s'%self.status)
+        if self.status>0:
+            print('Pressure: %g'%self.pressure)
+
+class GaugePirani(BaseGauge):
+    """
+    Class for Pirani gauge
+    """
+    pass
+
+class GaugeColdCathode(BaseGauge):
+    """
+    Class for Cold Cathode Gauge
+    """
+    enable = Cpt(EpicsSignalRO, ':ENB_DO_SW', kind='normal')
+    #this should be combined.
+    tripSetpointRB = Cpt(EpicsSignal, ':PSTATSPDES', kind='normal')
+    tripSetpoint = Cpt(EpicsSignalRO, ':PSTATSPRBCK', kind='normal')
+
+class GaugeSet(Device):
+    """
+    Class for a gauge set with a Pirani and Cold Cathode Gauge
+    argument is the base PV for the cold cathode gauge
+    """
+    gpi = Cpt(GaugePirani, self.prefix.replace('GCC','GPI'))
+    gcc = Cpt(GaugeColdCathode)
+    
+    def pressure(self):
+        if self.gcc.status > 0:
+            return self.gcc.pmon
+        else:
+            return self.gpi.pmon
+
--- a/pcdsdevices/gauge.py
+++ b/pcdsdevices/gauge.py
@@ -2,9 +2,39 @@
 Standard classes for LCLS Gauges
 """
 import logging
-from ophyd import EpicsSignal, EpicsSignalRO, Device, Component as Cpt
+from ophyd import EpicsSignal, EpicsSignalRO, Device
+from ophyd import Component as Cpt, FormattedComponent as FCpt
+from .doc_stubs import GaugeSet_base
 
 logger = logging.getLogger(__name__)
+
+
+class MKS937a(Device):
+    """
+    Vacuum gauge controller MKS637a)
+
+    A base class for an MKS637a controller
+
+    Parameters
+    ----------
+    prefix : ``str``
+        Full Gauge controller base PV
+
+    name : ``str``
+        Alias for the gauge controller
+    """
+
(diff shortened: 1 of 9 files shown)


pcdsdevices/gauge.py
```diff
@@ -0,0 +1,64 @@
...
+        Alias for the gauge
+    """
+    # 
+    pressure = Cpt(EpicsSignalRO, ':PMON', kind='normal')
+    status = Cpt(EpicsSignalRO, ':STATUSMON', kind='normal')
+    pressure_status = Cpt(EpicsSignalRO, ':PSTATMON', kind='normal')
+    pressure_status_enable = Cpt(EpicsSignal, ':PSTATMSP', kind='normal')
+
+    def __repr__(self):
+        print('Gauge is %s'%self.status)
```
teddyrendahl: A few notes here:

1. Just `self.status` and `self.pressure` will give you the `repr` of the signals. If you want the value you have to call `get`
2. It is worth having a `try / except` pattern here. If the object is disconnected not being able to represent it can be an issue
3. `repr` should return a string not `print`

pcdsdevices/gauge.py
```diff
@@ -0,0 +1,64 @@
...
+    #this should be combined.
+    tripSetpointRB = Cpt(EpicsSignal, ':PSTATSPDES', kind='normal')
+    tripSetpoint = Cpt(EpicsSignalRO, ':PSTATSPRBCK', kind='normal')
+
+class GaugeSet(Device):
+    """
+    Class for a gauge set with a Pirani and Cold Cathode Gauge
+    argument is the base PV for the cold cathode gauge
+    """
+    gpi = Cpt(GaugePirani, self.prefix.replace('GCC','GPI'))
```
teddyrendahl: You can't use `.self` calls in the class creation, you need to put this in the `__init__`. Ophyd has a way to do this with a subclass of `Component` called `FormattedComponent`. Here is an example that should map nicely to the issue here

https://github.com/pcdshub/pcdsdevices/blob/5925ccd77bcb1fc064ad51deb85c640a459feca3/pcdsdevices/mirror.py#L93

pcdsdevices/gauge.py
```diff
@@ -0,0 +1,64 @@
...
+    Class for Cold Cathode Gauge
+    """
+    enable = Cpt(EpicsSignalRO, ':ENB_DO_SW', kind='normal')
+    #this should be combined.
+    tripSetpointRB = Cpt(EpicsSignal, ':PSTATSPDES', kind='normal')
+    tripSetpoint = Cpt(EpicsSignalRO, ':PSTATSPRBCK', kind='normal')
+
+class GaugeSet(Device):
+    """
+    Class for a gauge set with a Pirani and Cold Cathode Gauge
```
teddyrendahl: Do you think replacing the Cold Cathode component with the Pirani is the cleanest? Why not have the prefix be the shared prefix and add them both? You should still be able to use the info contained in the comment below.

pcdsdevices/gauge.py
```diff
@@ -0,0 +1,64 @@
...
+    argument is the base PV for the cold cathode gauge
+    """
+    gpi = Cpt(GaugePirani, self.prefix.replace('GCC','GPI'))
+    gcc = Cpt(GaugeColdCathode)
+    
+    def pressure(self):
```
teddyrendahl: Probably want a docstring here to explain how we get this value?

pcdsdevices/pump.py
```diff
@@ -0,0 +1,76 @@
...
+    aomode = Cpt(EpicsSignal, ':AOMODEDES', write_pv=':AOMODE', kind='config')
+    calfactor = Cpt(EpicsSignal, ':CALFACTOREDES', write_pv=':CALFACTOR', kind='config')
+
+    def on(self):
+        self.state.put(1)
+
+    def off(self):
+        self.state.put(0)
+
+    def __repr__(self):
```
teddyrendahl: See above comments about `repr`

tests/test_gauge.py
```diff
@@ -0,0 +1,40 @@
...
+@pytest.fixture(scope='function')
+def fake_gauge_set():
+    FakeGaugeSet = make_fake_device(GaugeSetPirani)
+    gs = FakeGaugeSet('Test:Gauge', name='test', index='99')
+    gs.gcc.state.sim_put(1)
+    gs.gcc.pressure.sim_put(99)
+    gs.gpi.pressure.sim_put(98)
+    return gs
+
+@pytest.mark.timeout(5)
```
teddyrendahl: Don't think you should need to use this here.

tests/test_gauge.py
```diff
@@ -0,0 +1,40 @@
...
+    gs = fake_gauge_set
+    # Should return to original position on unstage
+    assert gs.pressure()==98
+    gs.gcc.state.sim_put(0)
+    assert gs.pressure()==99
+
+def test_gauge_factory():
+    m = GaugeSet('TST:MY', name='test_gauge', index='99')
+    assert isinstance(m, GaugeSetPirani)
+    m = GaugeSet('TST:MY', name='test_gauge', index='99', onlyGCC=1)
```
teddyrendahl: This looks like this test was written for a different version of your code?

pcdsdevices/pump.py
```diff
@@ -0,0 +1,145 @@
...
+        self.prefix_controller = prefix_controller
+        # Load Ion Pump itself
+        super().__init__(prefix, **kwargs)
+
+    def egu(self):
+        return self.controller.unit.get()
+
+
+# factory function for IonPumps
+def IonPump(prefix, *, name, **kwargs):
```
teddyrendahl: Add a docstring here for factory function

pcdsdevices/gauge.py
```diff
@@ -0,0 +1,184 @@
...
+    def __init__(self, prefix, *, name, index, prefix_controller,  **kwargs):
+        self.prefix_controller = prefix_controller
+        super().__init__(prefix, name=name, index=index, **kwargs)
+
+    def egu(self):
+        return self.controller.unit.get()
+
+
+# factory function for IonPumps
+def GaugeSet(prefix, *, name, index, **kwargs):
```
teddyrendahl: Docstring here

tests/test_gauge.py
```diff
@@ -0,0 +1,39 @@
...
+    gs = FakeGaugeSet('Test:Gauge', name='test', index='99')
+    gs.gcc.state.sim_put(1)
+    gs.gcc.pressure.sim_put(99)
+    gs.gpi.pressure.sim_put(98)
+    return gs
+
+def test_gauge_pressure(fake_gauge_set):
+    logger.debug('test_gauge_pressure')
+    gs = fake_gauge_set
+    # Should return to original position on unstage
```
teddyrendahl: Leftover comment?

PR #333, opened 2019-05-14
ENH: Add motor to state devices

### LINKED ISSUE #177: Fill in Motor Subdevices
## Expected Behavior

Positioner subdevices should be included in classes like `PIM` whenever we know what they are. This gives us direct control over the motor in case we need it without having to instantiate a separate device.

ZLLentz:
## Description

As of the newest `IMS` release I did, all state device records contain a convenient `:MOTOR` alias to the motor whose position is associated with the state. This PR replaces the previous janky `readback` signal with this `motor` subcomponent.

This PR adjusts the `IPM` and `LODCM` classes as well, because their previous structures were an unstable abuse of inheritance and doing this change broke the classes entirely. The previous structure was a state positioner that contained other state positioners, but this is not how the PVs are structured at all, giving an awkward class definition. These are now container devices that, while not positioners themselves, contain many subcomponent state positioners.

## Motivation and Context

This was the cleanest way to allow us to have state devices with direct access to the motor.
closes #177 

## How Has This Been Tested?

This worked on real IPMs that Silke has updated to the newest IMS release. Note that this will not work on devices without the newest IOC release, we need to update these as we find them.

Docstrings updated appropriately. Apologies for overtly breaking API.

### DIFF
--- a/pcdsdevices/lodcm.py
+++ b/pcdsdevices/lodcm.py
@@ -12,11 +12,10 @@ import logging
 import functools
 
 from ophyd.device import Device, Component as Cpt
-from ophyd.signal import EpicsSignal, EpicsSignalRO
 from ophyd.sim import NullStatus
 from ophyd.status import wait as status_wait
 
-from .doc_stubs import basic_positioner_init, insert_remove
+from .doc_stubs import insert_remove
 from .inout import InOutRecordPositioner
 
 logger = logging.getLogger(__name__)
--- a/pcdsdevices/lodcm.py
+++ b/pcdsdevices/lodcm.py
@@ -11,7 +11,7 @@ no readback into the device's alignment. This is intended for a future update.
 import logging
 import functools
 
-from ophyd import Component as Cpt, FormattedComponent as FCpt
+from ophyd.device import Device, Component as Cpt
 from ophyd.signal import EpicsSignal, EpicsSignalRO
 from ophyd.sim import NullStatus
 from ophyd.status import wait as status_wait
@@ -20,6 +20,13 @@ from .doc_stubs import basic_positioner_init, insert_remove
 from .inout import InOutRecordPositioner
 
 logger = logging.getLogger(__name__)
+
+
+class H1N(InOutRecordPositioner):
+    states_list = ['OUT', 'C', 'Si']
+    in_states = ['C', 'Si']
+    _states_alias = {'C': 'IN'}
+    _transmission = {'C': 0.8, 'Si': 0.7}
 
 
 class YagLom(InOutRecordPositioner):
@@ -52,7 +59,7 @@ class Foil(InOutRecordPositioner):
         super().__init__(prefix, *args, **kwargs)
 
 
-class LODCM(InOutRecordPositioner):
+class LODCM(Device):
     """
     Large Offset Dual Crystal Monochromator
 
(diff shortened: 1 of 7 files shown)


pcdsdevices/lodcm.py
```diff
@@ -62,31 +68,27 @@ class LODCM(InOutRecordPositioner):
...
     mono_line: ``str``, optional
         Name of the mono, double-bounce beamline.
     """
-    __doc__ = __doc__ % basic_positioner_init
-
-    state = Cpt(EpicsSignal, ':H1N', write_pv=':H1N:GO', kind='hinted')
-    readback = FCpt(EpicsSignalRO, '{self.prefix}:H1N:{self._readback}',
-                    kind='normal')
-
+    h1n = Cpt(H1N, ':H1N', kind='hinted')
```
teddyrendahl: I always forget the rules... a hint of my hinted component is my hint?

pcdsdevices/lodcm.py
```diff
@@ -11,17 +11,23 @@
...
 from .inout import InOutRecordPositioner
 
 logger = logging.getLogger(__name__)
 
 
+class H1N(InOutRecordPositioner):
+    states_list = ['OUT', 'C', 'Si']
+    in_states = ['C', 'Si']
+    _states_alias = {'C': 'IN'}
+    _transmission = {'C': 0.8, 'Si': 0.7}
```
teddyrendahl: The more I've seen these transmissions the more I dislike them. Think it might be worth just going to a binary of `transmissive` or not...

klauer: Based on the comment below:
```
    # TBH these are guessed. Please replace if you know better. These don't
    # need to be 100% accurate, but they should reflect a reasonable reduction
    # in transmission.
```
I think @teddyrendahl may have something with the change to a boolean.

pcdsdevices/state.py
```diff
@@ -392,12 +393,9 @@ class StateRecordPositioner(StatePositioner):
     settings, in the same order as the state enum. Unknown must be omitted.
     """
     state = Cpt(EpicsSignal, '', write_pv=':GO', kind='hinted')
-    readback = FCpt(EpicsSignalRO, '{self.prefix}:{self._readback}',
-                    kind='normal')
+    motor = Cpt(IMS, ':MOTOR', kind='normal')
```
klauer: 👏 for aliases!

pcdsdevices/state.py
```diff
@@ -406,7 +404,8 @@ def subscribe(self, cb, event_type=None, run=True):
         cid = super().subscribe(cb, event_type=event_type, run=run)
         if (event_type == self.SUB_READBACK and not
                 self._has_subscribed_readback):
-            self.readback.subscribe(self._run_sub_readback, run=False)
+            self.motor.user_readback.subscribe(self._run_sub_readback,
```
teddyrendahl: I could see the benefit of a more targeted error message suggesting the possibility of an IOC upgrade if this times out. On the other hand, lots of things could cause `TimeoutError`s...

pcdsdevices/lodcm.py
ZLLentz: I think all the subhints end up bubbling up somehow... I can't remember either.

pcdsdevices/lodcm.py
ZLLentz: Probably correct, for devices that are > 80% transmissive it isn't a relevant quantity. In the past this has mattered a lot for random devices that I didn't expect due to absorption edges, but that isn't even covered by ratios like these.

pcdsdevices/state.py
ZLLentz: I like this idea for our transition phase where we haven't updated everything. If the motor times out but nothing else does, then surely that means we didn't update.

ZLLentz:
Planned changes:
- simple warning for version upgrade if it isn't too hard to implement
- make the transmission binary

One more thought: we might have different transmissions down each exit of the LODCM due to diagnostics being in for the MONO line and nothing blocking the MAIN line. I think this isn't taken care of in the transmission/destinations API we set up for lightpath.

Some of the LODCM-specific stuff may belong in a follow-up PR because the goal here is simply to add the motor.

teddyrendahl:
> Some of the LODCM-specific stuff may belong in a follow-up PR because the goal here is simply to add the motor.

Agreed.

> make the transmission binary

Think this is out of scope as well. Will require changes to `lightpath`.

ZLLentz:
I'm going to merge this as-is for the tag I'm about to do and add issues for follow-up.

PR #334, opened 2019-05-21
CI: require 3.7 build passing

ZLLentz:
## Description

See title

## Motivation and Context

All the dependencies are available now, so we should be preparing for the long-delayed migration to 3.7 as the standard version


klauer:
Yes!

PR #341, opened 2019-10-23
ENH: Tab Completion Whitelist

ZLLentz:
## Description

Provide a curated `__dir__` method for users to tab through in `ipython`. This involves whitelists set up on a class-by-class basis to designate which attributes should be available in the basic view, as well a switch to change between the basic and engineering views.

## Motivation and Context

Tabbing through an ophyd object is extremely verbose to the point of being unhelpful.

## How Has This Been Tested?

Basic tests to make sure nothing broke and things are being whitelisted, plus feel tests to make sure it seems like it worked. This will inevitably need to be revisited because I refuse to write tests that specify which elements should be whitelisted or not.

Added to the docs pages


pcdsdevices/mv_interface.py
```diff
@@ -19,9 +20,89 @@
...
+                   parent class
+    """
+    tab_blacklist = ["OphydAttrList", "SUB_*", "_*", "actuate",
+                     "actuate_value", "clear_sub", "done", "done_value",
+                     "event_types", "get_device_tuple",
+                     "get_instantiated_signals", "hints", "kind", "root",
+                     "set", "setpoint", "stop_signal", "stop_value",
+                     "subscribe", "subscriptions", "trigger",
+                     "trigger_signals", "unsubscribe", "unsubscribe_all"]
+    tab_whitelist = ["__*__"]
```
klauer:
```suggestion
    tab_whitelist = ["__.*__"]
```

pcdsdevices/mv_interface.py
```diff
@@ -19,9 +20,89 @@
...
+
+    This class defines an API and some defaults for blacklisting tab-completion
+    results for new users to avoid confusion. The API involves setting the
+    following class variables on any subclasses:
+
+    tab_blacklist: list of string regex to omit from tab complete
+    tab_whitelist: list of string regex to include even if blacklisted by a
+                   parent class
+    """
+    tab_blacklist = ["OphydAttrList", "SUB_*", "_*", "actuate",
```
klauer:
```suggestion
    tab_blacklist = ["OphydAttrList", "SUB_.*", "_.*", "actuate",
```

ZLLentz:
I see, I didn't consider programatically combining the regex into one. Definitely going to do that.

pcdsdevices/mv_interface.py
ZLLentz: I'm actually going to remove this entirely because I don't think the base class should have a whitelist, but thanks for reminding me that I have to do the regexes properly

pcdsdevices/mv_interface.py
ZLLentz: I'm going to drop the leading underscore blacklist because ipython automatically hides these unless you insist on doing `device._` then tab

ZLLentz:
This is actually a pretty ok starting point for this feature now. I'm going to make this a normal PR instead of a draft and fill out the description properly.

pcdsdevices/mv_interface.py
```diff
@@ -19,9 +20,81 @@
...
+    """
+    Interface layer to attach to any Device for SLAC features.
+
+    This class defines an API and some defaults for filtering tab-completion
+    results for new users to avoid confusion. The API involves setting the
+    tab_whitelist attribute on any subclass of BaseInterface. When in
+    non-engineering mode, only elements on the whitelists will be displayed to
+    the user.
+
+    tab_whitelist: list of string regex to show in autocomplete for
```
klauer:
```suggestion
    Attributes
    ----------
    tab_whitelist: list
        list of string regex to show in autocomplete for
```

ZLLentz:
I'm doing a full once-over in the library to catch all the obvious places where I need to apply the interface. I'm also going to add a test and a doc entry before merging.

pcdsdevices/interface.py
```diff
@@ -0,0 +1,893 @@
...
+from bluesky.utils import ProgressBar
+from ophyd.device import Kind
+from ophyd.status import wait as status_wait
+
+from . import utils as util
+
+logger = logging.getLogger(__name__)
+engineering_mode = True
+
+OphydObject_whitelist = ["name", "connected", "check_value", "log"]
```
klauer: Completely unnecessary to refactor, but a thought: `whitelist_classes = {OphydObj: ['whitelist', ], Device: [...]}` - a dictionary of class to whitelist, that could easily be summed below.

pcdsdevices/interface.py
```diff
@@ -0,0 +1,893 @@
...
+    Attributes
+    ----------
+    tab_whitelist: list
+        list of string regex to show in autocomplete for non-engineering mode
+    """
+    tab_whitelist = (OphydObject_whitelist + BlueskyInterface_whitelist +
+                     Device_whitelist + Signal_whitelist +
+                     Positioner_whitelist)
+    _filtered_dir_cache = None
+
```
klauer: I haven't given this as much thought as you, but should maybe `tab_component_names` be `True` by default?

pcdsdevices/interface.py
```diff
@@ -0,0 +1,893 @@
...
+    def __init_subclass__(self):
+        string_whitelist = []
```
klauer: Might consider using a `set` here, though the performance implications are assuredly low even with duplicated `string_whitelist` entries

pcdsdevices/interface.py
```diff
@@ -0,0 +1,893 @@
...
+            return super().__dir__()
+        elif self._filtered_dir_cache is None:
+            self._init_filtered_dir_cache()
+        return self._filtered_dir_cache
+
+    def _init_filtered_dir_cache(self):
+        self._filtered_dir_cache = self._get_filtered_tab_dir()
+
+    def _get_filtered_tab_dir(self):
+        filtered = []
```
klauer: Unnecessary to refactor, but a list comprehension here would be clean + compact:
```python
return [elem
        for elem in super().__dir__()
        if self._tab_regex.fullmatch(elem)]
```

pcdsdevices/pump.py
```diff
@@ -5,17 +5,20 @@
...
 
-class TurboPump(Device):
+class TurboPump(Device, BaseInterface):
     """
     Vacuum Pump
     """
     atspeed = Cpt(EpicsSignal, ':ATSPEED_DI', kind='normal')
     start = Cpt(EpicsSignal, ':START_SW', kind='normal')
 
+    tab_whitelist = ['run', 'stop', 'atspeed']
```
klauer: This is a good example where hiding `start` saves a lot of potential confusion.

pcdsdevices/lens.py
```diff
@@ -41,6 +41,8 @@ class LensStack(PseudoPositioner):
 
     calib_z = Cpt(PseudoSingle)
 
+    tab_whitelist = ['x', 'y', 'z', 'calib_z', 'tweak', 'align']
```
klauer:
```suggestion
    tab_whitelist = ['tweak', 'align']
    tab_component_names = True
```

pcdsdevices/interface.py
ZLLentz: Yes, this probably would have been a cleaner way to do it

pcdsdevices/interface.py
ZLLentz: Maybe. I was seeing long lists of "omitted" components so I defaulted to `False`, but then later I switched the `tab_component_names` to ignore `omitted` components without revisiting the default.

ZLLentz: The other thing to consider is that, with the nature of whitelisting, the package is more consistent if showing anything on tab defaults to `False`.

pcdsdevices/interface.py
ZLLentz: Yeah, I don't see much of an impact either way. Is there a quick one-liner to add all elements from a list into a set?

pcdsdevices/interface.py
ZLLentz: `Unnecessary to refactor, but here's the code for you already refactored`

ZLLentz: In general I shy away from multi-line list comprehensions because they can quickly get ugly but this one seems pretty clean

pcdsdevices/lens.py
ZLLentz: For some reason I thought this would collect component names from parent classes, but it doesn't.

ZLLentz:
RE: decorator: the idea occurred to me while making this but I didn't implement it. I think it's a good idea but it wasn't easier for retroactively applying rules to classes everywhere in the repo.

pcdsdevices/interface.py
klauer: True!

pcdsdevices/interface.py
klauer: `set(iterable)`

pcdsdevices/lens.py
klauer: Huh, ok...

PR #342, opened 2019-10-23
FIX: Look for end transition instead of start transition

ZLLentz:
## Description

The event sequencer's `trigger` method will now correctly look for the sequence done transition instead of the sequence start transition

## Motivation and Context

We are not ready to read our detectors until the sequence is done, so the trigger should not return done until after the sequence

### DIFF
--- a/pcdsdevices/sequencer.py
+++ b/pcdsdevices/sequencer.py
@@ -254,7 +254,7 @@ class EventSequencer(Device, MonitorFlyerMixin, FlyerInterface):
 
         # Create our status
         def done(*args, value=None, old_value=None, **kwargs):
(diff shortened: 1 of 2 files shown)


ZLLentz:
```
$ caget -d 31 ECS:SYS0:6:PLSTAT
ECS:SYS0:6:PLSTAT
    Native data type: DBF_ENUM
    Request type:     DBR_CTRL_ENUM
    Element count:    1
    Value:            Stopped
    Status:           STATE
    Severity:         MAJOR
    Enums:            ( 3)
                      [ 0] Stopped
                      [ 1] Waiting
                      [ 2] Playing
```

PR #348, opened 2019-12-12
MNT: Clean Error Message for #347

### LINKED ISSUE #347: pcdsdevices/signal.py import can interfere with normal signal module
## Expected Behavior

Normal imports should be possible when running code while inside `pcdsdevices` folder to avoid confusion and lost time from pcds engineers.

## Current Behavior

Any code that imports the `signal` module (or that imports something that imports the `signal` module) breaks horribly and incomprehensibly if the current working directory is inside the `pcdsdevices` folder, and sometimes if the file being run is inside that folder (e.g. as a quick test script)

## Possible Solution

Rename the file at our next major version bump?
Or maybe add `from signal import *` inside `pcdsdevices/signal.py`?
Both of those sound bad...

1. cd `pcdsdevices/pcdsdevices`
2. Open the python interpreter
3. run python code that relies on the signal module, for example `import asyncio`

## Context

@ZryletTC came me today with a few incomprehensible python bugs and they ended up being this or various forms of this after we spent some time looking at it. This has also come up in the past and I expect it to come up again in the future.

ZLLentz:
## Description

Add a clean error message instead of the ugly one that usually happens.

## Motivation and Context

closes #347 

## How Has This Been Tested?

Hopefully the tests still pass, but I've interactively seen that this is enough to make it obvious what went wrong.

### DIFF
--- a/pcdsdevices/signal.py
+++ b/pcdsdevices/signal.py
@@ -1,6 +1,14 @@
 """
 Module to define ophyd Signal subclass utilities.
 """
+# Catch semi-frequent issue with scripts accidentally run from inside module
+if __name__ != 'pcdsdevices.signal':
+    raise RuntimeError(('A script tried to import pcdsdevices.signal '
+                        'instead of the signal built-in module. This '
+                        'usually happens when a script is run from '
+                        'inside the pcdsdevices directory and can cause '
+                        'extremely confusing bugs. Please run your script '
+                        'elsewhere for better results.'))
 import logging
 from threading import RLock, Thread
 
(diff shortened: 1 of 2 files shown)


pcdsdevices/signal.py
```diff
@@ -1,6 +1,14 @@
 """
 Module to define ophyd Signal subclass utilities.
 """
+# Catch semi-frequent issue with scripts accidentally run from inside module
+if __name__ != 'pcdsdevices.signal':
+    raise RuntimeError(('A script tried to import pcdsdevices.signal '
```
klauer: Double parentheses are unnecessary, FYI.

ZLLentz: My Python-fu is degrading. I'll fix by adding triple parentheses.

klauer: Quadruple or octuple would be preferable 🤓

PR #349, opened 2019-12-13
CI: Cleanup and BLD: noarch

ZLLentz:
## Description

I want to make one package noarch and py3.8 each day until we catch em all
I put the style checks and the docs build into their own jobs, and I also cleaned up the CI a bit, unfortunately working on a parallel path from @klauer and @n-wbrown on the ci-helpers stuff. Possibly going to contribute to that effort and then redirect this code to call stuff from there.

## Motivation and Context

noarch is more efficient and portable
doing docs build and style checks in their own jobs lets us easily see which is failing

## How Has This Been Tested?

It is tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,10 +39,12 @@ install:
   # Test conda build and create test environment
   - |
     if [[ $UNIT_TEST || $BUILD_DOCS ]]; then
+      echo "Building full environment"
       conda build -q conda-recipe --python $TRAVIS_PYTHON_VERSION --output-folder bld-dir
       conda config --add channels "file://`pwd`/bld-dir"
       conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION pcdsdevices --file dev-requirements.txt
     elif [[ $STYLE ]]; then
+      echo "Building minimal flake8 environment"
       conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION flake8
     fi
   # Launch Conda environment
@@ -50,6 +52,7 @@ install:
   # Add the docs packages
   - |
     if [[ $BUILD_DOCS ]]; then
+      echo "Adding docs requirements"
       conda install --file docs-requirements.txt
     fi
 
@@ -57,18 +60,21 @@ script:
   # Run tests
   - |
     if [[ $UNIT_TEST ]]; then
+      echo "Running tests"
       coverage run run_tests.py
       coverage report -m
     fi
   # Check style
   - |
     if [[ $STYLE ]]; then
+      echo "Checking style"
       flake8 pcdsdevices/*.py
       flake8 tests/*.py
     fi
   # Build docs
   - |
     if [[ $BUILD_DOCS ]]; then
+      echo "Building docs"
       set -e
       conda install --file docs-requirements.txt
       pushd docs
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -25,41 +43,74 @@ install:
...
+  # Test conda build and create test environment
+  - |
+    if [[ $UNIT_TEST || $BUILD_DOCS ]]; then
+      echo "Building full environment"
+      conda build -q conda-recipe --python $TRAVIS_PYTHON_VERSION --output-folder bld-dir
+      conda config --add channels "file://`pwd`/bld-dir"
+      conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION pcdsdevices --file dev-requirements.txt
+    elif [[ $STYLE ]]; then
+      echo "Building minimal flake8 environment"
+      conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION flake8
```
klauer: I think we can assume python + pip will be available in all of our travis matrices, so using conda isn't even required.

.travis.yml
```diff
@@ -25,41 +43,74 @@ install:
...
+  # Run tests
+  - |
+    if [[ $UNIT_TEST ]]; then
+      echo "Running tests"
+      coverage run run_tests.py
+      coverage report -m
+    fi
+  # Check style
+  - |
+    if [[ $STYLE ]]; then
```
klauer: I like the `$STYLE` variable for its simplicity but think it may fall short of what we'd want when breaking it out into our pcds-ci-helpers. That's why I went with the following pattern: https://github.com/pcdshub/pcds-ci-helpers/blob/master/travis/python_linter.sh

`init.sh` calls that script, so a copy/pasted `git clone`/`source init.sh` + a build matrix entry is sufficient to take care of a few scenarios.

.travis.yml
```diff
@@ -25,41 +43,74 @@ install:
...
+    if [[ $UPLOAD ]]; then
+      echo "Uploading to codecov"
+      codecov
+    fi
+  - |
+    if [[ $UPLOAD && $TRAVIS_PULL_REQUEST == false && $TRAVIS_REPO_SLUG == $OFFICIAL_REPO && $TRAVIS_BRANCH == $TRAVIS_TAG  && $TRAVIS_TAG != '' && $CONDA_UPLOAD_TOKEN_TAG != '' ]]; then
+      echo "Uploading to pcds_tag channel"
       export ANACONDA_API_TOKEN=$CONDA_UPLOAD_TOKEN_TAG
-      anaconda upload bld-dir/linux-64/*.tar.bz2
+      anaconda upload bld-dir/noarch/*.tar.bz2
```
klauer: 😄 :+1:

.travis.yml
ZLLentz: I agree, when I read through your PR I realized that I had forgotten that pip exists.

.travis.yml
ZLLentz: Yeah, this seems correct too. So for this repo `$LINT_PYTHON` would be set to `pcdsdevices/*.py tests/*.py`

ZLLentz: Or just set it to `pcdsdevices tests` and set the `flake8` ignores config properly

.travis.yml
ZLLentz: If we're going to do this in a uniform way in the ci-helpers lib, should this line look like:
`anaconda upload bld-dir/*/*.tar.bz2`?

ZLLentz:
I'm likely going to merge this and backport any good ideas into the ci helpers lib, then revisit this repo and replace it with ci helpers calls later

klauer:
Seems like a reasonable approach, @ZLLentz.

PR #351, opened 2020-01-07
CI: Fix cases where failed tests marked as passing

ZLLentz:
## Description

Make sure the tests are the last command run in the `if` blocks

## Motivation and Context

Only the last command in the `if` block has a return code that is elevated to the main test script. This means in the previous setup we could fail the run_tests and succeed on coverage report and the whole thing would be marked as good.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,6 +69,9 @@ script:
     if [[ $UNIT_TEST ]]; then
       echo "Running tests"
       coverage run run_tests.py
+    fi
+  - |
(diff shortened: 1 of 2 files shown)


klauer:
Also consider: https://stackoverflow.com/questions/19622198/what-does-set-e-mean-in-a-bash-script

ZLLentz:
`set -e` stops the scripts where it fails first, right? Sometimes I still want other tests to run if the first test fails.

PR #356, opened 2020-01-22
CI/DOCS: deploy key from someone who still works at SLAC

klauer:
Regenerated by @klauer 

We might have to pay him if he keeps committing our docs:
image


ZLLentz:
Is there an easy way to track these down? Only way I know is to wait for the push... I see that this should be fixed on `lightpath` as well

klauer:
Probably could figure it out through the GH API, but it'd likely be easier to just check the settings pages of each:
https://github.com/pcdshub/pcdsdevices/settings/keys

PR #364, opened 2020-02-05
ENH: PytmcEpicsSignal

ZLLentz:
## Description

Add `PytmcEpicsSignal` to connect to the PVs created by `pytmc` using the same arg as in the `pytmc` pragma.

## Motivation and Context

Should be really straightforward to make these pytmc pvs without picking between classes and typing _RBV in the right places

## How Has This Been Tested?

WIP

WIP


ZLLentz:
opened from wrong branch

PR #365, opened 2020-02-05
ENH: add PytmcSignal class

ZLLentz:
## Description

Add `PytmcSignal` to connect to the PVs created by `pytmc` using the same arg as in the `pytmc` pragma.

## Motivation and Context

Should be really straightforward to make these pytmc pvs without picking between classes and typing _RBV in the right places

## How Has This Been Tested?

Added basic tests, checked that the read and write PVs are what I expected, and connected to sample PVs in a live IOC

Included in the autodocs by default

### DIFF
--- a/pcdsdevices/signal.py
+++ b/pcdsdevices/signal.py
@@ -19,7 +19,7 @@ from pytmc.pragmas import normalize_io
 logger = logging.getLogger(__name__)
 
 
-class PytmcEpicsSignal(EpicsSignalBase):
+class PytmcSignal(EpicsSignalBase):
     """
     Class for a connection to a pytmc-generated EPICS record.
 
@@ -34,9 +34,9 @@ class PytmcEpicsSignal(EpicsSignalBase):
     def __new__(cls, prefix, *, io, **kwargs):
         norm = normalize_io(io)
         if norm == 'output':
-            return super().__new__(PytmcEpicsSignalRW)
+            return super().__new__(PytmcSignalRW)
         elif norm == 'input':
-            return super().__new__(PytmcEpicsSignalRO)
+            return super().__new__(PytmcSignalRO)
         else:
             # Should never get here unless pytmc's API changes
             raise ValueError(f'Invalid io specifier {io}')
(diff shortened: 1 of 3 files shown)


pcdsdevices/signal.py
```diff
@@ -13,11 +13,51 @@
...
+    the _RBV suffix and wrap the read/write PVs into the same signal object as
+    appropriate, and pick between a read-only signal and a writable one.
+
+    Under the hood this actually gives you the RW or RO version of the signal
+    depending on your io argument.
+    """
+    def __new__(cls, prefix, *, io, **kwargs):
+        # Same exact check done in pytmc
+        if 'o' in io:
+            return super().__new__(PytmcEpicsSignalRW)
```
klauer: Interesting! I like the idea.

Couple things:
* Should these pass kwargs through?
* The `io` normalization in pytmc got a bit more [complicated](https://github.com/slaclab/pytmc/blob/8dfb9408c7388bab17563f0a77869c3152759064/pytmc/pragmas.py#L288), but I suppose we could just support the one form here

ZLLentz: Strangely, when I pass any args/kwargs the thing errors out with something like "object new does not accept kwargs" (paraphrasing from memory). I think I'm doing something slightly wrong that may break later, though this definitely works as-is with the current stack. The result respects `isinstance`, etc. and the child classes follow their inheritance mro's properly.

Would it make sense to make pcdsdevices depend on pytmc and import this `normalize_io` function?

ZLLentz:
I think we should merge this soon and improve later. Having the API available to read your pragma and do:
`my_component = Cpt(PytmcEpicsSignal, 'STATE', io='io')`
instead of having to choose between:
`my_component = Cpt(EpicsSignalRO, 'STATE_RBV')`
and
`my_component = Cpt(EpicsSignal, 'STATE_RBV', write_pv='STATE')`

will be helpful.

klauer:
:+1: want to do the `pytmc.pragmas.normalize_io` call and then we merge?

ZLLentz:
Yes. I will accumulate a pytmc dependency very soon.

ZLLentz:
I'm ready to merge this if it still looks mostly ok, pairing with an added issue to investigate the intricacies of subclassing `__new__` like this

PR #367, opened 2020-02-11
Update IPM Ophyd class

ZryletTC:
Created new classes for IPIMB and Wave8 variations of the IPMs. Added diode x motion to IPMMotion. Still need to update X Motion PV once an alias is added to the IOC.

## Motivation and Context
Allows for better control of IPMs and their components through hutch-python.

## How Has This Been Tested?
Passes tests. Tested on XPP:SB3:IPM.

Docstrings have been created updated.


pcdsdevices/ipm.py
```diff
@@ -49,15 +96,276 @@ def removed(self):
...
+        Wave8 base PV
+
+    name : ``str``
+        Alias for the wave8
+    """
+    isum = Cpt(EpicsSignal, ':SUM', kind='normal')
+    xpos = Cpt(EpicsSignal, ':XPOS', kind='normal')
+    ypos = Cpt(EpicsSignal, ':YPOS', kind='normal')
+    evr_channel = Cpt(Trigger, ':TRIG:TRIG0', kind='normal')
+    ch0 = FCpt(Wave8Channel, '{self.prefix}', channel_index=0, kind='normal')
```
klauer: As no formatting is needed, these could all be `Cpt` with a suffix of `''`, i.e.,
```suggestion
    ch0 = Cpt(Wave8Channel, '', channel_index=0, kind='normal')
```

pcdsdevices/ipm.py
```diff
@@ -49,15 +96,276 @@ def removed(self):
...
+    __doc__ = __doc__ % (IPM_base) + basic_positioner_init
+
+    ipimb = FCpt(IPIMB, '{self.prefix_ipimb}', prefix_ioc='{self.prefix_ioc}')
+
+    def __init__(self, prefix, *, name, **kwargs):
+        self.prefix_ipimb = kwargs.pop('prefix_ipimb')
+        self.prefix_ioc = kwargs.pop('prefix_ioc', None)
+        super().__init__(prefix, name=name, **kwargs)
+
+    def isum(self):
```
klauer: Between `IPM_IPIMB`, `IPM_Wave8`, these utility methods all appear to be duplicated - only referencing a different sub-device (`self.ipimb` or `self.wave8`).

It seems like they could go into the base class for re-use (?) or a mixin (something like `HasXYPosAndISum` but actually sensibly named). Then you just refer to `self._xypos_device.isum.get()` instead of `self.ipimb`/`self.wave8`.

pcdsdevices/ipm.py
```diff
@@ -49,15 +96,276 @@ def removed(self):
...
+        return self.ipimb.isum.get()
+
+    def xpos(self):
+        return self.ipimb.xpos.get()
+
+    def ypos(self):
+        return self.ipimb.ypos.get()
+
+    def channel(self, i=0):
```
klauer: What's the benefit of `device.channel(3)` over `device.channels[3]`? The latter could be defined on `__init__`.

Also, `i` is not used in this method (should be `'ch%d' % i` instead of the current `'ch%i'`).

pcdsdevices/ipm.py
```diff
@@ -49,15 +96,276 @@ def removed(self):
...
+    # wait functionality?
+
+    def __init__(self, prefix, *, name, prefix_ioc, **kwargs):
+        if not prefix_ioc:
+            self._prefix_ioc = 'IOC:%s' % prefix
+        else:
+            self._prefix_ioc = prefix_ioc
+        super().__init__(prefix, name=name, **kwargs)
+
+    def screen(self):
```
klauer: I think the contents of the `screen` method should be pulled out into some reusable utility function that all of pcdsdevices could use.

pcdsdevices/ipm.py
```diff
@@ -25,14 +29,57 @@ class IPMTarget(InOutRecordPositioner):
...
+    """
+    Diode of a standard intensity position monitor.
+
+    This device has members for x and y motors, where x-motion is set up as an
+    IMS motor and y-motion is an InOutRecordPositioner that moves the diode
+    position to any of the four set positions, or out. There is also a member,
+    y, which points to the motor of the y-motion.
+    """
+
+    x = FCpt(IMS, '{self.x_prefix}', kind='normal')
```
ZLLentz: `self.x_prefix` is hard-coded in the init, so there is no need to use `FCpt`
```suggestion
    x = Cpt(IMS, ':XMOTOR', kind='normal')
```

pcdsdevices/ipm.py
```diff
@@ -25,14 +29,57 @@ class IPMTarget(InOutRecordPositioner):
...
+                                 wait=wait)
+
+    def remove(self, moved_cb=None, timeout=None, wait=False):
+        """Moves the diode out of the beam."""
+        return self.state.remove(moved_cb=moved_cb, timeout=timeout,
+                                 wait=wait)
+
+    remove.__doc__ += insert_remove
+
+    tab_whitelist = ['target', 'diode']
```
ZLLentz: settings like this are best kept near the top of the class def right under the component definitions so they are easy to find

pcdsdevices/ipm.py
```diff
@@ -49,15 +96,276 @@ def removed(self):
         """Returns ``True`` if target is removed. Diode never blocks."""
         return self.target.removed
 
-    def remove(self, moved_cb=None, timeout=None, wait=False):
-        """Moves the target out of the beam. Diode never blocks."""
-        return self.target.remove(moved_cb=moved_cb,
-                                  timeout=timeout,
-                                  wait=wait)
```
ZLLentz: I would argue that the `remove` method should not be removed so that we can still unambiguously move these out via lightpath or similar apps with a simple consistent interface (`device.remove() for device in devices`). The previous implementation worked under the assumption that the `IPMDiode` was never blocking and did not need to be removed, I do not know if that is a good assumption.

pcdsdevices/ipm.py
```diff
@@ -49,15 +96,276 @@ def removed(self):
...
+    delay = Cpt(EpicsSignal, ':TrigDelay', kind='config')
+    bias = Cpt(EpicsSignal, ':DiodeBias', kind='config')
+    ch0 = FCpt(IPIMBChannel, '{self.prefix}', channel_index=0, kind='normal')
+    ch1 = FCpt(IPIMBChannel, '{self.prefix}', channel_index=1, kind='normal')
+    ch2 = FCpt(IPIMBChannel, '{self.prefix}', channel_index=2, kind='normal')
+    ch3 = FCpt(IPIMBChannel, '{self.prefix}', channel_index=3, kind='normal')
+
+    # also be careful that you have to wait until trigger is enabled
+    # again before device can be used again.
+    # wait functionality?
```
ZLLentz: Are these comments planned code changes or are they just dangling? If not to tackle in this PR they belong in the github issues instead.

pcdsdevices/ipm.py
```diff
@@ -49,15 +96,276 @@ def removed(self):
...
+
+    def __init__(self, prefix, *, name, prefix_ioc, **kwargs):
+        if not prefix_ioc:
+            self._prefix_ioc = 'IOC:%s' % prefix
+        else:
+            self._prefix_ioc = prefix_ioc
```
ZLLentz: I've been considering doing something like this on all devices so we can quickly check ioc status or reboot from the python sessions. :+1: different use here but still good.

pcdsdevices/ipm.py
```diff
@@ -49,15 +96,276 @@ def removed(self):
...
+    def xpos(self):
+        return self.ipimb.xpos.get()
+
+    def ypos(self):
+        return self.ipimb.ypos.get()
+
+    def channel(self, i=0):
+        # thrown an exception for an invalid channel?
+        # if ( i >= 4 ): return None
+        return getattr(self.ipimb, 'ch%i').amplitude.get()
```
ZLLentz: Definitely should throw an exception here as per the first comment. The exception that will be thrown otherwise is not going to be helpful. Note that you should also catch negative channel inputs unlike the second comment.

pcdsdevices/ipm.py
ZLLentz: Why does this have a default arg?

pcdsdevices/ipm.py
```diff
@@ -49,15 +96,276 @@ def removed(self):
...
+        return self.wave8.isum.get()
+
+    def xpos(self):
+        return self.wave8.xpos.get()
+
+    def ypos(self):
+        return self.wave8.ypos.get()
+
+    def channel(self, i=0):
+        return getattr(self.wave8, 'ch%i').amplitude.get()
```
ZLLentz: Same questions/suggestions as the IPM variant

pcdsdevices/ipm.py
silkenelson: Largely historical: for PIM devices, only channel 0 has any data, the others only read noise.

pcdsdevices/ipm.py
```diff
@@ -49,15 +96,276 @@ def removed(self):
...
+        super().__init__(prefix, name=name, **kwargs)
+
+    def isum(self):
+        return self.wave8.isum.get()
+
+    def xpos(self):
+        return self.wave8.xpos.get()
+
+    def ypos(self):
+        return self.wave8.ypos.get()
```
ZLLentz: I'm getting a bit of deja vu with the methods here. There's likely a concise way to not have to copy/paste code between these two classes. You don't have to find it now, we can refactor later if someone figures out how to do it.

tests/test_ipm.py
```diff
@@ -1,56 +1,85 @@
...
+    """
+    return fake_ipm(IPMMotion)
+
+
+@pytest.fixture(scope='function', params=[IPM_Wave8, IPM_IPIMB])
+def fake_ipm_with_box():
+    """
+    Test IPM_Wave8 and IPM_IPIMB
+    """
+    return fake_ipm(request.param)
```
ZLLentz: Good concise test class setup :+1:

pcdsdevices/ipm.py
silkenelson: I opened a GitHub issue with expanding on what this comment tried to point out (was a reminder for me). This is likely not an issue or at least not a problem we are likely ro run into though.

pcdsdevices/ipm.py
silkenelson: The diode does not block if it's "IN" or "OUT" but it can block if it e.g. got stuck in an intermediate state.

tests/test_ipm.py
ZryletTC: Unfortunately, that way doesn't work so the nice conciseness is gone :(

teddyrendahl:
# [Codecov](https://codecov.io/gh/pcdshub/pcdsdevices/pull/367?src=pr&el=h1) Report
> Merging [#367](https://codecov.io/gh/pcdshub/pcdsdevices/pull/367?src=pr&el=desc) into [master](https://codecov.io/gh/pcdshub/pcdsdevices/commit/8299fff02d1d52248e521f7fdde40615d1608532&el=desc) will **decrease** coverage by `0.04%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/pcdshub/pcdsdevices/pull/367?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #367      +/-   ##
==========================================
- Coverage   93.45%   93.40%   -0.05%     
==========================================
  Files          34       34              
  Lines        2950     3080     +130     
==========================================
+ Hits         2757     2877     +120     
- Misses        193      203      +10     
```



------

[Continue to review full report at Codecov](https://codecov.io/gh/pcdshub/pcdsdevices/pull/367?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/pcdshub/pcdsdevices/pull/367?src=pr&el=footer). Last update [8299fff...dbbf3d4](https://codecov.io/gh/pcdshub/pcdsdevices/pull/367?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

pcdsdevices/ipm.py
ZryletTC: So should I put this method back in and have it remove both target and diode?

pcdsdevices/ipm.py
ZryletTC: Yeah, this was a placeholder while I was deciding whether it should be specified at instantiation or add the alias to the IOC. I'll probably go the IOC route and use this the way you suggested.

pcdsdevices/ipm.py
ZryletTC: @silkenelson Is there a reason to have both? If so, I can put it in `__init__` as suggested. If not, which do you prefer?

teddyrendahl:
# [Codecov](https://codecov.io/gh/pcdshub/pcdsdevices/pull/367?src=pr&el=h1) Report
> Merging [#367](https://codecov.io/gh/pcdshub/pcdsdevices/pull/367?src=pr&el=desc) into [master](https://codecov.io/gh/pcdshub/pcdsdevices/commit/8299fff02d1d52248e521f7fdde40615d1608532&el=desc) will **decrease** coverage by `0.07%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/pcdshub/pcdsdevices/pull/367?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #367      +/-   ##
==========================================
- Coverage   93.45%   93.37%   -0.08%     
==========================================
  Files          34       34              
  Lines        2950     3081     +131     
==========================================
+ Hits         2757     2877     +120     
- Misses        193      204      +11     
```



------

[Continue to review full report at Codecov](https://codecov.io/gh/pcdshub/pcdsdevices/pull/367?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/pcdshub/pcdsdevices/pull/367?src=pr&el=footer). Last update [8299fff...fd4dff9](https://codecov.io/gh/pcdshub/pcdsdevices/pull/367?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

pcdsdevices/ipm.py
ZryletTC: Where do you think it should go?

ZryletTC: Moved to utils.py

pcdsdevices/ipm.py
```diff
@@ -49,15 +100,274 @@ def removed(self):
...
+    ipimb = FCpt(IPIMB, '{self.prefix_ipimb}', prefix_ioc='{self.prefix_ioc}')
+
+    # IPIMB's have four channels
+    _num_channels = 4
+
+    def __init__(self, prefix, *, name, **kwargs):
+        self.prefix_ipimb = kwargs.pop('prefix_ipimb')
+        self.prefix_ioc = kwargs.pop('prefix_ioc', None)
+        super().__init__(prefix, name=name, **kwargs)
+        self.det = self.ipimb
```
ZLLentz: the `self.det` setattr here plus the new parent is a :+1: implementation of removing the duplicated code

PR #374, opened 2020-02-20
ENH: Update PIM classes

ZryletTC:
## Description
Changed hierarchy so that a PIM object has an InOutPositioner instead of being one.
Added classes for multiple configurations for PIMs which have LEDs, focus motors, or both.

Looking for feedback on a bunch of things. e.g. I'm not sure if the way I did some things is allowed/good practice. Also, please let me know if you have any suggestions for better class names or the current ones don't follow naming standards.

## Motivation and Context
Gives a better representation of what's actually on the beamline and allows for better control of the device.

## How Has This Been Tested?
Hasn't yet.

Docstrings

### DIFF
--- a/pcdsdevices/pim.py
+++ b/pcdsdevices/pim.py
@@ -1,23 +1,21 @@
 """
-Profile Intensity Monitor Classes
-
-This module contains all the classes relating to the profile intensity monitor
-classes at the user level. A PIM will usually have at least a motor to control
-yag position and a camera to view the yag.
+Module for the `PIM` profile intensity monitor classes
 """
 import logging
 
-from ophyd import FormattedComponent as FCpt
+from ophyd.device import Device, Component as Cpt, FormattedComponent as FCpt
+from ophyd.signal import EpicsSignal
 
+from .epics_motor import IMS
 from .areadetector.detectors import PCDSAreaDetector
 from .inout import InOutRecordPositioner
 
 logger = logging.getLogger(__name__)
 
 
-class PIMMotor(InOutRecordPositioner):
+class PIM_Y(InOutRecordPositioner):
     """
-    Standard position monitor motor.
+    Standard position monitor Y motor.
 
     This can move the stage to insert the yag
     or diode, or retract from the beam path.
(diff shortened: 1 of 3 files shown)


pcdsdevices/pim.py
```diff
@@ -1,23 +1,21 @@
 """
-Profile Intensity Monitor Classes
-
-This module contains all the classes relating to the profile intensity monitor
-classes at the user level. A PIM will usually have at least a motor to control
-yag position and a camera to view the yag.
```
ZLLentz: This becomes the header in the autodocs, so probably shouldn't be removed

tests/test_pim.py
```diff
@@ -27,27 +28,66 @@ def test_pim_stage(fake_pim):
...
+    pim.remove()
     assert pim.removed
-    pim.unstage()
+    pim.state.unstage()
     assert pim.inserted
 
 
 @pytest.mark.timeout(5)
-def test_pim_det():
+def test_pim_init():
```
ZLLentz: This test is :fire:
If you're bored you can try to parametrize it

tests/test_pim.py
```diff
@@ -56,8 +96,8 @@ def test_pim_subscription(fake_pim):
     logger.debug('test_pim_subscription')
     pim = fake_pim
     cb = Mock()
-    pim.subscribe(cb, event_type=pim.SUB_STATE, run=False)
-    pim.state.sim_put(2)
+    pim.state.subscribe(cb, event_type=pim.state.SUB_STATE, run=False)
+    pim.state.state.sim_put(2)
```
ZLLentz: I can't remember if there's a good reason to keep these top-level devices subscribeable. Maybe we were using that in `lightpath`? We should revisit in a later PR and do them uniformly if it's needed.

pcdsdevices/pim.py
```diff
@@ -30,39 +28,197 @@ class PIMMotor(InOutRecordPositioner):
...
+        """Pulls out the first two segments of the prefix PV, if not already
+           done"""
+        if not self._prefix_start:
+            self._prefix_start = '{0}:{1}:'.format(prefix.split(':')[0],
+                                                   prefix.split(':')[1])
+
+    @property
+    def prefix_start(self):
+        """Returns the first two segments of the prefix PV."""
+        return str(self._prefix_start)
```
ZLLentz: is this ever not a string already?

pcdsdevices/pim.py
```diff
@@ -30,39 +28,197 @@ class PIMMotor(InOutRecordPositioner):
...
+        """Returns ``True`` if the yag and diode are removed from the beam."""
+        return self.state.removed
+
+    @property
+    def inserted(self):
+        """Returns ``True`` if yag or diode are inserted."""
+        return self.state.inserted
+
+    def insert(self, moved_cb=None, timeout=None, wait=False):
+        """Moves the diode into the beam."""
```
ZLLentz: Is this true? Does insert pick the diode over the yag?

ZLLentz: If so I think the yag gets more use

pcdsdevices/pim.py
```diff
@@ -30,39 +28,197 @@ class PIMMotor(InOutRecordPositioner):
...
+        if prefix_zoom:
+            self._prefix_zoom = prefix_zoom
+        else:
+            self._prefix_zoom = self.prefix_start+'CLZ:01'
+
+        super().__init__(prefix, name=name, **kwargs)
+        self.y_motor = self.state.motor
+
+
+class PIM_withFocus(PIM):
```
ZLLentz: The python convention here would be something like:
```suggestion
class PIMWithFocus(PIM):
```
But that looks ugly so idk

pcdsdevices/pim.py
```diff
@@ -30,39 +28,197 @@ class PIMMotor(InOutRecordPositioner):
...
+    prefix_zoom : str, optional
+        The EPICS base PV of the zoom motor. If None, it will be attempted to
+        be inferred from `prefix`
+
+    prefix_focus : str, optional
+        The EPICS base PV of the focus motor. If None, it will be attempted to
+        be inferred from `prefix`
+    """
+    focus_motor = FCpt(IMS, '{self._prefix_focus}')
+
```
ZLLentz:
```suggestion
tab_whitelist = ['focus_motor']
```

pcdsdevices/pim.py
ZryletTC: No, it's just there to shallow copy

pcdsdevices/pim.py
```diff
@@ -30,39 +28,197 @@ class PIMMotor(InOutRecordPositioner):
...
+    prefix_zoom : str, optional
+        The EPICS base PV of the zoom motor. If None, it will be attempted to
+        be inferred from `prefix`
+
+    prefix_led : str, optional
+        The EPICS base PV of the LED. If None, it will be attempted to be
+        inferred from `prefix`
+    """
+    led = FCpt(EpicsSignal, '{self._prefix_led}')
+
```
ZLLentz:
```suggestion
tab_whitelist = ['led']

```

pcdsdevices/pim.py
```diff
@@ -1,23 +1,21 @@
...
 
+from .epics_motor import IMS
 from .areadetector.detectors import PCDSAreaDetector
 from .inout import InOutRecordPositioner
 
 logger = logging.getLogger(__name__)
 
 
-class PIMMotor(InOutRecordPositioner):
+class PIM_Y(InOutRecordPositioner):
```
ZLLentz: Here too, typically the underscore isn't used in class names by python convention
https://www.python.org/dev/peps/pep-0008/#class-names

ZLLentz:
Looks good overall, just quibbles

pcdsdevices/pim.py
klauer: Both are less than ideal - the other alternative could be `PimWithFocus`, but I like that even less. I suppose my vote would go with `PIMWithFocus` even if the camel-case name loses a bit of meaning.

pcdsdevices/pim.py
klauer: Agreed on the suggestion, but be careful accepting it as the indentation is wrong.

pcdsdevices/pim.py
```diff
@@ -30,39 +28,197 @@ class PIMMotor(InOutRecordPositioner):
...
+    prefix_led : str, optional
+        The EPICS base PV of the LED. If None, it will be attempted to be
+        inferred from `prefix`
+    """
+    led = FCpt(EpicsSignal, '{self._prefix_led}')
+
+    def __init__(self, prefix, *, name, prefix_led=None, **kwargs):
+        self.infer_prefix(prefix)
+
+        # Infer the illuminator PV from the base prefix
```
klauer: Just an idea, but I wonder if all of the infer steps be in a customized `infer_prefix` method in the subclasses.

e.g.,
```python
def __init__(self, prefix, ..., **kwargs):
    # Stash away all `prefix_*` kwargs by convention
    self._prefixes = {key.split('prefix_')[1]: kwargs.pop(key) for key in kwargs if key.startswith('prefix_')}
    self.infer_prefix()
    super().__init__(...)

def infer_prefix(self):
    super().infer_prefix()
    # Infer the illuminator PV from the base prefix
    self._prefixes['led']
```

Maybe it's useful as it seems like we repeat this sort of code a lot, but maybe it's too confusing...

pcdsdevices/pim.py
```diff
@@ -30,39 +28,197 @@ class PIMMotor(InOutRecordPositioner):
...
+    Parameters
+    ----------
+    prefix : str
+        The EPICS base of the PIM
+
+    name : str
+        A name to refer to the device
+
+    prefix_det : str, optional
+        The EPICS base PV of the detector. If None, it will be attemptted to be
```
klauer:
```suggestion
        The EPICS base PV of the detector. If None, it will be attempted to be
```

tests/test_pim.py
```diff
@@ -27,27 +28,66 @@ def test_pim_stage(fake_pim):
...
+    FakePIM('Test:Yag', name='test', prefix_det='potato', prefix_zoom='woosh',
+            prefix_focus='blur')
+    FakePIM('Test:Yag', name='test', prefix_det='potato', prefix_focus='blur')
+    FakePIM('Test:Yag', name='test', prefix_zoom='potato', prefix_focus='blur')
+    FakePIM('Test:Yag', name='test', prefix_focus='blurry')
+    FakePIM('Test:Yag', name='test', prefix_det='potato')
+    FakePIM('Test:Yag', name='test', prefix_zoom='potato')
+    FakePIM('Test:Yag', name='test')
+    FakePIM = make_fake_device(PIM_withBoth)
+    FakePIM('Test:Yag', name='test', prefix_det='potato', prefix_zoom='woosh',
```
klauer: 😆

tests/test_pim.py
klauer: Yeah, we could revisit that for lightpath if necessary.

pcdsdevices/pim.py
ZLLentz: my b

ZLLentz: also, I just remembered: you should be able to set `tab_component_names` to `True` to automatically get all of the components tabbable

pcdsdevices/pim.py
ZryletTC: No, it goes to YAG. Docstring was wrong. Now fixed.

pcdsdevices/pim.py
ZryletTC: Now been changed to PIMY

pcdsdevices/pim.py
ZryletTC: I'd like to do this, but I can't figure out a simple way to make it work.

pcdsdevices/pim.py
ZryletTC: Updated in new commit.

tests/test_pim.py
ZryletTC: Unfortunately, I don't have much time to be bored 😐

pcdsdevices/pim.py
ZryletTC: Will instead use tab_component_names.

pcdsdevices/pim.py
ZryletTC: Used tab_component_names in new commit.

pcdsdevices/pim.py
ZryletTC: Zach and I tried to figure it out to no avail and also noted that such a change would also make the code less readable.

klauer: :+1:

ZryletTC:
Tested on xpp_sb3_pim. Everything worked as expected other than unstage() did not work because that PIM's detector is disconnected.

PR #363, opened 2020-02-05
HXR Stubs

hhslepicka:
Some devices to spin the Home Screen. This was created to test the home screen...
Feel free to push to this branch as desired.

PR #377, opened 2020-02-24
Hxr stubs

slacAWallace:
An update to Hugo's original PR #363 

More stubs and a bit of reorg.


pcdsdevices/attenuator.py
```diff
@@ -351,3 +352,44 @@ def set_combined_attenuation(attenuation, *attenuators):
         else:
             attenuators[i].actuate_value()
 '''
+
+
+class FEESolidFilter(Device, BaseInterface):
```
ZLLentz:
```suggestion
class FEESolidAttenuator(Device, BaseInterface):
```
Because it feels wrong to refer to it four times in the docstring as an attenuator but then call it a filter

pcdsdevices/slits.py
```diff
@@ -304,3 +307,40 @@ def _aperture_changed(self, *args, **kwargs):
...
+    padding = max(range_)//10 + 2
+    defn = OrderedDict()
+    for i in range_:
+        attr = '{attr}{i}'.format(attr=attr_base, i=i)
+        suffix = ':RTD:{i}'.format(i=str(i).zfill(padding))
+        defn[attr] = (cls, suffix, kwargs)
+    return defn
+
+
+class RTD(Device, BaseInterface):
```
ZLLentz: Is there a functional difference between this `RTD` class and the `Thermocouple` class in `sensors.py`?

pcdsdevices/attenuator.py
klauer: Agreed

PR #379, opened 2020-02-25
MNT: changed PIPPLC class by adding state_rbv and SP_DI->HV_DI

slacAdpai:
## Description
 changed PIPPLC class by adding state_rbv and SP_DI->HV_DI

## Motivation and Context
Updating ion pump plc logic. The state variable has been addded to report whether or not the pump is off, starting, on, faulted. 

The SP_DI was changed to the HV_DI because the relay output is used to report the ion pump output state.


ZLLentz:
Some more context on this?

klauer:
I believe the PLC code was tweaked to better represent the internal state of the device.

klauer:
Confirmed by @slacAdpai via slack. Going to merge.

ZLLentz:
The original PR had no message content, makes sense with the edited message

PR #381, opened 2020-02-26
ENH: Filling in stubs for Beckhoff controlled mirror classes, both 1st and 2nd generation

jsheppard95:
## Description

Added components for each mirror PVs to `pcdsdevices.mirror.HardOffsetMirror`. These include the motor record of each axis (Yup, Ydwn, Xup, Xdwn, Pitch, Bender), gantry status/configuration (current Y/X gantry, coupling/decoupling signals), and RMS differences between encoder readback and axis setpoints (again for Y, X, Pitch, and Bender). 

## Motivation and Context

These changes are necessary for `lucid` to produce control screens using `typhos` for mirror systems MR1L0, MR2L0, and MR1K4 (1st generation Axilon systems refurbished to use Beckhoff controls)

## How Has This Been Tested?

Ran automated tests within this repository to ensure nothing was broken. Instantiated ophyd object/produced typhon screen through the following code snippet:
On `lfe-console` with `conda-dev` sourced:
```
import importlib.util # for loading dev version
import sys
import typhos
from qtpy.QtWidgets import QApplication

# Load dev version
spec = importlib.util.spec_from_file_location("pcdsdevices.mirror", "/reg/neh/home/sheppard/PCDSGitHub/pcdsdevices/pcdsdevices/mirror.py")
mirror = importlib.util.module_from_spec(spec)
spec.loader.exec_module(mirror)

# instantiate ophyd object
mr1l0 = mirror.HardOffsetMirror('MR1L0:LFE', name='mr1l0_lfe')

# Bring up Typhos screen
app = QApplication.instance() or QApplication(sys.argv)
typhos.use_stylesheet()
suite = typhos.TyphosSuite.from_device(mr1l0)
suite.show()
app.exec_()
```
## TODO/Questions
- We have two Beckhoff controlled mirror stubs: `HardOffsetMirror` (1st Generation Beckhoff controlled) and `XOffsetMirror` (2nd Generation Beckhoff controlled). However from the Epics Layer and up these should be the same (in the context of motion), i.e they will have the same PV structure. We currently do not have a 2nd generation system accessible via epics to confirm this will work, but once available I thought I should try to take this approach.

- What is the `value` keyword argument of `Component` used for?

### DIFF
--- a/pcdsdevices/mirror.py
+++ b/pcdsdevices/mirror.py
@@ -8,7 +8,6 @@ vertical gantries.
 """
 import logging
 
-from ophyd.signal import SignalRO
 import numpy as np
 from ophyd import (Device, EpicsSignal, EpicsSignalRO, Component as Cpt,
                    PVPositioner, FormattedComponent as FCpt)
@@ -16,6 +15,7 @@ from ophyd import (Device, EpicsSignal, EpicsSignalRO, Component as Cpt,
 from .doc_stubs import basic_positioner_init
 from .inout import InOutRecordPositioner
 from .interface import FltMvInterface, BaseInterface
+from .signal import PytmcSignal
 
 logger = logging.getLogger(__name__)
 
@@ -279,11 +279,11 @@ class PointingMirror(InOutRecordPositioner, OffsetMirror):
         return super().check_value(pos)
 
 
-class HardOffsetMirror(Device, BaseInterface):
+class XOffsetMirror(Device, BaseInterface):
     """
-    Hard X-Ray Offset Mirror
+    X-Ray Offset Mirror
 
-    1st gen Axilon design with LCLS-II Beckhoff motion architecture.
+    1st and 2nd gen Axilon designs with LCLS-II Beckhoff motion architecture.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


ZLLentz:
RE: the question at the bottom of your description:

`value` is not a `kwarg` of `Component`, but any additional `kwargs` here are used to construct the class when we instantiate the component. So the `value` can be used to set the starting value of a basic `Signal`.

ZLLentz:
Also: if the PV structure is exactly the same, they can be the same class.

pcdsdevices/mirror.py
```diff
@@ -293,8 +293,42 @@ class HardOffsetMirror(Device, BaseInterface):
...
+    # Motor components: can read/write positions
+    y_up = Cpt(EpicsSignal, ':MMS:YUP', name='y_up', kind='normal')
+    y_dwn = Cpt(EpicsSignal, ':MMS:YDWN', name='y_dwn', kind='config')
+    x_up = Cpt(EpicsSignal, ':MMS:XUP', name='x_up', kind='normal')
+    x_dwn = Cpt(EpicsSignal, ':MMS:XDWN', name='x_dwn', kind='config')
+    pitch = Cpt(EpicsSignal, ':MMS:PITCH', name='pitch', kind='normal')
+    bender = Cpt(EpicsSignal, ':MMS:BENDER', name='bender', kind='normal')
+
+    # Gantry components
+    gantry_x = Cpt(EpicsSignalRO, ':GANTRY_X_RBV', name='gantry_x',
```
ZLLentz: Do not supply names here, `ophyd` will automatically fill these in from the attribute name `gantry_x` on the left, prepending with the full device name e.g. `mr1l0_gantry_x`

pcdsdevices/mirror.py
```diff
@@ -293,8 +293,42 @@ class HardOffsetMirror(Device, BaseInterface):
...
+
+    # RMS Cpts:
+    y_enc_rms = Cpt(EpicsSignalRO, ':Y_ENC:RMS_RBV', name='y_enc_rms',
+                    kind='normal')
+    x_enc_rms = Cpt(EpicsSignalRO, ':X_ENC:RMS_RBV', name='x_enc_rms',
+                    kind='normal')
+    pitch_enc_rms = Cpt(EpicsSignalRO, ':PITCH_ENC:RMS_RBV',
+                        name='pitch_enc_rms', kind='normal')
+    bender_enc_rms = Cpt(EpicsSignalRO, ':BENDER_ENC:RMS_RBV',
+                         name='bender_enc_rms', kind='normal')
```
ZLLentz: Consider using `PytmcSignal` for all of the signals created by `pytmc`. e.g.
`gantry_x = Cpt(PytmcSignal, ':GANTRY_X', io='i')`
`decouple_y = Cpt(PytmcSignal, ':DECOUPLE_X', io='io')`

It fills in the `_RBV` for you and connects to both the non-RBV and the RBV PVs if applicable

pcdsdevices/mirror.py
```diff
@@ -293,8 +293,42 @@ class HardOffsetMirror(Device, BaseInterface):
...
+
+    def __init__(self, prefix, **kwargs):
+        super().__init__(prefix, **kwargs)
```
ZLLentz: You do not need to subclass `__init__` if you aren't changing anything

ZLLentz:
Regarding the leading colon, I just wanted the happi database to look nice... So I baked the colon into all my component names. 🤷‍♂

pcdsdevices/mirror.py
jsheppard95: Resolved in commit bb4a164

pcdsdevices/mirror.py
jsheppard95: Resolved in commit bb4a164

pcdsdevices/mirror.py
jsheppard95: Oh okay so now that I've removed this statement, all the parameters I pass to `XOffsetMirror` essentially go to `Device.__init__`? And `XOffsetMirror` doesn't have its own `__init__` because there aren't any parameters that I want to specifically passed to it?

ZLLentz: Exactly. If you don't define a particular method, it calls the method of the parent class instead. Here you had overwritten the `__init__` method to simply call the parent's `__init__` method with no functional changes, which is the same as if you didn't overwrite it at all.

Note that there were two non-functional changes you made by overriding the `__init__` method: you hid the parent's `__init__` docstring (if it existed) and you changed the outward-facing signature to have less args inspectable from `IPython`, e.g. with `__init__` redefined as above it's not clear that there is a mandatory `name` `kwarg`

jsheppard95: Makes sense, thanks its always good to have a refresher on class inheritance peculiarities.

jsheppard95:
Perfect, is it okay to go ahead and merge? I'd like to see how MR1L0 and MR2L0 look on the LFE home screen.

hhslepicka:
If you want to check at how a screen would look like and you are using the `dev_conda` environment then you can just do:

```bash
$ typhos 'package.ClassName[{"prefix":"MY:PREFIX"}]'
```

ZLLentz:
@jsheppard95 are you happy with this as-is? If so I will merge

jsheppard95:
Yes I'd say I'm pretty happy with it. Tried Hugo's comment and the screen looks good, I think we can go ahead and merge!

PR #382, opened 2020-02-28
ENH: specify typhos templates for some devices

ZLLentz:
## Description

Quick proposal for an interface to specify which `typhos` template should be associated with each device class. In `typhos` we can check if `typhos_detailed`, etc. are defined and pick a screen from the `typhos` path appropriately, or use the default templates if none is found. It would be up to us to include the template ui files in our `typhos` path. A nice thing about doing it this way is that we could take adventage of class inheritance to specify a screen for all motors, then override it as needed to add vendor-specific functions cleanly, etc.

I'm not really sure the best place to keep these template files. `typhos` will get bloated if we put them all there. The best option may be to put them in this repo to keep them linked to the device classes and then append to the `typhos` path on install, but that will make this repo very large. It will also pick up a `typhos` dependency and the whole stack below that. Another choice would be to stash them in a third repo that depends on both `pcdsdevices` and `typhos`, but then we're maintaining an extra repo. We could also spread them out into the various IOC and PLC repos as appropriate, but then they are scattered everywhere.

## Motivation and Context

Screens are looking good, but tweaked templates will make them even better. We should decide how we're specifying and deploying these.

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -58,6 +58,10 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
     tab_whitelist = ["set_current_position", "home", "velocity",
                      "enable", "disable"]
 
+    typhos_detailed = 'detailed_positioner.ui'
+    typhos_screen = 'positioner.ui'
+    typhos_embedded = 'embedded_positioner.ui'
+
     @property
     def low_limit(self):
         """
@@ -456,6 +460,10 @@ class BeckhoffAxis(EpicsMotorInterface):
     """
     __doc__ += basic_positioner_init
     tab_whitelist = ['clear_error']
+
+    typhos_detailed = 'detailed_beckhoff_positioner.ui'
+    typhos_screen = 'beckhoff_positioner.ui'
+    typhos_embedded = 'embedded_beckhoff_positioner.ui'
 
     plc = Cpt(BeckhoffAxisPLC, ':PLC:', kind='normal')
 
--- a/pcdsdevices/pim.py
+++ b/pcdsdevices/pim.py
@@ -83,6 +83,10 @@ class LCLS2ImagerBase(Device, BaseInterface):
     """
     tab_component_names = True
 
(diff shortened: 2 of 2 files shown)


hhslepicka:
I really like this proposal. Having the template close to the class definition is great.

Adding to your suggestion...

If no template is defined, what about fallback to look at TYPHOS_TEMPLATE_PATH for a specific format such as MyDeviceClass_detailed.ui, MyDeviceClass_embedded.ui, MyDeviceClass.ui before falling back to the defaults?

That would allow us to specify templates for Ophyd classes that we don’t have specific LCLS versions.

Also, what about make a convention of starting with class name first and type of template do web listed they will show close to each other...

ZLLentz:
I like your idea of using `MyDeviceClass` as the convention and as the fallback. Then you would never even put this on a class... Making this PR somewhat pointless. Would we also crawl up the `mro` and check if the parent device classes have templates as the fallback-fallback before using default?

E.g. we could rename the positioner templates in `typhos` to `PositionerBase.ui` etc. and all positioner subclasses would automatically get a starting point

hhslepicka:
@ZLLentz sorry that I did not think about it before... the mro is a fantastic idea! I like it... 
That would be great. And if no other option is found, fallback to the auto generated screen.

klauer:
👍 overall! Building on the above:

Order of template determination
-------------
1. User/developer -level override somehow?
2. `happi` entry template field
3. Class/instance-specified templates
   a. Fully qualified class path, e.g., `{pcdsdevices.whatever.ClassName}.{detailed}.ui`
   b. Bare class path, e.g., `{ClassName}.{detailed}.ui`
   c. Repeat the above for `cls.mro()[1:]`

Other
-------------
* Template file searching should likely be cached
* I also wonder if we could have any of the above alternate templates for screens swapped dynamically with, say, a context menu option. This could allow for the operator to fall back to a screen easily if we incorrectly specify it somehow.

ZLLentz:
I'm going to make an issue for this in `typhos` and close this PR

PR #387, opened 2020-03-03
implementing FEESolidAttenuator device for AT2L0 Hard x-ray solid att…

rajanplumley:
This is a quick-and-dirty point-to-point motion implementation of the system for early HXR commissioning.  I have tested this device and its autogenerated screen seperately in a script, but now I am porting it into pcdsdevices so it can be listed in Happi.

### DIFF
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -396,9 +396,6 @@ class FEESolidAttenuator(Device, BaseInterface):
     blade_18 = Cpt(BeckhoffAxis, ':MMS:18', kind='hinted')
     blade_19 = Cpt(BeckhoffAxis, ':MMS:19', kind='hinted')
 
-    def __init__(self, prefix, name='AT2L0', **kwargs):
-        super().__init__(prefix, name=name, **kwargs)
-
 
 class GasAttenuator(Device, BaseInterface):
     """
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -13,6 +13,7 @@ from ophyd.signal import Signal, SignalRO, EpicsSignal, EpicsSignalRO
 
 from .inout import InOutPositioner
 from .interface import FltMvInterface, BaseInterface
+from .epics_motor import BeckhoffAxis
 
 logger = logging.getLogger(__name__)
 MAX_FILTERS = 12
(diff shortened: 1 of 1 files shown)


pcdsdevices/attenuator.py
```diff
@@ -369,10 +370,34 @@ class FEESolidAttenuator(Device, BaseInterface):
     Notes
     ----------
     Solid attenuator variant from the LCLS-II XTES project.
-     Motorized, 16 blades.
+     Motorized, 18 filters + 1 inspection mirror.
```
ZLLentz: You put an extra space here, but don't bother fixing it

ZLLentz: Oh it wasn't you, it was already there before you started

pcdsdevices/attenuator.py
```diff
@@ -369,10 +370,34 @@ class FEESolidAttenuator(Device, BaseInterface):
...
+    blade_12 = Cpt(BeckhoffAxis, ':MMS:12', kind='hinted')
+    blade_13 = Cpt(BeckhoffAxis, ':MMS:13', kind='hinted')
+    blade_14 = Cpt(BeckhoffAxis, ':MMS:14', kind='hinted')
+    blade_15 = Cpt(BeckhoffAxis, ':MMS:15', kind='hinted')
+    blade_16 = Cpt(BeckhoffAxis, ':MMS:16', kind='hinted')
+    blade_17 = Cpt(BeckhoffAxis, ':MMS:17', kind='hinted')
+    blade_18 = Cpt(BeckhoffAxis, ':MMS:18', kind='hinted')
+    blade_19 = Cpt(BeckhoffAxis, ':MMS:19', kind='hinted')
+
+    def __init__(self, prefix, name='AT2L0', **kwargs):
```
ZLLentz: this isn't a valid name in happi because of the caps

ZLLentz: I suggest doing any one of:

- Remove the init entirely
- Change the name to `at2l0`
- Go all-in and set a default prefix

rajanplumley: I choose option 1

ZLLentz: fwiw I started an issue at https://github.com/pcdshub/typhos/issues/247 because I think `AT2L0` looks better than `at2l0` in context of the `Typhos` screen title

ZLLentz:
Merging to slap into the dev environment pronto

ZLLentz:
The `typhos` for this renders poorly on my small screen. I suspect it looks much better in the XPP control room.

PR #388, opened 2020-03-03
MNT: Fixed Valve.py VGC MPS status with correct PV, MPS_FAULT_OK_RBV

slacAdpai:
## Description
MNT: Fixed Valve.py VGC MPS status with correct PV, MPS_FAULT_OK_RBV

## Motivation and Context
I forgot to put in _RBV in the previous commit and therefore the expert screens don't show up correctly.


ZLLentz:
In the future you should just use `PytmcSignal` if it's a PLC variable. It will fill in the `_RBV` for you and connect to both the `_RBV` PV and the non-rbv PV if both exist. I will not ask you to refactor right now though.

PR #390, opened 2020-03-03
BUG: gauge protection setpoint should be treated writeable

n-wbrown:
## Description

Make GCC protection set point writable. 

## Motivation and Context

Without this cold cathode gauges cannot configure the point at which they enable. 

## How Has This Been Tested?

Unit tests pass.

### DIFF
--- a/pcdsdevices/gauge.py
+++ b/pcdsdevices/gauge.py
@@ -204,7 +204,7 @@ class GCCPLC(GaugePLC):
     high_voltage_disable = Cpt(EpicsSignalRO, ':HV_DIS_DO_RBV', kind='normal',
                                doc=('enables the high voltage on the cold '
                                     'cathode gauge'))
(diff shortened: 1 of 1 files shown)


pcdsdevices/gauge.py
```diff
@@ -204,7 +204,7 @@ class GCCPLC(GaugePLC):
     high_voltage_disable = Cpt(EpicsSignalRO, ':HV_DIS_DO_RBV', kind='normal',
                                doc=('enables the high voltage on the cold '
                                     'cathode gauge'))
-    protection_setpoint = Cpt(EpicsSignalRO, ':PRO_SP_RBV', kind='normal',
+    protection_setpoint = Cpt(EpicsSignalWithRBV, ':PRO_SP', kind='normal',
```
ZLLentz: In the future, we should use `PytmcSignal` for all of these instead of bouncing between using `EpicsSignalRO` and providing the `RBV` or using `EpicsSignalWithRBV` and omitting it.

PR #392, opened 2020-03-05
ENH: changed cmd_err_reset so its accessible from a typhos screen

jsheppard95:
## Description
Just changing the `kind` attribute of `epics_motor.BeckhoffAxisPLC.cmd_err_reset` from `omitted` to `config` so that we have access to it from a Typhos screen.

## Motivation and Context
One needs to write to this variable after an NC error to resume motion through epics and it is convenient to have it on the control screen.

## How Has This Been Tested?
After setting `PYTHONPATH` to look first for a local version, the following command generates a test screen with the reset parameter shown in the `config` window:
`typhos 'pcdsdevices.mirror.XOffsetMirror[{"prefix":"MR1L0:LFE"}]'`

## Screenshot:
typhos-cmd-err-reset

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -442,7 +442,7 @@ class BeckhoffAxisPLC(Device):
     status = Cpt(PytmcSignal, 'sErrorMessage', io='i', kind='normal',
                  string=True)
     err_code = Cpt(PytmcSignal, 'nErrorId', io='i', kind='normal')
-    cmd_err_reset = Cpt(PytmcSignal, 'bReset', io='o', kind='omitted')
+    cmd_err_reset = Cpt(PytmcSignal, 'bReset', io='o', kind='config')
 
 
 class BeckhoffAxis(EpicsMotorInterface):


ZLLentz:
This is set to `True` on the miscellaneous tab though

ZLLentz:
```
            <widget class="TyphosSignalPanel" name="misc_panel">
             <property name="toolTip">
              <string/>
             </property>
             <property name="whatsThis">
              <string>
    Panel of Signals for Device
    </string>
             </property>
             <property name="showHints" stdset="0">
              <bool>false</bool>
             </property>
             <property name="showNormal" stdset="0">
              <bool>false</bool>
             </property>
             <property name="showConfig" stdset="0">
              <bool>false</bool>
             </property>
            </widget>
```

klauer:
What template is that @ZLLentz?

ZLLentz:
It looks like checking and unchecking the box fixes the XML file... I think the class changed but the files were never updated to match. I will make a typhos PR

klauer:
Ah, OK 👍

ZLLentz:
It was `detailed_screen.ui`

PR #396, opened 2020-03-11
CI: ban defaults channel from travis build

ZLLentz:
## Description

Remove `defaults` from conda channels list in travis build.

## Motivation and Context

I'm planning to do this on the env builds because of exactly the issue here: https://travis-ci.org/github/pcdshub/pcdsdevices/jobs/661244275

There's some crossplay issues between `conda-forge` and `defaults` that we've been lucky to avoid up to this point. We can no longer ignore it, and we need `conda-forge`, so we're axing defaults.

## How Has This Been Tested?

It was not

It will not be

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,7 @@ install:
   - conda update -q conda conda-build
   - conda config --add channels pcds-tag
   - conda config --append channels conda-forge
+  - conda config --remove channels defaults
   # Useful for debugging any issues with conda
   - conda info -a
   # Paranoid check of what version we're using
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,7 @@ install:
   - conda config --append channels conda-forge
   - conda config --remove channels defaults
   - conda config --set channel_priority strict
+  - conda config --set update_dependencies false
   # Useful for debugging any issues with conda
   - conda info -a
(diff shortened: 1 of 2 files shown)


ZLLentz:
@ZryletTC hopefully this fixes the build

ZLLentz:
It did not fix the build

ZLLentz:
Local build works with this config:
```
channels:
  - pcds-tag
  - conda-forge
channel_priority: true
update_dependencies: false
```

klauer:
@ZLLentz can you paste the output of `conda info -a` from the working environment?

ZLLentz:
[conda_config.txt](https://github.com/pcdshub/pcdsdevices/files/4321709/conda_config.txt)

ZLLentz:
I spy a small incompatibility between matplotlib under 3 and cf_units on conda forge... I'm thinking I just resolve these issues that have bugged me for a while (stuck to under 3 plus I want to kill cf_units)

ZLLentz:
I'll push a "no pinning matplotlib" attempt and then go home

ZLLentz:
Old conda-build works

Don't ask questions, just merge things

klauer:
After the strict settings failing, I had a feeling it might've been the conda-build/conda versions... Good find! Will have to investigate later.

Issue #389, opened 2020-03-03
Refactor PLC Classes to use PytmcSignal


ZLLentz:
## Expected Behavior

Use `PytmcSignal` for all `pytmc`-originated variables

## Current Behavior

Many PLC classes made before `PytmcSignal` made

## Possible Solution

Take a deep breath and refactor

klauer:
👍 

I think we can start by listing the classes that are based on pytmc'd PLC IOCs. From there, converting won't be so bad. But we should also ensure that (PVs pre conversion) and (PVs post-conversion) are identical, per-device.

PR #397, opened 2020-03-12
BUG: VCN state must be write-able

n-wbrown:
## Description

The altered field should be writable from the Typhos generated screens.

### DIFF
--- a/pcdsdevices/valve.py
+++ b/pcdsdevices/valve.py
@@ -272,5 +272,5 @@ class VCN(Device):
                        doc='interlock ok status')
     open_command = Cpt(EpicsSignalWithRBV, ':OPN_SW', kind='normal',
                        doc='Epics command to Open valve')
-    state = Cpt(EpicsSignalWithRBV, ':STATE_RBV', kind='hinted', doc='Valve state')
+    state = Cpt(EpicsSignalWithRBV, ':STATE', kind='hinted', doc='Valve state')
     pos_ao = Cpt(EpicsSignalRO, ':POS_AO_RBV', kind='hinted')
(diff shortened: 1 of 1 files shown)


hhslepicka:
It seems wrong that users should be able to specify the current state of a device.
Also, this field is being used as the `ERROR` information for the widgets.

n-wbrown:
This isn't so much a interpreted state but a state that we set to determine the control mode for this valve.

pcdsdevices/valve.py
```diff
@@ -272,5 +272,5 @@ class VCN(Device):
                        doc='interlock ok status')
     open_command = Cpt(EpicsSignalWithRBV, ':OPN_SW', kind='normal',
                        doc='Epics command to Open valve')
-    state = Cpt(EpicsSignalRO, ':STATE_RBV', kind='hinted', doc='Valve state')
+    state = Cpt(EpicsSignalWithRBV, ':STATE_RBV', kind='hinted', doc='Valve state')
```
klauer:
```suggestion
    state = Cpt(EpicsSignalWithRBV, ':STATE', kind='hinted', doc='Valve state')
```

hhslepicka:
@n-wbrown in this case the widget is wrong I would say due to the fact that it is not the state of the device based on some feedback but based on what operator sets.
@ghalym would you mind to confirm, please?

pcdsdevices/valve.py
n-wbrown: Added. Thanks!

ZLLentz:
pcdsdevices travis is hopelessly broken right now, ignoring

merging with the :+1: review

sad that #389 is being deliberately ignored in a systematic way

ZLLentz:
After lockup I will be enforcing #389 in all PRs unless someone convinces me that the current practice is better

PR #400, opened 2020-03-20
ENH: New devices

### LINKED ISSUE #384: BaseInterface needs to be added as parent of ipm.py classes
And maybe others

ZryletTC:
## Description
Creates new classes for goniometers, von Hamos spectrometers, Beckhoff liquid jets, TimeTools, and PFLSes. Adds `CombinedInOutRecordPositioner` for use with X/Y combined state devices.
Also included minor other fixes.

## Motivation and Context
Need support for these devices in new python.
Also fixes #384.

## How Has This Been Tested?
Created tests for new devices. All tests pass.

Added docstrings. Their style will be fixed in upcoming PR.


ZryletTC:
Merged in pre-commit changes.

pcdsdevices/analog_signals.py
```diff
@@ -14,10 +14,10 @@ class Acromag(Device, BaseInterface):
 
     Parameters:
     -----------
-    prefix : str
+    prefix : ``str``
```
klauer, line 17: Backticks around builtin types is a bit weird. We generally follow the numpy docstring standard:
https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

ZryletTC: Ahh 👌 I'll remove them

ZryletTC: The only times I see backticks mentioned in that guide are for variables names and equations. Should I remove our other usages too?

klauer: It's not as clear cut in other scenarios, in my opinion. That jumps into sphinx territory from numpy docstring territory, and I'm not as much of a sphinx expert as I should be (so anyone feel free to correct me if I'm wrong).

Looking at the attenuator one, I see:
```
``Component``
```

Should really be:
```
:class:`ophyd.Component`
```

That one could probably be forgiven as it's obvious to us (PCDS).

References to arguments can be in backticks, such as:
```
``argument_name`` in :func:`function_name`.
```

Method name references should be like:

```
:meth:`class.method`
```

More here:
http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#the-python-domain
Note that we usually strip off the domain `:py` at the beginning by convention.

ZryletTC: Got it 👍
How about for values like ``True`` or ``filter1`` or ``-1``?
I didn't see any explicit mention about a convention for them in either sphinx or numpy docs although in some spots. It looks like they are not backticked but I can imagine cases when you'd want them to be.

ZryletTC: Never mind, using quotes around strings handles those situations so I guess values should never be backticked.

klauer: I forget, is there a `:keyword:` option? `:keyword:True` would be technically correct if so. I don't see a big problem with back-ticked `True` though, really.

I think it's fine to reference argument names by backticks or values you want to stand out.

Aside: sorry if I'm dragging down the PR with these stylistic things over functional things. Though maybe some good can come of it with one of us making a "docstring / sphinx style guide" of sorts on confluence...

ZryletTC: After this PR goes through, I'l try writing one up

klauer: That'd be really great!

ZryletTC:
I decided to make the docstring style consistent in the rest of the repo so I'll add that as a separate PR because there a lot of changes. Style changes for these files will be included there.

ZryletTC:
Docs aren't building, please don't merge until fixed.

pcdsdevices/ipm.py
```diff
@@ -174,6 +194,9 @@ class IPIMB(Device):
     delay: delay of trigger relative to input EVR
     trigger:
     """
+
+    tab_whitelist = ['isum', 'xpos', 'ypos']
+
     isum = Cpt(EpicsSignal, ':SUM', kind='hinted')
```
rajanplumley, line 200: should all these be digitizer signals `EpicsSignalRO`...?

ZryletTC: Yeah, they totally should :/

pcdsdevices/ipm.py
```diff
@@ -116,11 +133,11 @@ def remove(self, moved_cb=None, timeout=None, wait=False):
 
     @property
     def transmission(self):
-        """Returns the target's transmission value."""
-        return self.target.transmission
+        """Returns the combined transmission value of the target and diode"""
+        return self.target.transmission * self.diode.transmission
```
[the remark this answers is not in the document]
ZryletTC: This chooses 0.2 for unknown, it could be brought down to 0 if desired.
In and out are both full transmission.

pcdsdevices/pim.py
```diff
@@ -214,25 +214,25 @@ class PIMWithBoth(PIMWithFocus, PIMWithLED):
...
 
-    prefix_zoom : str, optional
+    prefix_zoom : ``str``, optional
         The EPICS base PV of the zoom motor. If None, it will be attempted to
         be inferred from `prefix`
 
-    prefix_focus : str, optional
+    prefix_focus : ``str``, optional
         The EPICS base PV of the focus motor. If None, it will be attempted to
         be inferred from `prefix`
```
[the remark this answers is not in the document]
ZryletTC: Opened as new issue #412

pcdsdevices/jet.py
```diff
@@ -0,0 +1,39 @@
+"""
```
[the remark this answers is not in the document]
ZryletTC: Yes, I will create the IMS version as part of my port of jet_tracking devices into pcdsdevices.
Ok, I'll have to learn more about the chemRIXS setup then I'll try to put that in too.

pcdsdevices/lens.py
```diff
@@ -44,6 +44,29 @@ def __init__(self, prefix, *, name, **kwargs):
...
+    prefix : ``str``
+        The EPICS base of the TimeTool
+
+    name : ``str``
+        A name to refer to the device
+    """
+
+    # Clear the default in/out state list. List will be populated during
+    # init to grab appropriate state names for each target.
+    states_list = []
```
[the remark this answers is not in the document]
ZryletTC: Will include during implementation of #411.

pcdsdevices/gon.py
```diff
@@ -0,0 +1,141 @@
...
+        The EPICS base PV of the sample-stage's tilt motor.
+
+    prefix_x : ``str``
+        The EPICS base PV of the sample-stage's x motor.
+
+    prefix_y : ``str``
+        The EPICS base PV of the sample-stage's y motor.
+
+    prefix_z : ``str``
+        The EPICS base PV of the sample-stage's z motor.
```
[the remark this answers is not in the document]
ZryletTC: This could be implemented either by moving these three motors to a separate class or by utilizing #407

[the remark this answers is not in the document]
ZLLentz: With #407 you would still want these to be in a separate class.

ZLLentz:
You don't have to fix the docs in this PR, we can do that in a separate PR.

pcdsdevices/ipm.py
```diff
@@ -103,6 +113,13 @@ def removed(self):
         return (self.target.removed and
                 (self.diode.removed or self.diode.inserted))
 
+    def insert(self, moved_cb=None, timeout=None, wait=False):
+        """Move both the target and diode in"""
+        return (self.target.insert(moved_cb=moved_cb, timeout=timeout,
```
ZLLentz, line 118: Can you remind me which target is selected when `insert` is called here? I remember there being four to choose between.

pcdsdevices/tt.py
```diff
@@ -0,0 +1,67 @@
+"""
```
ZLLentz, line 1: I think it's risky to make a module with such a short name. Can this be expanded to `timetool.py`?

pcdsdevices/ipm.py
ZryletTC: Should be the first one.

ZLLentz: Got it, and I assume it does nothing if we're already in an inserted state?

pcdsdevices/gon.py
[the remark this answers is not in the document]
ZryletTC: Ok, I'll move them now and can up the `UnrelatedComponent`s later

pcdsdevices/ipm.py
ZryletTC: Looks like it doesn't but I feel like it definitely should. This is inherited from InOutPositioner, is there any reason we'd want any other InOutPositioners to not have this behavior?

ZLLentz: I feel like the desired behavior is:
- always insert the diode
- move the target to T1 if it isn't at a specific target

pcdsdevices/ipm.py
ZryletTC: Talked with Mike and sorted it out. Will push new read-onlys plus an extra configuration PV for Wave8 soon.

pcdsdevices/ipm.py
ZryletTC: Can you clarify what you mean by a specific target?
> move the target to T1 if it isn't at a specific target

ZLLentz: `OUT` or `Unknown`

I assume calling `insert` and switching from `T4` to `T1` will cause some confusion

ZryletTC: Sorry, I'm getting a little confused by what you're saying. Simply, we want the below, right?
```suggestion
        if self.removed:
            return NullStatus()
        return (self.target.insert(moved_cb=moved_cb, timeout=timeout,
```

ZLLentz: I think the base class is correct to change here, it seems to have always been wrong. This is not the right part change though, we want to hit the `insert` method, not the `inserted` property. Making it look like `remove` is probably correct.

ZryletTC:
Doc build fix merged in.

pcdsdevices/ipm.py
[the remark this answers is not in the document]
silkenelson: I think I prefer 0 under the assumption that we are not going to use the transmission of this calculation as safety device, but only as a notification that something is (might be..) in the beam. This can be changed later.

pcdsdevices/gon.py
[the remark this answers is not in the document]
ZryletTC: Will be addressed by an implementation of #418.

ZryletTC:
> I think the XCS goniometer is now missing its XYZ motors. As they are not in base anymore, the "withArm" goniometer needs an XYZ stage object attached to it.

Is the XYZ stage not removable in XCS? Either way, I can add them in my implementation of #418.

ZryletTC:
Raj showed me a time when the XYZ stage was removed for an experiment so I will add them as with #418.

klauer:
🎉

PR #401, opened 2020-03-20
DEV: Pre-commit

ZryletTC:
## Description
Adds pre-commit to the repo with the following hooks:
- no-commit-to-branch
- trailing-whitespace
- end-of-file-fixer
- check-ast
- check-case-conflict
- check-json
- check-merge-conflict
- check-symlinks
- check-xml
- check-yaml
- debug-statements
- flake8
- isort

## Motivation and Context
Good tool to prevent bad style, syntax, commits, etc...

## How Has This Been Tested?
CI build passes.

It hasn't been

BTW: I accidentally created this branch off of my `newdevices` branch so I had to rebase and I'm not 100% sure it went alright.


klauer:
I like it. Can you bring these settings back to the pre-commit-hooks repo as our new recommended defaults?

We might ignore `meta.yaml` generally (without a path), but we could revise as that goes.

klauer:
I was thinking about that, @hhslepicka - I wonder if we should have a `pre-commit` job in our matrix. I think it could be a `pcds-ci-tools` thing eventually, but it could grow out of our testing here.

Something like:
```
matrix:
  include:
    name: pre-commit checks
    python: 3.7
    env:
      PRE_COMMIT=1


script:
    - |
      if [ "$PRE_COMMIT" ]; then
        pip install pre-commit
        pre-commit install
        pre-commit run --all-files  # or whatever it is
        exit $?  #  exit with pre-commits return code
      fi
```

Depending on what we have in the repo, we might want to xfail it.

klauer:
Let's gooo

PR #402, opened 2020-03-20
ENH: Use entry_points to vendor happi containers.

hhslepicka:
## Description

Create a new module to host happi containers specific to LCLS.
The containers will be discovered and loaded into Happi via the entry_points `happi.containers`.
See PR: https://github.com/pcdshub/happi/pull/126

## Motivation and Context

To move the containers closer to devices and get LCLS specific code outside of Happi.

## How Has This Been Tested?

Locally

Nowhere.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,14 +2,15 @@ from setuptools import find_packages, setup
 
 import versioneer
 
-setup(name='pcdsdevices',
-      version=versioneer.get_version(),
-      cmdclass=versioneer.get_cmdclass(),
-      license='BSD',
-      author='SLAC National Accelerator Laboratory',
-      packages=find_packages(),
-      description='IOC definitions for LCLS Beamline Devices',
-      entry_points={
-            'happi.containers': ['lcls = pcdsdevices.happi.containers'],
-      }
+setup(
+    name='pcdsdevices',
+    version=versioneer.get_version(),
+    cmdclass=versioneer.get_cmdclass(),
+    license='BSD',
+    author='SLAC National Accelerator Laboratory',
+    packages=find_packages(),
+    description='IOC definitions for LCLS Beamline Devices',
+    entry_points={
+        'happi.containers': ['pcdsdevices = pcdsdevices.happi.containers'],
+    }
 )
--- a/setup.py
+++ b/setup.py
@@ -9,4 +9,7 @@ setup(name='pcdsdevices',
       author='SLAC National Accelerator Laboratory',
       packages=find_packages(),
       description='IOC definitions for LCLS Beamline Devices',
-      )
+      entry_points={
+            'happi.containers': ['lcls = pcdsdevices.happi.containers'],
+      }
+)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,13 +4,14 @@ import warnings
 from pathlib import Path
 
 import pytest
+from pytmc.pragmas import normalize_io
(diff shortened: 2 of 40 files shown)


setup.py
```diff
@@ -8,4 +8,7 @@
       author='SLAC National Accelerator Laboratory',
       packages=find_packages(),
       description='IOC definitions for LCLS Beamline Devices',
-      )
+      entry_points={
+            'happi.containers': ['lcls = pcdsdevices.happi.containers'],
```
klauer: What does the `lcls` mean here? 

I talk somewhat frequently about using entrypoints, but I've admittedly never used it myself - so excuse my ignorance.

hhslepicka: name... but seems useless for what we have... :( 
I think it would only matter in the case of a clash if someone else had a `happi.containers` entry with `lcls`. I tried if we could just use it without the `lcls = ` but no deal.

klauer: Gotcha. How about making it the repo name by convention? `lcls` -> `pcdsdevices`

hhslepicka: Sounds good! Lemme fix that and rebase after the PR that you just merged.

hhslepicka: Fixed, rebased and force-pushed.

tests/conftest.py
```diff
@@ -4,13 +4,14 @@
 from pathlib import Path
 
 import pytest
+from pytmc.pragmas import normalize_io
+
```
ZLLentz: Empty line in the wrong place?

ZLLentz: There seem to be a bunch of instances of this scattered

ZLLentz:
I like how the setup looks on this side :+1:, don't want to merge into pcdsdevices until `happi` has the tag though since lots of people are making PRs to this repo and I don't want it to fail by default.

tests/conftest.py
klauer: Not sure why `isort` did that...

danielballan:
Nice to see another good use of entrypoints in the ecosystem. :-)

ZLLentz:
Now that `happi` has a tag, the only sticking point on this PR is that `isort` seems to have behaved incorrectly on many files here...

klauer:
`isort` master should resolve most of the issues that were of concern, I believe, though that's certainly far from ideal...

ZLLentz:
I'm just going to merge and worry about sort organization later. Any objections?

klauer:
Hold on, I think @hhslepicka has some thing he wanted to fix here, but I'm not sure.

In the meantime, let me try isort master and see if it does improve things as I had recalled.

hhslepicka:
I am still working on this at the Happi side and it may or may not result into changes here.

tests/conftest.py
ZryletTC: It seems to think pytmc is a third party package while ophyd is first party?
This can be fixed by adding a couple lines to setup.cfg:
https://github.com/timothycrosley/isort#configuring-isort

klauer: I traced it back to an issue in isort which is fixed in an as-of-yet unreleased version. Changes are now in c8574ae , which is now just separating the `import` and `from` blocks.

hhslepicka:
Gentle bump... @klauer @ZLLentz and @ZryletTC I think this one is good to go... 
Let me know if you think otherwise.

PR #404, opened 2020-03-23
FIX: "typo" in TwinCAT states class

ZLLentz:
## Description

At some point the TwinCAT Enum had a non-standard setup here, but it was fixed and the class was never adjusted.

## Motivation and Context

All devices that use this class with the newest PLC code do not work


rajanplumley:
blursed PR number

PR #407, opened 2020-04-09
ENH: Add UnrelatedComponent

ZLLentz:
## Description

Add `UnrelatedComponent` to help handle cases where we create "FrankenDevices" where many unrelated components (in the ophyd sense, with unrelated prefixes) are needed to be included in the same ophyd device. Functionally, the devices behave the same for the end-user, but can be implemented much more succinctly in the library here. See the example from the tests:

```
from ophyd.device import Component as Cpt
from pcdsdevices.component import UnrelatedComponent as UCpt

class Basic(Device):
    apple = UCpt(Device)
    sauce = UCpt(Device)
    empty = Cpt(Device, ':EMPTY')

    def __init__(self, prefix, *, name, **kwargs):
        UCpt.collect_prefixes(self, kwargs)
        super().__init__(prefix, name=name, **kwargs)

basic = Basic('PREFIX', name='basic', apple_prefix='APPLE', sauce_prefix='SAUCE')
```
This works recursively as well, e.g. you can have `UnrelatedComponent` point to devices that have `UnrelatedComponent` instances themselves and this will cascade the kwargs sensibly.

## Motivation and Context

I was looking at the CCM class and getting sad about the many levels of passed prefixes here. This should be automated to the extent possible.

## How Has This Been Tested?

I added tests, but need to add one more to verify 3 levels of recursion.

Not yet...

### DIFF
--- a/pcdsdevices/component.py
+++ b/pcdsdevices/component.py
@@ -40,7 +40,7 @@ class UnrelatedComponent(Component):
         # Needs to be non-None or it gets ignored
         if self.suffix is None:
             self.suffix = ''
-        add_prefix = kwargs.get('add_prefix', ['suffix'])
+        add_prefix = list(kwargs.get('add_prefix', ['suffix']))
 
         # Include subdevice UnrelatedComponent
         try:
--- a/pcdsdevices/component.py
+++ b/pcdsdevices/component.py
@@ -40,7 +40,13 @@ class UnrelatedComponent(Component):
         add_prefix = kwargs.get('add_prefix', ['suffix'])
 
         # Include subdevice UnrelatedComponent
-        for cpt_walk in self.cls.walk_components():
+        try:
+            walk_iterator = self.cls.walk_components()
+        except AttributeError:
+            # No subdevices
+            walk_iterator = ()
+
+        for cpt_walk in walk_iterator:
             if isinstance(cpt_walk.item, UnrelatedComponent):
                 name = cpt_walk.dotted_name.replace('.', '_') + '_prefix'
                 add_prefix.append(name)
--- a/pcdsdevices/component.py
+++ b/pcdsdevices/component.py
@@ -14,7 +14,7 @@ class UnrelatedComponent(Component):
     components.
     """
     @classmethod
-    def collect_prefixes(device, kwargs):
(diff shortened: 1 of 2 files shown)


klauer:
I like the idea! Ends up being pretty clean.

danielballan:
Me too.

ZLLentz:
I'm happy to ship this over to main ophyd if you think it has broader use (after refining this a bit more). Doing it in the main library would have some implementation advantages.

I assumed that the proliferation of devices that require 5+ unrelated prefixes was a SLAC fixation.

danielballan:
We've definitely run into this. Our usual recourse is `FormattedComponent`, but I can see some advantages to having an explicit `UnrelatedComponent`.

All of this makes me wish that `Component` itself did not assume concatenation and rather took template strings like `{prefix}...`. If only NSLS didn't put curly braces in PV names!

danielballan:
So, to be clear, :+1: for doing this upstream if you are game.

klauer:
@danielballan you could switch to your own NSLS-II `FormattedComponent` that does `template` string interpolation instead. https://docs.python.org/3/library/string.html#template-strings

It should come as no surprise that I disagree about interpolation vs concatenation. It's a feature that should act as a feedback source that drives proper naming of PVs. That it hasn't worked as intended is our own failing.

danielballan:
Actually, that is persuasive. It sounds like this may have come up before—I don’t remember—and it’s possible I’m more sympathetic to this view than I used to be. Taking the position, “Ophyd thinks PV names SHOULD reflect underlying hierarchy and optimizes for it, while allowing for exceptions,” is something we should document.

klauer:
@danielballan yeah, for sure. You might still consider my `string.Template` suggestion above. Had I known it existed in 2015, I would've used it in the first place to closer match EPICS macros.

ZLLentz:
This has been referenced a few times in other PRs/issues now on this repo, but my focus has shifted a bit and I haven't had a chance to continue working on it. Any objections to including this in `pcdsdevices` as-is and then working on a polished version for `ophyd` later?

klauer:
It's a reasonable approach. We can definitely use this now.

PR #414, opened 2020-04-10
BLD: Replace m2r with recommonmark

### LINKED ISSUE #410: m2r is broken
Docs aren't building because `m2r` is stale and fallen behind compatibility with the `sphinx` API.

Possible solutions:
- Could be fixed by [`m2r`#55](https://github.com/miyakogi/m2r/pull/55)
- If desired, can be bypassed for now by pinning `sphinx<3.0.0` or specifying [the hotfix branch of `m2r`](https://github.com/CrossNox/m2r)
- Realistically, that PR will never be merged as the repo has been inactive since 2018 so a better course of action would probably be to drop `m2r` and switch to `recommonmark` [as others are doing](https://github.com/sphinx-doc/sphinx/issues/7420)

_Originally posted by @ZryletTC in https://github.com/pcdshub/pcdsdevices/pull/400#issuecomment-611848170_

ZryletTC:
## Motivation and Context
Fixes #410

## How Has This Been Tested?
Docs build succeeds.
It's not. This PR and issue should be enough.


ZLLentz:
Is that really all it takes? It follows the same md include directive as m2r?

ZryletTC:
There's still an issue with replacing the `mdinclude::` in index.rst
Going to try to fix that before merge.

matthewfeickert:
> There's still an issue with replacing the `mdinclude::` in index.rst

@ZryletTC Over in pyhf (:wave: from other physics projects) we're facing the same issue ([pyhf Issue 818](https://github.com/scikit-hep/pyhf/issues/818) and [Sphinx Issue 7420](https://github.com/sphinx-doc/sphinx/issues/7420#issuecomment-610076853)). Finding a replacement for `mdinclude` seems to really be the last mile, but I haven't found anything that seems to work. If you end up getting somewhere on this is would be great to hear on the Sphinx Issue.

ZryletTC:
Hey, @matthewfeickert!
Unfortunately, I wasn't able to find any good replacement, the best I could do was #416, which is similar to what it looks like you were trying to do with `pandoc`, but it has the issue of still depending on `m2r`.
Ultimately, it looks like we're going the same way as pyhf and moving to `.rst` at least for now, but I'll let you know if I ever figure out a good solution.

PR #416, opened 2020-04-11
BLD: Fix m2r

### LINKED ISSUE #410: m2r is broken

ZryletTC:
## Description
So it turns out recommonmark isn't a full solution to our situation because it doesn't support include. The best you can do with it is use the markdown file inside a toctree but can't use it as a homepage. I'll leave up #414 in case anyone can figure out how to get it to display as index.html

As an alternative, I offer this (slightly messier) solution:
Since `m2r` is also available as a command line tool, I removed it from the `sphinx` extensions and instead added it as an extra step in the Makefile. Please let me know your thoughts.

## Motivation and Context
Fixes #410 

## How Has This Been Tested?
Tested locally and travis passes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ logs
 
 # Sphinx
 docs/source/generated/*
+docs/source/README.rst
 
 #pytest
 .pytest_cache
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -21,4 +21,5 @@ clean:
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
+	m2r ../README.md --dry-run > source/README.rst
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -43,8 +43,7 @@ extensions = ['sphinx.ext.autodoc',
               'sphinx.ext.napoleon',
               'IPython.sphinxext.ipython_directive',
               'IPython.sphinxext.ipython_console_highlighting',
-              'sphinx.ext.autosectionlabel',
-              'm2r'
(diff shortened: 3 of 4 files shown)


klauer:
To see if I fully grasp the extent that m2r is used for, is this it?
1. `README.md` -> `README.rst`
2. `mdinclude` in `index.rst` to include `README.md`

If that's really all, it seems like we should just drop `README.md`, switch to rst, rid ourselves of the dependency, and be done with it...

ZLLentz:
I agree with Ken

ZryletTC:
> To see if I fully grasp the extent that m2r is used for, is this it?
> 
> 1. `README.md` -> `README.rst`
> 2. `mdinclude` in `index.rst` to include `README.md`
> 
> If that's really all, it seems like we should just drop `README.md`, switch to rst, rid ourselves of the dependency, and be done with it...

In our repo, it really only did number 2.
So let's get rid `m2r` and leave the `README.md` as is for Github.
Then, we can either add a static `README.rst` to sources or change `index.rst` to something more like this: https://blueskyproject.io/ophyd/?

klauer:
Sounds good, @ZryletTC.  So long as we're all on the same page of removing the dependency, solve it how you'd like to solve it. 👍 

Moving toward a more useful index page would be nice for sure - if you feel so inspired!

ZryletTC:
I didn't put in the effort to make the index much more useful but the rest should be reflected in #417.

PR #424, opened 2020-04-27
MNT: Bring FakePytmcSignal into main module

### LINKED ISSUE #422: PytmcSignal breaks `ophyd.sim.make_fake_device`
## Expected Behavior
`ophyd.sim.make_fake_device` with a `PytmcSignal` should create a fake signal that does not try to talk over channel access.

## Current Behavior
`PytmcSignal` does not get replaced with a fake signal.

## Possible Solution
Add this fix on import of pcdsdevices PytmcSignal:
https://github.com/pcdshub/pcdsdevices/blob/6dd968b5bd323446b8ba13cd5190165df9bbdf16/tests/conftest.py#L22-L34

ZLLentz:
## Description

Extract `FakePytmcSignal` from `conftest.py` to `pcdsdevices/signal.py`

## Motivation and Context

Fake devices that include `PytmcSignal` don't work because `PytmcSignal` is not a suitable base class. The conftest patch fixes this for the tests here, but bringing it out allows us to use fake devices that include `PytmcSignal` in other contexts.
closes #422 

## How Has This Been Tested?

If the tests still pass, this worked.

Docstrings edited as appropriate.

### DIFF
--- a/pcdsdevices/signal.py
+++ b/pcdsdevices/signal.py
@@ -91,11 +91,12 @@ class FakePytmcSignal(FakeEpicsSignal):
         return super().__new__(new_cls)
 
     def __init__(self, prefix, io=None, **kwargs):
-        super().__init__(prefix, **kwargs)
+        super().__init__(prefix + '_RBV', **kwargs)
 
 
 class FakePytmcSignalRW(FakePytmcSignal, FakeEpicsSignal):
-    pass
+    def __init__(self, prefix, **kwargs):
+        super().__init__(prefix, write_pv=prefix, **kwargs)
 
 
 class FakePytmcSignalRO(FakePytmcSignal, FakeEpicsSignalRO):
--- a/pcdsdevices/signal.py
+++ b/pcdsdevices/signal.py
@@ -45,7 +45,7 @@ class PytmcSignal(EpicsSignalBase):
 
 
 def select_pytmc_class(io=None, *, prefix, write_cls, read_only_cls):
-    """Return the class to use for PytmcSignal's constructor"""
+    """Return the class to use for PytmcSignal's constructor."""
     if io is None:
         # Provide a better error here than "__new__ missing an arg"
         raise ValueError('Must provide an "io" argument to PytmcSignal. '
@@ -59,7 +59,7 @@ def select_pytmc_class(io=None, *, prefix, write_cls, read_only_cls):
 
 
 def pytmc_writable(io):
-    """Returns True if the pytmc io arg represents a writable PV"""
+    """Returns True if the pytmc io arg represents a writable PV."""
     norm = normalize_io(io)
     if norm == 'output':
         return True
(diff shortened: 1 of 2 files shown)


pcdsdevices/signal.py
```diff
@@ -39,36 +39,52 @@ def __new__(cls, prefix, io=None, **kwargs):
...
+# Make sure an acceptable fake class is set for PytmcSignal
+def FakePytmcSignal(prefix, *, io, **kwargs):
+    """Returns a suitable fake class for a PytmcSignal"""
+    if pytmc_writable(io):
+        return FakeEpicsSignal(prefix, **kwargs)
+    else:
+        return FakeEpicsSignalRO(prefix, **kwargs)
+
+
+fake_device_cache[PytmcSignal] = FakePytmcSignal
```
klauer, line 85:
```suggestion
# NOTE: This is an *on-import* update of the ophyd "fake" device cache
fake_device_cache[PytmcSignal] = FakePytmcSignal
```

ZLLentz:
This actually needs some editing to make it work properly with `typhos`, which freaks out over `FakePytmcSignal` being a function instead of a class- it can't tell if it is supposed to be a signal or a device.

pcdsdevices/signal.py
```diff
@@ -39,36 +39,52 @@ def __new__(cls, prefix, io=None, **kwargs):
...
-    """
+    """Read-only connection to a pytmc-generated EPICS record"""
     pass
 
 
+# Make sure an acceptable fake class is set for PytmcSignal
+def FakePytmcSignal(prefix, *, io, **kwargs):
+    """Returns a suitable fake class for a PytmcSignal"""
+    if pytmc_writable(io):
+        return FakeEpicsSignal(prefix, **kwargs)
```
klauer, line 80:
```suggestion
        return FakeEpicsSignal(prefix + '_RBV', write_pv=prefix, **kwargs)
```

ZryletTC:
> This actually needs some editing to make it work properly with `typhos`, which freaks out over `FakePytmcSignal` being a function instead of a class- it can't tell if it is supposed to be a signal or a device.

Could that be solved by stuffing that logic into the __new__ method of a class, like how PytmcSignal is set up?

ZLLentz:
Ever since writing that comment I've been trying to get it to work, using the strategy you described. Once I have it working I'll edit based on the review comments.

ZLLentz:
I think this is in an acceptable state now, though I'm not convinced I did it perfectly cleanly. I'll merge when I start tomorrow if Travis was happy.

klauer:
I like how you've broken out the functionality into reusable methods. Merge at will.

ZLLentz:
One small lesson learned from this PR, that maybe should have been apparent ahead of time: if we pass factory functions instead of classes for `Component`s, there is no way to inspect the device hierarchy. This will break `typhos` outright and cause more subtle bugs elsewhere.

klauer:
I wonder if ophyd should warn when `Component.cls` is merely callable and not a class?

ZLLentz:
I think you could even argue it should raise an exception outright... the `Component` docs specify that it should be a `class`: https://github.com/bluesky/ophyd/blob/1b0be51a095388376f90162fb08c5c91f8291175/ophyd/device.py#L70

So it wouldn't even be an API change per se... but a warning is probably more appropriate because we don't how how widely this "feature" has been used.

klauer:
Good point. 👍  Who wants to be the issue maker?

ZLLentz:
I'll make an issue, I have some things to do over in ophyd-land today anyway

PR #426, opened 2020-04-28
MNT: kluge to make tests pass for now

### LINKED ISSUE #425: Many Tests Fail with Latest Ophyd
## Expected Behavior

No regressions

## Current Behavior

Explosions

1. Update to ophyd=1.4.1
2. run the tests

## Context

Travis is still running the 1.4.0rc4 build and passing, but locally with ophyd 1.4.1 many tests don't pass. Locally with ophyd 1.4.0rc4 every tests passes.

ZLLentz:
## Description

Override `ophyd`'s `EpicsMotor` limit signals with `auto_monitor=True`.
Replace our signals with the new names decided upon in `ophyd`.

## Motivation and Context

`subscribe` on a non-auto_monitored signal blocks on connection, which is not ok during init. The subscribe calls were added in a recent `ophyd` PR. This PR would not be needed at all if https://github.com/bluesky/ophyd/issues/847 was solved. I will make an ophyd PR to similarly stopgap this on the ophyd end, but this PR will still be needed after that to keep providing some of the methods here. (and then we'd be able to drop the redundant component definitions)

closes #425

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -18,7 +18,7 @@ requirements:
 
   run:
     - python >=3.6
-    - ophyd 1.4.0rc4|>=1.4.0
+    - ophyd >=1.4.1
     - bluesky >=1.2.0
     - pyepics >=3.4.1
     - pytmc >=2.4.0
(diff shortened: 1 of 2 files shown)


ZryletTC:
Did you know if travis fails if pinned to new ophyd without the fix?

ZLLentz:
I did not run a travis test with just the pin to see if it fails. I'll set up a branch for that really quick to verify.

ZLLentz:
https://travis-ci.org/github/ZLLentz/pcdsdevices/builds/680689720
from the newly-minted "break-the-build" branch

ZLLentz:
Confirmed that it fails without the edits to epics_motor here

ZryletTC:
Cool 👍
So it wasn't just me having a slow computer 😆

PR #427, opened 2020-05-12
FIX: importability on win32

klauer:
This is not necessary to actually merge, as we have no _real_ use case for Windows instantiations of pcdsdevices.

However, for typhos testing I wanted to show the same device without copy/pasting it out of the library, and these were the fixes necessary to do so.

Here's what typhos + pcdsdevices on Windows looks like, for those interested:
image

### DIFF
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -1,7 +1,6 @@
 """
 Module for defining bell-and-whistles movement features
 """
-import fcntl
 import logging
 import numbers
 import re
@@ -20,6 +19,11 @@ from ophyd.ophydobj import OphydObject
 from ophyd.status import wait as status_wait
 
 from . import utils as util
+
+try:
+    import fcntl
+except ImportError:
+    fcntl = None
 
 logger = logging.getLogger(__name__)
 engineering_mode = True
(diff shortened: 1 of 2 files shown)


ZLLentz:
If it helps with testing, then it is probably worth including the patch

klauer:
OK, that seems fair. Is our testing broken here?

ZLLentz:
Yes, the tests look super broken... I'm going to re-run the latest master branch test to confirm

ZLLentz:
Master actually fails worse than this one does... Somehow we've regressed on the style check, of all things. https://travis-ci.org/github/pcdshub/pcdsdevices/builds/680754216

hhslepicka:
@ZLLentz that is due to the latest Flake8 which got more strict.

klauer:
Huh, that's surprising. We've been seeing the new flake8 be more strict about certain things, but those specific errors are all old, clear-cut flake8 violations.

ZLLentz:
Gotcha. In any case, this PR does not introduce any new failures.

ZLLentz:
The test fix was merged into master

ZLLentz:
Seems like it doesn't disrupt anything, so if this will be helpful for testing I would like to merge.

klauer:
GitHub didn't accept the approval while it was a draft - mind re-approving @ZLLentz?

PR #430, opened 2020-05-15
MNT: Ophyd 1.5 Changes

### LINKED ISSUE #423: obj.value is deprecated (+maybe others)
## Expected Behavior
Tests pass without errors or warnings 

## Current Behavior
Lots of tests pass but use deprecated attributes like `.value` and therefore spout off warnings like:
```
tests/test_valve.py::test_valve_motion
  /Users/pennebak/Documents/GitHub/pcdsdevices/tests/test_valve.py:113: UserWarning: You have called obj.value on FakeEpicsSignal(name='test_valve_command', parent='test_valve', value=1, timestamp=1587761949.875202) (test_valve_command.command) which is a non-monitored signal.
  We are falling back to calling `.get` and interrogating the underlying control system, however this may cause several other problems:
     1. This property access may take an arbitrarily long time
     2. This property access, which you expect to be read only may change other state in the Signal.
  Your options to fix this are:
    - do not use obj.value.
      - If you are using this is in a plan you like want to be using bps.read, bps.rd, bpp.reset_positions_decorator, bpp.reset_positions_wrapper, bpp.relative_set_decorator, or bpp.relative_set_wrapper
      - if you are doing this in an ophyd method use `self.get`
    - set up the Signal to monitor
  
  This behavior will likely change in the future.
    assert valve.command.value == valve.commands.open_valve.value
```

## Possible Solution
Change these according to the suggestions offered in the warnings.
e.g. `.value` -> `.get`

## Your Environment
Local checkout of pcdsdevices
ophyd=1.4.1

### LINKED ISSUE #429: Incompatibility with Ophyd=1.5.0
## Expected Behavior

Tests still pass

## Current Behavior

Tests fail

ZLLentz:
## Description

Update the library for ophyd=1.5.0

- [x] Fix the test suite
- [x] `Signal.value` -> `Signal.get()` (careful to not break `Enum.value`)
- [x] `Status._finished` -> `Status.set_finished` and related changes

The test suite was broken for two main reasons:

- `FakeEpicsSignal` is no longer a usable stand-in for `EpicsSignal` in `EpicsMotor` after the limit-updating patch. Long term I believe this class should be replaced with a true `sim` control layer at `ophyd`.
- `Status` objects are no longer guaranteed to be ready immediately when checked from the same thread that calls `_finished/set_finished/set_exception`... or at least I think that's what's going on. Adding waits fixed it.

Additionally, there was one test that was simply broken and I could not figure out how it originally passed. I adjusted this test appropriately.

## Motivation and Context

The tests are broken, there are new deprecation warnings, and there are new toys.
closes #423 
closes #429 

## How Has This Been Tested?

The tests still pass after my mucking about, and the number of deprecation warnings has been minimized. A few remain but I do not think they originate from this repo.

These are internals that will likely only be documented via this PR and the release notes.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -18,7 +18,7 @@ requirements:
 
   run:
     - python >=3.6
-    - ophyd >=1.4.1
+    - ophyd >=1.5.0
     - bluesky >=1.2.0
     - pyepics >=3.4.1
     - pytmc >=2.4.0
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -5,7 +5,6 @@ import logging
 import time
 
 import numpy as np
-
 from ophyd.device import Component as Cpt
 from ophyd.device import Device
 from ophyd.device import FormattedComponent as FCpt
@@ -199,7 +198,7 @@ class AttBase(FltMvInterface, PVPositioner):
                 self._original_vals[filt.state] = filt.out_states[0]
             # Otherwise, remember so we can restore
             else:
-                self._original_vals[filt.state] = filt.state.value
+                self._original_vals[filt.state] = filt.state.get()
         return super().stage()
 
     def _setup_move(self, position):
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -384,7 +384,9 @@ class IMS(PCDSMotorBase):
         # Check that we need to actually set the flag
         if flag_is_cleared(value=self.bit_status.get()):
             logger.debug("%s flag is not currently active", flag)
-            return DeviceStatus(self, done=True, success=True)
+            st = DeviceStatus(self)
+            st.set_finished()
+            return st
 
         # Issue our command
         logger.info('Clearing %s flag ...', flag)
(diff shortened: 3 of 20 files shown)


ZLLentz:
The next thing that seems to show up everywhere in the test suite is that `FakeEpicsSignal` has fallen out of sync with `EpicsSignal`. I will rework `FakeEpicsSignal` tomorrow and based on my rework make an upstream PR to `ophyd`. I'm leaning towards scrapping the old class entirely and making `FakeEpicsSignal` a subclass of `EpicsSignal` with a hard-coded dummy cl, plus some testing utilities that match the current fake class API. This way it will not fall out of sync again.

klauer:
Would you be against absorbing commits from #427 into this PR?

ZLLentz:
It's probably better to just merge #427, right?

klauer:
~If you're OK with the failing tests, that's probably the preferable alternative. My assumption was you'd get everything passing here, and then merge.~

Actually, scratch that - just ignore me. Get this PR working, then we rebase/merge that piddly PR after it passes. Easy enough.

ZLLentz:
It turns out that doing this the "correct" way is a lot of effort, and probably needs some additional thought to pick the best way to proceed. I'm going to look for a "quick" way so I can move towards getting a new working environment soon.

ZLLentz:
The main lingering issue here is the `ophyd`/`pyepics` cleanup bug. This is pretty annoying when running the tests locally, hopefully it finishes soon enough for a Travis build to mark it as passing...

ZLLentz:
A quick summary because I think it is a good API change:
Instead of calling `Status._finished(success=True)` or `Status._finished(success=False)`, the new API is to call `Status.set_finished()` (implying success) and `Status.set_exception(exc)`. This gets rid having a "hidden" method that the application developer definitely needs to call and it allows us to pass in the precise reason why a status failed via an exception. This exception will be raised if we use the new `Status.wait()` method to wait on the status, or returned if we use the new `Status.exception()` method instead.

PR #436, opened 2020-05-26
ENH: Add support for MODS zoom telescope

slactjohnson:
## Description
Adds in support for the zoom telescope that will be used in the MODS. 

## Motivation and Context
This allows for control of the zoom telescope via python, and generates a decent screen via Typhos.

## How Has This Been Tested?
Tested this on a zoom telescope IOC running in the Heinz lab. 

Comments within the code.


ZLLentz:
Does this belong in something like a `mods` module? Or do you expect it to be used as a de-facto zoom telescope in other places?

pcdsdevices/zoomtelescope.py
```diff
@@ -0,0 +1,95 @@
...
+    _max_zoom = Cpt(EpicsSignalRO, ':CAL:MAX_ZOOM', kind='config')
+
+    # Control parameters
+    _req_zoom = Cpt(EpicsSignal, ':REQ_ZOOM', kind='normal')
+
+    _tweak_plus = Cpt(EpicsSignal, ':TWEAK_ZOOM_PLUS.PROC', kind='normal')
+    _tweak_minus = Cpt(EpicsSignal, ':TWEAK_ZOOM_MINUS.PROC', kind='normal')
+    _tweak_val = Cpt(EpicsSignal, ':TWEAK', kind='normal')
+
+    _permission = Cpt(EpicsSignalRO, ':PERMISSION_STATUS', kind='normal')
```
ZLLentz, line 60: Is it your intention to make every component hidden from the user? This is unusual for an ophyd device.

pcdsdevices/zoomtelescope.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @property
+    def tweak_val(self):
+        """The amount to tweak the magnification when calling tweak_plus or
+        tweak_minus."""
+        tweak = self._tweak_val.get()
```
ZLLentz, line 66: We need to be careful of doing EPICS gets in attribute access, this can cause various problems because getattr is expected to be short/instant.

slactjohnson: So the alternative would be skip the properties and just use <component>.get()/.put(<val>)? 

I like using properties for this because I get tired of adding .put()/.get() at the end of everything (especially interactively), but if it's problematic then I can switch to this method.

pcdsdevices/zoomtelescope.py
slactjohnson: The idea was to use properties to access the most useful bits, and hide the configuration parameters for debugging purposes. Most of those PVs a typical user should never touch (I debated in the comments as to whether I should include them, but erred on the side of having more information).

slactjohnson:
@ZLLentz I debated as to whether this and other MODS devices should go in a MODS module, but I think that the majority of the devices that we have built for the MODS could end up in HXR hutches as well. They're already talking about putting a couple Tuttifrutti and other MODS assemblies in XPP for incoupling from beam transport. I think I could go either way on this. If you think that these belong in their own module then I can do it that way. But I also anticipate a MODS module for the TILE ophyd devices, so then MODS module would be getting pretty monolithic at that point. I guess something like a "pcdslaserdevices" repo is an option as well.

silkenelson:
I like pcdslaserdevices.

ZLLentz:
We should be cautious about making something like `pcdslaserdevices` because it becomes really hard to track down every device for systematic updates if we split too far. I'm repeatedly having issues getting environments deployed because `hxrsnd` and `transfocate` fall out of date and need to be updated in exactly the same ways as done here. But maybe this is a knock on `ophyd` for causing me too many downstream issues.

I would recommend instead making these subfolders in this module, e.g. we would be able to import from `pcdsdevices.lasers.zoomtelescope`

In fact the current pcdsdevices module is a bit messy and needs reorganizing as above. So if you don't want to do that now I'll hit it in a future update.

slactjohnson:
I like that better. I might as well organize them that way now.

pcdsdevices/zoomtelescope.py
ZLLentz: I'm going to merge this with the getattr/setattr as long as you're sure it's what you want with these devices. So take what I'm saying here (and in most other comments) as a suggestion rather than a mandate.

I agree that the getattr/setattr is convenient. It becomes confusing though as some things are getattr/setattr and some things are not. Does the user need to memorize which devices have this set up?

A long getattr causes the most frequent issues with tab completion in ipython. This is usually why the tab will be slow. Perhaps there's no reason to care about the slow tabs when your device is disconnected though.

klauer: Properties are _really_ convenient - there's no arguing there. I was a big proponent of using them everywhere in ophyd and devices in general, but have slowly shifted my opinion over time.

Consider the following:
* Properties hide `args` and `kwargs` that might get passed to `.get` and `.set` (or `.put`)
* Properties may raise `TimeoutError` making `device.tweak_val` take a few seconds in places where you wouldn't expect it to
* Properties don't get mapped into the bluesky data model (that is, `.read()` and `.describe()` won't know about them)
* Properties won't return a status object indicating completion in the background (as `.set()` will)

The solution for most of these is "get used to calling `.get`/`.set`", but also to consider adding components (making them visible to bluesky) with alternative signal classes such as `ophyd.signal.DerivedSignal` or `ophyd.signal.AttributeSignal`. Unfortunately, documentation of these is nonexistent outside of the source code. ☹️

slactjohnson: Thank you both for your sage advice. Consider me a convert. :) I'll make this change. 

I will take a look at DerivedSignal and Attribute signal as well. Thanks!

ZryletTC:

slactjohnson:
@ZLLentz Yes, I am.

PR #437, opened 2020-05-26
ENH: Add support for Thorlabs Elliptec stages

slactjohnson:
## Description
This adds support for the Thorlabs elliptec stages which will be used throughout the MODS system. 

## Motivation and Context
This provides control of the Thorlabs elliptec stages via python, and generates reasonable screens via Typhos.

## How Has This Been Tested?
I tested this on elliptec stages setup in the Heinz lab. 

Some documentation in comments.


pcdsdevices/elliptec.py
```diff
@@ -0,0 +1,62 @@
...
+from ophyd import Device
+from ophyd.signal import EpicsSignal, EpicsSignalRO
+
+
+class ELL_Base(Device):
+    """
+    Base class for Elliptec stages.
+
+    Sufficient for control of ELL17/20 (28/60mm linear stages) and ELL14/18
+    (rotation) stages.
```
ZLLentz, line 16: This class looks very much like an `ophyd` `PVPositioner`. If that class fits, I suggest we use it as the base class so we can move this like a motor.

pcdsdevices/elliptec.py
```diff
@@ -0,0 +1,62 @@
...
+Classes for ThorLabs Elliptec motors.
+"""
+
+
+from ophyd import Component as Cpt
+from ophyd import Device
+from ophyd.signal import EpicsSignal, EpicsSignalRO
+
+
+class ELL_Base(Device):
```
ZLLentz, line 11: The Python convention for class names from PEP-8 are CamelCase without underscores.

pcdsdevices/elliptec.py
```diff
@@ -0,0 +1,62 @@
...
+    _Current_EGU = Cpt(EpicsSignal, ':CURPOS.EGU', name='Current EGU',
+                       kind='config')
+    _Status = Cpt(EpicsSignalRO, ':STATUS', name='Status', kind='normal')
+    _From_Addr = Cpt(EpicsSignal, ':FROM_ADR', name='From Address',
+                     kind='config')
+    _To_Addr = Cpt(EpicsSignal, ':TO_ADR', name='To Address',
+                   kind='config')
+    _Save_Addr = Cpt(EpicsSignal, ':SAVE', name='Save Data', kind='config')
+    _Command = Cpt(EpicsSignal, ':CMD', name='Command', kind='config')
+    _Response = Cpt(EpicsSignalRO, ':RESPONSE', name='Response', kind='config')
```
ZLLentz, line 42: Is it also your intention here to make all of the components hidden from the user?

pcdsdevices/elliptec.py
```diff
@@ -0,0 +1,62 @@
...
+                           kind='normal')
+    _Target_Precision = Cpt(EpicsSignal, ':MOVE.PREC', name='Target Precision',
+                            kind='config')
+    _Target_EGU = Cpt(EpicsSignal, ':MOVE.EGU', name='Target EGU',
+                      kind='config')
+    _Current_Position = Cpt(EpicsSignalRO, ':CURPOS', name='Current Position',
+                            kind='normal')
+    _Current_Precision = Cpt(EpicsSignal, ':CURPOS.PREC',
+                             name='Current Precision', kind='config')
+    _Current_EGU = Cpt(EpicsSignal, ':CURPOS.EGU', name='Current EGU',
```
ZLLentz, line 33: Renaming signals here can have undesirable consequences in `bluesky` scans, where any two devices of this type will have exactly the same data keys for the corresponding values.

ZLLentz:
Does this one also belong in a `mods` module?

pcdsdevices/elliptec.py
slactjohnson: I'll try that out and see how that works.

pcdsdevices/elliptec.py
slactjohnson: Got it. I'll fix that here and elsewhere.

pcdsdevices/elliptec.py
slactjohnson: Many of these the casual user will never need. But, I should expose the most useful ones, or provide methods to access them. I'll add this.

ZryletTC:

pcdsdevices/lasers/elliptec.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class Ell9(Ell6):
+    """
+    Class for Thorlabs ELL9 4 position filter slider.
+    """
+
+    # Names for slider positions
+    Name2 = Cpt(EpicsSignal, ':NAME2', kind='config')
+    Name3 = Cpt(EpicsSignal, ':NAME3', kind='config')
```
ZLLentz, line 58: The standard naming for python attributes (including `ophyd` components) is lowercase, with underscores if necessary. Can we make this file conform to the pep8 standards? That's the only thing left before I approve and merge.

PR #438, opened 2020-05-26
ENH: Add support for EK9000 + EL3174, and MODS environmental sensors

slactjohnson:
## Description
This adds support for EL3174 AI cards attached to EK9000 as well as the laser environmental sensor board which uses this interface. 

## Motivation and Context
This allows for the creation of a decent screen via Typhos for these devices. 

## How Has This Been Tested?
Tested on a EK9000 running in the Heinz lab. 

Some documentation in comments.


pcdsdevices/ek9000.py
```diff
@@ -0,0 +1,39 @@
+"""
+Classes for cards attached to EK9000 bus couplers.
+"""
+
+from ophyd import Component as Cpt
+from ophyd import Device
+from ophyd.signal import EpicsSignal, EpicsSignalRO
+
+
+class EL3174_AI_Ch(Device):
```
ZLLentz, line 10: The Python convention for class names from PEP-8 is CamelCase without underscores.

pcdsdevices/ek9000.py
```diff
@@ -0,0 +1,39 @@
...
+        The PV base of the card.
+    """
+    _Measured = Cpt(EpicsSignalRO, '.VAL', name='Converted Signal',
+                    kind='normal')
+    _Raw_ADC = Cpt(EpicsSignalRO, '.RVAL', name='Raw Signal', kind='config')
+    _EGU = Cpt(EpicsSignal, '.EGU', name='Signal Units', kind='config')
+    _EGU_Max = Cpt(EpicsSignal, '.EGUF', name='EGU Max', kind='config')
+    _EGU_Min = Cpt(EpicsSignal, '.EGUL', name='EGU Min', kind='config')
+    _Conversion = Cpt(EpicsSignal, '.LINR', name='Conv. Method', kind='config')
+    _Precision = Cpt(EpicsSignal, '.PREC', name='Precision', kind='config')
```
ZLLentz, line 28: Same comment about hidden components, `name=` from the other PR

pcdsdevices/ek9000.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    EL3174 analog input card channel.
+
+    Can be used to digitize a variety of sensors.
+
+    Parameters
+    ----------
+    prefix : str
+        The PV base of the card.
+    """
```
ZLLentz, line 20: The PVs here look very generic. Is this a more general class than just for an EL3174? Could it potentially be for the whole EL3XXX series?

pcdsdevices/ek9000.py
```diff
@@ -0,0 +1,39 @@
...
+    _EGU_Min = Cpt(EpicsSignal, '.EGUL', name='EGU Min', kind='config')
+    _Conversion = Cpt(EpicsSignal, '.LINR', name='Conv. Method', kind='config')
+    _Precision = Cpt(EpicsSignal, '.PREC', name='Precision', kind='config')
+
+
+class EnvironmentalMonitor(Device):
+    """
+    Class for the MODS environmental monitoring system, which typically
+    consists of three measurements: P, T, and %RH.
+    """
```
ZLLentz, line 35: If we do make a `mods` module, this would belong in it, importing the analog class from this file.

pcdsdevices/ek9000.py
klauer: Seems like switching these from `name=` to `doc=` would be appropriate.

slactjohnson: Got it, I'll switch to doc then.

pcdsdevices/ek9000.py
slactjohnson: It probably does work for the whole EL3XXX series, but I only have an EL3174 to test with. :)

I can rename this to something more generic.

ZLLentz: Makes sense! We can leave it as-is and modify later if we find that it is generic.

ZryletTC:

pcdsdevices/lasers/ek9000.py
```diff
@@ -0,0 +1,38 @@
...
+
+class EnvironmentalMonitor(Device):
+    """
+    Class for the MODS environmental monitoring system, which typically
+    consists of three measurements: P, T, and %RH.
+    """
+
+    Pressure = Cpt(El3174AiCh, ':1')
+    Humidity = Cpt(El3174AiCh, ':2')
+    Temperature = Cpt(El3174AiCh, ':3')
```
ZLLentz, line 38: The standard naming for python attributes (including `ophyd` components) is lowercase, with underscores if necessary. Can we make this file conform to the pep8 standards? That's the only thing left before I approve and merge.

PR #444, opened 2020-06-05
SXR Test Absorber addition

slacAdpai:
## Summary
Added sxr_test_absorber.py and SxrTestAbsorber class. There are one off components. Although the name is an absorber, its function is a sacrifical stopper for the 1Mhz sxr beam and used when they are ramping up. 

## Motivation and Context
Add the SXR test absorber to pcdsdevices so a gui can appear.

### DIFF
--- a/pcdsdevices/sxr_test_absorber.py
+++ b/pcdsdevices/sxr_test_absorber.py
@@ -19,4 +19,4 @@ class SxrTestAbsorber(Device):
 
     tab_component_names = True
 
-    absorber_vert = Cpt(BeckhoffAxis, ':absorber_vert', kind='normal')
+    absorber_vert = Cpt(BeckhoffAxis, ':MMS:01', kind='normal')
--- a/pcdsdevices/sxr_test_absorber.py
+++ b/pcdsdevices/sxr_test_absorber.py
@@ -1,17 +1,13 @@
 """
-Module for the various absorbers.
+Module for the SXR Test Absorbers.
 """
 from ophyd.device import Component as Cpt
 from ophyd.device import Device
 
 from .epics_motor import BeckhoffAxis
 
-# from ophyd.device import FormattedComponent as FCpt
 
-# from .interface import BaseInterface
-
-
-class TestAbsorber(Device):
+class SxrTestAbsorber(Device):
     """
     SXR Test Absorber: Used for testing the sxr beamline at high pulse rates.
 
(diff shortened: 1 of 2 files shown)


slacAdpai:
Hello Tyler,

Can you help me rebase? I have tried a few times and did not get it to work correctly

ZryletTC:
@slacAdpai Sure, I'll take a look at it after the tailgate

pcdsdevices/sxr_test_absorber.py
```diff
@@ -0,0 +1,22 @@
...
+
+    This device has 1 main members: the stopper/absorber (diamond stopper).
+
+    The vertical motor moves stopper in and out of the beam path in
+    the +/- y direction.
+    """
+
+    tab_component_names = True
+
+    absorber_vert = Cpt(BeckhoffAxis, ':absorber_vert', kind='normal')
```
ZLLentz, line 22: Does this class have anything that isn't just a motor? If not, t may be better to just use `BeckhoffAxis` directly rather than make a special class for it.

slacAdpai: As of right now no. There may be more additions in the future. There are Thermocouples that may be added in the future. The hardware is there, but the specifications were very lax for installation.

slacAdpai: Also, there was no mps requirement for this device but we should still have it because it can block beam.

klauer: I can see both sides of this (keep the class, or just use `BeckhoffAxis`), but some thoughts in favor of keeping it:
* It's a place for documentation
* It's a place where further signals can be added, if needed
* Its special class name allows for a custom screen to be easily created by filename (though certainly in the one motor case, the axis in happi could have this screen information)

ZLLentz: If there are other things to add later then :+1:

ZLLentz, line 22:
```suggestion
    absorber_vert = Cpt(BeckhoffAxis, ':MMS', kind='normal')
```
Since the PV here is `{attribute 'pytmc' := 'pv: ST1K4:TEST:MMS'}`
And I assume the base PV of the device is `ST1K4:TEST`

klauer:
(Closing and re-opening this PR as we're trying to fix our TravisCI Continuous Integration settings)

PR #445, opened 2020-06-09
ENH/WIP: add start of "variety" concept + schema validation

klauer:
## Description
* Proposal for varieties is here, with the schemas written based on this (anyone at Stanford should be able to comment): https://docs.google.com/spreadsheets/d/1nQQQvR-uqtAohpzqdLGxb8heDeT_sttbo3yS93yjE58/edit?ts=5ed199cc#gid=0
* Utilizes the library [schema](https://github.com/keleshev/schema), which seems to work reasonably well. That said, it would not be too difficult to refactor this to be validated with jsonschema or even code (removing the dependency), though this may be more error-prone.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -15,7 +15,6 @@ requirements:
   build:
     - python >=3.6
     - setuptools
-
   run:
     - python >=3.6
     - ophyd >=1.5.0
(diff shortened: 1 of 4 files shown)


klauer:
Premature PR (darn alt+enter); let me retry...

PR #452, opened 2020-06-18
DEV: PIPPLC added AI Offset and AutoOn. Also added VPCNAME and PUMPSI…

slacAdpai:
…ZE. Someone might get unhappy. Bill Scholetter wanted it implemented.

## Description
For PIPPLC, added AI offset and AutoOn  for function block. Also put in non plc but very important parameters for someone viewing the ion pump expert screens. Bill Scholetter wanted it implemented.

### DIFF
--- a/pcdsdevices/pump.py
+++ b/pcdsdevices/pump.py
@@ -173,7 +173,7 @@ class PIPPLC(Device):
                               doc='epics command to switch on the '
                               'high voltage')
     plc_ai_offset = Cpt(EpicsSignalWithRBV, ':AI_Offset', kind='config',
-                        doc=('analog input offset must match ion pump '
+                        doc=('Analog input offset must match ion pump '
                              'analog ouput offset. Default: 13'))
     auto_on = Cpt(EpicsSignalWithRBV, ':Auto_On', kind='config',
                   doc=('Ion pump with automatically turn on when '
--- a/pcdsdevices/pump.py
+++ b/pcdsdevices/pump.py
@@ -172,6 +172,17 @@ class PIPPLC(Device):
     high_voltage_switch = Cpt(EpicsSignalWithRBV, ':HV_SW', kind='config',
                               doc='epics command to switch on the '
                               'high voltage')
+    plc_ai_offset = Cpt(EpicsSignalWithRBV, ':AI_Offset', kind='config',
+                        doc=('analog input offset must match ion pump '
+                             'analog ouput offset. Default: 13'))
+    auto_on = Cpt(EpicsSignalWithRBV, ':Auto_On', kind='config',
+                  doc=('Ion pump with automatically turn on when '
+                       'reference gauge pressure is below protection '
+                       'setpoint'))
+    qpc_name = Cpt(EpicsSignalRO, ':VPCNAME', kind='config',
+                   doc='Quad Ion Pump Controller Name')
+    qpc_pumpsize = Cpt(EpicsSignalRO, ':PUMPSIZE', kind='config',
+                       doc='Ion Pump Size (L/s)')
 
 
 class PTMPLC(Device):
--- a/pcdsdevices/pump.py
+++ b/pcdsdevices/pump.py
@@ -176,7 +176,7 @@ class PIPPLC(Device):
                         doc=('Analog input offset must match ion pump '
                              'analog ouput offset. Default: 13'))
     auto_on = Cpt(EpicsSignalWithRBV, ':Auto_On', kind='config',
-                  doc=('Ion pump with automatically turn on when '
+                  doc=('Setting to automatically turn on the ion pump when the'
                        'reference gauge pressure is below protection '
                        'setpoint'))
     qpc_name = Cpt(EpicsSignalRO, ':VPCNAME', kind='config',


slacAdpai:
Is there a more optimized way to capitalize? I am fine with edits on the doc sections as well

ZLLentz:
You have sentences that start with lowercase letters :shrug:

ZLLentz:
Ok, actually just one sentence. The other is from a previous PR but I got confused.

pcdsdevices/pump.py
```diff
@@ -172,6 +172,17 @@ class PIPPLC(Device):
     high_voltage_switch = Cpt(EpicsSignalWithRBV, ':HV_SW', kind='config',
                               doc='epics command to switch on the '
                               'high voltage')
+    plc_ai_offset = Cpt(EpicsSignalWithRBV, ':AI_Offset', kind='config',
+                        doc=('analog input offset must match ion pump '
```
ZLLentz, line 176:
```suggestion
                        doc=('Analog input offset must match ion pump '
```

slacAdpai:
Thank you Zach!

ZLLentz:
Just waiting on the travis to pass to merge as usual (just in case there is a syntax error typo or some equivalent)
I'll trust that you double-checked the PV names here, because if you didn't you'll just find your way back to making a follow-up PR

pcdsdevices/pump.py
```diff
@@ -172,6 +172,17 @@ class PIPPLC(Device):
...
+                        doc=('Analog input offset must match ion pump '
+                             'analog ouput offset. Default: 13'))
+    auto_on = Cpt(EpicsSignalWithRBV, ':Auto_On', kind='config',
+                  doc=('Ion pump with automatically turn on when '
+                       'reference gauge pressure is below protection '
+                       'setpoint'))
+    qpc_name = Cpt(EpicsSignalRO, ':VPCNAME', kind='config',
+                   doc='Quad Ion Pump Controller Name')
+    qpc_pumpsize = Cpt(EpicsSignalRO, ':PUMPSIZE', kind='config',
+                       doc='Ion Pump Size (L/s)')
```
ZryletTC, line 185: Is it size or max flow?

pcdsdevices/pump.py
```diff
@@ -172,6 +172,17 @@ class PIPPLC(Device):
     high_voltage_switch = Cpt(EpicsSignalWithRBV, ':HV_SW', kind='config',
                               doc='epics command to switch on the '
                               'high voltage')
+    plc_ai_offset = Cpt(EpicsSignalWithRBV, ':AI_Offset', kind='config',
+                        doc=('Analog input offset must match ion pump '
+                             'analog ouput offset. Default: 13'))
+    auto_on = Cpt(EpicsSignalWithRBV, ':Auto_On', kind='config',
+                  doc=('Ion pump with automatically turn on when '
```
ZryletTC, line 179:
```suggestion
                  doc=('Setting to automatically turn on the ion pump when the'
```

pcdsdevices/pump.py
```diff
@@ -172,6 +172,17 @@ class PIPPLC(Device):
...
                               doc='epics command to switch on the '
                               'high voltage')
+    plc_ai_offset = Cpt(EpicsSignalWithRBV, ':AI_Offset', kind='config',
+                        doc=('Analog input offset must match ion pump '
+                             'analog ouput offset. Default: 13'))
+    auto_on = Cpt(EpicsSignalWithRBV, ':Auto_On', kind='config',
+                  doc=('Setting to automatically turn on the ion pump when the'
+                       'reference gauge pressure is below protection '
+                       'setpoint'))
+    qpc_name = Cpt(EpicsSignalRO, ':VPCNAME', kind='config',
```
ZLLentz, line 182: Hang on, is this supposed to be QPCNAME?
E.g.
```suggestion
    qpc_name = Cpt(EpicsSignalRO, ':QPCNAME', kind='config',
```

pcdsdevices/pump.py
slacAdpai: The pump size is measured in liters per second. The reason why these items show up in the pipplc is because that is the object that is on the icon. My thought process, which may be convinced otherwise is that the name pipplc should change to a more generic one like qpcpip. The reason is because in operation, the person on the screen would like to find all relevant information for an ion pump instead of looking through the physical representation of the ion pump controller (which is a ion pump connected to a Quad Ion Pump Controller that controls 4 different ion pumps which is connected to a plc by a dsub and also by ethernet for added functionality). By having a screen that condenses all the complexity onto a screen, hopefully it will be easier to operate. The complexity is for the controls engineer to work through. Maybe there is another way to approach it.

pcdsdevices/pump.py
slacAdpai: According to the vacuum support ioc it is currently VPCNAME. I checked if before added the name in. Yes, it is confusing

ZLLentz: Nice

pcdsdevices/pump.py
ZryletTC: Interesting. Are all PLC PIPs connected through a QPC?

slacAdpai: Yes, all ion pumps in the FEE and EBD are connected through the QPC first. The parts of the PVs come through the plc ioc and other luxurious PVs come through another ioc called vacuum support that talks to the QPC through ethernet.

slacAdpai: The complexity of transducing the pressure to an analog signal is done by the controller so that the plc does not have to do all the lower level sections. On the other hand, Beckhoff is doing the lower level complexity of driving the dumb motors we have for motion

PR #456, opened 2020-06-23
ENH: add concept of "tags" to variety metadata

klauer:
## Description
Adds the 'tags' key to all variety metadata schemas.
'tags' is a `set()` of curated string values that classify components.

An example would be that which is presented in this PR: a "protected" signal. If a GUI client sees this tag, it may require a password or mode-dependent (expert vs user) verification before proceeding.

Additional tags will require follow-up PRs, along with single-line explanations as to what the tag means.
Explanations should be written in a way that they can be used consistently from higher-level applications.
`The component: (tag_reason=* May require additional privileges to change.)`

### DIFF
--- a/pcdsdevices/tags.py
+++ b/pcdsdevices/tags.py
@@ -1,0 +1,13 @@
+tag_to_explanation = {
+    'protected': 'May require additional privileges to change.',
+}
+
+
+def get_valid_tags():
+    """Return a list of all valid variety tags."""
+    return list(tag_to_explanation)
+
+
+def explain_tag(tag_name):
+    """Return a user-friendly explanation of a tag's meaning."""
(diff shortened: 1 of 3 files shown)


pcdsdevices/tags.py
```diff
@@ -0,0 +1,13 @@
+tag_to_explanation = {
+    'protected': 'May require additional privileges to change.',
+}
+
+
+def get_valid_tags():
+    """Return a list of all valid variety tags."""
+    return list(tag_to_explanation)
```
ZLLentz, line 8: I know it's a small thing, but for some reason this collection of tags being a list and every other collection of tags being a set busts my brain

klauer: No, that's definitely valid; fixed up in bada93c

PR #441, opened 2020-05-29
ENH: Add Valve Fast Shutter class

slacAdpai:
## Description
Create new Valve Fast Shutter (VFS) class by inheriting VRC. Added VFS closed, VFS open, vfs vac fault ok, vfs vac fault reset, vfs mps ok, vfs trigger, and vfs close command. 

## Motivation and Context
This allows for the generation for fast shutter expert screens.

## How Has This Been Tested?
This is the first iteration. The pragmas are all based from the vacuum lib library.

## Where Has This Been Documented?
Not sure where

image

PR #458, opened 2020-06-24
DEV: In valve.py, added VFS class.

slacAdpai:
## Description
In valve.py, added VFS class.

## Motivation and Context
Valve Fast Shutter does not have an expert screen yet so this is the first version of it.

### DIFF
--- a/pcdsdevices/valve.py
+++ b/pcdsdevices/valve.py
@@ -243,7 +243,7 @@ class VFS(Device):
                               doc=('Epics Command to force open the valve in'
                                    'override mode'))
     gfs_trigger = Cpt(EpicsSignalRO, ':TRIG_RBV', kind='normal',
-                      doc='Guage Fast Sensor Input Trigger')
+                      doc='Gauge Fast Sensor Input Trigger')
 
     position_close = Cpt(EpicsSignalRO, ':IS_CLOSED_RBV', kind='normal',
                          doc='Fast Shutter Closed Valve Position')
--- a/pcdsdevices/valve.py
+++ b/pcdsdevices/valve.py
@@ -231,7 +231,7 @@ class VFS(Device):
     """Class for Fast Shutter Valve."""
     request_close = Cpt(EpicsSignalWithRBV, ':CLS_SW', kind='normal',
                         doc=('Request Fast Shutter to Close. When both close'
-                            'and open are requested, VFS will close.'))
+                             'and open are requested, VFS will close.'))
     request_open = Cpt(EpicsSignalWithRBV, ':OPEN_SW', kind='normal',
                        doc=('Request Fast Shutter to Open. When both close and'
                             'open are requested, VFS will close.'))
--- a/pcdsdevices/valve.py
+++ b/pcdsdevices/valve.py
@@ -227,6 +227,34 @@ class VGC(VRC):
                     doc=('individual valve MPS state for debugging'))
 
 
+class VFS(Device):
+    """Class for Fast Shutter Valve."""
+    request_close = Cpt(EpicsSignalWithRBV, ':CLS_SW', kind='normal',
+                        doc='Request Fast Shutter to Close')
+    request_open = Cpt(EpicsSignalWithRBV, ':OPEN_SW', kind='normal',
+                       doc='Request Fast Shutter to Open')
+    reset_vacuum_fault = Cpt(EpicsSignalWithRBV, ':FAULT_RESET', kind='normal',
+                             doc=('Reset Fast Shutter Vacuum Faults: fast'
+                                  'sensor triggered, fast sensor turned off'))
+    override_status = Cpt(EpicsSignalWithRBV, ':VFS_OVRD_ON', kind='normal',
+                          doc='Epics Readback on Override mode')
+    override_force_open = Cpt(EpicsSignalWithRBV, ':VFS_FORCE_OPN',
+                              kind='normal',
+                              doc=('Epics Command to force open the valve in'
+                                   'override mode'))
+    gfs_trigger = Cpt(EpicsSignalRO, ':TRIG_RBV', kind='normal',
+                      doc='Guage Fast Sensor Input Trigger')
+
+    position_close = Cpt(EpicsSignalRO, ':IS_CLOSED_RBV', kind='normal',
+                         doc='Fast Shutter Closed Valve Position')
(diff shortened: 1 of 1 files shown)


slacAdpai:
#441

pcdsdevices/valve.py
```diff
@@ -227,6 +227,34 @@ class VGC(VRC):
...
+class VFS(Device):
+    """Class for Fast Shutter Valve."""
+    request_close = Cpt(EpicsSignalWithRBV, ':CLS_SW', kind='normal',
+                        doc='Request Fast Shutter to Close')
+    request_open = Cpt(EpicsSignalWithRBV, ':OPEN_SW', kind='normal',
+                       doc='Request Fast Shutter to Open')
+    reset_vacuum_fault = Cpt(EpicsSignalWithRBV, ':FAULT_RESET', kind='normal',
+                             doc=('Reset Fast Shutter Vacuum Faults: fast'
+                                  'sensor triggered, fast sensor turned off'))
+    override_status = Cpt(EpicsSignalWithRBV, ':VFS_OVRD_ON', kind='normal',
```
hhslepicka, line 239: If it is a readback shouldn’t it be an EpicsSignalRO?

slacAdpai: For Reset_vacuum Fault and override status, they are both settable values. Override status, is some overarching override, should the name get changed?

hhslepicka: probably name and docstrings.

ZLLentz: There's just a mismatch between `EpicsSignalWithRBV` and `doc=Epics Readback on Override mode`. Perhaps the correct thing to do is fix the docstring if this is a settable?

ZLLentz: My suggestion for a rename if you are considering one is `override_mode`

pcdsdevices/valve.py
```diff
@@ -227,6 +227,34 @@ class VGC(VRC):
...
+                          doc='Epics Readback on Override mode')
+    override_force_open = Cpt(EpicsSignalWithRBV, ':VFS_FORCE_OPN',
+                              kind='normal',
+                              doc=('Epics Command to force open the valve in'
+                                   'override mode'))
+    gfs_trigger = Cpt(EpicsSignalRO, ':TRIG_RBV', kind='normal',
+                      doc='Guage Fast Sensor Input Trigger')
```
ZLLentz, line 246:
```suggestion
                      doc='Gauge Fast Sensor Input Trigger')
```

pcdsdevices/valve.py
```diff
@@ -227,6 +227,34 @@ class VGC(VRC):
                     doc=('individual valve MPS state for debugging'))
 
 
+class VFS(Device):
+    """Class for Fast Shutter Valve."""
+    request_close = Cpt(EpicsSignalWithRBV, ':CLS_SW', kind='normal',
+                        doc='Request Fast Shutter to Close')
+    request_open = Cpt(EpicsSignalWithRBV, ':OPEN_SW', kind='normal',
+                       doc='Request Fast Shutter to Open')
```
ZLLentz, line 235: There's nothing wrong with how you've set these up, but in the other valve classes we call these `close_command` and `open_command`. You can decide if it's better for this device to look the same, or if it is better to keep it looking different because it is different.

pcdsdevices/valve.py
```diff
@@ -227,6 +227,34 @@ class VGC(VRC):
...
+                      doc='Guage Fast Sensor Input Trigger')
+
+    position_close = Cpt(EpicsSignalRO, ':IS_CLOSED_RBV', kind='normal',
+                         doc='Fast Shutter Closed Valve Position')
+    position_open = Cpt(EpicsSignalRO, ':IS_OPEN_RBV', kind='normal',
+                        doc='Fast Shutter Open Valve Position')
+    vfs_vac_fault_ok = Cpt(EpicsSignalRO, ':FAULT_OK_RBV', kind='normal',
+                           doc='Fast Shutter Vacuum Fault OK')
+    vfs_mps_ok = Cpt(EpicsSignalRO, ':FFO_OK_RBV', kind='normal',
+                     doc='Fast Shutter Fast Fault Output OK')
```
ZLLentz, line 255: This one is another judgement call: is it better to have the redundant `vfs_` in these two components, or not?

slacAdpai: The vfs has been taken out and the doc string has been revised to provide a clearer explanation.

pcdsdevices/valve.py
slacAdpai: Thank you Zach, `override_mode` is much better. Doc strings have been updated

pcdsdevices/valve.py
```diff
@@ -227,6 +227,36 @@ class VGC(VRC):
...
+                        doc=('Request Fast Shutter to Close. When both close'
+                            'and open are requested, VFS will close.'))
+    request_open = Cpt(EpicsSignalWithRBV, ':OPEN_SW', kind='normal',
+                       doc=('Request Fast Shutter to Open. When both close and'
+                            'open are requested, VFS will close.'))
+    reset_vacuum_fault = Cpt(EpicsSignalWithRBV, ':FAULT_RESET', kind='normal',
+                             doc=('Reset Fast Shutter Vacuum Faults: fast'
+                                  'sensor triggered, fast sensor turned off'))
+    override_mode = Cpt(EpicsSignalWithRBV, ':VFS_OVRD_ON', kind='normal',
+                          doc='Epics Command to set Override mode')
```
ZLLentz, line 242: flake8 failing here
```suggestion
                        doc='Epics Command to set Override mode')
```

pcdsdevices/valve.py
```diff
@@ -227,6 +227,36 @@ class VGC(VRC):
                     doc=('individual valve MPS state for debugging'))
 
 
+class VFS(Device):
+    """Class for Fast Shutter Valve."""
+    request_close = Cpt(EpicsSignalWithRBV, ':CLS_SW', kind='normal',
+                        doc=('Request Fast Shutter to Close. When both close'
+                            'and open are requested, VFS will close.'))
```
ZLLentz, line 234: flake8
```suggestion
                             'and open are requested, VFS will close.'))
```

pcdsdevices/valve.py
slacAdpai: Made updates to the doc string. Will keep the current names for now.

ZryletTC:
🎉

PR #459, opened 2020-06-24
MNT: sync XPIM classes with updated PVs

ZLLentz:
## Description

I added some features to the PLC/IOC and this change to the class reflects the additions.

- LED auto-shutoff after a timeout
- LED auto on/off if we go to the relevant state position
- Ability to lock out zoom/focus motors

## Motivation and Context

Feature requests from the engineers and scientists

## How Has This Been Tested?

I instantiated the XPIM device for IM4L0, called `.get()`, and got a valid readback from every single PV.

### DIFF
--- a/pcdsdevices/pim.py
+++ b/pcdsdevices/pim.py
@@ -343,6 +343,22 @@ class XPIMFilterWheel(StatePositioner):
 class XPIMLED(Device):
     """
     Controllable illumination with auto-on, auto-off, and shutdown timer.
+
+    The power signal will read back the current LED power state, and can be
+    used to set the power state if we are not in auto_mode.
+
+    The power_timeout signal can be used to configure the LED shutdown in
+    units of minutes. If power_timeout is zero, there will not be a timeout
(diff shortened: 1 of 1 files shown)


pcdsdevices/pim.py
```diff
@@ -340,6 +340,19 @@ class XPIMFilterWheel(StatePositioner):
     error_message = Cpt(PytmcSignal, ':ERR:MSG', io='i', kind='omitted')
 
 
+class XPIMLED(Device):
+    """
+    Controllable illumination with auto-on, auto-off, and shutdown timer.
+    """
```
ZryletTC, line 346:
```suggestion
    """Controllable illumination with auto-on, auto-off, and shutdown timer."""
```

ZLLentz: I dislike how one-line docstrings look so much that I'm going to make the docstring longer instead

ZryletTC: Lmao ok

PR #461, opened 2020-06-24
MNT: Updated valve.py, FSV class, vac_fault_ok PV to correct one.

slacAdpai:
## Description
Fixed incorrect PV in FSV class.

## Motivation and Context
In the FSV class, I used the incorrect PV for vac_fault_ok. Here is the fix.

### DIFF
--- a/pcdsdevices/valve.py
+++ b/pcdsdevices/valve.py
@@ -250,7 +250,7 @@ class VFS(Device):
                          doc='Fast Shutter Closed Valve Position')
     position_open = Cpt(EpicsSignalRO, ':IS_OPEN_RBV', kind='normal',
                         doc='Fast Shutter Open Valve Position')
-    vac_fault_ok = Cpt(EpicsSignalRO, ':FAULT_OK_RBV', kind='normal',
+    vac_fault_ok = Cpt(EpicsSignalRO, ':VAC_FAULT_OK_RBV', kind='normal',
                        doc=('Fast Shutter Vacuum Fault OK. To open VFS, this'
                             'needs to be reset to TRUE after a vacuum event'))
     mps_ok = Cpt(EpicsSignalRO, ':FFO_OK_RBV', kind='normal',
--- a/pcdsdevices/valve.py
+++ b/pcdsdevices/valve.py
@@ -253,7 +253,7 @@ class VFS(Device):
     position_open = Cpt(EpicsSignalRO, ':IS_OPEN_RBV', kind='normal',
                         doc='Fast Shutter Open Valve Position')
     vac_fault_ok = Cpt(EpicsSignalRO, ':VAC_FAULT_OK_RBV', kind='normal',
-                       doc=('Fast Shutter Vacuum Fault OK Readback')
+                       doc=('Fast Shutter Vacuum Fault OK Readback'))
     mps_ok = Cpt(EpicsSignalRO, ':FFO_OK_RBV', kind='normal',
                  doc='Fast Shutter Fast Fault Output OK')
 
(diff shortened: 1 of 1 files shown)


pcdsdevices/valve.py
```diff
@@ -250,7 +250,7 @@ class VFS(Device):
                          doc='Fast Shutter Closed Valve Position')
     position_open = Cpt(EpicsSignalRO, ':IS_OPEN_RBV', kind='normal',
                         doc='Fast Shutter Open Valve Position')
-    vac_fault_ok = Cpt(EpicsSignalRO, ':FAULT_OK_RBV', kind='normal',
+    vac_fault_ok = Cpt(EpicsSignalRO, ':VAC_FAULT_OK_RBV', kind='normal',
                        doc=('Fast Shutter Vacuum Fault OK. To open VFS, this'
                             'needs to be reset to TRUE after a vacuum event'))
```
ZryletTC, line 255: I'm a little confused by the doc for this component as it says it needs to be reset making it seems someone needs to write TRUE to it but it's a readonly, so it can't be written to. Can you clarify?

klauer:
I was independently trying to verify your PVs because of a new pytmc utility I'm poking around with. It seems like the vacuum library TMC is not up-to-date (or my utility isn't working 😦 )?

Source: https://pcdshub.github.io/lcls-twincat-vacuum/master/L2SIVacuumLib_L2SIVacuum_source.html#pou-fb-vfs

pcdsdevices/valve.py
slacAdpai: Ahh, 

> We merged this last time under the assumption that all the PVs have been checked. Can you do a quick check before we merge this one?
> 
> ```
> source pcds_conda
> cd <your pcdsdevices clone>
> ipython
> from pcdsdevices.valve import VFS
> vfs = VFS('<your prefix>', name='vfs')
> vfs.get()
> ```
> 
> This should work successfully if all the PV names are valid.

Thank for this PV check, I'll add it to my toolbox! Yes the PV check works properly for the VFS class.

slacAdpai:
> I was independently trying to verify your PVs because of a new pytmc utility I'm poking around with. It seems like the vacuum library TMC is not up-to-date (or my utility isn't working  )?
> 
> Source: https://pcdshub.github.io/lcls-twincat-vacuum/master/L2SIVacuumLib_L2SIVacuum_source.html#pou-fb-vfs

Hello Ken, This is actually pointing to a VFS_interface instead. The pou-fb-vfs is running on a faster task on the plc and is linked to the pou-fb-vfs-interface, which is on the main/slower task. The ioc runs only runs on the main plc project task. Is there another way to do it? The goal was to reduce the overhead on the vfs fast task so it can react to closing the fast shutter quickly.
Source: https://pcdshub.github.io/lcls-twincat-vacuum/master/L2SIVacuumLib_L2SIVacuum_source.html#pou-fb-vfs-interface

pcdsdevices/valve.py
slacAdpai: I have updated the doc strings accordingly. Thank you all for your discerning eyes!

pcdsdevices/valve.py
```diff
@@ -250,9 +252,8 @@ class VFS(Device):
                          doc='Fast Shutter Closed Valve Position')
     position_open = Cpt(EpicsSignalRO, ':IS_OPEN_RBV', kind='normal',
                         doc='Fast Shutter Open Valve Position')
-    vac_fault_ok = Cpt(EpicsSignalRO, ':FAULT_OK_RBV', kind='normal',
-                       doc=('Fast Shutter Vacuum Fault OK. To open VFS, this'
-                            'needs to be reset to TRUE after a vacuum event'))
+    vac_fault_ok = Cpt(EpicsSignalRO, ':VAC_FAULT_OK_RBV', kind='normal',
+                       doc=('Fast Shutter Vacuum Fault OK Readback')
```
ZLLentz, line 256:
```suggestion
                       doc=('Fast Shutter Vacuum Fault OK Readback'))
```

ZLLentz:
Sorry, I missed the comment where you ran the check since it was in a reply on a different thread.

PR #462, opened 2020-06-24
DOC: Annotate LCLS2 Imagers and Related Classes

ZLLentz:
## Description

For LCLS2 imagers and related classes:
- Try out the shiny `variety` schemas on a subset of the component signals
- Fill in the empty `doc` kwargs on many components

## Motivation and Context

- I felt weird poring over `doc=` tags in the new PRs while all of my devices don't have these defined
- Hopefully `doc=` becomes mouse-over help text in a screen somewhere
- Variety will be used in `typhos` at some point, and it will be nice to have some users ahead of that

## How Has This Been Tested?

Tests still pass

This is docs

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -17,6 +17,7 @@ from .doc_stubs import basic_positioner_init
 from .interface import FltMvInterface
 from .pseudopos import DelayBase
 from .signal import PytmcSignal
+from .variety import set_metadata
 
 logger = logging.getLogger(__name__)
 
@@ -443,11 +444,15 @@ class PMC100(PCDSMotorBase):
 
 
 class BeckhoffAxisPLC(Device):
-    """Debug PVs from the Beckhoff Axis PLC code"""
+    """Error handling for the Beckhoff Axis PLC code."""
     status = Cpt(PytmcSignal, 'sErrorMessage', io='i', kind='normal',
-                 string=True)
-    err_code = Cpt(PytmcSignal, 'nErrorId', io='i', kind='normal')
-    cmd_err_reset = Cpt(PytmcSignal, 'bReset', io='o', kind='config')
+                 string=True, doc='PLC error or warning')
+    err_code = Cpt(PytmcSignal, 'nErrorId', io='i', kind='normal',
+                   doc='Current NC error code')
+    cmd_err_reset = Cpt(PytmcSignal, 'bReset', io='o', kind='config',
+                        doc='Command to reset an active error')
+
+    set_metadata(cmd_err_reset, dict(variety='command', value=1))
 
 
 class BeckhoffAxis(EpicsMotorInterface):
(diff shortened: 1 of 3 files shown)


pcdsdevices/pim.py
```diff
@@ -334,10 +366,17 @@ class XPIMFilterWheel(StatePositioner):
...
 
-    state = Cpt(EpicsSignal, ':GET_RBV', write_pv=':SET', kind='normal')
+    state = Cpt(EpicsSignal, ':GET_RBV', write_pv=':SET', kind='normal',
+                doc='Control of the filter wheel state by preset '
+                    'transmission percentages.')
+
+    reset_cmd = Cpt(PytmcSignal, ':ERR:RESET', io='i', kind='omitted',
+                    doc='Command to reset a filter wheel error.')
+    error_message = Cpt(PytmcSignal, ':ERR:MSG', io='i', kind='omitted',
+                        string=True,
```
klauer, line 376: Good catch :+1:

pcdsdevices/epics_motor.py
```diff
@@ -443,11 +444,16 @@ def home(self, *args, **kwargs):
...
-                 string=True)
-    err_code = Cpt(PytmcSignal, 'nErrorId', io='i', kind='normal')
-    cmd_err_reset = Cpt(PytmcSignal, 'bReset', io='o', kind='config')
+                 string=True, doc='PLC error or warning')
+    err_code = Cpt(PytmcSignal, 'nErrorId', io='i', kind='normal',
+                   doc='Current NC error code')
+    cmd_err_reset = Cpt(PytmcSignal, 'bReset', io='o', kind='config',
+                        doc='Command to reset an active error')
+
+    set_metadata(err_code, dict(variety='scalar', display_format='hex'))
```
klauer, line 455: 🎆

pcdsdevices/pim.py
ZLLentz: It's one of a handful that looks very silly in typhos right now because it is an array instead of a string. Should be better after this.

ZLLentz:
@klauer one thing I had trouble identifying as I went through this was: which signals are worth putting the variety metadata on? All of them? All of them definitely fall into one of the categories, but the majority would just end up as `variety=enum` for the bools and `variety=scalar` for the rest. I ended up just picking the ones with meaningful information to start.

I also found myself struggling to pick between `command-enum` and `enum` for things like position state signals that can be both set and read.

klauer:
Good questions. I think adding metadata is only strictly necessary when there's some benefit from the differentiation between varieties (or there's some info beyond the variety itself you want to add). For basic scalars with nothing special to say about them, I think it's fine - perhaps even preferred - to leave the metadata off. It's more code to maintain after all.

As far as when to choose what... I'm not sure just yet. We need to refine the varieties and the metadata that can go with them before rolling this out everywhere. Some of these varieties should be removed/renamed/modified significantly.

Out of the gate in https://github.com/pcdshub/typhos/pull/357, I got lost on what the difference between a `tweakable` vs a `scalar-range` was supposed to be. They might be combinable.

ZLLentz:
It may be that a `tweakable` is just a `scalar-range` that is writable; just like a `command-enum` is just an `enum` that is writable. I'm sure we'll find others.

hhslepicka:
`tweakable` was intended for things like jog... `scalar-range` was for the slider.

PR #463, opened 2020-06-25
REF: restructure and rework variety metadata: use nested schema, expand dotted dictionaries

klauer:
This PR:
* Pairs with https://github.com/pcdshub/typhos/pull/357
* Adds dotted dictionary support ([example](https://github.com/pcdshub/typhos/blob/91989d836fb76b335900148b80c4f7cc12435fbc/tests/variety_ioc.py#L18-L23))
* Dotted schema dictionaries are automatically expanded prior to schema validation, requiring less nesting on definition.

The vast majority of this PR is the dotted dictionary support, which may well be moved to (e.g.) pcdsutils at some point. The remainder is changes to the schema that are required for https://github.com/pcdshub/typhos/pull/357 and appear to generally be a good thing, based solely on my own use/testing.

### DIFF
--- a/pcdsdevices/variety.py
+++ b/pcdsdevices/variety.py
@@ -76,6 +76,9 @@ common_schema = {
 }
 
 
+_default_bitmask_style = dict(shape='rectangle', on_color='green',
+                              off_color='gray')
+
 schema_by_category = {
     'command': schema.Schema({
         'variety': schema.Or(*varieties_by_category['command']),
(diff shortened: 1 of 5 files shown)


klauer:
Schema verification shows some promise: `schema.SchemaError: Key 'range' error: (0, 100) should be instance of 'dict'`
which was in the PIM device. This was expected as this PR changes the schema - hopefully for the better.

klauer:
I take it back; this isn't quite ready yet. Forgot entirely about the test suite 👀

pcdsdevices/variety.py
```diff
@@ -70,83 +70,166 @@ def validate(value):
...
+        Optional('display_format', default='default'): schema.Or(
             'default', 'string', 'decimal', 'exponential', 'hex', 'binary'),
+
+        Optional('range'): schema.Schema({
+            Optional('value'): _length_validate(2, 2, (float, int)),
+            Optional('source', default='use_limits'): schema.Or(
+                'use_limits', 'value'),
+        }),
+
+        Optional('delta'): schema.Schema({
```
ZLLentz, line 129: I'm having trouble interpreting what a scalar's delta would be. Can you give an example?

klauer: The full name would be "delta [tweak] value". So, basically what is a reasonable "delta" to change the scalar by? And what is a valid range of that delta?

A sample value would be for a 0-100% value, a standard delta of 0.1%, with an adjustable range of `-1% <= delta <= 1%` if it's for some reason advisable to not tweak by larger than 1% (🤷 maybe unnecessary?)

ZLLentz: Gotcha

PR #464, opened 2020-06-26
DOC: Display interlocking vacuum device in Typhos screen for VGC, PIP

n-wbrown:
## Description

Add the :ILK_DEVICE* field to the VGC and PIP so it will display in a Typhos screen. This variable tracks the component the VGC or PIP is using to validate its pressure interlocks.

## Motivation and Context

This will allow users to determine which gauges are required for this device to function from the Typhos screen. @slacAWallace 

## How Has This Been Tested?

Standard tests and style checks succeed.

The component docstrings have been updated indicating the goal for this field.

### DIFF
--- a/pcdsdevices/valve.py
+++ b/pcdsdevices/valve.py
@@ -235,7 +235,6 @@ class VGC(VRC):
                                       'interlocking this valve')
 
 
-
 class VFS(Device):
     """Class for Fast Shutter Valve."""
     request_close = Cpt(EpicsSignalWithRBV, ':CLS_SW', kind='normal',
--- a/pcdsdevices/valve.py
+++ b/pcdsdevices/valve.py
@@ -225,6 +225,15 @@ class VGC(VRC):
                 doc='Error Present')
     mps_state = Cpt(EpicsSignalRO, ':MPS_FAULT_OK_RBV', kind='omitted',
                     doc=('individual valve MPS state for debugging'))
+    interlock_device_upstream = Cpt(EpicsSignalRO, ':ILK_DEVICE_US_RBV',
+                                    kind='config',
+                                    doc='Upstream vacuum device used for'
+                                    'interlocking this valve')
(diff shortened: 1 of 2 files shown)


pcdsdevices/valve.py
```diff
@@ -225,6 +225,15 @@ class VGC(VRC):
...
                     doc=('individual valve MPS state for debugging'))
+    interlock_device_upstream = Cpt(EpicsSignalRO, ':ILK_DEVICE_US_RBV',
+                                    kind='config',
+                                    doc='Upstream vacuum device used for'
+                                    'interlocking this valve')
+    interlock_device_downstream = Cpt(EpicsSignalRO, ':ILK_DEVICE_DS_RBV',
+                                      kind='config',
+                                      doc='Downstream vacuum device used for'
+                                      'interlocking this valve')
+
```
ZLLentz, line 236: flake8 is failing on `./pcdsdevices/valve.py:239:1: E303 too many blank lines (3)`
```suggestion
```

ZLLentz:
Same thought as Tyler, just confirm that you've done the good ol get test (or some equivalent) to verify the PVs:
```
source pcds_conda
cd <your pcdsdevices clone>
ipython
from pcdsdevices.<module> import <class>
test = <class>('<your prefix>', name='test')
test.get()
```

klauer:
Using only the read-only RBV intentionally?
image
ref https://pcdshub.github.io/lcls-twincat-vacuum/master/L2SIVacuumLib_L2SIVacuum_epics.html

n-wbrown:
@ZryletTC, it's tested now. During testing I discovered a few issues with components that were RBV but should have been RO. The fix is contained in my latest commit. 
@klauer, the RO is intentional. These fields are generated on the PLC from the string name of the device used for interlocks which I think should only ever be determined by the PLC. I'm a bit concerned that these TwinCAT variables are writable through pytmc/epics. 
@ZLLentz, style checks should pass now.

PR #465, opened 2020-06-30
DEV: Updated VFS class in valve.py for new VFS interface. Backend VFS…

slacAdpai:
…_FB was updated and here are the changes to the pragmas.

## Description
Updated VFS class in valve.py for new VFS interface. Backend VFS_FB was updated and here are the changes to the pragmas.
## Motivation and Context
image (8)

### DIFF
--- a/pcdsdevices/valve.py
+++ b/pcdsdevices/valve.py
@@ -237,6 +237,8 @@ class VGC(VRC):
 
 class VFS(Device):
     """Class for Fast Shutter Valve."""
+    valve_position = Cpt(EpicsSignalRO, ':POS_STATE', kind='normal',
+                         doc='Ex: OPEN, CLOSED, MOVING, INVALID, OPEN_F')
     vfs_state = Cpt(EpicsSignalRO, ':STATE_RBV', kind='normal',
                     doc='Fast Shutter Current State')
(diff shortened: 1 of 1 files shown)


pcdsdevices/valve.py
```diff
@@ -237,32 +237,38 @@ class VGC(VRC):
...
                         doc='Epics Command to set Override mode')
-    override_force_open = Cpt(EpicsSignalWithRBV, ':VFS_FORCE_OPN',
+    override_force_open = Cpt(EpicsSignalWithRBV, ':FORCE_OPN',
                               kind='normal',
-                              doc=('Epics Command to force open the valve in'
-                                   'override mode'))
+                              doc=('Epics Command to force open'
+                                   'the valve in override mode'))
+    gfs_name = Cpt(EpicsSignalRO, ':GFS_RBV', kind='normal',
+                   doc='Gauge Fast Sensor Name')
```
hhslepicka, line 262: This should be `string` otherwise it will present the widget as a waveform plot. Please fix this.

ZLLentz: `string=True` is the kwarg needed here

slacAdpai: Attempt to fix in #467

PR #471, opened 2020-07-06
MNT: fix issue with failed subscription callback.

ZLLentz:
## Description

Switch None-check to TypeError try/except to skip comparison regardless of the failure mode here.

## Motivation and Context

If this subscription fails, the motor will have lots of issues like the `position` property failing to work.
Recently the default value was changed from `None` to `object()` for epics signals. This fixes it so we can bypass an error regardless of why the comparison check fails here.

## How Has This Been Tested?

Works locally in xcs python. Probably passes tests, I did not check.


ZryletTC:
> Reasonable - though the `object()` change is weird...

Zach gave a good explanation on why it's good to use as a sentinel value as `object()` creates an immutable object that a distinguishable from the other initial values and useful for comparison when debugging.

Issue #473, opened 2020-07-14
Add valve.py, class VGC ALM_RST pv into the expert gui screens


slacAdpai:
## Steps to Reproduce (for bugs)

1.
2.
3.

slacAdpai:
Fixed with #475

Issue #474, opened 2020-07-15
Add pump.py, class pipplc add `:FORCE_START` pv 


slacAdpai:
## Current Behavior
No override button, so need to go directly to plc to override and turn on pump.

## Steps to Reproduce (for bugs)

1.
2.
3.

slacAdpai:
Fixed with #475

PR #475, opened 2020-07-15
DEV: pump.py/PIPPLC added override_force_on. valve.py/ValveBase added…

slacAdpai:
… error_reset.

## Description
added override_force_on and error_Reset for vacuum devices


slacAdpai:
Fixes #473  and #474

PR #478, opened 2020-07-17
ENH: swap out cf-units for pint for portability

ZLLentz:
## Description

`cf-units` -> `pint`

## Motivation and Context

`cf-units` is not platform-independent because it relies on udunits being installed
`pint` is pure python

## How Has This Been Tested?

Tests still pass

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -22,7 +22,7 @@ requirements:
     - pyepics >=3.4.1
     - pytmc >=2.4.0
     - pyyaml
-    - cf-units
+    - pint
     - scipy
     - matplotlib <3
     - periodictable
--- a/pcdsdevices/utils.py
+++ b/pcdsdevices/utils.py
@@ -4,7 +4,7 @@ import shutil
 import sys
 import time
 
-from cf_units import Unit
+import pint
 
 try:
     import tty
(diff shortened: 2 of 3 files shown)


ZLLentz:
I'm going to use the codecov output to check

ZLLentz:
```
In [2]: convert_unit(100, 'meters', 'centimeters')
Out[2]: 10000.0

In [3]: convert_unit(100, 'ns', 'sec')
Out[3]: 1.0000000000000001e-07
```

ZLLentz:
Terminal coverage snippit since I remembered I can do that
```
Name                                    Stmts   Miss  Cover   Missing
---------------------------------------------------------------------
pcdsdevices/pseudopos.py                   55      0   100%
pcdsdevices/utils.py                       63     13    79%   12-14, 60, 128-137, 156
```

ZLLentz:
Ok and codecov double confirms the diff coverage is 100%. I'm confident that this is OK.

A cool thing about pint is that we can register custom units. So that's kind of neat if we have that sort of need.

hhslepicka:
Just beware that Pint has some issues with constants and wrong values. I did not experience this myself. This comes from feedback from a different project.

ZLLentz:
Thanks for the disclaimer @hhslepicka. I hope we don't run into the same issue... I haven't dipped into the constants in `pint` yet. Right now we're just converting length and time quantities so hopefully we can coast along until things are updated/fixed upstream.

Shouldn't we use scipy constants for that anyway? I'm having trouble remembering how these packages are laid out.
At any rate, I'm sick of the various issues with using `cf-units` and will be welcoming of a new set of issues.

PR #480, opened 2020-07-22
Qmini Spectrometer

slactjohnson:
## Description

Provides support for the RGB Photonics/Broadcom Qmini spectrometer. 
## Motivation and Context

Adds support for a device we need. Generates reasonable Typhos screens. 
## How Has This Been Tested?

Tested using a spectrometer in the Heinz lab.

### DIFF
--- a/pcdsdevices/lasers/qmini.py
+++ b/pcdsdevices/lasers/qmini.py
@@ -1,0 +1,83 @@
+import logging
+
+from ophyd import Device, EpicsSignal, EpicsSignalRO, Component as Cpt
+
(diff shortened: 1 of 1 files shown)


pcdsdevices/lasers/qmini.py
```diff
@@ -0,0 +1,81 @@
...
+    Parameters
+    ----------
+    prefix : str
+        Base prefix of the Qmini
+
+    name : str
+        Name of Qmini object
+
+    Examples
+    --------
```
ZLLentz, line 23: Empty `Examples` section should probably be filled or removed

PR #482, opened 2020-07-23
ENH: Add new detector class specifically for creating areaDetector Typhos screens

slactjohnson:
## Description

This adds a new class, PCDSAreaDetectorTyphos, for creating areaDetector Typhos screens. This cuts out much of the extra areaDetector plugins and PVs that are not often used, and creates a fairly slim screen for simple AD camera configuration and operation. Includes a pop-up screen for camera viewing. 
## Motivation and Context

Other PCDSAreaDetector classes have a lot of PVs that are not needed most of the time. The other classes also did not allow for viewing the image. 
This
## How Has This Been Tested?

I tested this in the Heinz lab on a Basler camera. 

Not much to document since this is mainly just a pile of PVs. 

See screen shot with the generated screen, compared to the camera viewer. 
typhos_gige

## Notes
I will note one annoyance with this screen right now. When you first open the image viewer, the image is not immediately visible. It appears that the default color map (Magma) does not work with this viewer. In fact, I only got a few of the color maps to work: Jet, Monochrome, and Hot. Once you have set the color map, if you close the image viewer it will retain the color map if you open it back up again.

### DIFF
--- a/pcdsdevices/areadetector/detectors.py
+++ b/pcdsdevices/areadetector/detectors.py
@@ -8,13 +8,17 @@ import logging
 import warnings
 
 from ophyd.areadetector import cam
-from ophyd.areadetector.base import ADComponent
+from ophyd.areadetector.base import ADComponent, DDC_EpicsSignalRO, NDDerivedSignal
 from ophyd.areadetector.detectors import DetectorBase
 from ophyd.device import Component as Cpt
+from ophyd import Device
+from ophyd.signal import EpicsSignal, EpicsSignalRO, AttributeSignal
 
 from .plugins import (ColorConvPlugin, HDF5Plugin, ImagePlugin, JPEGPlugin,
                       NetCDFPlugin, NexusPlugin, OverlayPlugin, ProcessPlugin,
                       ROIPlugin, StatsPlugin, TIFFPlugin, TransformPlugin)
+
+import numpy as np
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


hhslepicka:
The colormap issue was addressed at the PyDM level... it was an issue due to new pyqtgraph.
@slactjohnson do you mind to confirm the pydm version that you have with `pydm --version` ?

slactjohnson:
@hhslepicka I was using pcds-3.2.0 in hutch python, which appears to use pydm version 1.9.0

hhslepicka:
@slactjohnson could you give it a try with dev conda?

ZLLentz:
I like this a lot. I want to do a follow-up PR that slaps this onto all the imager devices.

slactjohnson:
@hhslepicka Switching to dev conda (or anything above pcds-3.2.0) breaks laspython for some reason. I'm working on figuring out why. I'll let you know when I try it out.

ZLLentz:
Does @klauer want to take a look before I merge?

pcdsdevices/areadetector/detectors.py
```diff
@@ -161,3 +163,51 @@ class PCDSAreaDetector(PCDSAreaDetectorEmbedded):
...
+    manufacturer = Cpt(EpicsSignalRO, 'Manufacturer_RBV', kind='config')
+    camera_model = Cpt(EpicsSignalRO, 'Model_RBV', kind='normal')
+    sensor_size_x = Cpt(EpicsSignalRO, 'MaxSizeX_RBV', kind='config')
+    sensor_size_y = Cpt(EpicsSignalRO, 'MaxSizeY_RBV', kind='config')
+    data_type = Cpt(EpicsSignal, 'DataType', kind='config')
+    data_type_RBV = Cpt(EpicsSignalRO, 'DataType_RBV', kind='config')
+
+    # Acquisition settings
+    exposure = Cpt(EpicsSignal, 'AcquireTime', kind='config')
+    exposure_rbv = Cpt(EpicsSignalRO, 'AcquireTime_RBV', kind='config')
```
klauer, line 184: Any reason to not combine the setpoint and readback (i.e., using `EpicsSignalWithRBV`)?

slactjohnson: @klauer Is "I didn't notice that signal type before" a good reason? :)

Let me try that out; that could clean this device (and the Qmini, for that matter) up quite a bit.

klauer: A very good reason indeed! 😄

slactjohnson: Swapped to EpicsSignalWithRBV; this looks much better now. Thanks!

ZLLentz:
The travis will pass and then I will merge

PR #485, opened 2020-07-24
MNT: do not put to PVs until the user asks

ZLLentz:
## Description

Lock the slits callbacks to not put to live PVs until there is a move request from this python object

## Motivation and Context

Multiple inactive clients should not be putting to random PVs

### DIFF
--- a/pcdsdevices/slits.py
+++ b/pcdsdevices/slits.py
@@ -383,16 +383,22 @@ class BeckhoffSlits(SlitsBase):
     north = Cpt(BeckhoffAxis, ':MMS:NORTH', kind='normal')
     south = Cpt(BeckhoffAxis, ':MMS:SOUTH', kind='normal')
 
+    def __init__(self, prefix, *, name, **kwargs):
+        self._started_move = False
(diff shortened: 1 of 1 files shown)


ZLLentz:
Wish I was clever enough to catch it ahead of time. Lucid startup:
```
[2020-07-24 10:01:38,894] [ERROR   ] - Subscription value callback exception (Signal(name='sl1l0_done_all', parent='sl1l0', value=True, timestamp=1595610098.8572037))
  File "/reg/g/pcds/pyps/apps/dev/pythonpath/ophyd/ophydobj.py", line 462, in inner
    cb(*args, **kwargs)
  File "/reg/g/pcds/pyps/apps/dev/pythonpath/pcdsdevices/slits.py", line 397, in _reset_exec_move
    self.exec_move.put(0)
  File "/reg/g/pcds/pyps/apps/dev/pythonpath/ophyd/signal.py", line 1625, in put
    raise ReadOnlyError('No write access to underlying EPICS PV')
ophyd.utils.errors.ReadOnlyError: No write access to underlying EPICS PV

```

PR #489, opened 2020-07-30
ENH: add initial attenuator calculator devices - separate from the motion components for now

klauer:
Pairs with https://github.com/pcdshub/hxr-attenuator caproto IOC work

<details>
<summary>attempt 1</summary>
image
</details>

image

### DIFF
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -446,13 +446,34 @@ class GasAttenuator(Device, BaseInterface):
 
 
 class AttenuatorCalculatorFilter(Device, BaseInterface):
-    material = Cpt(EpicsSignal, 'Material', kind='hinted')
-    thickness = Cpt(EpicsSignal, 'Thickness', kind='hinted')
-    is_stuck = Cpt(EpicsSignal, 'IsStuck', kind='hinted')
-    closest_energy = Cpt(EpicsSignalRO, 'ClosestEnergy_RBV', kind='config')
-    transmission = Cpt(EpicsSignal, 'Transmission_RBV', kind='normal')
-    transmission_3omega = Cpt(EpicsSignal, 'Transmission3Omega_RBV',
-                              kind='normal')
+    material = Cpt(
+        EpicsSignal, 'Material', kind='hinted', string=True,
+        doc='The material formula (e.g., Si, C)'
+    )
+    thickness = Cpt(
+        EpicsSignal, 'Thickness', kind='hinted',
+        doc='Thickness in micron',
+    )
+    is_stuck = Cpt(
+        EpicsSignal, 'IsStuck', kind='hinted',
+        doc='Is the filter stuck / unusable?',
+    )
+    closest_energy = Cpt(
+        EpicsSignalRO, 'ClosestEnergy_RBV', kind='config',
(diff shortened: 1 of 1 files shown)


pcdsdevices/attenuator.py
```diff
@@ -441,3 +443,245 @@ class GasAttenuator(Device, BaseInterface):
...
+    )
+    is_stuck = Cpt(
+        EpicsSignal, 'IsStuck', kind='hinted',
+        doc='Is the filter stuck / unusable?',
+    )
+    closest_energy = Cpt(
+        EpicsSignalRO, 'ClosestEnergy_RBV', kind='config',
+        doc='Closest tabulated energy available to the requested one',
+    )
+    transmission = Cpt(EpicsSignal, 'Transmission_RBV', kind='normal',
```
ZLLentz, line 465: it's kind of cool you're doing this per-filter, the previous calculation IOCs didn't have this

pcdsdevices/attenuator.py
```diff
@@ -441,3 +443,245 @@ class GasAttenuator(Device, BaseInterface):
...
+        EpicsSignal, ':SYS:ApplyConfiguration', kind='config',
+        doc='Apply the best configuration (i.e., move the filters)',
+    )
+    set_metadata(apply_config, dict(variety='command-proc', value=1))
+
+    moving = Cpt(
+        EpicsSignal, ':SYS:Moving_RBV', kind='config',
+        doc='Are filters being moved in/out?',
+    )
+    set_metadata(moving, dict(variety='bitmask', bits=1))
```
ZLLentz, line 599: Does this give us a little light up box? If so it's a good trick I'd like to bring to other classes

klauer: Yes:

image

(Although in pasting this, I realized I noticed that this should be `EpicsSignalRO` so it only generates the readback widget)

pcdsdevices/attenuator.py
```diff
@@ -441,3 +443,245 @@ class GasAttenuator(Device, BaseInterface):
...
+
+    def get_best_config(self, **kwargs):
+        """Get the calculated (best) filter configuration."""
+        return self._bitmask_to_list(self.best_config_bitmask.get(**kwargs))
+
+    def get_moving_status(self, **kwargs):
+        """Get the filter motion status."""
+        return self._bitmask_to_list(self.filters_moving_bitmask.get(**kwargs))
+
+    def calculate(self, transmission, *, energy=None, use_floor=True):
```
ZLLentz, line 625: :+1: very useful helper method

pcdsdevices/attenuator.py
ZLLentz: Still beats `moving      1`

klauer:
I'll make an issue for combining this with the motion device, but will go with this version for the time being.

No major objections with the naming (however verbose and ugly I made it here), I hope?

ZLLentz:
RE: the naming, the only thing that bugs me about the `attenuator` module naming in general is that some things are not verbose enough... There are several different implementations of attenuator IOCs and you have to read the docstrings to figure out which one you actually want. +1 to verbosity.

klauer:
👍 Sounds good. Here we go...

PR #493, opened 2020-08-03
Update Elliptec 

slactjohnson:
## Description

Updated the elliptec classes to use formatted components to make dealing with some of the elliptec IOC PV weirdness easier. 
Restructured the code to clean up some of the screens. 
Added more docstrings. 
## Motivation and Context

This fixes some of the components, and makes the screens cleaner.
## How Has This Been Tested?

Tested in on stages in the Heinz lab. 

Docstrings.

### DIFF
--- a/pcdsdevices/lasers/elliptec.py
+++ b/pcdsdevices/lasers/elliptec.py
@@ -3,7 +3,7 @@ Classes for ThorLabs Elliptec motors.
 """
 
 
-from ophyd import Component as Cpt
+from ophyd import FormattedComponent as FCpt
 from ophyd import Device
 from ophyd.signal import EpicsSignal, EpicsSignalRO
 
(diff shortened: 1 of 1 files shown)


pcdsdevices/lasers/elliptec.py
```diff
@@ -3,56 +3,137 @@
...
 from ophyd import Device
 from ophyd.signal import EpicsSignal, EpicsSignalRO
 
 
 class EllBase(Device):
     """
     Base class for Elliptec stages.
+    """
+
+    set_position = FCpt(EpicsSignal, '{prefix}:M{self._channel}:CURPOS',
```
klauer, line 16: Not necessary to change, but FYI with recent versions of ophyd you no longer need the `self.` in formatted components:
```suggestion
    set_position = FCpt(EpicsSignal, '{prefix}:M{_channel}:CURPOS',
```

ZLLentz, line 16: Not important to change, but you can omit all `self.` in the formatted component strings:
`'{prefix}:M{_channel}:CURPOS'`

pcdsdevices/lasers/elliptec.py
```diff
@@ -3,56 +3,137 @@
...
+        The motor channel on the controller (0-F)
+    port : str
+        The port of the Elliptec controller (typically 0)
+
+    Examples
+    --------
+    ell14 = EllRotation('LM1K4:COM_DP1_TF1_ROT1:ELL', port=0, channel=1,
+                         name='ell14')
+    """
+    # Currently no difference between rotation implementation and linear
```
klauer, line 137: Good future-proofing 👍

pcdsdevices/lasers/elliptec.py
klauer: Darn it @ZLLentz 😆

ZLLentz: so lightning does strike twice

Issue #495, opened 2020-08-03
Combined device for HXR solid attenuator calculator + motion components


klauer:
Currently:
* Motion components: `FEESolidAttenuator` containing `BeckhoffAxis` (will be state positioners soon)
* Filter calculator: `AttenuatorCalculator_AT2L0` containing `AttenuatorCalculatorFilter`

Package HXR solid attenuator calculator + motion components together:
* Either in the same device or a new device

PR #496, opened 2020-08-03
ENH: State positioners for New HXR FEE attenuator

ZLLentz:
## Description

Add the newly added state positioners into the ophyd device for the new hxr fee attenuator.
This PR also contains some backend work on the lightpath interface features that were needed for this device.

## Motivation and Context

Companion to #489 
Needed before we address #495 

## How Has This Been Tested?

The blades move. All PVs connect. The typhos screen looks OK (about as good as a screen with 19 motors and 19 state positioners could look). The lightpath that I'm working on looks good.

### DIFF
--- a/pcdsdevices/state.py
+++ b/pcdsdevices/state.py
@@ -552,16 +552,6 @@ class TwinCATStateConfigAll(Device):
     state03 = Cpt(TwinCATStateConfigOne, ':03', kind='omitted')
     state04 = Cpt(TwinCATStateConfigOne, ':04', kind='omitted')
     state05 = Cpt(TwinCATStateConfigOne, ':05', kind='omitted')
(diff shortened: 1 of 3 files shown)


pcdsdevices/state.py
```diff
@@ -552,16 +552,6 @@ class TwinCATStateConfigAll(Device):
     state03 = Cpt(TwinCATStateConfigOne, ':03', kind='omitted')
     state04 = Cpt(TwinCATStateConfigOne, ':04', kind='omitted')
     state05 = Cpt(TwinCATStateConfigOne, ':05', kind='omitted')
-    state06 = Cpt(TwinCATStateConfigOne, ':06', kind='omitted')
```
klauer, line 555: 👏

PR #505, opened 2020-08-13
ENH: include calculator in AT2L0 object

### LINKED ISSUE #495: Combined device for HXR solid attenuator calculator + motion components

klauer:
Closes #495 

There's a slight issue here as I'm modifying an existing object with an existing happi entry.

Two options that seem obvious to me:
1. Merging/using this will have to wait for a tag as I modified the prefix of the `FEESolidAttenuator` object (was: `ATL0:XTES`, now: `AT2L0`)
2. Or, if needed before a tag, these could become 'UnrelatedComponents' and have two separate prefixes (ignoring the fact that `AT2L0:CALC` and `AT2L0:XTES` do share a base PV)

Opinions?

Screen:
image

### DIFF
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -432,61 +432,6 @@ class FEESolidAttenuatorBlade(Device, BaseInterface, LightpathInOutMixin):
     motor = Cpt(BeckhoffAxis, '')
 
 
-class FEESolidAttenuator(Device, BaseInterface, LightpathInOutMixin):
-    """
-    Solid attenuator variant from the LCLS-II XTES project.
-
-    Motorized, 18 filters + 1 inspection mirror.
-
-    This is a quick-and-dirty Ophyd device for controlling AT2L0 motion and
-    generating a PyDM control screen.
-
-    Parameters
-    ----------
-    prefix : str
-        Full Solid Attenuator base PV.
-
-    name : str
-        Alias for the Solid Attenuator.
-    """
-
-    # QIcon for UX
-    _icon = 'fa.barcode'
-
-    # Register that all blades are needed for lightpath calc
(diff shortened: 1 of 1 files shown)


ZryletTC:
Sorry I'm a little unclear why this needs a release to deploy. Can we not run it in dev from the time being?

klauer:
The environments share the same happi database, right?

If I change the prefix for the development device as in this PR, it'll break the released config, and merging this without changing it would break the development device.

ZryletTC:
But where is this device being accessed from? Anywhere other than hutch_python? Because the pcdsdevices run by hutch-python is running from a dev checkout of the repository, which can just be pulled once your change is merged.

klauer:
OK, it's entirely possible I don't understand how the production configuration is setup - I've shielded myself from it for a long while 😁 

I'd imagine these are largely being accessed from LUCID/typhos screens currently.

klauer:
Without moving the classes around, the motor would have a forward-reference to the calculator class.

ZryletTC:
Ahh 👌 I didn't notice the calculator hidden in between there.
As for deployment, if we're using them with Typhos/LUCID, you're probably right that we should wait for a release, which actually might not be too far away :)

klauer:
Refactored due to fears of deployment confusion (b81cdf1) and back-compatibility'd the old class name. Hard-coded prefix as it is for one specific device.

Very simple to do thanks to the Zach's great `UnrelatedComponent` work 👍

PR #506, opened 2020-08-14
ENH: add LookupTablePositioner and LxePositioner

klauer:
## Description
Adds a general `LookupTablePositioner` (only 2 columns supported for now)
Adds a specific `LaserEnergyPositioner` to close https://github.com/pcdshub/Bug-Reports-and-Requests/issues/11
Fixes a small consistency issue with the move interface - I think it should always access `.wm()` (motivated by my pseudopositioner patch of `.wm` here)

## Motivation and Context
Calibration data in file format that looks something like:
```
150.0000    4.0998
160.0000    5.0318
170.0000    6.1180
180.0000    7.3984
190.0000    8.9290
200.0000   10.7843
210.0000   13.0599
```

## How Has This Been Tested?
Locally, so far

```
In [16]: motor.umv(10)

In [17]: motor
Out[17]:

motor
-----
position: LaserEnergyPositionerPseudoPos(energy=10.000000000000002)

energy
------
  position: 10.000000000000002

motor
-----
  position: 195.77265132323615
  user_setpoint: 195.77265132323615
```

Plotting
--------
Plotting each move works, if enabled by way of `motor.enable_plotting = True`:
image


klauer:
Plotting each move works, if enabled:

image

silkenelson:
This looks good to me, it's only missing the 'notepad PV' feedback, but that is tracked elsewhere.

klauer:
@silkenelson Great 👍 thanks for taking a look.

So then we just need to be able to add their calibration settings filename to happi for it to be loaded without users having to specify it each session.

klauer:
👍 I'm unclear if that plotting is still something that the user needs/wants. If it ends up not, we should remove it down the line.

silkenelson:
It's good to see that you have loaded the calibration file you wanted to see loaded.
As far as putting this in happi: we usually put these files into the <hutch>opr/experiments/ directory so that you can see what was used for a given experiment. If a calibration file does not exist, the motor does not get created. Would that work for happi?

klauer:
The `safe_load` functionality in hutch-python should bail but not cause things to fail if a given file does not exist.
@ZLLentz  - do we have any devices that reference per-experiment config files that you know of?

@silkenelson Asking due to my ignorance about these laser energy positioners: are previous calibration files useful, or are they just for reference? If just for reference, then it seems like a `/path/to/latest/calibration.txt`-style path could be used to soft-link to the latest one in some experiment directory.

silkenelson:
I would have to check with the scientists, but for lxe I think old values are useless as often as they are ok to use: it depends on wether the laser setup for the experiment changed. But if we have a softlink and and we have a different motor connected to the controller where the ND-filter usually sits we would see an lxe that will do weird things. The lxe-motors are 'semi-permanent': I don't think they could live in happi, I would define them in the beamline file and if the calibration files exists, then add the pseudo-positioner. If they switch experiments AND the file continue to be good, I do get calls asking what happened to the lxe. I prefer that though to the problems we might invite by doing the 'wrong' thing.

ZLLentz:
@klauer I don't know of any existing devices that reference per-experiment config files, though in principle we could assign a filepath with a substitution point and require the session to specify an experiment number (which it will always have access to).

klauer:
OK, based on @ZLLentz's response, in principle it seems that we could solve this generally. I think this could be done in a follow-up PR if everyone's OK with that.

In short: if you have the config file in the directory, it loads the motor. If not, the loading fails and the motor doesn't exist in the environment. Then it's up to the scientists to keep those files up-to-date. I think that addresses your concerns, @silkenelson?

klauer:
OK, I'm pretty sure we can resolve this in a follow-up PR. There's also been a recent tag, so I'll self-merge.

PR #508, opened 2020-08-14
ENH: Signal annotations for auto-generated notepad IOC support

klauer:
Overview
----------
Paired with https://github.com/pcdshub/pcdsdevices_notepad, this is almost a full-fledged implementation, despite only taking a bit over an hour or two.

The concept here is that we could add in soft signals that have no control system backing and automatically generate autosaved PVs for them.

A common name for PCDS has been the "PV Notepad IOC", as it's a place where you can just dump information.

How
----
You'll note that the `NotepadLinkedSignal` is no more than an `EpicsSignal` with its own set of metadata.

1. That metadata is discovered by happi and https://github.com/pcdshub/pcdsdevices_notepad/blob/master/notepad_finder.py
2. Which in turn generates a file like: https://github.com/pcdshub/pcdsdevices_notepad/blob/master/config.json
3. Which in turn is consumed on-the-fly by a caproto IOC that's watching for it: https://github.com/pcdshub/pcdsdevices_notepad/blob/master/ioc.py

What if caproto breaks
------------------------
~There's a sufficient set of interfaces where we can break this up, should we need to swap out, say, the caproto IOC for an actual pvNotepad IOC down the line.~

_Down the line_ is here; we don't want caproto IOCs for now.

Traditional pvNotepad-generating scripts are now in `pcdsdevices_notepad`: https://github.com/pcdshub/pcdsdevices_notepad/pull/5


klauer:
If the new signal class is undesirable, we could alternatively use a method like `set_metadata`, where we just look for a specific attribute. In my opinion, the signal class makes sense, but I'm open to suggestions.

pcdsdevices/pseudopos.py
```diff
@@ -167,4 +168,23 @@ class SimDelayStage(DelayBase):
...
+        NotepadLinkedSignal, ':OphydReadback',
+        notepad_metadata={'record': 'ai', 'default_value': 0.0},
+    )
+
+    def __init__(self, prefix='', parent=None, **kwargs):
+        if not prefix:
+            # PseudoSingle generally does not get a prefix. Fix that here,
+            # or 'notepad_setpoint' and 'notepad_readback' will have no
+            # prefix.
+            prefix = parent.prefix
```
ZLLentz, line 186: We'll need to add prefixes to the PseudoPositioners with multiple pseudo axes to avoid name conflicts here

pcdsdevices/pseudopos.py
```diff
@@ -167,4 +168,23 @@ class SimDelayStage(DelayBase):
...
-    pass
+    notepad_setpoint = Cpt(
+        NotepadLinkedSignal, ':OphydSetpoint',
+        notepad_metadata={'record': 'ao', 'default_value': 0.0},
+    )
+
+    notepad_readback = Cpt(
+        NotepadLinkedSignal, ':OphydReadback',
+        notepad_metadata={'record': 'ai', 'default_value': 0.0},
+    )
```
ZLLentz, line 179: The implication here is that all of our PseudoSingles will have PVs associated with them, which I think is good. It does mean that these devices will load poorly (`connected=False`) until we run the notepad IOC though.

pcdsdevices/pseudopos.py
```diff
@@ -167,4 +168,23 @@ class SimDelayStage(DelayBase):
...
+    def __init__(self, prefix='', parent=None, **kwargs):
+        if not prefix:
+            # PseudoSingle generally does not get a prefix. Fix that here,
+            # or 'notepad_setpoint' and 'notepad_readback' will have no
+            # prefix.
+            prefix = parent.prefix
+
+        super().__init__(prefix=prefix, parent=parent, **kwargs)
+
+        # TODO: update the notepad PVs in this class
```
ZLLentz, line 190: It's not clear to me that there is a great way to implement this here, since it is the parent PseudoPositioner class that actually controls all the set/get logic

ZLLentz:
I think the approach here is solid

pcdsdevices/pseudopos.py
klauer: Good call. Adding on `:{self.attr_name}` would probably be the best option.

pcdsdevices/pseudopos.py
klauer: Definitely - without this IOC in place, things will break.

I'm hoping that the `find` tool will be run periodically either through a cron job or in a spawned subprocess from the IOC itself, making integrating these pretty seamless. (That's where the periodic checks for reloading that file come in, even if it's not really used currently.)

ZLLentz: Will we need to run an instance of this IOC on every subnet?

klauer: I hadn't considered breaking it up between subnets, rather assuming it was truly global. 

If we do split it up, we'd want to be selective about which PVs were on which IOC. Can that be done cleanly? Beamline metadata in happi, I assume?

ZLLentz: If it's global, where do we run the IOC so that all consumers can get their respective data, and so that all python sessions can put to the global PV? Just things to think about.

Some combination of happi metadata and PV naming should be enough to decide where these go if we split it up. Ideally we want to hit the same subnet as the associated real motor PVs.

ZryletTC: I don't think either way will be perfectly clean but I vote for splitting it up. I think the 'beamline' field in happi should probably be enough to tell you each IOC which objects it needs.

Doing it that way will still lead to lead to some gateway issues for shared objects (off the top of my head, I'm thinking of the CXI/XCS LODCM), but that's easier and/or cleaner than any method that I could imagine for setting up the gateway with this running globally.

pcdsdevices/pseudopos.py
ZryletTC, line 186: Is there any protection against someone trying to instantiate a `PseudoSingleInterface` on its own, without a parent?

ZryletTC: Well I guess that should be done in `PseudoSingle`, not here anyways

pcdsdevices/pseudopos.py
ZryletTC: Think we need a subclass of PseudoPositioner to handle these?

pcdsdevices/pseudopos.py
klauer: Resolved in 90b479a1a8

pcdsdevices/pseudopos.py
klauer: OK, will add in parameters for splitting up the config in the IOC's `notepad_finder` helper.

pcdsdevices/pseudopos.py
klauer: I'll take a look and see what's necessary for this - 1st pass was "see if they like the notepad idea and show how it could be integrated"; 2nd pass will be "actually make it work" 😄

pcdsdevices/pseudopos.py
klauer: You're absolutely right there should be some checks to ensure it's a component in a pseudopositioner device hierarchy - but I think the risk is pretty low here, as even ophyd-aware people are confused by what these PseudoSingles and such are 😀

klauer:
There's also the old, unsolved mystery of dealing with multiple clients trying to use these notepad PVs from multiple IPython sessions. I don't have any good plans for that here.

ZryletTC:
> There's also the old, unsolved mystery of dealing with multiple clients trying to use these notepad PVs from multiple IPython sessions. I don't have any good plans for that here.

What issues would this raise? As long as they don't both try to change it at the same time, each should update when the PV is updated by the other user, right?

klauer:
If we only have the notepad setpoint PVs, these will be updated only on `move()` requests.

If we keep the readback PVs as I'm proposing here, they could fight each other. Perhaps we should just avoid that entirely by not adding them?

ZLLentz:
Presumably, the multiple sessions would always agree on what the readback PV should be? So long as they are running the same pcdsdevices version?

ZLLentz:
Worst-case we have the server running the pseudo-positioner classes itself just to do the readback calculations in one centralized place

silkenelson:
I agree: these notepad PV essentially translate the inputs which hopefully are the same for all sessions.....

klauer:
Yes, though it may be a race to update values and things may vary based on per-class settings.

I think my concern may be overblown, but it's worth noting as we go through this.

In any case, I believe as of 02d4e83 all pseudopositioners have been updated to use the linked PVs, and the pcdsdevices subclass class writes to the notepad PVs on update. Once I get the notepad IOC up and running, I could use some help giving this a spin if one of you has time later.

pcdsdevices/pseudopos.py
klauer: Resolved in the reasonably minimal subclass of 02d4e83

silkenelson:
We already have the notepad IOC with all the required PVs running. Do you want something else?

klauer:
@silkenelson this PR proposes a different method, where one doesn't need to manually create a notepad IOC configuration - they're picked up dynamically by a separate process.

If we're happy with typing the full PV names, I can redo this to try and use the old method.

ZryletTC:
I don't have much time coming up soon, but if all goes well with ops, I'll be available to help test later this week.

I vote for new method!

klauer:
The hutch/beamline separation is done in the `pcdsdevices_notepad` IOC and finder utilities. Archive files are generated, as are the epicsArch/DAQ files.

I'm going to look into deploying them similar to how ioc-lfe-at2l0-calc was done (https://github.com/pcdshub/hxr-attenuator/blob/master/ioc-lfe-at2l0-calc/iocBoot/st.cmd).

klauer:
OK, I think we're at a point where we can move forward with this, trialing at one beamline for starters.

I've generated a "traditional" pvNotepad IOC configuration file for each beamline but have yet to start it up and see what happens.

klauer:
@ZryletTC I don't have access to XCS to even try deploying this thing. Would you be willing to give it a go?

`/reg/neh/home/klauer/Repos/xcs-notepad`
and
`/reg/neh/home/klauer/Repos/xcs-notepad/build/iocBoot/ioc-xcs-pcdsdevices`

klauer:
Other generated configs can be found here, by the way, `/reg/neh/home/klauer/Repos/pcdsdevices/notepad/config`

And this PR was merged on top of master a couple times, so it should be glanced at by someone other than me.

ZLLentz (approved):
This looks solid to me. Is this still stuck in the "chicken and egg" twilight zone?

klauer:
Since I have the notepad configurations generated based on this PR, I think we should be safe for now. (With the important note that the IOCs are not yet running - hoping to get @ZryletTC 's help with testing on one beamline before rolling out to others).

For the future:
* Searching pcdsdevices again for notepad signals will happen in the currently activated environment
* The old configuration won't be overwritten but rather updated - so old PVs will not be lost as long as the files aren't deleted.

I think this is about the best we can do, leaving the configuration file juggling up to us.

klauer:
Follow-up:
* @ZryletTC's preliminary testing at XCS went well, except for a race condition at startup:
<details>
<summary>Exception details</summary>

```
ERROR    Subscription value callback exception (NotepadLinkedSignal(read_pv='XCS:LAS:MMN:01:OphydOffset', name='lxt_ttc_delay_user_offset', parent='lxt_ttc_delay', value=0.0, timestamp=631152000.0, auto_monitor=False, string=False, write_pv='XCS:LAS:MMN:01:OphydOffset', limits=False, put_complete=False))
  File "/u1/conda/pcds-3.3.4/lib/python3.6/site-packages/ophyd/ophydobj.py", line 462, in inner
    cb(*args, **kwargs)
  File "/reg/g/pcds/pyps/apps/hutch-python/common/dev/devpath/pcdsdevices/pseudopos.py", line 299, in _offset_changed
    self._update_position()
  File "/reg/g/pcds/pyps/apps/hutch-python/common/dev/devpath/pcdsdevices/pseudopos.py", line 119, in _update_position
    position = super()._update_position()
  File "/u1/conda/pcds-3.3.4/lib/python3.6/site-packages/ophyd/pseudopos.py", line 666, in _update_position
    real_cur_pos = self.real_position
  File "/u1/conda/pcds-3.3.4/lib/python3.6/site-packages/ophyd/device.py", line 1197, in __getattr__
    raise AttributeError(name)
AttributeError: real_position
```
</details>

* Rollout of new notepad IOC to other hutches is held up by my lack of access to SVN (requested last week, no updates yet)
* Once those IOCs are deployed, we can merge this and move forward

klauer:
Continuing this verbose PR, 4636fb6 brings in support for _completely optional_ notepad signals. 

If this works well enough, there should be no chicken/egg problem in the future. Just unhappy users when our notepad IOC doesn't get updated quickly enough.

PR #509, opened 2020-08-16
BUG: Fix typo in PV name of BeckhoffJet slits

ZryletTC:
## Description
IOC had ":SS", class had ":SLIT". Fixed the class to match.

## Motivation and Context
Devices work better when their PVs actually exist.

## How Has This Been Tested?
Ran xcs3, and `ljh` shows as connected now.

Nowhere


josephmlev:
Not sure where `:SLIT` came from, it is `:SS` on the confluence page. Looks good, especially if it works :D

Issue #488, opened 2020-07-27
Missing LODCM, CCM Features 


ZLLentz:
## Expected Behavior

- Energy motor
- Refer to old beamline python

## Current Behavior

- LODCM has no energy calculation
- CCM has an energy calculation, but it isn't easy to find
- Not sure if full set from old python is reflected in the CCM, need to research.

## Context

Silke is reviving the old code because it doesn't exist here, rather than making it exist here.

ZLLentz:
The current behavior of the `CCM` class is as follows:
- `ccm.calc.energy` is a motor that does calculations and moves the `alio` appropriately
- `ccm.theta2fine` is a motor that moves the fine theta2 control
- `ccm.x` is a motor that moves the upstream and downstream x motors at the same time.
- `ccm.y` is a motor that moves the upstream and downstream y motors at the same time.
- ccm `inserted` and `removed` states are defined by `in_pos` and `out_pos` passed in parameters, which check the x state. The `insert` and `remove` functions do the correct things.

I am going to instantiate a dummy CCM and make sure it behaves like I expect.

ZLLentz:
I am having trouble getting my dummy object to scan... but otherwise it looks ok

ZLLentz:
```
In [2]: fake_ccm.calc.alio.kind = 'hinted'

In [3]: RE(scan([fake_ccm.calc.alio], fake_ccm.calc.energy, 7, 8, 10))


Transient Scan ID: 2     Time: 2020-08-05 14:51:28
Persistent Unique Scan ID: '0d039684-5af4-4795-bef6-a7b55187b7e8'
New stream: 'primary'
+-----------+------------+----------------------+--------------------+
|   seq_num |       time | fake_ccm_calc_energy | fake_ccm_calc_alio |
+-----------+------------+----------------------+--------------------+
|         1 | 14:51:32.4 |                7.000 |              5.894 |
|         2 | 14:51:32.5 |                7.111 |              4.826 |
|         3 | 14:51:32.5 |                7.222 |              3.792 |
|         4 | 14:51:32.5 |                7.333 |              2.791 |
|         5 | 14:51:32.6 |                7.444 |              1.821 |
|         6 | 14:51:32.6 |                7.556 |              0.881 |
|         7 | 14:51:32.6 |                7.667 |             -0.031 |
|         8 | 14:51:32.6 |                7.778 |             -0.916 |
|         9 | 14:51:32.6 |                7.889 |             -1.775 |
[...]

PR #513, opened 2020-08-19
ENH: CCM with Vernier

### LINKED ISSUE #498: PVPositioner[Comparator] based on CCMMotor implementation
An idea that we could generalize and re-use this PVPositioner class from https://github.com/pcdshub/pcdsdevices/blob/a5e803867bfb3fa365d499a6b4aa23c29a9aa4a9/pcdsdevices/ccm.py#L30-L73

Requiring either atol or a callable that takes in setpoint/readback or something?

ZLLentz:
## Description

Add vernier integration into the CCM class.
This includes a beam requests pv positioner and a vernier with ccm energy motor.
Also adds new positioner base classes whose utility seems to crop up multiple times.

The vernier is something the accelerator crew can use to adjust the beam energy (frequency/color, not intensity).
We manipulate the vernier when it has been agreed upon that we can do so with ACR.
We do this by putting to an energy request PV, then ACR satisfies the request.
We need to move the energy during energy scans that are larger than the SASE bandwidth.

Contributes to #488 
Closes #498

Some unanswered questions:
- Should this really be called the vernier if it is a generic PV that the accelerator chooses how to satisfy?
- Should the vernier move to exactly our energy value? Or is it better to move a little bit past it so we can scan the full SASE?

## Motivation and Context

Large-range energy scans need accelerator adjustments.

## How Has This Been Tested?

Added tests for all permutations of vernier moves

Docstrings only

### DIFF
--- a/pcdsdevices/beam_stats.py
+++ b/pcdsdevices/beam_stats.py
@@ -1,8 +1,10 @@
 from ophyd.device import Component as Cpt
 from ophyd.device import Device
-from ophyd.signal import AttributeSignal, EpicsSignalRO
+from ophyd.device import FormattedComponent as FCpt
+from ophyd.signal import AttributeSignal, EpicsSignal, EpicsSignalRO
 
 from .interface import BaseInterface
+from .pv_positioner import PVPositionerIsClose
 from .signal import AvgSignal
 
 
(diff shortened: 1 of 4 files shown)


klauer:
Those utility classes are great - we'll get a lot of use out of them.

pcdsdevices/beam_stats.py
```diff
@@ -26,3 +28,63 @@ class SxrGmd(Device):
 
     def __init__(self, prefix='', name='sxr_gmd', **kwargs):
         super().__init__(prefix=prefix, name=name, **kwargs)
+
+
+class BeamEnergyRequest(PVPositionerIsClose):
+    """
+    Positioner to request beam color changes from ACR in eV.
+
+    It is up to ACR how to and whether to fulfil these requests. This is often
```
klauer, line 37:
```suggestion
    It is up to ACR how to and whether to fulfill these requests. This is often
```

pcdsdevices/beam_stats.py
```diff
@@ -26,3 +28,63 @@ class SxrGmd(Device):
...
+    fulfilled by moving the Vernier but can also be a more involved process.
+
+    Motion is considered "done" when the photon energy readback is close enough
+    to the requested value. The default tolerance here is 30 eV, but this can
+    be changed on a per-instance basis by passing atol into the init, or on a
```
klauer, line 42:
```suggestion
    be changed on a per-instance basis by passing ``atol`` into the initializer, or on a
```

pcdsdevices/beam_stats.py
ZLLentz: TIL fulfil/fulfill has two spellings https://www.grammarly.com/blog/fulfil-fulfill/

pcdsdevices/beam_stats.py
ZLLentz: I think this rides me over an 80-length line, I'll make the equivalent change in a different commit

pcdsdevices/beam_stats.py
ZLLentz: We will go with the US spelling as suggested

klauer: 2nded that TIL... huh. It jumped out with the `fulfilled` below

ZLLentz:
I'm going to pull this up and make sure we connect to all PVs for both XPP and XCS CCMs before merging

ZLLentz:
Everything looks good on the connect. There is some oddity with some of the values not having units associated with them, but I have to assume that either the EPICS PVs have incomplete metadata or that the gateway is eating it. Either way, probably not something for the Pythons to worry about.

ZLLentz:
Ah! Actually, the problem is that the status prints use the `position` attribute, which has no units data on it! This needs some revising.

ZLLentz:
But this PR is done until we get blank time with real hardware to try it on. I suspect it will work but it may need some tweaking.

klauer:
Ahh, that makes sense - maybe that's where my frustrated getattr/getitem solution had come in... 🤷‍♂️

PR #514, opened 2020-08-19
ENH: Add support for Thorlabs WFS-40 detector

slactjohnson:
## Description
Adds in a new device for the Thorlabs WFS40 detector. 

## Motivation and Context

This device is used in the laser control system. 
## How Has This Been Tested?

Tested using a WFS-40 in the B40 Heinz lab. The component PVs work, but Typhos hangs when attempting to create a screen. This appears to be related to the parent class PCDSAreaDetectorTyphos, rather than the WFS class, but that can be fixed in a different PR. 

It hasn't, outside of the source. However, I think the component names are pretty self-explanatory.

### DIFF
--- a/pcdsdevices/lasers/thorlabsWFS.py
+++ b/pcdsdevices/lasers/thorlabsWFS.py
@@ -4,10 +4,12 @@ from ophyd.device import Component as Cpt
 from ophyd.device import Device 
 from ophyd.signal import EpicsSignal, EpicsSignalRO
 
-from pcdsdevices.areadetector.detectors import PCDSAreaDetectorBase
+from pcdsdevices.areadetector.detectors import PCDSAreaDetectorTyphos
 
-class ThorlabsWfs40(PCDSAreaDetectorBase):
+class ThorlabsWfs40(PCDSAreaDetectorTyphos):
     """Class to implement a Thorlabs WFS40 Wavefront sensor."""
+
+    beam_status = Cpt(EpicsSignalRO, 'StatHighAmbientLight', kind='normal')
 
     pupil_centroid_x = Cpt(EpicsSignal, 'PupilCentroidX', kind='config')
     pupil_centroid_y = Cpt(EpicsSignal, 'PupilCentroidY', kind='config')
--- a/pcdsdevices/lasers/thorlabsWFS.py
+++ b/pcdsdevices/lasers/thorlabsWFS.py
@@ -1,10 +1,12 @@
 import logging
 
 from ophyd.device import Component as Cpt
-from ophyd.device import Device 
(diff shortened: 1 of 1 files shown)


pcdsdevices/lasers/thorlabsWFS.py
```diff
@@ -0,0 +1,49 @@
...
+    pupil_diameter_y = Cpt(EpicsSignal, 'PupilDiameterY', kind='config')
+
+    pupil_centroid_x_rbv = Cpt(EpicsSignal, 'PupilCentroidX_RBV',
+                               kind='normal')
+    pupil_centroid_y_rbv = Cpt(EpicsSignal, 'PupilCentroidY_RBV',
+                               kind='normal')
+    pupil_diameter_x_rbb = Cpt(EpicsSignal, 'PupilDiameterX_RBV',
+                               kind='normal')
+    pupil_diameter_y_rbv = Cpt(EpicsSignal, 'PupilDiameterY_RBV',
+                               kind='normal')
```
ZLLentz, line 28: Are these RBV PVs are meant to be `EpicsSignalRO`?

pcdsdevices/lasers/thorlabsWFS.py
```diff
@@ -0,0 +1,49 @@
...
+    pupil_centroid_x = Cpt(EpicsSignal, 'PupilCentroidX', kind='config')
+    pupil_centroid_y = Cpt(EpicsSignal, 'PupilCentroidY', kind='config')
+    pupil_diameter_x = Cpt(EpicsSignal, 'PupilDiameterX', kind='config')
+    pupil_diameter_y = Cpt(EpicsSignal, 'PupilDiameterY', kind='config')
+
+    pupil_centroid_x_rbv = Cpt(EpicsSignal, 'PupilCentroidX_RBV',
+                               kind='normal')
+    pupil_centroid_y_rbv = Cpt(EpicsSignal, 'PupilCentroidY_RBV',
+                               kind='normal')
+    pupil_diameter_x_rbb = Cpt(EpicsSignal, 'PupilDiameterX_RBV',
```
ZLLentz, line 25: typo rbb

slactjohnson: D'oh!

pcdsdevices/lasers/thorlabsWFS.py
slactjohnson: No, that would be giving me too much credit. I wrote most of this prior to learning about that signal type, and I forgot to update it here.

slactjohnson: Yes, they should be RO.

PR #518, opened 2020-08-20
ENH: add UnitConversionDerivedSignal

klauer:
## Description
Adds a new signal class to aid in simple unit conversions.
This is the first part of the upcoming `lxt` motor support.

## Motivation and Context
Though it's preferable to do these at the IOC level so all clients - not just hutch-python/typhos - benefit, it's not always possible.

## How Has This Been Tested?
A simple unit test.

Docstring

### DIFF
--- a/pcdsdevices/signal.py
+++ b/pcdsdevices/signal.py
@@ -306,7 +306,6 @@ class UnitConversionDerivedSignal(DerivedSignal):
     Custom units may be specified for the original signal, or if specified, the
     original signal's units may be retrieved upon first connection.
 
-
     Parameters
     ----------
     derived_from : Signal or str
@@ -314,6 +313,13 @@ class UnitConversionDerivedSignal(DerivedSignal):
         attribute name that indicates a sibling to use.  When used in a
         ``Device``, this is then simply the attribute name of another
         ``Component``.
+
+    derived_units : str
+        The desired units to use for this signal.
+
+    original_units : str, optional
+        The units from the original signal.  If not specified, control system
+        information regarding units will be retrieved upon first connection.
 
     write_access : bool, optional
         Write access may be disabled by setting this to ``False``, regardless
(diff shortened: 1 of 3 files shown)


ZLLentz:
This is pretty snazzy

pcdsdevices/signal.py
```diff
@@ -293,3 +297,63 @@ def put(self, value, *, timestamp=None, force=False):
...
+    write_access : bool, optional
+        Write access may be disabled by setting this to ``False``, regardless
+        of the write access of the underlying signal.
+
+    name : str, optional
+        The signal name.
+
+    parent : Device, optional
+        The parent device.  Required if ``derived_from`` is an attribute name.
+    """
```
ZLLentz, line 327: Looks like this documents the *args and **kwargs, should also add in the `derived_units` and `original_units` too

tests/test_signal.py
```diff
@@ -46,3 +48,39 @@ def test_avg_signal():
...
+        event.set()
+
+    converted.subscribe(callback, run=False)
+    orig.put(20, wait=True)
+    event.wait(1)
+    cb.assert_called_once()
+
+    args, kwargs = cb.call_args
+    assert kwargs['value'] == 20_000
+    assert converted.get() == 20000
```
ZLLentz, line 86: I totally forgot that `20_000` was a thing, but it kind of bugs me you used it on one check and not the other

klauer: OCD issues resolved in 4110b0e  ;)

pcdsdevices/signal.py
klauer: D'oh, thanks - good catch. Lazy docstring writing.

pcdsdevices/signal.py
```diff
@@ -293,3 +297,72 @@ def put(self, value, *, timestamp=None, force=False):
...
+    def _derived_metadata_callback(self, *, connected, **kwargs):
+        super()._derived_metadata_callback(connected=connected, **kwargs)
+        if connected and 'units' in kwargs:
+            if self.original_units is None:
+                self.original_units = kwargs['units']
+
+    def describe(self):
+        full_desc = super().describe()
+        desc = full_desc[self.name]
+        desc['units'] = self.derived_units
```
ZryletTC, line 363: Hey Ken, I notice you put this here but didn't update the value of units in metadata. Any reason why not and/or any clue on a good way to do so? I tried a couple things but they all either didn't work or broke other things.

klauer: Let's make an issue out of it and revisit. I'm sure we can shove it in the metadata with some additional work.

PR #519, opened 2020-08-20
ENH: add LaserTiming PVPositioner

### LINKED ISSUE #516: Add Vitara device / lxt motor
## Expected Behavior
* Add device as found in the old hutch python
* Figuring out what the device _is_ would be nice, too

## Current Behavior
Does not appear to exist, unless under a different name

klauer:
## Description
Adds "lxt" motor.

* Closes #516 
* Hopefully closes https://github.com/pcdshub/Bug-Reports-and-Requests/issues/27
* Previous PR and PVPositioner made this pretty simple, fortunately!
* Unclear as to whether or not we should implement the `stop` signal here. We know that there's a motor record sitting underneath, but the old code did not touch it (other than using its DMOV flag in a very convoluted way).
* Apologies for the dumb 'lxe' module name. I think this will have to be reorganized, but I don't really know where to put it.

## Motivation and Context
Required from old hutch python config.

## How Has This Been Tested?
Simple unit tests. Also locally with a simple caproto IOC:

```python
    mms_ph = SubGroup(FakeMotor, velocity=3., precision=2, prefix=':MMS:PH')
    fs_tgt_time = pvproperty(value=0.0, name=':VIT:FS_TGT_TIME')

    @fs_tgt_time.putter
    async def fs_tgt_time(self, instance, value):
        await self.mms_ph.motor.write(value / 1e9)
```

### DIFF
--- a/pcdsdevices/lxe.py
+++ b/pcdsdevices/lxe.py
@@ -214,7 +214,7 @@ class LaserEnergyPositioner(LookupTablePositioner, FltMvInterface):
         return super().wm[0]
 
 
-class LaserEnergyTimingLxt(PVPositioner, FltMvInterface):
+class LaserTiming(PVPositioner, FltMvInterface):
     """
     "lxt" motor, which may also have been referred to as Vitara.
 
--- a/pcdsdevices/lxe.py
+++ b/pcdsdevices/lxe.py
@@ -235,4 +235,9 @@ class LaserEnergyTimingLxt(PVPositioner, FltMvInterface):
     done_value = 1
 
     def __init__(self, prefix='', *, egu=None, **kwargs):
+        if egu not in (None, 's'):
+            raise ValueError(
+                f'{self.__class__.__name__} is pre-configured to work in units'
+                f' of seconds.'
+            )
         super().__init__(prefix, egu='s', **kwargs)
--- a/pcdsdevices/lxe.py
+++ b/pcdsdevices/lxe.py
@@ -35,11 +35,13 @@ import typing
 import numpy as np
 
 from ophyd import Component as Cpt
+from ophyd import EpicsSignal, PVPositioner
 
 from .epics_motor import EpicsMotorInterface
 from .interface import FltMvInterface
 from .pseudopos import (LookupTablePositioner, PseudoSingleInterface,
                         pseudo_position_argument)
+from .signal import UnitConversionDerivedSignal
 
 if typing.TYPE_CHECKING:
     import matplotlib  # noqa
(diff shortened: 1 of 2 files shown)


pcdsdevices/lxe.py
```diff
@@ -210,3 +212,27 @@ def move(self, position, wait=True, timeout=None, moved_cb=None):
...
+                   original_units='ns',
+                   )
+
+    # A motor (record) will be moved after the above record is touched, so
+    # use its done motion status:
+    done = Cpt(EpicsSignal, ':MMS:PH.DMOV', auto_monitor=True)
+    done_value = 1
+
+    def __init__(self, prefix='', *, egu=None, **kwargs):
+        super().__init__(prefix, egu='s', **kwargs)
```
ZLLentz, line 238: These two lines are kind of strange, the default prefix surely never works and the user's egu argument is ignored if passed.

pcdsdevices/lxe.py
```diff
@@ -210,3 +212,27 @@ def move(self, position, wait=True, timeout=None, moved_cb=None):
...
+    """
+    "lxt" motor, which may also have been referred to as Vitara.
+
+    Conversions for Vitara FS_TGT_TIME nanoseconds <-> seconds are done
+    internally, such that the user may work in units of seconds.
+    """
+
+    _fs_tgt_time = Cpt(EpicsSignal, ':VIT:FS_TGT_TIME', auto_monitor=True)
+    setpoint = Cpt(UnitConversionDerivedSignal,
+                   derived_from='_fs_tgt_time',
```
ZLLentz, line 227: I should use derived signals more, they seem handy

pcdsdevices/lxe.py
klauer: The `prefix=''` is a carryover from the superclass `__init__`.

I'll add a check for `egu` to complain if it's set, as I don't plan on having it customizable. Users appear to want to work with seconds here, so adding configuration is overkill.

pcdsdevices/lxe.py
```diff
@@ -210,3 +212,27 @@ def move(self, position, wait=True, timeout=None, moved_cb=None):
     def wm(self) -> float:
         # Remove the PseudoPosition tuple for FltMvInterface compatibility
         return super().wm[0]
+
+
+class LaserEnergyTimingLxt(PVPositioner, FltMvInterface):
```
ZLLentz, line 217: I think this deserves at least a quick look from @silkenelson to confirm that this class name makes sense

tests/test_pseudopos.py
```diff
@@ -145,3 +148,26 @@ def no_op(*args, **kwargs):
...
+            pv_positioner.done.sim_put(1 - pv_positioner.done_value)
+            pv_positioner.done.sim_put(pv_positioner.done_value)
+        return st
+
+    lxt = make_fake_device(LaserEnergyTimingLxt)('prefix', name='lxt')
+    lxt._fs_tgt_time.sim_set_limits((0, 4e9))
+    lxt._fs_tgt_time.sim_put(0)
+
+    # A basic dependency sanity check...
+    np.testing.assert_allclose(convert_unit(1, 's', 'ns'), 1e9)
```
ZLLentz, line 167: :rocket:

pcdsdevices/lxe.py
klauer: Resolved in 86b14de

pcdsdevices/lxe.py
klauer: Name and high-level functionality if possible, too! (@silkenelson)

klauer:
Renamed to `LaserTiming` as the previous name was apparently incorrect.

klauer:
👍 We _may_ rename this one again, but this wouldn't be the first instance of that happening... Merging

PR #520, opened 2020-08-20
ENH: add units to the handy IPython repr

### LINKED ISSUE #507: Add engineering units to IPython repr summary?
It'd be nice to know the units of items, if available

Instead of:
```
energy
------
  position: 10.00
```

Including units:
```
energy
------
  position: 10.00 uJ
```

klauer:
Closes #507 

Could use further testing, but in brief poking around the result looks something like:
```
In [22]: class MyDevice(pcdsdevices.interface.BaseInterface, ophyd.Device):
    ...:     cpt = ophyd.Component(ophyd.EpicsSignal, 'LCLS:HXR:BEAM:EV')
    ...:

In [23]: dev = MyDevice(name='dev')

In [24]: dev
Out[24]:

dev
---
cpt: 9500 [eV]
```

where `dev.cpt.metadata['units'] == 'eV'`

### DIFF
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -14,7 +14,6 @@ from types import MethodType, SimpleNamespace
 from weakref import WeakSet
 
 import yaml
-
 from bluesky.utils import ProgressBar
 from ophyd.device import Device
 from ophyd.ophydobj import Kind, OphydObject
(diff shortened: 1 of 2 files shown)


ZLLentz:
Tests failed on the mint umvr race condition

pcdsdevices/interface.py
```diff
@@ -241,6 +243,17 @@ def get_value(signal):
     return None
 
 
+def get_units(signal):
+    attrs = ('units', 'egu', 'derived_units')
```
ZLLentz, line 247: I never realized we had so many places to store this info

klauer: I _think_ for the most part it'll get it right on the first `metadata['units']` attempt, but to be honest I just shrugged and thought _meh, assume nothing, try all the places_.

PR #529, opened 2020-08-24
ENH: log calls to mv, umv at level info.

ZLLentz:
## Description

Log what we're moving, where we're moving from, and where we're moving to in the `mv` family of functions.

## Motivation and Context

- Tim said this would be helpful, and it seemed easy to implement.
- Will not spam during scans because those don't use `mv`

## How Has This Been Tested?

I moved some motors

### DIFF
--- a/docs/source/upcoming_release_notes/0529-mv-info.rst
+++ b/docs/source/upcoming_release_notes/0529-mv-info.rst
@@ -1,0 +1,31 @@
+529 Mv Info
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- Moves using mv and umv will log their moves at info level for interactive
+  use to keep track of the sessions.
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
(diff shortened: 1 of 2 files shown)


pcdsdevices/interface.py
```diff
@@ -389,6 +389,11 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._mov_ev = Event()
 
+    def _log_move(self, position):
+        template = 'Moving {} from {} to {}'
+        msg = template.format(self.name, self.wm(), position)
+        logger.info(msg)
```
ZryletTC, line 395:
```suggestion
        logger.info('Moving {} from {} to {}', self.name, self.wm(), position)
```

docs/source/upcoming_release_notes/0529-mv-info.rst
```diff
@@ -0,0 +1,31 @@
...
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- Moves using mv and umv will log their moves at info level for interactive
+  use to keep track of the sessions.
```
klauer, line 11: 👏

klauer:
CI was bugged a bit here; Travis both passed and was hung up starting to run 🤷‍♂️

PR #533, opened 2020-08-25
ENH/FIX: user offsets for lxt motor and general UnitConversionDerivedSignal

### LINKED ISSUE #531: Offset for LaserTiming
## Expected Behavior
User offset can be set to create a user readback offset from the dial value.

## Current Behavior
No offset possible.

## Possible Solution
Implement an offset PV.

## Context
Scientists want to zero the lxt motor.

### LINKED ISSUE #532: LXT has no hinted signals, preventing LiveTable
## Expected Behavior

Hinted readback

## Current Behavior

No hints = LiveTable crash

## Possible Solution

IDK, livetable doesn't display 1e-9 very well anyway.
We used the hidden signal in the xcs beamline because the units were better for the table

klauer:
## Description

  support.
  ``LiveTable`` support.

Tweaks the developer utility script, adding release note files to `git` if the editor reported success.

## Motivation and Context
Closes #531 
Closes #532 

## How Has This Been Tested?
Included unit tests.

Docstrings have been updated.

### DIFF
--- a/pcdsdevices/lxe.py
+++ b/pcdsdevices/lxe.py
@@ -257,6 +257,5 @@ class LaserTiming(PVPositioner, FltMvInterface):
             The new current position.
         '''
         self.user_offset.put(0.0)
-
-        new_offset = self.setpoint.get() - position
+        new_offset = self.setpoint.get() + position
         self.user_offset.put(new_offset)
--- a/pcdsdevices/lxe.py
+++ b/pcdsdevices/lxe.py
@@ -33,9 +33,9 @@ import types
 import typing
 
 import numpy as np
-
 from ophyd import Component as Cpt
 from ophyd import EpicsSignal, PVPositioner
+from ophyd.signal import AttributeSignal
 
 from .epics_motor import EpicsMotorInterface
 from .interface import FltMvInterface
@@ -222,16 +222,20 @@ class LaserTiming(PVPositioner, FltMvInterface):
     internally, such that the user may work in units of seconds.
     """
 
-    _fs_tgt_time = Cpt(EpicsSignal, ':VIT:FS_TGT_TIME', auto_monitor=True)
+    _fs_tgt_time = Cpt(EpicsSignal, ':VIT:FS_TGT_TIME', auto_monitor=True,
+                       kind='omitted')
     setpoint = Cpt(UnitConversionDerivedSignal,
                    derived_from='_fs_tgt_time',
                    derived_units='s',
                    original_units='ns',
+                   kind='hinted',
                    )
+    user_offset = Cpt(AttributeSignal, attr='setpoint.user_offset',
+                      kind='config')
 
     # A motor (record) will be moved after the above record is touched, so
     # use its done motion status:
-    done = Cpt(EpicsSignal, ':MMS:PH.DMOV', auto_monitor=True)
+    done = Cpt(EpicsSignal, ':MMS:PH.DMOV', auto_monitor=True, kind='omitted')
     done_value = 1
 
     def __init__(self, prefix='', *, egu=None, **kwargs):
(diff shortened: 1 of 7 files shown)


pcdsdevices/signal.py
```diff
@@ -335,21 +341,56 @@ class UnitConversionDerivedSignal(DerivedSignal):
         Keyword arguments are passed to the superclass.
     """
 
+    derived_units: str
+    original_units: str
```
ZLLentz, line 345: I've never seen this before. Is this how you type hint a class attribute?

klauer: Yes, it's the new-style class instance attribute type hint annotation style.

Apologies for inserting these here in the first place; my side-project coding style has been leaking out of its own context recently.

pcdsdevices/signal.py
```diff
@@ -335,21 +341,56 @@ class UnitConversionDerivedSignal(DerivedSignal):
...
+    @property
+    def user_offset(self) -> typing.Optional[typing.Any]:
+        """A user-specified offset in *derived*, user-facing units."""
+        return self._user_offset
+
+    @user_offset.setter
+    def user_offset(self, offset):
+        self._user_offset = offset
+        self._recalculate_position()
+
```
ZLLentz, line 380: I suspect a very quick request once a user gets their hands on this will be to ask for a method that updates the offset such that the current position matches set_position, e.g. lxt.set_position(0) could be used to zero the position value.

I'm not sure what to call this method or whether it is worth doing before someone asks.

klauer: I think that's a "calibrate"-style operation. Reasonable name `calibrate_zero_position` or just `set_zero_position`?

pcdsdevices/signal.py
klauer: To be clear, I don't mind removing these at your discretion as these don't really fit with the code style here.

pcdsdevices/signal.py
klauer: OK, will add.

pcdsdevices/lxe.py
```diff
@@ -241,3 +245,18 @@ def __init__(self, prefix='', *, egu=None, **kwargs):
...
+        ``position`` as the new current position.
+
+        Parameters
+        ----------
+        position
+            The new current position.
+        '''
+        self.user_offset.put(0.0)
+
+        new_offset = self.setpoint.get() - position
```
ZryletTC, line 261: Tbh, I was a little confused by your 'added when setting, subracted when reading' note but it makes sense to me now that I'm looking at it with a clear mind and this equation helps to clarify.

Is this convention consistent with EpicsMotor offsets and motor record offsets? Personally, I would expect the offset to work the other way around: Dial + Offset = RBV

klauer: I think you're probably right. When there's two options like this, I generally get it wrong 😆 

Checking the motor record docs: `userVAL = DialVAL * DIR + OFFset`
Indicates that: `user value = dial value + offset`
And if my advanced algebra skills haven't failed me: `dial value = user_value - offset`

I'll mark this as draft again and fix it up. Thanks for being thorough here 👍

klauer:
OK, I think this is more consistent.

Also tacked on a couple extras: removing the unit conversion deprecation warning using equivalent code, testing limits after using `set_current_position`.

pcdsdevices/signal.py
```diff
@@ -322,11 +322,16 @@ class UnitConversionDerivedSignal(DerivedSignal):
...
-        signal, and *subtracted* when calculating the derived value.
+        An optional user offset that will be *subtracted* when updating the
+        original signal, and *added* when calculating the derived value.
         This offset should be supplied in ``derived_units`` and not
         ``original_units``.
 
+        For example, if the original signal updates to a converted value of
+        500 ``derived_units`` and the ``user_offset`` is set to 100, this
+        ``DerivedSignal`` will show a value of 600.  When providing a new
+        setpoint, the ``user_offset`` will be subtracted.
```
ZryletTC, line 333: 🥇

PR #534, opened 2020-08-25
TST: LookupTable limit testing

### LINKED ISSUE #530: PV put during LookupTablePositioner init
## Expected Behavior

No puts

## Current Behavior

Can do EPICS puts, which either succeed and change the beamline or fail and cause the object to not be created.
https://github.com/pcdshub/pcdsdevices/blob/b59c9ade3f4e5007851528c54c7be942aeefdcd2/pcdsdevices/pseudopos.py#L240
Setting a limits attribute on a positioner is not safe as it can be linked to a PV.

klauer:
## Description
Does not include original proposed functionality, but retains additional tests from it.

Discussion/agreement on path forward in #530

Should probably merge after https://github.com/pcdshub/pcdsdevices/pull/536

No changelog item as this is really test-only.

## How Has This Been Tested?
See included unit tests.


klauer:
Removed functionality, updated description, marking as draft pending #536

klauer:
Kicking travis to re-run on master; I think this should be good to go.

PR #536, opened 2020-08-25
MNT: local limit overrides, rely more on base class after updates

ZLLentz:
## Description

- Provide mechanisms for overriding the epics limits in software without changing them
- Remove all epics gets/puts from limit properties in our `EpicsMotorInterface`
- Use base class's mechanisms for getting epics limits. We had overrided these, but upstream ophyd has fixed this long ago (and handled it better than we were)

## Motivation and Context

https://github.com/pcdshub/pcdsdevices/issues/530#issuecomment-680276993

## How Has This Been Tested?

Added relevant test cases.

Going to write it up right now

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,6 +21,8 @@ MODULE_PATH = Path(__file__).parent
 # Needs to not pass tons of kwargs up to Signal.put
 warnings.filterwarnings('ignore',
                         message='Signal.put no longer takes keyword arguments')
+
+
 # Other temporary patches to FakeEpicsSignal
 def check_value(self, value):
     if value is None:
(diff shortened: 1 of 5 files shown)


ZLLentz:
This was an absolute nightmare because `FakeEpicsSignal` needs a full rewrite to handle metadata properly.

tests/conftest.py
```diff
@@ -21,8 +22,27 @@
 warnings.filterwarnings('ignore',
                         message='Signal.put no longer takes keyword arguments')
 # Other temporary patches to FakeEpicsSignal
+def check_value(self, value):
```
klauer, line 25: Should upstream this and others. Looks good though.

ZLLentz: Yeah, I kind of want to upstream this and I kind of want to avoid it because I'll fall down the "make it actually good" rabbit hole...

klauer: Fair!

PR #538, opened 2020-08-25
FIX: tab completion + inheritance

klauer:
## Description
  been changed.
  added.

## Motivation and Context
Class hierarchy issues caused tab completion to break. The root cause was slightly more complicated than one might have expected, necessitating updating the majority of devices.

## How Has This Been Tested?
Added a fairly extensive regression test + tab completion test.

Indicates that there are still some classes that may need updating (marked with `SKIPPED`).

Release notes + docstrings + helper exception messages.

### DIFF
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -41,7 +41,7 @@ Positioner_whitelist = ["settle_time", "timeout", "egu", "limits", "move",
                         "position", "moving"]
 
 
-class BaseInterface(OphydObject):
+class BaseInterface:
     """
     Interface layer to attach to any Device for SLAC features.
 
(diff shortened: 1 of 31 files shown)


pcdsdevices/interface.py
```diff
@@ -72,7 +84,8 @@ def __init_subclass__(cls, **kwargs):
                 for cpt_name in parent.component_names:
                     if getattr(parent, cpt_name).kind != Kind.omitted:
                         string_whitelist.append(cpt_name)
-        cls._tab_regex = re.compile("|".join(string_whitelist))
+
+        cls._tab_regex = re.compile("|".join(sorted(set(string_whitelist))))
```
ZryletTC, line 88: 👍👍👍

tests/test_interface.py
```diff
@@ -184,3 +186,22 @@ def test_dir_whitelist_basic(fast_motor):
...
+
+
+@pytest.mark.parametrize('cls', conftest.all_device_classes)
+def test_per_class_whitelist(cls):
+    if BaseInterface not in cls.mro():
+        ignore_list = {'.areadetector.', }
+        fully_qualified_name = f'{cls.__module__}.{cls.__name__}'
+        if any(name in fully_qualified_name for name in ignore_list):
+            pytest.skip()
+        pytest.skip(f'{cls} does not inherit from the interface')
```
ZryletTC, line 198: Probably doesn't matter but I kinda feel like the ignore_list check should be outside/before the BaseInterface if branch.

klauer: Good call - refactored as such.

klauer:
Thanks, @ZLLentz 👍 

@ZryletTC - despite the tests that were added, a big PR like this makes me a bit nervous. Would you mind taking this for a spin and see that this still works on the beamline and fixes tab completion for you?

ZryletTC:
Yeah, that's understandable. I'll test it out today and let you know if I run into any issues.

ZryletTC:
Seems to be working great! Thanks for fixing it :)

klauer:
🎉

PR #549, opened 2020-08-27
MNT: rework beam energy positioner to be setpoint-only

### LINKED ISSUE #546: BeamEnergyRequest should not have any links to the actual beam Energy
This is causing issues with waiting, unexpected move requests, etc.

From Silke:

"I'm not sure we have a super reliable energy we get from AD: it varies from shot-to-shot. And I'm not sure of the PMPS number takes the vernier into account"

ZLLentz:
## Description

The beam energy positioner only has a setpoint. The readback I had originally picked was not appropriate for the use case for various reasons. I have resolved this by making a new base class.

## Motivation and Context

closes #546 

## How Has This Been Tested?

Passes old tests

Oops, I'll go write the pre-release docs

### DIFF
--- a/docs/source/upcoming_release_notes/546-beam-req.rst
+++ b/docs/source/upcoming_release_notes/546-beam-req.rst
@@ -1,0 +1,34 @@
+IssueNumber Title
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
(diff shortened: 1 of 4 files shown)


ZLLentz:
At some point, `PVPositioner` did not require a `done`! (I think)

PR #552, opened 2020-08-28
MNT: fix issue with epics setpoint limits being (None, None)

ZLLentz:
## Description

Extend the check to also check `(None, None)` and not just `None`

## Motivation and Context

LXE is not initializing because its newport is initializing to limits `(None, None)`

## How Has This Been Tested?

LXE initializes now

Patches bug that was introduced after to the last tag, so I'll overlook it


ZLLentz:
I will merge, and add more in a future PR if there are other reasons why these devices are failing.

PR #554, opened 2020-08-28
MNT: Fix umv and tweak issues

### LINKED ISSUE #281: Tweak should be more aggressive about clearing lines
I'm not sure how the old python got around this, but sometimes you can get a longer readback on one line during a tweak and the `/r` character followed by the new print isn't enough to fully clear the previous line, so you can get malformed text on the screen

### LINKED ISSUE #481: Race Condition in test_umv
This test fails only sometimes:
```
================================================================================================== FAILURES ==================================================================================================
__________________________________________________________________________________________________ test_umv __________________________________________________________________________________________________

slow_motor = SlowMotor(, name=sim_slow)

    @pytest.mark.timeout(5)
    def test_umv(slow_motor):
        logger.debug('test_umv')
        slow_motor._set_position(5)
>       slow_motor.umvr(2)

tests/test_interface.py:41:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pcdsdevices/interface.py:299: in umvr
pcdsdevices/interface.py:280: in umv
    status_wait(status)
/reg/neh/home/zlentz/miniconda3/envs/dev/lib/python3.6/site-packages/ophyd/status.py:894: in wait
    return status.wait(timeout)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = MoveStatus(done=True, pos=sim_slow, elapsed=0.1, success=False, settle_time=0.0), timeout = None

    def wait(self, timeout=None):
        """
        Block until the action completes.

        When the action has finished succesfully, return ``None``. If the
        action has failed, raise the exception.

        Parameters
        ----------
        timeout: Union[Number, None], optional
            If None (default) wait indefinitely until the status finishes.

        Raises
        ------
        WaitTimeoutError
            If the status has not completed within ``timeout`` (starting from
            when this method was called, not from the beginning of the action).
        StatusTimeoutError
            If the status has failed because the *timeout* that it was
            initialized with has expired.
        Exception
            This is ``status.exception()``, raised if the status has finished
            with an error.  This may include ``TimeoutError``, which
            indicates that the action itself raised ``TimeoutError``, distinct
            from ``WaitTimeoutError`` above.
        """
        if not self._event.wait(timeout=timeout):
            raise WaitTimeoutError("Status has not completed yet.")
        if self._exception is not None:
>           raise self._exception
E           ophyd.utils.errors.UnknownStatusFailure: The status MoveStatus(done=False, pos=sim_slow, elapsed=0.1, success=False, settle_time=0.0) has failed. To obtain more specific, helpful errors in the future, update the Device to use set_exception(...) instead of _finished(success=False).
```

### LINKED ISSUE #527: Tweak is Janky Right now
## Expected Behavior

- Display should be intuitive, and should be something like
```
<position>, <step size>
```
At all times, even right at startup
- Progress bars are not required and not especially desirable.

## Current Behavior

- No readback at startup
- between moves we get [In progress. No progress bar available.] in the readback like:
```
In [8]: test_motor.tweak()
INFO     Started tweak of IMS(XCS:USR:MMS:33, name=test_motor)
 -2.700000 [In progress. No progress bar available.]
```
- These seem to be from accumulated changes from dependent libraries

## Possible Solution

- Clean it up

## Context

- Useful function for user ops is not being so useful

### LINKED ISSUE #553: Motor umv(r)
**Feature Request**
Move a motor and print the current position until move is completed.

**Existing Alternatives**
use old python.

**Context**
This can give a lot of comfort when you don't trust a stage and/or just want to see how long the move will take.

ZLLentz:
## Description

- umv displays position at the end of the move
- umv displays complete instead of not complete at the end of the move
- umv and tweak agree on an output format
- tweak doesn't tweak out, it has measured, consistent output
- tweak accepts + and - for increasing or decreasing the scaling
- tweak properly clears lines between prints
- tweak code refactored a bit for cleanliness

## Motivation and Context

closes #553 (umv)
The progress bar / position bar was not guaranteed an update on status completion. This guarantees such an update. Otherwise, you can get stuck with a progress bar that says "move in progress, no real information, blah blah blah"

closes #527 (tweak)
closes #281 (tweak)

Partially addresses https://github.com/pcdshub/Bug-Reports-and-Requests/issues/16 by adding the + and - keystrokes

also closes #481. I tracked down the race condition because it was extremely annoying.

## How Has This Been Tested?

Locally
```
In [4]: mot.umv(100)
mot (100.000) [Complete.]
```

In the pre-release docs section

### DIFF
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -1173,14 +1173,25 @@ def tweak_base(*args):
     print()
 
 
+loglist = []
+
+
 class AbsProgressBar(ProgressBar):
     """Progress bar that displays the absolute position as well."""
+    def __init__(self, status_objs, delay_draw=1.0):
+        self._last_position = None
+        super().__init__(status_objs, delay_draw=delay_draw)
+
     def update(self, *args, name=None, current=None, **kwargs):
-        if None not in (name, current):
-            super().update(*args, name='{} ({:.3f})'.format(name, current),
-                           current=current, **kwargs)
-        else:
-            super().update(*args, name=name, current=current, **kwargs)
+        try:
+            name = '{} ({:.3f})'.format(name, current)
+            self._last_position = current
+        except Exception:
+            name = name or 'motor'
+
+        current = current or self._last_position
+        loglist.append((name, current, kwargs))
+        super().update(*args, name=name, current=current, **kwargs)
 
 
 class LightpathMixin(OphydObject):
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -9,7 +9,7 @@ import signal
 import time
 from contextlib import contextmanager
 from pathlib import Path
-from threading import Event, Thread
+from threading import Event
 from types import MethodType, SimpleNamespace
 from weakref import WeakSet
 
@@ -20,7 +20,7 @@ from ophyd.ophydobj import Kind, OphydObject
 from ophyd.signal import AttributeSignal, Signal
 from ophyd.status import wait as status_wait
 
-from . import utils as util
+from . import utils
 from .signal import NotImplementedSignal
 
 try:
@@ -405,7 +405,10 @@ class MvInterface(BaseInterface):
     def _log_move(self, position):
         logger.info('Moving %s from %s to %s', self.name, self.wm(), position)
 
-    def mv(self, position, timeout=None, wait=False):
+    def _log_move_end(self):
+        logger.info('%s reached position %s', self.name, self.wm())
+
+    def mv(self, position, timeout=None, wait=False, log=True):
         """
         Absolute move to a position.
 
@@ -422,15 +425,21 @@ class MvInterface(BaseInterface):
         wait : bool, optional
             If `True`, wait for motion completion before returning.
             Defaults to :keyword:`False`.
+
+        log : bool, optional
+            If `True`, logs the move at INFO level.
         """
-        self._log_move(position)
+        if log:
+            self._log_move(position)
         self.move(position, timeout=timeout, wait=wait)
+        if wait and log:
+            self._log_move_end()
 
     def wm(self):
         """Get the mover's current positon (where motor)."""
         return self.position
 
-    def __call__(self, position=None, timeout=None, wait=False):
+    def __call__(self, position=None, timeout=None, wait=False, log=True):
         """
         Dispatches to :meth:`mv` or :meth:`wm` based on the arguments.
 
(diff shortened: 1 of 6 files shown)


pcdsdevices/interface.py
```diff
@@ -1175,12 +1177,36 @@ def movement(scale, direction):
...
+    def __init__(self, status_objs, **kwargs):
+        self._last_position = None
+        self._name = None
+        self._no_more = False
+        super().__init__(status_objs, **kwargs)
+
+        # Extra call when status is complete
+        for i, obj in enumerate(status_objs):
+            if not obj.done:
+                obj.add_callback(functools.partial(self._status_cb, i))
```
klauer, line 1189: What's the significance of passing in the index? Is this just what `super()`/`ProgressBar` is expecting?

ZLLentz: Yes, that's all it is.

ZLLentz: I think this part should be written slightly differently, I realized that there is a potential race condition here, and that the indices are not guaranteed to be in this order for multiples statuses... But the index does need to be passed in.

ZLLentz:
I did all of my testing with `sim.SlowMotor`, which led to resolving the race condition.
I would like to repeat with a "real" Beckhoff motor at the very least, and same for an IMS.

klauer:
Sounds good. Merge then test, or test then merge?

ZLLentz:
Slight rewrite, some testing, then merge

pcdsdevices/interface.py
klauer: OK, I see. 👍

ZLLentz:
I noticed that this completely breaks the `tweak` method, so I'm just going to expand scope... and fill out the umv/tweak/etc. laundry list

ZLLentz:
Hmm. I'd have to think a bit for how to handle the "needing a newline" vs "please don't make a newline" better...

I'm going see about using a real motor now to be thorough.

ZLLentz:
Works great with the real deal, going to merge and move on

Issue #550, opened 2020-08-27
To document/fix: calculations involving `EpicsMotor` `egu` property can be slow


klauer:
We use this at least in `DelayBase` for unit conversion.

Upstream should fix this at some point, but we should be aware of it:
https://github.com/bluesky/ophyd/blob/9a69acd5c016127855d953ab7148ec27db449f8f/ophyd/epics_motor.py#L114-L117

Without auto-monitor, this results in a `.get()`-call with a network hit each time:

```
In [1]: %timeit delay_stage.forward(0)
339 ms ± 7.54 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

ZLLentz:
Right, we should probably ensure that every property access is "fast", that is, if it's important enough to be a property it should be monitored/cached in some way.

klauer:
Largely resolved by #558. Closing for now.

Issue #526, opened 2020-08-22
Waiting on delay base extremely slow in a scan


ZLLentz:
## Expected Behavior

Waiting finishes at the right time

## Current Behavior

Waiting takes about 10 seconds
Set seems to wait

ZLLentz:
May be related to #550, and therefore might be simpler than expected

ZLLentz:
I've confirmed that this happens to the `SimDelayStage`, which should be completing instantaneously. I will investigate deeper.

ZLLentz:
I'm going to close this due to #558 and the fact that the scans work now

PR #558, opened 2020-08-28
MNT: establish ureg only once, it is a slow process

ZLLentz:
## Description

Create the `pint.UnitRegistry` at first use only instead of at every conversion call.

## Motivation and Context

The function call is slow, and only needs to be done once.
Contributes to #526, might be related to #550 

Here is the line_profiler output for this method from a software-only motor move:
```
Total time: 6.7158 s
File: /u/lu/zlentz/pydev/pcdsdevices/utils.py
Function: convert_unit at line 91

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    91                                           def convert_unit(value, unit, new_unit):
    92                                               """
    93                                               One-line unit conversion.
    94
    95                                               Parameters
    96                                               ----------
    97                                               value : float
    98                                                   The starting value for the conversion.
    99
   100                                               unit : str
   101                                                   The starting unit for the conversion.
   102
   103                                               new_unit : str
   104                                                   The desired unit for the conversion.
   105
   106                                               Returns
   107                                               -------
   108                                               new_value : float
   109                                                   The starting value, but converted to the new unit.
   110                                               """
   111
   112         8       6705.0    838.1     99.8      ureg = pint.UnitRegistry()
   113         8          4.1      0.5      0.1      expr = ureg.parse_expression(unit)
   114         8          6.7      0.8      0.1      return (value * expr).to(new_unit).magnitude
```
## How Has This Been Tested?

Locally only with simulated hardware- all slowdowns after this are due to the hardware classes and should be investigated too.

Pre-release notes added

### DIFF
--- a/docs/source/upcoming_release_notes/558-slow-ureg.rst
+++ b/docs/source/upcoming_release_notes/558-slow-ureg.rst
@@ -1,0 +1,31 @@
+558 Slow Ureg
+#############
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- Fix issue where converting units could incur time penalties of up to
+  7 seconds. This should take around 10ms now.
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- zlentz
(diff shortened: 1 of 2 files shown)


ZLLentz:
One more thought: this does add a 0.8s slowdown to the module import- I probably should've run this at first conversion rather than at import.

ZLLentz:
Switched to doing it at first use rather than at first import- I think this is the expected way to implement that, but it's been a while. Waiting on Mr. Travis.

PR #563, opened 2020-09-03
ENH: allow waiting on in-progress moves without caching status object

ZLLentz:
## Description

After a call to any move command, you will be able to call `.wait()` later as a proxy to `status.wait()` for the motor's most recently generated `MoveStatus` object, bypassing the need to keep track of this object yourself.

## Motivation and Context

The weird interactive move shortcuts that were requested of us do not return status objects, so if you don't pass `wait=True` you miss your chance forever. This is used in the "move the whole beamline between pink and mono" script I was asked to write, which heavily leverages the presets system, which doesn't return status objects.

An alternate approach here may have been to add extra presets methods that do return status objects... but this seemed easier to figure out.

## How Has This Been Tested?

Locally

## How Has This Been Documented
You got me there

### DIFF
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -410,8 +410,8 @@ class MvInterface(BaseInterface):
     def _log_move_end(self):
         logger.info('%s reached position %s', self.name, self.wm())
 
-    def move(self, position, timeout=None, **kwargs):
-        st = super().move(position, timeout=timeout, **kwargs)
+    def move(self, *args, **kwargs):
+        st = super().move(*args, **kwargs)
         self._last_status = st
         return st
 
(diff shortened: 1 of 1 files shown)


ZryletTC (approved):
I like it 😃

ZryletTC:
I guess that ccm error is due to a combination of #538 and this. We want pseudopositioners to inherit their move methods from pseudopositioner but we want to inherit from the interface first for the tab_complete stuff. Can we do something weird like `class CCMCalc(BaseInterface, PseudoPositioner, FltMvInterface)`?

pcdsdevices/interface.py
```diff
@@ -399,15 +399,25 @@ class MvInterface(BaseInterface):
...
+        self._last_status.set_finished()
+        super().__init__(*args, **kwargs)
 
     def _log_move(self, position):
         logger.info('Moving %s from %s to %s', self.name, self.wm(), position)
 
     def _log_move_end(self):
         logger.info('%s reached position %s', self.name, self.wm())
 
+    def move(self, position, timeout=None, **kwargs):
```
klauer:
```suggestion
    def move(self, *args, **kwargs):
```

Test is breaking due to naming arguments here.

klauer (approved):
Needs the minor fix, but looks good to me.

PR #565, opened 2020-09-03
BUG: Fix pulsepicker so it gets its PV like it says it does

ZryletTC:
## Description
The `PulsePickerInOut` class is meant to grab the first two segments of its PV to use for its in-out motion. At it is currently written, it only grabs the first segment of the PV instead of the first two.

## Motivation and Context
Wanted to implement a `PulsePickerInOut`. It didn't work.

## How Has This Been Tested?
```
In [40]: pp.prefix.split(':')[:1]
Out[40]: ['XCS']

In [41]: pp.prefix.split(':')[:2]
Out[41]: ['XCS', 'SB2']
```

🤷‍♂️

### DIFF
--- a/docs/source/upcoming_release_notes/565-pp-inout-pvfix.rst
+++ b/docs/source/upcoming_release_notes/565-pp-inout-pvfix.rst
@@ -1,0 +1,31 @@
+565 PP InOut PVFix
+##################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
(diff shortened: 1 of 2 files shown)


ZLLentz (approved):
oops

ZLLentz (changes requested):
Should add a pre-release docs page before merging

ZLLentz (approved):

PR #566, opened 2020-09-03
MNT: fix done running race condition in EventSequencer class

ZLLentz:
## Description

Add a check to make sure that the "done" transitions happen at a timestamp after the last put to `play_control`.

## Motivation and Context

In functions like `trigger`, we return a `SubscriptionStatus` that is to be marked as finished when we transition from state "playing" (2) to state "not playing" (0).

If the python code is running particularly fast (or the network/IOC particularly slow), it can get to the point where it sets up the `SubscriptionStatus` with local function `done`, `run=True` prior to getting an update that the sequencer has started playing (transition from 0 to 2). What happens in this case is that we get a callback with the details from the previous transition, (typically from 2 to 0), and since `run=True` we mark ourselves as `finished` immediately- after all, we just got an update that the sequencer stopped playing!

`run=True` is mandatory in case the function runs extremely slowly, so that we don't completely miss the done transition and sit there forever.

We always start the sequencer by putting 1 to the `play_control` signal, which updates the `timestamp` metadata field of that signal. Since all the timestamps are created in the python process, and since the sequencer cannot start until we ask it to, a "good" done transition must have a `timestamp` later than our last `play_control` start.

## How Has This Been Tested?

only locally

Added release notes

### DIFF
--- a/docs/source/upcoming_release_notes/566-seq-race.rst
+++ b/docs/source/upcoming_release_notes/566-seq-race.rst
@@ -1,0 +1,31 @@
+566 Seq Race
+############
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- Fixed a race condition in the EventSequencer device's status objects. Waiting
+  on these statuses will now be more reliable.
+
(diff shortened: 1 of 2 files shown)


klauer (approved):
Nice find and fix.

ZLLentz:
This seemed to work, so I'll merge it

PR #551, opened 2020-08-27
ENH: lxt_ttc, txt, tt_delay / LaserTimingCompensation

klauer:
Closes #535

Notes
-------
* `lxt_ttc`: time stage series compensation using stages `lxt` and `txt` (appears to be like a `SyncAxes`-style thing)
* `lxt` is already implemented as the `LaserTiming` class, a virtual motor based on vitara positions
* `txt` is a virtual motor based on `tt_delay` (`xppTTdelay`/`xcsTTdelay`), which uses delay stage to time (perhaps could use `DelayBase`) to convert the `tt_delay` stage position
* `tt_delay` is a motor record type `xps8p` (``pcdsdevices`` ``Newport``)

XCS
---

```python
lxt_ttc = pcdsdevices.lxe.LaserTimingCompensation(
    '',
    delay_prefix='XCS:LAS:MMN:01',
    laser_prefix='LAS:FS4',
    name='lxt_ttc',
)
```

XPP
---

```python
lxt_ttc = pcdsdevices.lxe.LaserTimingCompensation(
    '',
    delay_prefix='XPP:LAS:MMN:16',
    laser_prefix='LAS:FS3',
    name='lxt_ttc',
)
```

Additional notes:
* `lxt` - `LaserTiming` / Vitara
* `ttc` - stage that moves the time tool signal to keep
* `lxt_ttc` moves `lxt` and also `ttc` to compensate so that we can get the
  true laser x-ray delay by using the `lxt`-value and the result of the
  analysis of the time tool data and we don't count anything double. (Silke)
* `ttc` has a smaller range, so `lxt_ttc` will move both when it can. When
  `ttc` is out of range, it will only move `lxt` until we are back in the
  range where `ttc` can move. (Silke)

Conversion check
----------------

Note that the original ``xcsTTdelay`` conversion utilities don't take into
account the number of bounces and require external multiplication/division by
that.

TODO
-------
- [x] `lxt`
- [x] `tt_delay` (class `Newport`)
- [x] `txt` (`DelayNewport`, depends on `tt_delay`, relies on motor.EGU being set appropriately)
- [x] `lxt_ttc` (`SyncAxes`, depends on `lxt`, `txt`)
- [x] Capture additional notes from Silke + Zach in docstrings
- [x] `txt` unit conversion manual check
- [x] `txt` unit conversion check in testing
- [x] Invert `lxt` direction
- [x] Check if this does what it's supposed to
- [x] Add release notes
- [x] Tab whitelist
[...]

PR #575, opened 2020-09-16
535 lxt ttc alt

klauer:
Alternate to #551, for your consideration.

Limit handling is probably wrong here.

### DIFF
--- a/docs/source/upcoming_release_notes/535-lxt-ttc.rst
+++ b/docs/source/upcoming_release_notes/535-lxt-ttc.rst
@@ -1,0 +1,37 @@
+535 lxt_ttc
+###########
+
+API Changes
+-----------
+- N/A
+
+Features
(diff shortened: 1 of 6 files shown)


ZLLentz:
This one forgoes the combined device and does the synchronous move in a separate function call?

klauer:
Yes, at the end of `pcdsdevices/lxe.py` (`lxt_ttc`).

ZLLentz:
I think both formulations are good. It's hard for me to approve either of them when the other exists though, since I don't understand the beamline case for doing it one way vs the other.

klauer:
I readily admit I do not understand the requirements - so this is a something along the lines of: try multiple options, throw them at (over?) the wall and see what sticks.

klauer:
Seems like #551 will stick

PR #579, opened 2020-09-16
REF: tab completion state in a separate class

### LINKED ISSUE #576: Preset tab completion in Engineering Mode
Currently, the motor presets are not added to the tab whitelist, making them pretty useless. We need to make the interface class tab completion mechanisms preset-aware so these methods can be easily found.

klauer:
An alternative to holding tab completion state in `BaseInterface`, pulling it out to allow for class or instance-level modification.

If this is deemed acceptable, we can build functionality on top of it for https://github.com/pcdshub/pcdsdevices/issues/576

If not, we can patch in that requested functionality in each subclass.

8d64558 closes #576

### DIFF
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -43,6 +43,10 @@ Positioner_whitelist = ["settle_time", "timeout", "egu", "limits", "move",
 
 
 class _TabCompletionHelper:
+    """
+    Base class for `TabCompletionHelperClass`, `TabCompletionHelperInstance`.
+    """
+
     _includes: typing.Set[str]
     _regex: typing.Optional[typing.Pattern]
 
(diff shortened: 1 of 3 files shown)


pcdsdevices/interface.py
```diff
@@ -1020,6 +1119,8 @@ def _remove_methods(self):
                 delattr(obj, method_name)
             except AttributeError:
                 pass
+            if hasattr(obj, '_tab'):
+                obj._tab.remove(method_name)
```
ZLLentz, line 1123: Rather than managing these individually, we probably could have done blanket regexes e.g. `wm_.*`

ZLLentz (approved):
This makes it easier to manage the tabs and I like it

klauer:
Cleaned, added tests, release notes. Moving forward based on previous approval.

PR #584, opened 2020-09-17
DOC: release note extender

ZLLentz:
An uncoordinated simultaneous effort with a similar goal as #583

### DIFF
--- a/docs/release_notes.py
+++ b/docs/release_notes.py
@@ -1,0 +1,93 @@
+import sys
+import time
+from collections import defaultdict
+from pathlib import Path
+
+# find the pre-release directory and release notes file
+THIS_DIR = Path(__file__).resolve().parent
+PRE_RELEASE = THIS_DIR / 'source' / 'upcoming_release_notes'
+TEMPLATE = PRE_RELEASE / 'template-short.rst'
+RELEASE_NOTES = THIS_DIR / 'source' / 'releases.rst'
+
+# Set up underline constants
+TITLE_UNDER = '#'
+RELEASE_UNDER = '='
+SECTION_UNDER = '-'
+
+def parse_pre_release_file(path):
+    """
+    Return dict mapping of release notes section to lines.
+
+    Uses empty list when no info was entered for the section.
+    """
+    print(f'Checking {path} for release notes.')
+    with path.open('r') as fd:
+        lines = fd.readlines()
+
+    section_dict = defaultdict(list)
+    prev = None
+    section = None
+
+    for line in lines:
+        if prev is not None:
+            if line.startswith(SECTION_UNDER * 2):
+                section = prev.strip()
+                continue
+            if section is not None and line[0] in ' -':
+                notes = section_dict[section]
+                if len(line) > 6:
+                    notes.append(line)
+                section_dict[section] = notes
+        prev = line
+
+    return section_dict
+
+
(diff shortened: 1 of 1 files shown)


klauer:
Looks pretty clean to me. I like the goal of it being more of a real command-line tool instead of a helper script.

I have no issue whatsoever with closing my PR. Still it would be interesting to compare the output as a logic check for both of ours :+1:

ZLLentz:
To compare with the other PR:
```
Writing out release notes file


v2.0.0 (2020-09-17)
===================

API Changes
-----------
  been changed.
- :class:`~pcdsdevices.interface.Presets` ``add_hutch`` (and all similar
  ``add_*``) methods no longer require a position.  When unspecified, the
  current position is used.

Features
--------
- Moves using mv and umv will log their moves at info level for interactive
  use to keep track of the sessions.
  support.
- Epics motors can now have local limits updated per-session, rather than
  only having the option of the EPICS limits. Setting limits attributes will
  update the python limits, putting to the limits PVs will update the limits
  PVs.
- Add PVPositionerDone, a setpoint-only PVPositioner class that is done moving
  immediately. This is not much more useful than just using a PV, but it is
  compatibile with pseudopositioners and has a built-in filter for ignoring
  small moves.

Device Updates
--------------
- CCM energy limited to the range of 4 to 25 keV
- CCM theta2fine done moving tolerance raised to 0.01
- Beam request default move start tolerance dropped to 5eV

Bugfixes
--------
- Fix bug on beam request where you could not override the tolerance
  via init kwarg, despite docstring's indication.
  added.
  ``LiveTable`` support.
- Fix issue where converting units could incur time penalties of up to
  7 seconds. This should take around 10ms now.
- Fix issue where putting to the limits property would update live PVs,
  contrary to the behavior of all other limits attributes in ophyd.
- Fix issue where doing a getattr on the limits properties would fetch
  live PVs, which can cause slowdowns and instabilities.
- Rework BeamEnergyPositioner to be setpoint-only to work properly
  with the behavior of the energy PVs.
- FltMvPositioner.wm will now return numeric values if the position
  value is a tuple. This value is the first element of the tuple, which
  for pseudo positioners is a value that can be passed into move and have
  it do the right thing. This resolves consistency issues and fixes bugs
  where mvr and umvr would fail.
- Tweak will feel less "janky" now and give useful feedback.
- Tweak now accepts + and - as valid inputs for changing the step size.
- Tweak properly clears lines between prints.
- umv will now properly display position and completion status after a move.
- Fix bug in PulsePickerInOut where it would grab only the first section of
  of the PV instead of the first two
- Fixed a race condition in the EventSequencer device's status objects. Waiting
  on these statuses will now be more reliable.

Maintenance
-----------
- Establish DOC conventions for accumulating release notes from every
  pull request.
- Use more of the built-in ophyd mechanisms for limits rather than
  relying on local overrides.
- Tweak refactored for maintainability.

Contributors
------------
- zrylettc
- zlentz
- klauer
- zllentz
- pennebak

```

ZLLentz:
A quick scan suggests the following diffs:
- I added a header
- You sorted the contributors

klauer:
Looks good! Let's go with it.

Randomly sorting contributors may be more fair than alphabetical order here for you guys with `z*` names...

ZLLentz:
I'm not sure if it's random or first-come-first-serve...
I think we should sort to make it easier to pick out typos

PR #585, opened 2020-09-18
HACK: fixed-length reads for the sequencer

klauer:


klauer:
Reads back:
```
In [1]: seq.sequence.read()
Out[1]:
OrderedDict([('seq_sequence_ec_array',
              {'value': array([170,   0,   0, ...,   0,   0,   0], dtype=int32),
               'timestamp': 1600432914.390497}),
             ('seq_sequence_bd_array',
              {'value': array([4, 0, 0, ..., 0, 0, 0], dtype=int32),
               'timestamp': 1600433310.489143}),
             ('seq_sequence_fd_array',
              {'value': array([0, 0, 0, ..., 0, 0, 0], dtype=int32),
               'timestamp': 631152000.0}),
             ('seq_sequence_bc_array',
              {'value': array([0, 0, 0, ..., 0, 0, 0], dtype=int32),
               'timestamp': 631152000.0})])
```

ZLLentz:
We should probably try this patch in MEC

klauer:
I haven't tried writes (rather: can't), but I don't _think_ there'll be issues there...

Is Python really the only client that matters here? We're fortunate if so...

slactjohnson:
@klauer Yeah, I had planned on doing something like this if we were unable to fix the underlying issue by the beamtime. It's still not clear to me that our writes are properly reaching the sequencer though. Silke was going to take a look from mcc-login; not sure of her results.

klauer:
Gateway issue resolved

PR #592, opened 2020-09-25
target_in for IPM

### LINKED ISSUE #544: IPM's should have a `target_in` function that takes an argument to insert the specified target
Copied from card on run 18 board

IPM Motion devices can be moved to specific states using the `move` function. I remember state 0 being unknown, states 1-4 being targets 1-4, and 5 being out

cristinasewell:
## Description

A `target_in` function has been added to the IPMMotion Class

## Motivation and Context

IPM Motion devices can be driven to specific states using the move function - `target_in` is taking as a parameter a "target state" and it is moving the target to that state. 
Trying to close #544

## How Has This Been Tested?

Has only been tested using the `fake_imp` device

N/A

### DIFF
--- a/pcdsdevices/ipm.py
+++ b/pcdsdevices/ipm.py
@@ -5,7 +5,6 @@ from ophyd.device import Component as Cpt
 from ophyd.device import Device
 from ophyd.device import FormattedComponent as FCpt
 from ophyd.signal import EpicsSignal, EpicsSignalRO
-from ophyd.status import wait as status_wait
 
 from .doc_stubs import IPM_base, basic_positioner_init, insert_remove
 from .epics_motor import IMS
@@ -140,7 +139,7 @@ class IPMMotion(BaseInterface, Device):
     def target_in(self, target_num, moved_cb=None, timeout=None, wait=False):
         """
         Moves the target to one of the target positions. There are 4 targets
-        with different thinkness and absorbtion/signal.
+        with different thickness and absorption/signal.
         The targets move vertically. To drive them in, use presets:
         ipm.target_in(x), where x = target number
 
@@ -157,7 +156,7 @@ class IPMMotion(BaseInterface, Device):
 
         timeout : float, optional
             Maximum time for the motion. If `None` is given, the default value
-            of this positioners is used.
+            of this positioner is used.
 
         wait : bool
             If `True`, block until move is completed.
@@ -166,21 +165,8 @@ class IPMMotion(BaseInterface, Device):
         -------
         status: MoveStatus
         """
-        status = self.target.move(target_num, moved_cb=moved_cb,
-                                  timeout=timeout, wait=wait)
-
-        # Add our callback if one was given
-        if moved_cb is not None:
-            status.add_callback(moved_cb)
-
-        # Wait if instructed to do so. Stop the motors if interrupted
-        if wait:
-            try:
-                status_wait(status)
-            except KeyboardInterrupt:
-                self.target.stop()
-                raise
-        return status
+        return self.target.move(target_num, moved_cb=moved_cb,
+                                timeout=timeout, wait=wait)
 
     @property
     def transmission(self):
--- a/pcdsdevices/ipm.py
+++ b/pcdsdevices/ipm.py
@@ -5,6 +5,7 @@ from ophyd.device import Component as Cpt
 from ophyd.device import Device
 from ophyd.device import FormattedComponent as FCpt
 from ophyd.signal import EpicsSignal, EpicsSignalRO
+from ophyd.status import wait as status_wait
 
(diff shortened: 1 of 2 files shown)


cristinasewell:
@ZLLentz I might be very off here, but I won't know until I get some feedback :) 

For the `States` - should I specifically handle the cases where an invalid state was provided? Like, should I `set_exception()` or something like that?

pcdsdevices/ipm.py
```diff
@@ -136,6 +137,51 @@ def remove(self, moved_cb=None, timeout=None, wait=False):
...
+        Returns
+        -------
+        status: MoveStatus
+        """
+        status = self.target.move(target_num, moved_cb=moved_cb,
+                                  timeout=timeout, wait=wait)
+
+        # Add our callback if one was given
+        if moved_cb is not None:
+            status.add_callback(moved_cb)
```
ZLLentz, line 174: This and some of the other things here may be redundant with `target.move`'s handling of the keyword arguments

pcdsdevices/ipm.py
```diff
@@ -136,6 +137,51 @@ def remove(self, moved_cb=None, timeout=None, wait=False):
...
+            Maximum time for the motion. If `None` is given, the default value
+            of this positioners is used.
+
+        wait : bool
+            If `True`, block until move is completed.
+
+        Returns
+        -------
+        status: MoveStatus
+        """
```
ZLLentz, line 168: Very good to have a thorough docstring

pcdsdevices/ipm.py
```diff
@@ -136,6 +137,51 @@ def remove(self, moved_cb=None, timeout=None, wait=False):
             return (rmstatus & self.diode.remove(moved_cb=moved_cb,
                                                  timeout=timeout, wait=wait))
 
+    def target_in(self, target_num, moved_cb=None, timeout=None, wait=False):
+        """
+        Moves the target to one of the target positions. There are 4 targets
+        with different thinkness and absorbtion/signal.
```
ZLLentz, line 143:
```suggestion
        with different thickness and absorption/signal.
```

pcdsdevices/ipm.py
```diff
@@ -5,6 +5,7 @@
 from ophyd.device import Device
 from ophyd.device import FormattedComponent as FCpt
 from ophyd.signal import EpicsSignal, EpicsSignalRO
+from ophyd.status import wait as status_wait
```
ZLLentz, line 8: This is actually deprecated in favor of `Status.wait`

tests/test_ipm.py
```diff
@@ -116,6 +116,21 @@ def test_ipm_motion(fake_ipm):
     # Remove IPM Diode
     ipm.diode.remove()
     assert ipm.diode.state.state.get() == 2
+    # Target_in
+    status = ipm.target_in(4)
+    assert ipm.target.state.get() == 4
+    status.wait(timeout=1)
```
ZLLentz, line 122: technically we should wait before checking the result, not after. This could matter if the fake device implementation changes later

```suggestion
    status.wait(timeout=1)
    assert ipm.target.state.get() == 4
```

ZLLentz:
I think this is exactly what the request was asking for.

> For the States - should I specifically handle the cases where an invalid state was provided? Like, should I set_exception() or something like that?

`Status.set_exception` is a handler for when the status fails in some way. I think it is a great thing to use in general, I'm not sure how to implement that here for invalid state selections since the `move` command should raise an exception and the function will never return a `Status` in the first place. I don't remember how clear or unclear the parent class's `ValueError` for this situation here.

pcdsdevices/ipm.py
cristinasewell: Ohh ok, I should not need to handle the `moved_cb` and `wait` here...

pcdsdevices/ipm.py
cristinasewell: ok good to know, thank you

tests/test_ipm.py
cristinasewell: sounds good, thank you - fixed this in 371615e

pcdsdevices/ipm.py
cristinasewell: thanks for the corrections, I need to get better at catching my misspelled comments - cause I do this all the time :( 
do you know of any plugins that I can install that checks for grammar :D ?

ZLLentz: It's hard to do this generically for code... I don't have a spellcheck plugin and I also make these kinds of typos all the time, especially in docstrings where I don't read it very closely again after writing it once.

ZLLentz (approved):
I say we merge this and move on to other things. Looks great.

ZLLentz:
If this doesn't work on the real device, the class has deeper issues than the features introduced in the PR.

PR #596, opened 2020-10-01
MNT: revert to old ccm energy calculation

### LINKED ISSUE #570: Revert to old CCM Energy calculation
## Expected Behavior

- Use exactly the same calculation as the old Python for `alio_to_theta` and `theta_to_alio`
- Switch the alio calculation test to expected fail

## Current Behavior

- Tries to be clever (a mistake, always)

## Possible Solution

Sample code from old python:
```
def thetaToAlio(theta):
  """ Function that converts theta angle (deg) to alio position (mm) """
  t_rad = (theta-gTheta0)*n.pi/180.
  x = gR*(1/n.cos(t_rad)-1)+gD*n.tan(t_rad)
  return x
def alioToTheta(alio):
  """ Function that converts alio position (mm) to theta angle (deg) """
  return gTheta0 + 180/n.pi*2*n.arctan( (n.sqrt(alio**2+gD**2+2*gR*alio)-gD) / (2*gR+alio) )
```
Note that the new python changes some of the symbols to make it more clear what is going on, but the calculations should be able to be the same.

## Context

- CCM calibrated with old calculations
- Requested by Matthieu

cristinasewell:
## Description

The def `theta_to_alio` and def `alio_to_theta` functions have been reverted to use the `old` calculations.
## Motivation and Context

Trying to close #570
## How Has This Been Tested?

Has only been tested with a `fake_ccm`.

### DIFF
--- a/pcdsdevices/ccm.py
+++ b/pcdsdevices/ccm.py
@@ -282,16 +282,9 @@ def alio_to_theta(alio, theta0, gr, gd):
     """
     Converts alio position (mm) to theta angle (rad).
 
-    This is an empirical inversion via binary search. If you decide to spend
-    time trying to find the analytic solution here, please update this
-    docstring, either to indicate your success or to increment the hours
-    counter below.
-
     Conversion function
     theta_angle = f(x) = 2arctan * [(sqrt(x^2 + D^2 + 2Rx) - D)/(2R + x)]
     Note that for x = −R, θ = 2 arctan(−R/D)
-
-    total hours spent here: 2
     """
     return theta0 + 2 * np.arctan(
          (np.sqrt(alio ** 2 + gd ** 2 + 2 * gr * alio) - gd) / (2 * gr + alio)
--- a/pcdsdevices/ccm.py
+++ b/pcdsdevices/ccm.py
@@ -15,12 +15,11 @@ from .pv_positioner import PVPositionerIsClose
 logger = logging.getLogger(__name__)
 
 # Constants
-
 si_111_dspacing = 3.1356011499587773
 si_511_dspacing = 1.0452003833195924
 
 # Defaults
-default_theta0 = 14.9792
+default_theta0 = 14.9792 * np.pi/180
 default_dspacing = si_111_dspacing
 default_gr = 3.175
 default_gd = 231.303
@@ -266,14 +265,16 @@ def theta_to_alio(theta, theta0, gr, gd):
     Delta_Theta:   the effective scattering angle (adjusted with Alio stage)
     R = 0.003175m: radius of the sapphire ball connected to the Alio stage
     D = 0.232156m: distance between the Theta_B rotation axis and the center
-                   of the saphire sphere located on the Alio stage
+                   of the saphire sphere located on the Alio stage.
+                   note: The current value that we're using for D is 0.231303 -
+                   possibly measured by metrology
 
     Theta_B = Theta_0 + Delta_Theta
     Conversion formula:
     x = f(Delta_Theta) = D * tan(Delta_Theta)+(R/cos(Delta_Theda))-R
     Note that for ∆θ = 0, x = R
     """
-    t_rad = (theta - theta0) * np.pi / 180.0
+    t_rad = theta - theta0
     return gr * (1 / np.cos(t_rad) - 1) + gd * np.tan(t_rad)
 
 
@@ -292,19 +293,19 @@ def alio_to_theta(alio, theta0, gr, gd):
 
     total hours spent here: 2
     """
-    return theta0 + 180 / np.pi * 2 * np.arctan(
-        (np.sqrt(alio ** 2 + gd ** 2 + 2 * gr * alio) - gd) / (2 * gr + alio)
-    )
+    return theta0 + 2 * np.arctan(
+         (np.sqrt(alio ** 2 + gd ** 2 + 2 * gr * alio) - gd) / (2 * gr + alio)
+     )
 
 
 def wavelength_to_theta(wavelength, dspacing):
     """Converts wavelength (A) to theta angle (rad)."""
-    return 180.0 / np.pi * np.arcsin(wavelength / 2 / dspacing)
+    return np.arcsin(wavelength/2/dspacing)
 
 
 def theta_to_wavelength(theta, dspacing):
     """Converts theta angle (rad) to wavelength (A)."""
-    return 2 * dspacing * np.sin(theta / 180 * np.pi)
+    return 2*dspacing*np.sin(theta)
 
 
 def energy_to_wavelength(energy):
--- a/pcdsdevices/ccm.py
+++ b/pcdsdevices/ccm.py
@@ -1,5 +1,4 @@
 import logging
-import time
 
 import numpy as np
 from ophyd.device import Component as Cpt
@@ -16,11 +15,12 @@ from .pv_positioner import PVPositionerIsClose
 logger = logging.getLogger(__name__)
 
 # Constants
+
 si_111_dspacing = 3.1356011499587773
 si_511_dspacing = 1.0452003833195924
 
 # Defaults
-default_theta0 = 14.9792 * np.pi/180
+default_theta0 = 14.9792
 default_dspacing = si_111_dspacing
 default_gr = 3.175
 default_gd = 231.303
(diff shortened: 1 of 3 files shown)


pcdsdevices/ccm.py
```diff
@@ -270,33 +286,25 @@ def alio_to_theta(alio, theta0, gr, gd):
...
-    return theta_guess
+    return theta0 + 180 / np.pi * 2 * np.arctan(
+        (np.sqrt(alio ** 2 + gd ** 2 + 2 * gr * alio) - gd) / (2 * gr + alio)
+    )
 
 
 def wavelength_to_theta(wavelength, dspacing):
     """Converts wavelength (A) to theta angle (rad)."""
-    return np.arcsin(wavelength/2/dspacing)
+    return 180.0 / np.pi * np.arcsin(wavelength / 2 / dspacing)
```
cristinasewell, line 302: OK this is a weird one here..... when I implemented the old function back for `alio_to_theta` and `theta_to_alio` - these two functions: `wavelength_to_theta` and `theta_to_wavelenght` were failing the tests.... usually we would probably change the tests, but for curiosity I changed these functions to use the `old` implementations, mainly added the conversion from radians to degrees back and it seems to work fine now... 
Simply putting back the `default_theta0 = 14.9792 * np.pi/180` was not fixing the problem - it has to do with the calculation of `theta` as well....
Can anybody please give me a bit of insight here ?? 
Was this part of the new requirements?

pcdsdevices/ccm.py
```diff
@@ -270,33 +286,25 @@ def alio_to_theta(alio, theta0, gr, gd):
...
 
 
 def theta_to_wavelength(theta, dspacing):
     """Converts theta angle (rad) to wavelength (A)."""
-    return 2*dspacing*np.sin(theta)
+    return 2 * dspacing * np.sin(theta / 180 * np.pi)
```
cristinasewell, line 307: Same as comment above...

pcdsdevices/ccm.py
```diff
@@ -257,8 +257,24 @@ def _state(self, value):
...
-    return gr * (1/np.cos(theta)-1) + gd * np.tan(theta - theta0)
+    """
+    Converts theta angle (rad) to alio position (mm).
+
+    Theta_B:       scattering angle, the angle reduces when rotating clockwise
+                   (Bragg angle)
+    Theta_0:       scattering angle offset of the Si (111) first crystal
+    Delta_Theta:   the effective scattering angle (adjusted with Alio stage)
+    R = 0.003175m: radius of the sapphire ball connected to the Alio stage
+    D = 0.232156m: distance between the Theta_B rotation axis and the center
```
cristinasewell, line 268: Based on the paper https://docs.google.com/viewer?a=v&pid=sites&srcid=ZGVmYXVsdGRvbWFpbnx4cHB3aWtpfGd4OjQ0ZDA5ZGFkYTM0MjFhZDk the `D` value is slightly different - I did not change the number, but I should probably change the docstring here to reflect the actual number we are using - out of curiosity though - did this change because we have new hardware/devices?  
I got the paper from (https://sites.google.com/site/xppwiki/beamline-components/ccm)

pcdsdevices/ccm.py
```diff
@@ -257,8 +257,24 @@ def _state(self, value):
 
 # Calculations between alio position and energy, with all intermediates.
 def theta_to_alio(theta, theta0, gr, gd):
```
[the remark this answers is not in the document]
klauer: I think one comment can answer most of your questions here: the current API should remain the same (including units - if a function takes degrees now, it should take degrees after this PR), but the internal algorithm should be reverted to what's shown in the issue (with fixes for unit conversion, if necessary).

pcdsdevices/ccm.py
klauer: It may be worth reaching out to the requester to see why the constant is different, and if it's for a good reason, documenting it here.

pcdsdevices/ccm.py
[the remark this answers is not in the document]
cristinasewell: Ok got it, i'll change it back to reflect what we had then - thank you

pcdsdevices/ccm.py
ZLLentz: I wrote these tests to verify that the calculations were reversible. The old calculations are not reversible (which is why I updated them). Therefore, the tests fail.

ZLLentz: I suggest just putting an xfail on this test

cristinasewell: It is not really that they're failing the assertion, it is more like failing the calculation: 
` ValueError: Position is outside of pseudo single limits: fake_ccm_calc_energy, 4 < -5.02697921205642 < 25`
```
FAILED tests/test_ccm.py::test_ccm_calc - ValueError: Position is outside of pseudo single limits: fake_ccm_calc_energy, 4 < -5.02697921205642 < 25
FAILED tests/test_ccm.py::test_vernier - ValueError: Position is outside of pseudo single limits: fake_ccm_calc_energy, 4 < -5.02697921205642 < 25
```

I need to look into it more...

pcdsdevices/ccm.py
```diff
@@ -257,8 +256,26 @@ def _state(self, value):
...
+                   of the saphire sphere located on the Alio stage.
+                   note: The current value that we're using for D is 0.231303 -
+                   possibly measured by metrology
+
+    Theta_B = Theta_0 + Delta_Theta
+    Conversion formula:
+    x = f(Delta_Theta) = D * tan(Delta_Theta)+(R/cos(Delta_Theda))-R
+    Note that for ∆θ = 0, x = R
+    """
+    t_rad = theta - theta0
```
cristinasewell, line 277: assuming here that `theta` and `theta0` are radians

pcdsdevices/ccm.py
```diff
@@ -270,23 +287,15 @@ def alio_to_theta(alio, theta0, gr, gd):
...
-        theta_guess = (low+high)/2
-        alio_calc = theta_to_alio(theta_guess, theta0, gr, gd)
-        if np.isclose(alio, alio_calc):
-            break
-        elif alio_calc > alio:
-            high = theta_guess
-        else:
-            low = theta_guess
-    return theta_guess
+    return theta0 + 2 * np.arctan(
```
cristinasewell, line 296: took out the  `180 / n.pi` from the old function, because we are now returning radians instead of degrees.

old_func:
```
  return gTheta0 + 180 / n.pi * 2 * n.arctan(
        (n.sqrt(alio ** 2 + gD ** 2 + 2 * gR * alio) - gD) / (2 * gR + alio)
    )
```

pcdsdevices/ccm.py
cristinasewell: Ok I had some conversion issues - that's why I was getting these (thanks @klauer for pointing that our for me on slack)

But.... wasn't it supposed to fail??? Am I still doing something wrong?? the test seems to work fine ... maybe i should wait until passes the checks though

klauer: It may just be that there's one value that is reversible here? Not sure, but it seems a good thing anyway.

klauer: On the note of testing, it might not be a bad idea to use the old code and pick a few sets of calculation results that we can keep as regression test cases.

For example, if we were testing `cos(x)` we could say `test_cases = [(0, 1), (np.pi / 2, 0), ...]``

klauer (approved):
Calculations look correct to me, compared to what you pasted. 👍

pcdsdevices/ccm.py
cristinasewell: @klauer that would be great - where in the old code I can I get a few sets of calculation results from? - I am looking the `old-hutch-configs` repo, but can't find these - I might be looking for the wrong things though

ZLLentz (changes requested):
A few more things before merging:
1. Can you add an entry to the pre-release docs by calling the `docs/pre-release-notes.sh` script? These are used to compile the release notes.
2. Can you remove my snarky comments from 3 years ago from the docstring of `alio_to_theta`?
3. I'm going to check that the calculations behave as expected in a python session

ZLLentz:
Spot check results, querying for the ccm energy (e.g. the energy that the crystal permits through the monochromator):
```
Before this PR: 8.644099048074112 
After this PR:  8.630370238006924 
Old Old Python: 8.6303702380043426 
```
I think we're within a floating point error now and should be good with this fix

ZLLentz (approved):
LGTM Let's merge it

cristinasewell:
> A few more things before merging:
> 
> 1. Can you add an entry to the pre-release docs by calling the `docs/pre-release-notes.sh` script? These are used to compile the release notes.
>> did that in c94d9fd - please let me know if I need to add more info there
> 2. Can you remove my snarky comments from 3 years ago from the docstring of `alio_to_theta`?
>> did that in def49ad 
> 3. I'm going to check that the calculations behave as expected in a python session

ZLLentz:
Just waiting for one more turn of the travis wheel before merging

PR #598, opened 2020-10-01
ENH: add aerotechmotor from hxrsnd

klauer:
While cleaning up hxrsnd, I came across https://github.com/pcdshub/hxrsnd/issues/73

Closes https://github.com/pcdshub/hxrsnd/issues/73

Related: https://github.com/pcdshub/sim-ioc/pull/3

Further cleanup necessary:
- [ ] additional status handling probably unnecessary
- [ ] keyboardinterrupt utility?

### DIFF
--- a/pcdsdevices/aerotech.py
+++ b/pcdsdevices/aerotech.py
@@ -1,0 +1,693 @@
+"""
+Aerotech devices.
+"""
+import logging
+
+import numpy as np
+from ophyd import Component as Cpt
+from ophyd.signal import EpicsSignal, EpicsSignalRO, Signal
+from ophyd.utils import LimitError
+
+from .epics_motor import EpicsMotorInterface, MotorDisabledError
+from .utils import stop_on_keyboardinterrupt
+
+logger = logging.getLogger(__name__)
+
+
+class AerotechMotorError(Exception):
+    """
+    Exception raised when an action requiring the motor not have an error is
+    requested.
+    """
+    pass
+
+
+class AerotechMotorFaulted(AerotechMotorError):
+    """
+    Exception raised when an action requiring the motor not be faulted is
+    requested.
+    """
+    pass
+
+
+class AerotechMotorStopped(AerotechMotorError):
+    """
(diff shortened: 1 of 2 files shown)


ZLLentz:
Also need to see how much of this source was lifted from `pcdsdevices/epics_motor.py`

pcdsdevices/aerotech.py
```diff
@@ -0,0 +1,693 @@
...
+        """
+        # Print to console or just to the log
+        if print_set:
+            log_level = logger.info
+        else:
+            log_level = logger.debug
+
+        log_level("'%s' previous position: %s, offset: %s", self.position,
+                  self.offset)
+        self.offset += position_des - self.position
```
klauer, line 383: This does nothing in the current class hierarchy; I wonder if this relied on old pcdsdevices API?

pcdsdevices/aerotech.py
```diff
@@ -0,0 +1,693 @@
...
+
+class AerotechMotorStopped(AerotechMotorError):
+    """
+    Exception raised when an action requiring the motor to not be stopped is
+    requested.
+    """
+    pass
+
+
+class AerotechMotor(EpicsMotorInterface):
```
klauer, line 41: Once this is pared down, seems like it should go in `epics_motor.py` so the correct class can be chosen based on prefix.

pcdsdevices/aerotech.py
```diff
@@ -0,0 +1,504 @@
...
+    #     print_msg : bool, optional
+    #         Prints that the screen is being launched.
+    #     """
+    #     path = absolute_submodule_path(
+    #         "hxrsnd/screens/motor_expert_screens.sh")
+    #     if print_msg:
+    #         logger.info("Launching expert screen.")
+    #     os.system(f"{path} {self.prefix} aerotech &")
+
+    def status(self, status="", offset=0, print_status=True, newline=False,
```
klauer, line 453: This should be redone as `format_status_info`

PR #604, opened 2020-10-03
ENH: Target Stage Module

slactjohnson:
## Description

Starts a module for collecting common target stage stacks. 
Includes a convenience function for bundling stages together into a single object via either PVs or previously instantiated motor objects.
Defines a XY stage stack intended for use with a regularly spaced grid of targets, including compensation of target skew. 
Re-factors epics_motor.py slightly to create a function for determining the appropriate motor class for a PV. (While writing this I realized that this function is not strictly necessary for what I'm doing in the StageStack class, but I think it's useful, so I'm leaving it in.)

## Motivation and Context

The XYTargetGrid class will be used in LV25 in MEC.
Closes pcdshub/Bug-Reports-and-Requests#23

## How Has This Been Tested?

Tested using simulated motors since I don't have access to the hardware right now. 

Docstrings and comments.


pcdsdevices/targets.py
```diff
@@ -0,0 +1,167 @@
...
+    def move(self, nxspaces, nyspaces, wait=False):
+        """
+        Move to a specific target position given by (xspaces, yspaces)
+        from the defined initial position. Includes compensation if defined.
+        """
+        xpos = self.x_init + self.x_spacing*nxspaces + self.x_comp*nyspaces
+        ypos = self.y_init + self.y_spacing*nyspaces + self.y_comp*nxspaces
+
+        self.x.mv(xpos, wait=wait)
+        self.y.mv(ypos, wait=wait)
```
ZLLentz, line 167: note that `wait=True` here will move x, then wait for x, then move y, then wait for y

pcdsdevices/targets.py
```diff
@@ -0,0 +1,167 @@
...
+        curr_pos = self.stage.wm()
+        next_pos = curr_pos + direction*(nspaces*self.spacing)
+        self.stage.mv(next_pos, wait=wait)
+
+
+class XYTargetGrid():
+    """
+    Class methods for managing a target grid oriented normal to the beam, with
+    regular X-Y spacing between targets.
+    """
```
ZLLentz, line 81: I think this is missing a `Parameters` section for the init variables

pcdsdevices/targets.py
```diff
@@ -0,0 +1,167 @@
...
+
+    dev = cls()
+
+    return dev
+
+
+class GridAxis():
+    """
+    Class for axes that move in regularly spaced intervals.
+    """
```
ZLLentz, line 59: This is also missing a `Parameters` section, though it looks like a purely internal class?

pcdsdevices/targets.py
```diff
@@ -0,0 +1,167 @@
...
+                           "Skipping axis {}.".format(mitem, mname))
+    cls_name = name + '_StageStack'
+    cls = type(cls_name, (object,), cpts)
+
+    dev = cls()
+
+    return dev
+
+
+class GridAxis():
```
ZLLentz, line 56: There's no need to use parentheses in a class definition if you aren't inheriting from anything. No need to change it but I thought I should point it out.

pcdsdevices/targets.py
```diff
@@ -0,0 +1,167 @@
...
+        assert isinstance(spacing, float), "Specify a float target spacing"
+        self.stage = stage
+        self.spacing = spacing
+
+    def advance(self, nspaces, direction, wait=False):
+        """
+        Move the grid axis in the specified direction by (n_spacings * spacing)
+        """
+        assert direction in [1, -1], "Direction must be 1 or -1"
+        assert nspaces >= 1, "n_targets must be >= 1"
```
ZLLentz, line 70: Note that there are situations where Python skips assert statements, though I don't think we're running into this in our current python envs:
https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

ZLLentz (commented):
Should also run `docs/pre-release-notes.sh` to put in some pre-release notes. The code here looks solid.

docs/source/upcoming_release_notes/23-target-stages.rst
```diff
@@ -0,0 +1,30 @@
...
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- N/A
```
ZLLentz, line 30:
```suggestion
- tjohnson
```

docs/source/upcoming_release_notes/23-target-stages.rst
```diff
@@ -0,0 +1,30 @@
...
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
```
ZLLentz, line 18:
```suggestion
- Add XYTargetGrid, a interactive utility class for managing a target grid
  oriented normal to the beam, with regular X-Y spacing between targets.
```

PR #607, opened 2020-10-07
REF/FIX: Noisy limits and EpicsMotor limit clamping

### LINKED ISSUE #555: EpicsMotorInterface: when EPICS limits are set to `(0, 0)`, Python limits are clamped
## Expected Behavior
EPICS limits of `(0, 0)` means "no soft limits".

That means that `EpicsMotorInterface` should be able to be set to any value.

## Current Behavior
Limits are clamped to (0, 0) in an awkward way.

## Possible Solution
Check for the `(0, 0)` condition and do not clamp.

### LINKED ISSUE #591: Move interface: consider catching limit-related exceptions to not overwhelm the user with traceback messages

klauer:
## Description
```
- User-facing move functions will not be able to catch the
  :class:`~ophyd.utils.LimitError` exception.  These interactive methods are
  not meant to be used in scans, as that is the role of bluesky.
- Catch :class:`~ophyd.utils.LimitError` in all
  :class:`pcdsdevices.interface.MvInterface` moves, reporting a simple error by
  way of the interface module-level logger.
- Allow setting of :class:`~ophyd.EpicsMotor` limits when unset in the motor
  record (i.e., ``(0, 0)``) when using
  :class:`~pcdsdevices.epics_motor.EpicsMotorInterface`.
```

## Motivation and Context
Closes #555 
Closes #591 

## How Has This Been Tested?
Interactively.

Release notes.

### DIFF
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -535,10 +535,16 @@ class MvInterface(BaseInterface):
     def _log_move_end(self):
         logger.info('%s reached position %s', self.name, self.wm())
 
-    def move(self, position, *args, **kwargs):
+    def move(self, *args, **kwargs):
         try:
-            st = super().move(position, *args, **kwargs)
+            st = super().move(*args, **kwargs)
         except ophyd.utils.LimitError as ex:
+            # Pick out the position either in kwargs or args
+            try:
+                position = kwargs['position']
+            except KeyError:
+                position = args[0]
+
             self._log_move_limit_error(position, ex)
             raise
 
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -14,6 +14,7 @@ from threading import Event
 from types import MethodType, SimpleNamespace
 from weakref import WeakSet
 
+import ophyd
 import yaml
 from bluesky.utils import ProgressBar
 from ophyd.device import Device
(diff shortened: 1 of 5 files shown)


pcdsdevices/interface.py
```diff
@@ -524,14 +525,23 @@ def __init__(self, *args, **kwargs):
...
         logger.info('%s reached position %s', self.name, self.wm())
 
-    def move(self, *args, **kwargs):
-        st = super().move(*args, **kwargs)
+    def move(self, position, *args, **kwargs):
+        try:
+            st = super().move(position, *args, **kwargs)
+        except ophyd.utils.LimitError as ex:
+            self._log_move_limit_error(position, ex)
+            raise
```
ZLLentz, line 543: This will cause `bluesky` scans to miss the `LimitError`, since for subclasses of `PositionerBase`, `set` and `move` are the same thing: https://github.com/bluesky/ophyd/blob/4440bc74f226fccac6bb59a17753b9c517cbe864/ophyd/positioner.py#L46-L88

ZLLentz (approved):
This all looks good to me. I thought something was wrong at first, but I wasn't reading the code correctly, so I retracted that comment.

pcdsdevices/interface.py
```diff
@@ -524,14 +525,23 @@ def __init__(self, *args, **kwargs):
...
+
     def _log_move(self, position):
         logger.info('Moving %s from %s to %s', self.name, self.wm(), position)
 
     def _log_move_end(self):
         logger.info('%s reached position %s', self.name, self.wm())
 
-    def move(self, *args, **kwargs):
-        st = super().move(*args, **kwargs)
+    def move(self, position, *args, **kwargs):
```
ZLLentz, line 538: forcing a positional arg named position is breaking tests for the ccm psuedomotor

klauer: _Bah_, thanks.

klauer:
Have more limits-related tests to fix up here. _You should have split the PR, @klauer!_

klauer:
I saw 6 failed tests and got a bit scared, but then I saw it was one parametrized test. 😌

PR #608, opened 2020-10-08
DOC: add release notes for v3.0.0

ZLLentz:
## Description

- add release notes using the script
- add a pre-merge-checklist to the issue template

I thought the API changes were enough to warrant a v3.0.0
As usual, reviewers are everyone who contributed to the tag.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -18,3 +18,9 @@
 <!--
 ## Screenshots (if appropriate):
 -->
+
+## Pre-merge checklist
+- [ ] Test suite passes locally
+- [ ] Test suite passes on travis
+- [ ] Code works interactively
+- [ ] Ran docs/pre-release-notes.sh and created a pre-release documentation page
--- a/docs/source/releases.rst
+++ b/docs/source/releases.rst
@@ -2,6 +2,80 @@ Release History
 ###############
 
 
+v3.0.0 (2020-10-07)
+===================
+
+API Changes
+-----------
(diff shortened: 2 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -18,3 +18,9 @@
 <!--
 ## Screenshots (if appropriate):
 -->
+
+## Pre-merge checklist
+- [ ] Test suite passes locally
+- [ ] Test suite passes on travis
+- [ ] Code works interactively
+- [ ] Ran docs/pre-release-notes.sh and created a pre-release documentation page
```
klauer, line 26: A ➕

klauer (approved):
Bigger release than I recalled. Nice - v3.0.0.

.github/PULL_REQUEST_TEMPLATE.md
cristinasewell: 👏

cristinasewell (commented):
A bit late for the review, but looks great and It is good to know what goes into the release notes 👍 
One thing that I think was wrong on my part, is to say that there were `API changes` related to the `alio_to_theta` and `theta_to_alio` calculations, because theoretically we still have the same inputs and outputs - sorry about that...

ZLLentz:
I think the output is different enough for it to count as an API change. As in, if you ran analysis with the calculation code somehow, then you'd get a different result. So I thought the categorization was :+1:

This release note amalgamator is a bit of an experiment right now, I think it's working nicely.

Also, consider release note review requests as a "hey, your code is being released!" more than anything else. (especially here, where we're essentially copy/pasting already-reviewed documentation)

PR #628, opened 2020-10-18
ENH: Add camera viewer to PCDSAreaDetectorTyphos

### LINKED ISSUE #627: Cameras should have the camera viewer available from Typhos screens
## Expected Behavior
It would be great if we could access the camera viewer from camera Typhos screens. 

## Current Behavior
The small 'cam_image' viewer available in the PCDSAreaDetectorTyphos class is OK, but is pretty limited and doesn't have the features that are required by the laser system.  

## Possible Solution
A method to open the camera viewer for an instantiated camera is pretty easy. It would be great if there was a way to have Typhos make a button out of class methods using variety metadata. 

Alternative solution (which I will probably use for the short term) is to create a Typhos function panel with a camviewer function(s) for the camera(s) of interest and add this to a TyphosSuite for a particular laser TILE. 

## Context
I am trying to make Typhos screens to aid in laser commissioning coming up soon. The EDM screens we have been using are getting clunky given the large number of control points/diagnostics that we have. 

## Your Environment
Any hutch python.

slactjohnson:
## Description
I added a method to PCDSAreaDetectorTyphos to launch the camera viewer for that camera.  This method is available on Typhos screens via a button. 

## Motivation and Context
The camera viewer is pretty fully featured, and the LCLS staff are used to it. It would be nice to have this easily accessible from automatically generated Typhos screens. 

Closes #627.

## How Has This Been Tested?
Tested on a camera in the Heinz lab. This method uses the get_hutch_name script from engineering_tools, which in its latest release does not include the las or hpl subnets. There is a PR in the rectify this, and will be fixed with the next release of engineering tools. The code still works, though on the las/hpl subnets the viewer is instantiated as 'unknown_hutch' for the --instrument flag. This is fine for now, and will be fixed on the next release of engineering_tools.

The code. 

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/areadetector/detectors.py
+++ b/pcdsdevices/areadetector/detectors.py
@@ -19,6 +19,8 @@ from ophyd.signal import EpicsSignal, EpicsSignalRO
 
 from pcdsdevices.variety import set_metadata
 
+from pcdsdaq.ext_scripts import hutch_name
+
 from .plugins import (ColorConvPlugin, HDF5Plugin, ImagePlugin, JPEGPlugin,
                       NetCDFPlugin, NexusPlugin, OverlayPlugin, ProcessPlugin,
                       ROIPlugin, StatsPlugin, TIFFPlugin, TransformPlugin)
@@ -213,19 +215,21 @@ class PCDSAreaDetectorTyphos(Device):
                     num_dimensions='ndimensions',
                     kind='normal')
 
+    # Eventually I want to typhos-ify into a button. I couldn't figure out
+    # how to do that, so I will settle for this for now. 
     def viewer(self):
+        """
+        Launch the python camera viewer for this camera.
+        """
         arglist = ['/reg/g/pcds/pyps/apps/camviewer/latest/run_viewer.sh',
                    '--instrument',
-                   'las',
-                   '--oneline']
-
-        oneline = 'GE:16,{0}:IMAGE1;{0},,{0}'
-
-        # Remove terminating colon weirdness of AD ophyd device prefix
-        arglist.append(oneline.format(self.prefix[0:-1]))
+                   '{0}'.format(get_hutch().lower()),
+                   '--oneline',
+                   'GE:16,{0}:IMAGE1;{0},,{0}'.format(self.prefix[0:-1])]
 
         subprocess.run(arglist, stdout=subprocess.PIPE,
                        stderr=subprocess.PIPE, check=True)
+
 
 class PCDSAreaDetectorTyphosBeamStats(PCDSAreaDetectorTyphos):
     """
--- a/pcdsdevices/areadetector/detectors.py
+++ b/pcdsdevices/areadetector/detectors.py
@@ -19,7 +19,7 @@ from ophyd.signal import EpicsSignal, EpicsSignalRO, AttributeSignal
 
 from pcdsdevices.variety import set_metadata
 
-#from pcdsutils.ext_scripts import get_hutch_name
+from pcdsutils.ext_scripts import get_hutch_name
 
 from .plugins import (ColorConvPlugin, HDF5Plugin, ImagePlugin, JPEGPlugin,
                       NetCDFPlugin, NexusPlugin, OverlayPlugin, ProcessPlugin,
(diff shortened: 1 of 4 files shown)


slactjohnson:
Hoping for some feedback from @ZLLentz, @klauer and @hhslepicka on how to best approach Typhos-ifying a class method.

klauer:
Teddy had some preliminary work done for implementing methods dynamically in Typhos panels, but our approach changed a bit over time since he left. I think we're missing a hook for that sort of functionality in the way we handle devices.

A short-term hacky fix could be making an `AttributeSignal` (with `command-proc` variety metadata), that when writing to an associated property, spawns the screen.

```python
open_screen = Cpt(AttributeSignal, '_my_property')
set_metadata(...)

@property
def _my_property(self):
    return 0

@_my_property.setter
def _my_property(self, value):
    self.open_display()
```

If we do this enough before enhancing Typhos, it may be appropriate to make a mix-in helper class for this.

pcdsdevices/areadetector/detectors.py
```diff
@@ -209,6 +215,21 @@ class PCDSAreaDetectorTyphos(Device):
...
+    # how to do that, so I will settle for this for now. 
+    def viewer(self):
+        """
+        Launch the python camera viewer for this camera.
+        """
+        arglist = ['/reg/g/pcds/pyps/apps/camviewer/latest/run_viewer.sh',
+                   '--instrument',
+                   '{0}'.format(get_hutch().lower()),
+                   '--oneline',
+                   'GE:16,{0}:IMAGE1;{0},,{0}'.format(self.prefix[0:-1])]
```
ZLLentz, line 228: Thoughts on `IMAGE1` vs `IMAGE2`? I know that sometimes `IMAGE1` is the data stream, sometimes it's a viewer.

ZLLentz:
@klauer's `AttributeSignal` suggestion is currently the easiest way to get this into the screen. When we get back to doing `typhos` things we'll want to be providing some convention for marking up functions for this purpose.

pcdsdevices/areadetector/detectors.py
```diff
@@ -15,6 +17,10 @@
 from ophyd.device import Component as Cpt
 from ophyd.signal import EpicsSignal, EpicsSignalRO
 
+from pcdsdevices.variety import set_metadata
+
+from pcdsdaq.ext_scripts import hutch_name
```
ZLLentz, line 22: `pcdsdevices` does not have a `pcdsdaq` dependency, nor should it. This is why the travis is breaking.
Should we move these utilties to `pcdsutils`?

klauer: Sounds good to me.

ZLLentz: Ok I will transplant `ext_scripts` to `pcdsutils`

pcdsdevices/areadetector/detectors.py
slactjohnson: Yeah, I've seen both as well. My reason for choosing IMAGE1 is that some stripped down areaDetector IOCs only have IMAGE1, the Thorlabs WFS IOC, a candidate for this class, being the one I can think of off the top of my head. The VISAR IOC may be this way as well. This may not be the best reason, but it's all I have. :)

We could make this method a little smarter and have it figure out which stream to use if we find that we have a mix of cameras using IMAGE1 vs IMAGE2. It would be nice if we could standardize the purpose of the image streams internally to avoid this kind of thing.

slactjohnson:
@klauer Nice, thanks! I'm going to play around with this right now. :)

pcdsdevices/areadetector/detectors.py
ZLLentz: That's pretty good reasoning. Let's stick with IMAGE1 and maybe revisit it later.

klauer:
(Docs build failing not due to this PR's contents; its conda env Python is unpinned and building docs on 3.9)

klauer:
Cycling the PR, hoping to pick up pcds-ci-helpers changes and not have docs build fail

ZLLentz:
`pcdsutils` `v0.4.0` will have the ext_scripts once it finishes uploading

slactjohnson:
@klauer @ZLLentz I think this is basically where I want it to be. I'll clean up the PR and then I'll be ready for review. 

One question though: Do either of you have thoughts on removing the simple 'cam_image' viewer that I originally wrote into the class? The python camera viewer seems better than the cam_image viewer in basically every way. @ZLLentz, it's my understanding that this was included in the PPM class, and perhaps elsewhere. Did anyone find this feature useful? Or did people ignore it and just use a separate TMO/FEE/whatever camviewer to look at cameras? One potential use case I see for this signal would be as a thumbnail image widget, perhaps with the 'array-image' variety?

ZLLentz:
@slactjohnson the `cam_image` actually got disabled in dev last week for imagers because it broke `lucid` in conjunction with loading ~5 PPMs each on broken 0.1G links. In my experience it gets basically no use (and nobody complained that it was missing)

Coincidentally, there IS a `THUMBNAIL` stream on many of these cameras...

klauer:
PR is looking good. As far as images and area detector go: we have a long way toward getting "good" support in typhos/lucid.  It's not really even functional in its current form.

It'd be nice to have a thumbnail inlined in the display, some quick way of viewing the full image with/without spawning a new camera viewer instance, and ... many other things. Topic for another day, I suppose.

slactjohnson:
@ZLLentz @klauer OK, thanks for the feedback. Let's mothball the cam_image widget for now. 

Ask Ken said though, an in-line thumbnail would be nice at some point as a sanity check for cameras.

slactjohnson:
@klauer @ZLLentz Looks like this is failing due to pcdsutils missing.  I'm not familiar with our Travis build system; do I need to add this to requirements.txt to make this available?

klauer:
@slactjohnson - yes, please add it to `requirements.txt` and also `conda-recipe/meta.yaml` (requirements/run block)

ZLLentz (approved):
This looks good and I think it is ready to merge, can you confirm there's nothing else you want to add here @slactjohnson ?

slactjohnson:
@ZLLentz Yep, I'm all done with this for now. Thanks!

PR #633, opened 2020-10-19
Be lens classes

### LINKED ISSUE #577: Be lens class documentation, etc.
Original card:
- combined motor working in spot-size as position
- code believed to be present, needs name to be known (potentially aliases), test functionality

My take:
- Better documentation on how to use `LensStack.calib_z` and `LensStack.beam_size`
- Testing with real hardware
- Consider changing API to be more like old python
- Consider refactoring code to make it easier to read
- Also look at #541

cristinasewell:
This is a work in progress PR - will update this soon

## Motivation and Context

Trying to close #577 

## Pre-merge checklist


pcdsdevices/lens.py
```diff
@@ -286,31 +341,294 @@ def _make_safe(self):
             filt.insert()
             time.sleep(0.01)
         if filt.inserted:
-            print("REMINDER: Beam stop attenuator moved in!")
```
cristinasewell, line 289: Do we want this as `prints` instead of loggings - i think there might be a reason for it, if so i'll change it back?

pcdsdevices/lens.py
```diff
@@ -286,31 +341,294 @@ def _make_safe(self):
...
+        Object that gets PVs from lcls (for energy)
+    mono_obj
+        Object that gets energy from monochromator
+    beamsize_unfocused : float, optional
+        Radial size of x-ray beam before focusing.
+    path : str
+        Path to the file that defines which lenses are being used.
+
+    Examples
+    --------
```
cristinasewell, line 383: Ok so here is where I think I went a bit too far and provided too much information in the Examples docstring... please let me know

pcdsdevices/lens.py
```diff
@@ -286,31 +341,294 @@ def _make_safe(self):
...
+
+    >>> sets_list = [[3, 0.0001, 1, 0.0002],
+                     [1, 0.0001, 1, 0.0003, 1, 0.0005],
+                     [2, 0.0001, 1, 0.0005]]
+    >>> be.set_lens_set_to_file(sets_list, path)
+
+    Create the LensStack() object by providing the x, y and z `prefixes`, as
+    well as all the other parameters:
+
+    >>> att = Attenuator('att', 4, name='att')
```
cristinasewell, line 404: When calling `move()` we need to provide an attenuator object, how do scientists go about? Is this somewhat they way we do it (of course providing a real prefix instead of a dummy string)

pcdsdevices/lens.py
```diff
@@ -286,31 +341,294 @@ def _make_safe(self):
...
+
+    >>> be_stack = lens.LensStack(path=path, x_prefix='X:PREF',
```
cristinasewell, line 406: when we create the LensStack object, do we get the Energy  coming from? In the old code we have something like this:
https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xpp/blinst/crl.py#L102

Is this a good place to apply what we discussed here:
https://github.com/pcdshub/pcdscalc/pull/8#discussion_r509793381

Basically create the subscription in the LCLS object  to get the right E and have LensStack use that info?

pcdsdevices/lens.py
```diff
@@ -286,31 +341,294 @@ def _make_safe(self):
...
+
+    For calculations, one set at the time will be used. Use the `set_lens_set`
+    method to choose what set to use. Ex. to use the second set:
+
+    >>> sim.set_lens_set(2)
+
+    >>> sim.lens_set
+    [1, 0.0001, 1, 0.0003, 1, 0.0005]
+
+    Now you can try to tweak the motors:
```
cristinasewell, line 458: Here is where the docstring information could be way off, because I don't know exactly how we do it - any help here would be very appreciated

pcdsdevices/lens.py
```diff
@@ -286,31 +341,294 @@ def _make_safe(self):
...
+    FWHM at lens   : 5.000e-04
+    waist          : 3.113e-07
+    waist FWHM     : 3.666e-07
+    rayleigh_range : 1.965e-03
+    focal length   : 2.680e+00
+    size           : 2.092e-04
+    size FWHM      : 2.463e-04
+    0.00024626624937199417
+
+    Use `calib_z` and `beam_size` to to move the z motor:
```
cristinasewell, line 510: Is this correct? we use `calib_z` and `beam_size` to move the z motor?

pcdsdevices/lens.py
```diff
@@ -286,31 +341,294 @@ def _make_safe(self):
...
+
+    >>> sim.calib_z(sim.x.position, sim.y.position, sim.z.position)
+    >>> sim.beam_size(sim.x.position, sim.y.position, sim.z.position)
+
+    Calculated the pseudo position:
+
+    >>> sim.inverse(sim.x.position, sim.y.position, sim.z.position)
```
cristinasewell, line 517: Same thing with these two, how exactly are we using them? `inverse` and `forward`

pcdsdevices/lens.py
```diff
@@ -286,31 +341,294 @@ def _make_safe(self):
...
+            Pack of lens sets.
+        """
+        lens_pack = calcs.get_lens_set(None, self.path, get_all=True)
+        if print_only:
+            print(lens_pack)
+        else:
+            self.lens_pack = lens_pack
+            return self.lens_pack
+
+    # TODO: i think i'll need this one here or something similar
```
cristinasewell, line 572: I'm leaving this here for now because I don't know yet how we want to get the set, for now I have a method `get_lens_set(index)` to manually do it

pcdsdevices/lens.py
ZLLentz: I think it's good to include verbose docs in a docstring :+1:
The alternative is to move it into the docs/source in a tutorial section or something like that

pcdsdevices/lens.py
klauer: We almost always want logging. Only potential exception in my mind is if it's _solely_ a high-level function displaying interactive information to the user that we don't care to have logged.

pcdsdevices/lens.py
klauer: Agreed! These should render just fine in the sphinx-generated docs.

pcdsdevices/lens.py
klauer: `.tweak()` seems fine, though `.umvr()` (relative motion, effectively a "tweak") could be more compact for an example here.

pcdsdevices/lens.py
ZLLentz: Logging is better here

pcdsdevices/lens.py
ZLLentz: This is a huge problem with the existing class- there isn't a reasonable way to connect these objects together. In the hutch python session there will be an existing attenuator object, my constructor in xpp's beamline file looks like:
```
with safe_load('xpp_lens'):
    from pcdsdevices.lens import LensStack
    from xpp.db import xpp_attenuator
    _folder = "/reg/g/pcds/pyps/apps/hutch-python/xpp/"
    xpp_lens = LensStack("XPP:SB2:MMS:13",  # X
                         "XPP:SB2:MMS:14",  # Y
                         "XPP:SB2:MMS:15",  # Z
                         name='xpp_lens',
                         z_offset=3.852,   # lifted from old
                         z_dir=-1,         # lifted from old
                         E=9.5,            # keV, not automatic
                         att_obj=xpp_attenuator,  # breaks happi
                         beamsize_unfocused=500e-6,  # from old
                         path=_folder + "lens/sets_Be")
```

pcdsdevices/lens.py
ZLLentz: I'm not immediately sure how to implement here- maybe we need to register the `BeamStats` and/or `LCLS` objects that are created and have methods that set up the subscription callbacks? Maybe a little more thought needed.

pcdsdevices/lens.py
ZLLentz: Right now, if you move `calib_z` it will move the x, y, and z motors at the same time based on the last time we ran the `align` method. `beam_size` will do a calculation on top of that and move the `calib_z` motor underneath, so you can do `beam_size.move(size)` and it will move x, y, and z appropriately to get to the correct spot size at the sample.

pcdsdevices/lens.py
ZLLentz: The user should probably never need to call these directly. Internally, the `PseudoPositioner` base class uses these methods to get the positions of the calculated motors and to move the real motors based on the calculated motor setpoints.

pcdsdevices/lens.py
ZLLentz: I think `get_lens_set(index)` is a good way to handle this. Another approach could be to combine with the XFLS class and read off the y motor state, but that's not necessarily guaranteed to be consistent (what if we're between states? e.g. for a CCM experiment, which shifts the beam along the y axis)

pcdsdevices/lens.py
```diff
@@ -20,7 +18,10 @@
                         pseudo_position_argument, real_position_argument)
 from .sim import FastMotor
 
-LENS_RADII = [50e-6, 100e-6, 200e-6, 300e-6, 500e-6, 1000e-6, 1500e-6]
+logger = logging.getLogger(__name__)
+
+# Set of Be lenses with thicknesses.
+LENS_RADII = calcs.LENS_RADII
```
ZLLentz, line 24: Is this used in the file now that the calcs are all in pcdscalcs?

cristinasewell: Ohh good point, we're not using this here anymore, ('ill double check) but we should not need to use it at least - thank you

pcdsdevices/lens.py
```diff
@@ -80,17 +81,61 @@ def __init__(self, prefix, *, name, **kwargs):
...
     beam_size = Cpt(PseudoSingleInterface)
 
-    tab_whitelist = ['tweak', 'align']
+    tab_whitelist = ['tweak', 'align', 'calib_z', 'beam_size', 'create_lens',
+                     'read_lens']
     tab_component_names = True
 
+    # TODO: do we want these guys to be None? it seems like if they
+    # are None calculations fail - at least z_offset and z_dir should
+    # probably not be None
```
ZLLentz, line 138: I agree. These were probably making some assumptions about how the underlying calculation functions handled a `None` input argument, and should be changed along with the change of the underlying calculation functions.

I suspect any of these that are configurations in pcdscalcs (beamsize_unfocused) should also be removed or reworked- note how if you make this object with default args you'll ignore the configured value for that parameter.

pcdsdevices/lens.py
cristinasewell: Ok i see, so scientists are able to just use the existing attenuator then - but there isn't much I can do here for now right? I'm just going to update the docstring

ZLLentz: Makes sense

pcdsdevices/lens.py
cristinasewell: Yeah I took the `beamsize_unfocused` out (in d3c2985)- because we can configure this default value through `pcdscalc.be_lens_calcs` - i think this was the only one that needed to be removed for now

pcdsdevices/lens.py
cristinasewell: took it out in d3c2985

pcdsdevices/lens.py
cristinasewell: I am not sure how to deal with this, it seems like there are two pieces to it:

In Crl old object we have something like this:
https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xpp/blinst/crl.py#L102-L136

where both the `monoObj` and `lclsObj` are involved in getting the Energy. The lclsObj would just be similar to what we said above, but with `monoObj` is a bit more involved...

Is the old `monoObj` - the current `LODCM` object?

pcdsdevices/lens.py
cristinasewell: @ZLLentz can I assume that in the python session there will be an existing lcls object too?

ZLLentz: yes, there will be an existing lcls object

pcdsdevices/lens.py
```diff
@@ -398,13 +424,23 @@ class LensStack(LensStackBase):
     >>> from pcdsdevices.attenuator import Attenuator
     >>> att = Attenuator('att', 4, name='att')
 
+    To be able to get the current Beam Energy import the lcls object eg.:
+
+    >>> from xpp.db import xpp_lcls
```
ZLLentz, line 429:
```suggestion
    >>> from xpp.db import lcls
```

pcdsdevices/lens.py
ZLLentz: oh that's interesting

Yes the mono obj is the LODCM usually- except when it is the CCM... XPP and XCS each have two monochromators. So deciding which mono object to use is experiment-specific... and the lodcm calcs aren't even set up yet in the python 3 libraries

cristinasewell: Ok we can use the LclsObject's energy for now, and then later we can add the LODCM's into play too - at least we should probably do this in a separate Issue - or what do you think?

pcdsdevices/lens.py
cristinasewell: thank you - fixed it in a675677

pcdsdevices/lens.py
ZLLentz: Yes I'd consider adding LODCM/CCM into this as a separate issue. In practice it should be OK to just override the energy with your planned energy for the experiment.

pcdsdevices/lens.py
klauer: This `print` is gone but a few other `print()` statements seem to remain in error

cristinasewell: Ok thanks Ken. 
The only one that's left now is the one from here:
https://github.com/pcdshub/pcdsdevices/blob/f771f2a1fbc4b3a1daa3dc48d1bccb3e47fdcf8f/pcdsdevices/lens.py#L254-L260

I don't think we mean to log it though - should I still make it a `logger.debug()` or something like this?

ZLLentz (approved):
This looks good to me, will need to re-run travis when we tag the calcs PR

ZLLentz:
I've restarted the travis build because the pcdscalc tag is up on conda forge now

klauer:
Tests are passing. Are we ready here?

cristinasewell:
> Tests are passing. Are we ready here?

I think I am as ready as I can be here for now, until we get some feedback from the scientists... or from somebody who has more experience about how this class is supposed to behave

ZLLentz:
Merging!

PR #635, opened 2020-10-22
Von hamos beckhoff

jsheppard95:
## Description
- Added class `vonhamos_motion.VonHamos`. Currently, this just contains a single vertical axis, with the rest of the Von Hamos system using LCLS-I style dumb motors with MForce Chassis.
- Currently feels like overkill to have separate class for one Beckhoff axis, but eventually all LCLS-I dumb motors will be converted to Beckhoff axes and can be included in this class

## Motivation and Context
- Need to control Von Hamos Y axis. This originally was a dumb motor but it needed to be replaced. Thus, the decision was made to start the Beckhoff conversion with this single axis.

## How Has This Been Tested?
- Ran `pcdsdevices` test suite, no failed tests.
- Instantiated object in Ipython session/inspected components
- Opened test typhos screen

- Docstrings

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/636-VonHamos_motion_system_class.rst
+++ b/docs/source/upcoming_release_notes/636-VonHamos_motion_system_class.rst
@@ -1,0 +1,30 @@
+636 VonHamos_motion_system_class
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- Von Hamos Spectrometer (HXX-VON_HAMOS)
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- sheppard
--- a/pcdsdevices/vonhamos_beckhoff_motion.py
+++ b/pcdsdevices/vonhamos_beckhoff_motion.py
@@ -1,0 +1,34 @@
+"""
+LAMP Motion Classes
+
+This module contains classes related to the TMO-LAMP Motion System
+"""
+
+from ophyd import Component as Cpt
+from ophyd import Device
+
+from .epics_motor import BeckhoffAxis
+from .interface import BaseInterface
+
+
+class VonHamosBeckhoff(BaseInterface, Device):
+    """
+    Von Hamos Beckhoff Motion Class
+
+    This class controls Beckhoff motors fixed to the Von Hamos Spectrometer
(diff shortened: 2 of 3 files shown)


ZLLentz:
A bit confused here:
- Is it not good enough to just make a BeckhoffAxis in xpp's python?
- Is it not possible to include the other axes? There is an IMS class.

klauer:
Makes sense to include the IMS axes and then change them over to Beckhoff if/when the hardware switchover happens.

jsheppard95:
Yes I think you are both right. For the LV68 experiment, there's no reason I couldn't just add this axis to xpp3. I could even make a happi entry for it and set the class as `epics_motor.BeckhoffAxis` so we can have a simple typhos screen as well. And yes I should definitely have the other axes in here as well, and just set them as IMS for now.

ZLLentz:
Ok, which direction do you want to go then?
1. fix travis style complaints and merge PR essentially as-is
2. table PR for now, get xpp working without
3. update PR to have all the axes

I'll be happy with any of these or another suggestion

jsheppard95:
I ended up going with option 2 for now just to get to get XPP going. I just added the axis to `xpp3` and then made a happi entry for the motor itself with the name `hxx_von_hamos_y` and class `pcdsdevices.epics_motor.BeckhoffAxis`, just to have a simple GUI as well.

So tabling the PR for now, but it would be awesome to have a permanent class. I realized though its a bit tricky with some of the axes as IMS, since we run these through the MForce Chassis PPL's and thus the PVs change depending on which hutch/channels are chosen. So far its only Y that's fixed since its running on the Beckhoff panel. Does it make sense to specify the IMS PVs in the `__init__` for now?

ZLLentz:
You can totally specify all the PVs in the ``__init__`` if you like- it's a bit annoying to do but we put some magic in with `UnrelatedComponent` to make it easier. See example usage here: https://github.com/pcdshub/pcdsdevices/blob/dc9237ce539da7de76bf31371652b9506ff3ef99/pcdsdevices/sample_delivery.py#L277-L400

This is probably the way to go if you want a permanent class with migrating IOCs, if you think that's a useful thing to have.

ZLLentz:
I'm going to close this PR, assuming you have other things you'd prefer to be spending your time on @jsheppard95. Feel free to re-open if I was wrong.

jsheppard95:
Yes closing this PR was appropriate, definitely other things I'd prefer to work on, but if a greater need arises I'll return/reopen.

Issue #647, opened 2020-10-26
Add import tests for all modules


ZLLentz:
## Expected Behavior

Test that every class's class definition works on import. For bonus points, test that every class can be instantiated.

## Current Behavior

https://github.com/pcdshub/pcdsdevices/pull/646#issuecomment-716656907

## Possible Solution

Example: https://github.com/pcdshub/pcdsdevices/blob/03a1dc2cdd08b8a39198e4bed9834b436c445732/tests/test_wfs.py#L1-L13

## Context

Can catch bugs that are easy to miss if we're going fast

klauer:
Top-level modules actually do not have this issue. I think the issue was that this happened in a submodule of pcdsdevices.

Why? I did a quick `echo 'garbage string' > pcdsdevices/foobar.py` and `test_interface` fails:

```
================================================================ ERRORS =================================================================
_______________________________________________ ERROR collecting tests/test_interface.py ________________________________________________
test_interface.py:219: in <module>
    for cls in conftest.find_all_device_classes()
conftest.py:103: in find_all_device_classes
    __import__(f'{pkgname}.{module}')
../pcdsdevices/foobar.py:1: in <module>
    jasldfkjalksdjfaoi
E   NameError: name 'jasldfkjalksdjfaoi' is not defined
```

ZLLentz:
I don't see how this could happen then, since the `find_all_device_classes` mechanism looks like it is iterating properly through submodules

klauer:
We now do best-effort instantiation of all device classes found in pcdsdevices, including those in sub-packages.

I think this is good enough, but feel free to re-open if you don't.

PR #657, opened 2020-10-27
FIX: import laser devices; implicitly test interface through find_all_device_classes

klauer:
## Description
* Add laser imports to `pcdsdevices.device_types`
* `test_interfaces` will now pick up laser classes and check their tab completion settings

It may be worthwhile to do an alternative PR to this, moving laser classes to the top level. I think we at least discussed that possibility or we may have had an issue on that at some point?

## Motivation and Context
Related to #647 , but does not close it.

## How Has This Been Tested?
Locally, noting that laser devices are imported and run through `test_interfaces`.

Release notes

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/647-import_lasers.rst
+++ b/docs/source/upcoming_release_notes/647-import_lasers.rst
@@ -1,0 +1,31 @@
+647 import lasers
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- Add laser imports to :mod:`pcdsdevices.device_types`.  Test fixtures now
+  verify imported laser devices' tab completion settings.
+
+Contributors
+------------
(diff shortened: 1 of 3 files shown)


ZLLentz:
Rather than move laser classes to the top level, I want to move more devices into folders. When there are too many files in a directory I find it hard to find the one that I'm looking for. Open to opinions on this.

ZLLentz:
Another alternative to this would be to rework the `find_all_device_classes` mechanism to recursively search through submodules. It looks like there is a `pkgutil.walk_packages` for this?

klauer:
Yeah, that could be a good option @ZLLentz, if that's the direction you want to take it 👍

klauer:
As far as an opinion goes - there's pluses and minuses of flat vs structured, of course. I could be convinced we're at the threshold where making submodules is sensible, so I wouldn't argue against it.

ZLLentz (approved):
I see no reason not to merge this either way

ZLLentz:
Except the lack of pre-release notes and the deleted checklist

klauer:
Will add it tomorrow 👍

ZLLentz:
I think this is good to merge

PR #658, opened 2020-10-29
WIP: MODS TILE Devices

### LINKED ISSUE #639: ENH: Add MODS TILE, MODS Table Classes
## Expected Behavior
We should have classes for MODS TILEs. 

## Current Behavior
We do not have classes for MODS TILEs. 

## Possible Solution
Write classes for MODS TILEs. 

Base Classes: 
1. BaseTILE
2. BaseMODS

Shared Classes:
1. Injection
2. Compressor
3. Harmonics
4. ATM

Unique Classes:
1. TMO Ejection
2. ChemRIXS Ejection
3. XPP Injection?
4. TMO MODS Table
5. ChemRIXS MODS Table

## Context
We would like to use TILE classes to bundle the MODS devices for screen generation as well as interactive use. 

## Your Environment
Laspython.

slactjohnson:
## Description
Adds in Ophyd devices for various MODS laser TILEs, as well as complete MODS Table devices.

Will add more details later. 

## Motivation and Context
Closes #639 

## How Has This Been Tested?
WIP

WIP

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/lasers/mods.py
+++ b/pcdsdevices/lasers/mods.py
@@ -1,0 +1,288 @@
+"""
+Module for the L2SI Laser MODS Table TILEs and Table classes.
+"""
+from ophyd import Device
+from ophyd import Component as Cpt
+
+from pcdsdevices.epics_motor import SmarAct, SmarActTipTilt
+from pcdsdevices.areadetector.detectors import LasBasler
+
+from pcdsdevices.lasers.tuttifrutti import TuttiFrutti, TuttiFruttiCls
+from pcdsdevices.lasers.zoomtelescope import ZoomTelescope
+from pcdsdevices.lasers.ek9000 import EnvironmentalMonitor, El3174AiCh
+from pcdsdevices.lasers.elliptec import EllLinear, EllRotation, Ell9
+
+
+### Generic TILE Devices
+class TileBase(Device):
+    """
+    Base class for MODS TILEs.
+    """
+    pass
+
+
+class InjectionTile(TileBase):
+    """
+    Class for L2SI MODS injection TILE. Common to several installations.
+
+    Parameters
+    ----------
+    prefix : str
+        Five character prefix for particular MODS installation, e.g. LM1K4
+        (TMO).
+
+    name : str
+        Name for the device, e.g. tmo_mods_inj_tile
+
+    Examples
+    --------
+    # TMO Injection TILE
+    tmo_inj = InjectionTile('LM1K4', 'tmo_inj')
+    # ChemRIXS Injection TILE
+    crixs_inj = InjectionTile('LM2K2', 'crixs_inj')
+    """
+    zoom_telescope = Cpt(ZoomTelescope, ':INJ_MP1_ZOO1', kind='normal')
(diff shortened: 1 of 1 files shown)


pcdsdevices/lasers/mods.py
```diff
@@ -0,0 +1,288 @@
...
+
+    Examples
+    --------
+    # TMO Injection TILE
+    tmo_inj = InjectionTile('LM1K4', 'tmo_inj')
+    # ChemRIXS Injection TILE
+    crixs_inj = InjectionTile('LM2K2', 'crixs_inj')
+    """
+    zoom_telescope = Cpt(ZoomTelescope, ':INJ_MP1_ZOO1', kind='normal')
+    waveplate1 = Cpt(SmarAct, ':INJ_MP1_ATT1_WP1', kind='normal')
```
klauer, line 45: This may be part of your upcoming commits (and if so, feel free to ignore this comment) - but I'm curious about the nomenclature. 

`:INJ_MP1_ATT1_WP1` Is there meaning in the order of those, like `INJ -> MP1 -> ATT1...`? What's `MP`, etc.?

slactjohnson: Hi Ken, this is part of the PV nomenclature that we have decided to use for the laser system. We were able to get the laser group on board with giving everything in the system a PV, including things that don't usually have PVs like static optics and optomechanics. 

This is documented somewhere (I can send to you if you're interested), but the short version is that the names are given by: <TILE>_<BeamPath>_<Assembly>_<Subassembly_or_Component>_<Component>. MP is a "Main Path", e.g. a primary path for beam propagation. This contrasts with DP or "Diagnostic Path" which is for delivering the beam to a diagnostic station. You'll see here that the PVs aren't completely sequential, e.g. MR1 (Mirror), MR2, MR3 etc. because there are static mirrors and other stuff in between that get their own name in sequence. 

It's a little convoluted, but it's what we came up with. There can be over 100 components in a TILE, so we had to come up with some way to keep things straight, and we decided that the language of choice would be PVs, both active and imaginary. This has a nice benefit in that if we decide to motorize or otherwise EPICS-ify a thing, we don't have to change any names.

klauer: Ah, very interesting - thanks for the info, Tyler! 

I think I may have ~inadvertently~ successfully baited you into making a nice module docstring detailing the naming here too 😁

slactjohnson: Very sneaky. Well done. :)

ZLLentz:
Hi Tyler, are you planning to complete this one? Or is it secretly done?

slactjohnson:
Hi Zach, I'm not sure what to do with this one. Now that we're putting everything into Happi for LUCID and IOC stuff, I'm not sure it makes sense to use these big, monolithic classes for the MODS. The original plan was to use these classes to generate screens, but now we're leaning on LUCID for that.

ZLLentz:
Ok, feel free to either close or let it stay open then. I'm just sifting through because I noticed we had a lot of stale PRs.

PR #662, opened 2020-10-31
Mpod channels

cristinasewell:
Added a MPODChannel object. 

## Motivation and Context

Trying to implement and close https://github.com/pcdshub/Bug-Reports-and-Requests/issues/50
## How Has This Been Tested?

Locally only.

Doctring.

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/662-MPOD_Channels.rst
+++ b/docs/source/upcoming_release_notes/662-MPOD_Channels.rst
@@ -1,0 +1,30 @@
+662 MPOD Channels
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- `MPODChannelHV`, and `MPODChannelLV` new Devices have been added.
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- cristinasewell
--- a/pcdsdevices/device_types.py
+++ b/pcdsdevices/device_types.py
@@ -39,4 +39,4 @@ from .spectrometer import Kmono, VonHamos4Crystal
 from .timetool import Timetool, TimetoolWithNav
 from .valve import GateValve, Stopper
 from .wfs import WaveFrontSensorTarget
-from .mpod import MPODChannelHV, MPODChannelLV
+from .mpod import HVChannel, LVChannel
--- a/pcdsdevices/device_types.py
+++ b/pcdsdevices/device_types.py
@@ -39,4 +39,4 @@ from .spectrometer import Kmono, VonHamos4Crystal
 from .timetool import Timetool, TimetoolWithNav
 from .valve import GateValve, Stopper
 from .wfs import WaveFrontSensorTarget
-from .mpod import HVChannel, LVChannel
+from .mpod import MPODChannelHV, MPODChannelLV
--- a/pcdsdevices/device_types.py
+++ b/pcdsdevices/device_types.py
@@ -39,3 +39,4 @@ from .spectrometer import Kmono, VonHamos4Crystal
 from .timetool import Timetool, TimetoolWithNav
 from .valve import GateValve, Stopper
 from .wfs import WaveFrontSensorTarget
+from .mpod import MPODChannelHV, MPODChannelLV
--- a/pcdsdevices/mpod.py
+++ b/pcdsdevices/mpod.py
@@ -1,0 +1,106 @@
+import logging
+
+from ophyd.device import Component as Cpt
+from ophyd.device import Device
+from ophyd.signal import EpicsSignal, EpicsSignalRO
+
+from .interface import BaseInterface
+
+logger = logging.getLogger(__name__)
+
+
+class MPODChannel(BaseInterface, Device):
+    """
+    MPOD Channel Object.
+
+    Parameters
+    ----------
+    prefix : str
+        The EPICS base of the MPOD Channel.
+
+    name : str
+        A name to refer to the device.
+    """
+    # TODO - might have too many Signals here!!
+    voltage = Cpt(EpicsSignal, ':GetVoltageMeasurement',
+                  write_pv=':SetVoltage', kind='normal',
+                  doc='MPOD Channel Voltage Measurement [V]')
+    max_voltage = Cpt(EpicsSignalRO, ':GetMaxVoltage', kind='normal',
+                      doc='MPOD Channel Maximum Voltage [V]')
+    terminal_voltage = Cpt(EpicsSignalRO, ':GetTerminalVoltageMeasurement',
+                           kind='normal', doc='MPOD Terminal Voltage [V]')
+    current = Cpt(EpicsSignalRO, ':GetCurrentMeasurement', kind='normal',
+                  doc='MPOD Channel Current Measurement [A]')
+    max_current = Cpt(EpicsSignalRO, ':GetMaxCurrent', kind='normal',
+                      doc='MPOD Channel Max Current [A]')
+    temperature = Cpt(EpicsSignalRO, ':GetTemperature', kind='normal',
+                      doc='MPOD Temperature [C]')
+    status_string = Cpt(EpicsSignalRO, ':GetStatusString',
+                        kind='normal', doc='MPOD Channel Status String')
+    # TODO - should i make this string=True?
+    state = Cpt(EpicsSignal, ':GetSwitch', write_pv=':SetSwitch',
+                kind='normal', doc='MPOD Channel State [Off/On]')
+    # 0 means no EPICS high limit.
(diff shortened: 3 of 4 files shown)


ZLLentz:
I did a bit of snooping on the XPP screens as a first-pass check. Here is what their control screen looks like on my screen:
image

The top is a low voltage module, the bottom is a high voltage module. Each has 8 channel objects associated with it. The following sample PVs are common between the two:
```
XPP:R39:MPD:CH:0:GetVoltage.DESC
XPP:R39:MPD:CH:0:SetSwitch Off
XPP:R39:MPD:CH:0:SetSwitch On
XPP:R39:MPD:CH:0:SetSwitch Emer
XPP:R39:MPD:CH:0:SetSwitch Reset
XPP:R39:MPD:CH:0:SetSwitch Clr
XPP:R39:MPD:CH:0:GetSwitch
XPP:R39:MPD:CH:0:SetVoltage
XPP:R39:MPD:CH:0:GetVoltageMeasurement
XPP:R39:MPD:CH:0:GetTerminalVoltageMeasurement
XPP:R39:MPD:CH:0:SetCurrent
XPP:R39:MPD:CH:0:GetCurrentMeasurement
XPP:R39:MPD:CH:0:GetTemperature
```
The low voltage has the additional rise/fall rates per channel:
```
XPP:R39:MPD:CH:0:SetVoltageRiseRate
XPP:R39:MPD:CH:0:SetVoltageFallRate
```
The high voltage has one voltage rise/fall for the whole card... so for all 8 channels, one rate governs the full set, and it has a different PV prefix, for example compare the channel 100 temperature with the module 10 voltage rise PV:
```
XPP:R39:MPD:CH:100:GetTemperature
XPP:R39:MPD:MOD:10:SetVoltageRiseRate
```

cristinasewell:
Ohh thanks @ZLLentz - I need to adjust a few things then: 
Should I have `SetSwitch` as `print=True` and set all these with string?
```
XPP:R39:MPD:CH:0:SetSwitch Off
XPP:R39:MPD:CH:0:SetSwitch On
XPP:R39:MPD:CH:0:SetSwitch Emer
XPP:R39:MPD:CH:0:SetSwitch Reset
XPP:R39:MPD:CH:0:SetSwitch Clr
```
or should I use their corresponding values:
```
    Enums:            ( 5)
                      [ 0] Off
                      [ 1] On
                      [ 2] Reset
                      [ 3] EmerOff
                      [ 4] ClrEvnt
```
Just to make sure here, I'm going to have a few more common methods correct: `reset`, `emer_off` and `clear_event` (is emer - emergency) besides the `on` and `off`?

And for the `Low Voltage` I should have two functions probably:
`set_voltage_rise_rate` and `set_voltage_fall_rate` ?


Also, `:GetVoltage` is different than `:GetVoltageMeasurement` - what is the difference here? I am currently using this: https://github.com/pcdshub/pcdsdevices/blob/309285be459cf45258c2204bcc0420598669c5fb/pcdsdevices/mpod.py#L25-L27 for `voltage` - (based on this info https://github.com/pcdshub/Bug-Reports-and-Requests/issues/50#issuecomment-715477125) - is that fine? - or should I be using `:GetVoltage` for voltage and have `:GetVoltageMeasurement` into a different signal?

One more question though, am I on the right path with having the Class per Channel instead of a Class per HV module, and Class per LV module? (we can have modules with more channels, not only 8 correct?)

ZLLentz:
Oh shoot some of my notes were mixed in with that... sorry! Only the PV names actually meant something, I was scribbling over a notepad with the button labels and copy/pasted the whole thing in.

Answering each line item:

- Yes I think `string=True` is best so that people can read off directly what state the MPOD is in. Your enums are correct, my noted scribbles are not.

- Regarding functions/methods, you only need to implement the ones Silke asked for, though I'm sure if you implemented all the variants for the `SetSwitch` PV they would get some use.

- I have absolutely no idea what the difference between the two get voltages are! I was simply harvesting a screen (and then I got distracted). You'll note in my screenshot they are always the same number, I suspect we can ignore the second one.

- I think having one class per channel is the best approach because different channels are typically used for very different things.

- One thing I meant to point out about the HV more closely is the differing prefix for the rise/fall rate... so that channel needs to be supplied two prefixes if we want any HV channel to able to configure the rise/fall rate

cristinasewell:
That's great, thank you. 
Regarding this- to double check:
- ***One thing I meant to point out about the HV more closely is the differing prefix for the rise/fall rate... so that channel needs to be supplied two prefixes if we want any HV channel to able to configure the rise/fall rate***

I should be having something like this in the base `MPODChannel()` class:
```
  def __init__(self, prefix, prefix_ch,  name='MPOD', **kwargs):
```
And when we create an mpod channel object we'll have to always provide the prefix for the channel separately? 

hv_mpod = MPODChannelHV(prefix='XPP:R39:MPD:', prefix_ch='100')

something like this?

silkenelson:
Actually, we tend to use GetVoltageMeasurement everywhere: the mpod has a 'Sense' which you read the GetVoltageMeasurement and one that gives you the voltage at the 'output'. We bind them together, which is why in practice it does not matter right now.

cristinasewell:
> Actually, we tend to use GetVoltageMeasurement everywhere: the mpod has a 'Sense' which you read the GetVoltageMeasurement and one that gives you the voltage at the 'output'. We bind them together, which is why in practice it does not matter right now.

ok that sounds good, thank you @silkenelson - i'll keep using the `:GetVoltageMeasurement` for now then.

ZLLentz:
Yes, that's the right idea with the doubling of the prefixes, but I think the channel prefix is the main one (since it's a channel object) and the mpod card prefix is the secondary one... I think I'd want to pass something like:
`HVChannel('XPP:R39:MPD:CH:100', card_prefix='XPP:R39:MPD:MOD:10', name='xpp_mpod_ch100')`

For consistency with the LV channels, and so we don't have many objects sharing the same main prefix.

I understand yours is shorter- I think it's important for clarity and simplicity of implementation for the main prefix to be the full channel prefix. Open to differing opinions of course.

cristinasewell:
> Yes, that's the right idea with the doubling of the prefixes, but I think the channel prefix is the main one (since it's a channel object) and the mpod card prefix is the secondary one... I think I'd want to pass something like:
> `HVChannel('XPP:R39:MPD:CH:100', card_prefix='XPP:R39:MPD:MOD:10', name='xpp_mpod_ch100')`
> 
> For consistency with the LV channels, and so we don't have many objects sharing the same main prefix.
> 
> I understand yours is shorter- I think it's important for clarity and simplicity of implementation for the main prefix to be the full channel prefix. Open to differing opinions of course.

I see.., I was thinking a bit differently but your suggestion is much better. 
Please let me know if this is what you meant: 5bdc777

ZLLentz:
Yeah, that seems pretty good. Straightforward and clear.

You didn't need to remove MPOD from all the classnames- I just couldn't remember what you had used and was trying to save time while writing my comment. I have no idea whether `MPODChannelLV` or `LVChannel` is preferable in any sense (in my head they are basically the same).

There's a small amount of code duplication here but it doesn't bother me much.

silkenelson:
I'm leaning to leave MPOD in the name as we have other power supplies we support controls for.
I would think that we might want crate/card objects that are nested carte <- has cards <- have channels might be nice, but for questionnaire should create for refers to the single channels.

cristinasewell:
> I'm leaning to leave MPOD in the name as we have other power supplies we support controls for.
> I would think that we might want crate/card objects that are nested carte <- has cards <- have channels might be nice, but for questionnaire should create for refers to the single channels.

Sounds good, i will add that back - thank you

cristinasewell:
> Yeah, that seems pretty good. Straightforward and clear.
> 
> You didn't need to remove MPOD from all the classnames- I just couldn't remember what you had used and was trying to save time while writing my comment. I have no idea whether `MPODChannelLV` or `LVChannel` is preferable in any sense (in my head they are basically the same).
> 
> There's a small amount of code duplication here but it doesn't bother me much.

Ohh and I can totally put in the `MPODChannel` class at least the `set_voltage_rise_rate` and `set_voltage_fall_rate` functions so I don't duplicate them - i don't know if i can put anything else - thank you

cristinasewell:
> > I'm leaning to leave MPOD in the name as we have other power supplies we support controls for.
> > I would think that we might want crate/card objects that are nested carte <- has cards <- have channels might be nice, but for questionnaire should create for refers to the single channels.
> 
> Sounds good, i will add that back - thank you

Any preferences though? 
```
MPODHVChannel
MPODHvChannel
MpodHvChannel
MPODChannelHV
```
or others - please let me know :)

ZLLentz:
Appending to the end makes the classes sort together alphabetically, for whatever that's worth:
`MPODChannel`
`MPODChannelHV`
`MPODChannelLV`

ZLLentz (approved):
This looks ready to go to me, completes the linked issue cleanly and in full.

klauer (approved):
👍 merge at will

PR #664, opened 2020-11-04
Acromag channels prefixed

cristinasewell:
Added `AcromagChannelInput` and `AcromagChannelOutput` devices - they will help create an `Acromag` object per channel.

Still trying to figure out what to do here.....

## Motivation and Context

Partially addresses https://github.com/pcdshub/Bug-Reports-and-Requests/issues/51

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/analog_signals.py
+++ b/pcdsdevices/analog_signals.py
@@ -61,6 +61,56 @@ class Acromag(BaseInterface, Device):
     tab_component_names = True
 
 
+class AcromagChannelInput(BaseInterface, Device):
+    """
+    Class for an Acromag Channel Input.
+
+    Parameters
+    ----------
+    prefix : str
+        The base EPICS PV for the Acromag. E.g.: `XPP:USR:ai1`
+    channel : str
+        The channel number. E.g.: `1` [0-15]
+    name : str
+        A name to refer to the Acromag Input Channel.
+    """
+    channel = FCpt(EpicsSignalRO, '{self._prefix}:{self._channel}',
+                   kind='normal')
+
+    tab_component_names = True
+
+    def __init__(self, prefix, channel, **kwargs):
+        self._channel = channel
+        self._prefix = prefix
+        self._name = ''.join(['ai1_', self._channel])
+        super().__init__(prefix, name=self._name)
+
+
+class AcromagChannelOutput(BaseInterface, Device):
+    """
+    Class for an Acromag Channel Output.
+
+    Parameters
+    ----------
+    prefix : str
+        The base EPICS PV for the Acromag. E.g.: `XPP:USR:ao1`
+    channel : str
+        The channel number. E.g.: `1` [0-15]
+    name : str
+        A name to refer to the Acromag Output Channel.
+    """
+    channel = FCpt(EpicsSignal, '{self._prefix}:{self._channel}',
+                   kind='normal')
+
+    tab_component_names = True
+
+    def __init__(self, prefix, channel, **kwargs):
+        self._channel = channel
+        self._prefix = prefix
+        self._name = ''.join(['ao1_', self._channel])
+        super().__init__(prefix, name=self._name)
+
+
 class Mesh(BaseInterface, Device):
     """
     Class for Mesh HV Supply that is connected to Acromag inputs and outputs.
--- a/pcdsdevices/analog_signals.py
+++ b/pcdsdevices/analog_signals.py
@@ -79,10 +79,10 @@ class AcromagChannelInput(BaseInterface, Device):
 
     tab_component_names = True
 
-    def __init__(self, prefix, channel, **kwargs):
+    def __init__(self, prefix, channel, name=None, **kwargs):
         self._channel = channel
         self._prefix = prefix
-        self._name = ''.join(['ai_', self._channel])
+        self._name = name or ''.join(['ai_', self._channel])
         super().__init__(prefix, name=self._name)
 
 
--- a/pcdsdevices/analog_signals.py
+++ b/pcdsdevices/analog_signals.py
@@ -61,75 +61,30 @@ class Acromag(BaseInterface, Device):
     tab_component_names = True
 
 
-class AcromagChannelInput(BaseInterface, Device):
+def AcromagChannel(prefix, channel, *, signal_class=None, name=None, **kwargs):
     """
-    Class for an Acromag Channel Input.
+    This is a factory function for creating an Acromag output or input signal.
 
     Parameters
     ----------
     prefix : str
         The base EPICS PV for the Acromag. E.g.: `XPP:USR:ai1`
-    channel : str
+    channel : int or str
         The channel number. E.g.: `1` [0-15]
-    name : str
-        A name to refer to the Acromag Input Channel.
+    signal_class : str, optional
+        Signal Class to use. E.g.: `EpicsSignalRO`, `EpicsSignal`. If `None`
+        provided, it will use the prefix to try to generate the appropriate
+        signal class.
+    name : str, optional
+        A name to refer to the Acromag Channel, if `None` it will use the
+        prefix and channel to generate a name. E.g.: `ai_1` or `ao_1`.
     """
-    channel = FCpt(EpicsSignalRO, '{self._prefix}:{self._channel}',
-                   kind='normal', doc='Analog Input [Volts]')
-
-    tab_component_names = True
-    tab_whitelist = ['get_voltage']
-
-    def get_voltage(self):
-        """Get the voltage."""
-        return self.channel.get()
-
-    def __init__(self, prefix, channel, name=None, **kwargs):
-        self._channel = channel
-        self._prefix = prefix
-        self._name = name or ''.join(['ai_', self._channel])
-        super().__init__(prefix, name=self._name)
-
-
-class AcromagChannelOutput(BaseInterface, Device):
-    """
-    Class for an Acromag Channel Output.
-
-    Parameters
-    ----------
-    prefix : str
-        The base EPICS PV for the Acromag. E.g.: `XPP:USR:ao1`
-    channel : str
-        The channel number. E.g.: `1` [0-15]
-    name : str
-        A name to refer to the Acromag Output Channel.
-    """
-    channel = FCpt(EpicsSignal, '{self._prefix}:{self._channel}',
-                   kind='normal', doc='Analog Output [Volts]')
-
-    tab_component_names = True
-    tab_whitelist = ['get_voltage', 'set_voltage']
-
-    def get_voltage(self):
-        """Get the voltage."""
-        return self.channel.get()
-
-    def set_voltage(self, voltage):
-        """
-        Set the voltage in Volts.
-
-        Parameters
-        ----------
-        voltage : number
-            Voltage in Volts.
-        """
-        return self.channel.put(voltage)
-
-    def __init__(self, prefix, channel, name=None, **kwargs):
-        self._channel = channel
-        self._prefix = prefix
-        self._name = name or ''.join(['ao_', self._channel])
-        super().__init__(prefix, name=self._name)
+    if signal_class is None:
+        signal_class = EpicsSignalRO if ':ai' in prefix else EpicsSignal
+    name_prefix = 'ai_' if ':ai' in prefix else 'ao_'
+    name = name or ''.join([name_prefix, str(channel)])
+    prefix = ''.join([prefix, ':', str(channel)])
+    return signal_class(prefix, name=name, kind='normal')
 
 
 class Mesh(BaseInterface, Device):
--- a/pcdsdevices/analog_signals.py
+++ b/pcdsdevices/analog_signals.py
@@ -83,8 +83,8 @@ def acromag_factory_func(prefix, channel, *, signal_class=None, name=None,
     if signal_class is None:
         signal_class = EpicsSignalRO if ':ai' in prefix else EpicsSignal
     name_prefix = 'ai_' if ':ai' in prefix else 'ao_'
-    name = name or ''.join([name_prefix, str(channel)])
-    prefix = ''.join([prefix, ':', str(channel)])
+    name = name or f'{name_prefix}{channel}'
+    prefix = f'{prefix}:{channel}'
     return signal_class(prefix, name=name, kind='normal')
 
 
--- a/pcdsdevices/analog_signals.py
+++ b/pcdsdevices/analog_signals.py
@@ -61,8 +61,8 @@ class Acromag(BaseInterface, Device):
     tab_component_names = True
 
 
-def acromag_factory_func(prefix, channel, *, signal_class=None, name=None,
-                         **kwargs):
+def acromag_ch_factory_func(prefix, channel, *, signal_class=None, name=None,
+                            **kwargs):
     """
     This is a factory function for creating an Acromag output or input signal.
 
@@ -86,6 +86,9 @@ def acromag_factory_func(prefix, channel, *, signal_class=None, name=None,
     name = name or f'{name_prefix}{channel}'
     prefix = f'{prefix}:{channel}'
     return signal_class(prefix, name=name, kind='normal')
+
+
+AcromagChannel = acromag_ch_factory_func
 
 
 class Mesh(BaseInterface, Device):
--- a/pcdsdevices/analog_signals.py
+++ b/pcdsdevices/analog_signals.py
@@ -75,7 +75,7 @@ class AcromagChannelInput(BaseInterface, Device):
         A name to refer to the Acromag Input Channel.
     """
     channel = FCpt(EpicsSignalRO, '{self._prefix}:{self._channel}',
-                   kind='normal')
+                   kind='normal', doc='Analog Input [Volts]')
 
     tab_component_names = True
 
@@ -100,7 +100,7 @@ class AcromagChannelOutput(BaseInterface, Device):
         A name to refer to the Acromag Output Channel.
     """
     channel = FCpt(EpicsSignal, '{self._prefix}:{self._channel}',
-                   kind='normal')
+                   kind='normal', doc='Analog Output [Volts]')
 
     tab_component_names = True
 
--- a/pcdsdevices/analog_signals.py
+++ b/pcdsdevices/analog_signals.py
@@ -78,6 +78,11 @@ class AcromagChannelInput(BaseInterface, Device):
                    kind='normal', doc='Analog Input [Volts]')
 
     tab_component_names = True
+    tab_whitelist = ['get_voltage']
+
+    def get_voltage(self):
+        """Get the voltage."""
+        return self.channel.get()
 
     def __init__(self, prefix, channel, name=None, **kwargs):
         self._channel = channel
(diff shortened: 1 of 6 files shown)


pcdsdevices/analog_signals.py
```diff
@@ -61,6 +61,56 @@ class Acromag(BaseInterface, Device):
...
+    """
+    channel = FCpt(EpicsSignalRO, '{self._prefix}:{self._channel}',
+                   kind='normal', doc='Analog Input [Volts]')
+
+    tab_component_names = True
+
+    def __init__(self, prefix, channel, **kwargs):
+        self._channel = channel
+        self._prefix = prefix
+        self._name = ''.join(['ai1_', self._channel])
```
ZryletTC, line 85: Will it always be 1? Same with output

pcdsdevices/analog_signals.py
```diff
@@ -61,6 +61,56 @@ class Acromag(BaseInterface, Device):
...
+    Parameters
+    ----------
+    prefix : str
+        The base EPICS PV for the Acromag. E.g.: `XPP:USR:ai1`
+    channel : str
+        The channel number. E.g.: `1` [0-15]
+    name : str
+        A name to refer to the Acromag Input Channel.
+    """
+    channel = FCpt(EpicsSignalRO, '{self._prefix}:{self._channel}',
```
ZryletTC, line 77: I'm wondering about the possibility of the class inheriting from EpicsSignal instead of having one as a component. I guess that would make the class a `Signal`, not a `Device`. Does that hurt it?

I'm also still not fully set on these classes being necessary at all, since they really only contain a single Signal.

If that's not the right path, could we instead have methods for setting and getting the voltage on the device level so we can do `ao1.put()` instead of `ao1.channel.put()`. It looks like a small change but I can already hear the scientists unhappy about having to type the extra word.

tests/test_analog_signals.py
```diff
@@ -64,6 +83,11 @@ def test_acromag_readback(fake_acromag):
     assert fake_acromag.ai1_13.get() == 1.0
 
 
+def test_acromag_ch_readback(fake_acromag_in, fake_acromag_out):
+    assert fake_acromag_in.channel.get() == 0.23
+    assert fake_acromag_out.channel.get() == 0.33
```
ZryletTC, line 88: Probably also good to try a `.put()` on the output channel and test that it changed.

tests/test_analog_signals.py
```diff
@@ -64,6 +83,11 @@ def test_acromag_readback(fake_acromag):
     assert fake_acromag.ai1_13.get() == 1.0
 
 
+def test_acromag_ch_readback(fake_acromag_in, fake_acromag_out):
+    assert fake_acromag_in.channel.get() == 0.23
```
ZryletTC, line 87: Do you need an is_close or smth to deal with the float equivalence? I'm not sure

pcdsdevices/analog_signals.py
ZryletTC: Also, maybe it could be better to accept a name in the args

ZryletTC (commented):
Couple notes. It feels like I haven't worked with this sort of code in a good while so take everything I commented with a grain of salt.

pcdsdevices/analog_signals.py
cristinasewell: Yeah this is a great point - thank you, I don't think I need to figure out the name here at all, I'm just going to give it a default name when initializing it.  👍

tests/test_analog_signals.py
cristinasewell: I am assuming I should not need an `is_close` or something of this sort, because i'm testing it against the exact value that I put it, but I could be wrong here

ZryletTC: Right, I guess there is no calculation going on, so you should me good 👍

pcdsdevices/analog_signals.py
cristinasewell: I am not sure here how I would do it with inheriting from EpicsSignal... but I guess an alternative is to have a factory method like you guys mentioned before, instead of having two devices here, but this will be a bit more tricky as I need to differentiate from an Input/Output channel, and also I need to create a device with only One channel instead of all of them..... my understanding was that the scientist won't use these two devices outside the questionnaire? 

Need to think about this more though... thank you for the suggestions 😊

klauer: A factory function would be the alternative here. We briefly had chatted about the options here and I seem to recall we landed on "Device", but maybe I'm misremembering.

I like that we can use `FCpt` here - that's the nicest part of the Device option.

cristinasewell: Yeah I thought we landed on 'Device' too... but I also thought that I could have misinterpreted it hahah, so it's not bad to revise this and re-think about the other options pros&cons 😊 ...
I'll leave it as is for now then. 
Also, I can do a `get()` on the `a` instead of `a.channel.get()` - but the output does not look as pretty, i'm also assuming i can do an `a.put()` but won't try that. - 
```
In [2]: a = AcromagChannelInput(prefix='XPP:USR:ai1', channel='1')

In [3]: a
Out[3]:

ai
--
channel: -0.008544921875 [Volts]

In [4]: a.get()
Out[4]: AcromagChannelInputTuple(channel=-0.0079345703125)

In [5]: a.channel.get()
Out[5]: -0.008544921875

```
@ZryletTC did you mean that I should have something like this:??
```
def get_voltage(self):
    self.channel.get()

def set_voltage(self, voltage):
    self.channel.put(voltage)
```

tests/test_analog_signals.py
cristinasewell: Added one in 1381498  - thank you

pcdsdevices/analog_signals.py
cristinasewell: changed it to `ai_` and `ao_`   in 1381498 - hopefully this is fine

pcdsdevices/analog_signals.py
silkenelson: Not sure where best to comment: 
In the general beamline python, I will likely have an object for my acromag with my AI card and its channels and similar for the AI and DI/DO,... cards. I use this to e.g. define LaserShuttters of other devices that take AO channels in input which are permanently installed. For experiment setup that changes (where requests are put in through the questionnaire), I have two things happening: for an AI channels, I might need to read the value in some function that would write for a given experiment (i.e. wait until signal reaches XXX level, then runs the sequencer). For an AO channel, I might want to write to that AO to e.g start a device, close a shutter,... in some script combined with other stuff or just by itself. 
Currently, entering channels in he questions creates a while acromag object (or at least the whole card) which is definitely not what I want. Maybe all that should happen is to link/alias the corresponding channel on my beamline acromag to the name that is defined in the questionnaire, so that e.g. 'LED.put(5)' will set xpp.acromag.ao1.channelXXX to 5 V (note: I obviously have not looked up what exactly the my acromag is called)

cristinasewell: Ok maybe i need to do a bit more digging here because i'm not sure anymore if i understood correctly the requirements for this issue. 
So before when loading the devices form the experiment we'd get something like this: 
```
 'acromag':
acromag
-------
ao1_0: None
ao1_1: None
ao1_2: None
ao1_3: None
ao1_4: None
ao1_5: None
ao1_6: None
ao1_7: None
ao1_8: None
ao1_9: None
ao1_10: None
ao1_11: None
ao1_12: None
ao1_13: None
ao1_14: None
ao1_15: None
ai1_0: None
ai1_1: None
ai1_2: None
ai1_3: None
ai1_4: None
ai1_5: None
ai1_6: None
ai1_7: None
ai1_8: None
ai1_9: None
ai1_10: None
ai1_11: None
ai1_12: None
ai1_13: None
ai1_14: None
ai1_15: None}
```
and now we're getting something like this:
```
 'ao1_6':
ao_6
----
channel: 0.0 [Volts]}
```

Where from the questionnaire run details we had valid info only for `ao_1` - 
```
 'pcdssetup-ao-6-device': '',
 'pcdssetup-ao-6-name': '',
 'pcdssetup-ao-6-pvbase': '',
 'pcdssetup-ao-1-pvbase': 'XPP:USR:ao1',
 'pcdssetup-ao-4-purpose': '',
 'pcdssetup-ao-5-purpose': '',
 'pcdssetup-ao-5-device': '',
 'pcdssetup-ao-4-device': '',
 'pcdssetup-ao-5-name': '',
 'pcdssetup-ao-4-name': '',
 'pcdssetup-ao-4-pvbase': '',
 'pcdssetup-ao-5-pvbase': '',
 'pcdssetup-ao-1-purpose': 'Evo Shutter 1',
 'pcdssetup-ao-1-channel': '6',
 'pcdssetup-ao-2-channel': '7',
 'pcdssetup-ao-3-channel': '9',
 'pcdssetup-ao-3-purpose': '',
 'pcdssetup-ao-2-purpose': '',
 'pcdssetup-ao-2-device': '',
 'pcdssetup-ao-3-device': '',
 'pcdssetup-ao-2-name': '',
 'pcdssetup-ao-3-name': '',
 'pcdssetup-ao-3-pvbase': '',
 'pcdssetup-ao-2-pvbase': '',
 'pcdssetup-ao-1-name': 'acromag',
```
Am I on the wrong path here?

klauer: @silkenelson I believe is saying she doesn't want the entire Acromag device, but the individual channels as you are working on. So, I think you're overall on the right path here, @cristinasewell.

As to whether or not those individual signals are borrowed from the full `Acromag` device or are created one-by-one (through either a factory function or Device): well, these are implementation details where we have some wiggle room as long as we deliver what the scientists want.

silkenelson:
Yes, indeed: you are on the right path: in the end I want an object with the name that I'd given it in the questionnaire that I can put, set and get as applicable - which is what you have. Like Tyler,  also feel that whatever I get from the questioinnaire should not require 'drilling down', meaning I don't want <object>.channel.get but instead <object>.get. etc
How exactly that is implemented (link or creations of a new object) does not matter as much. The acromag does not have a whole lot of common PVs that I might want access to (I'm not aware of any)

pcdsdevices/analog_signals.py
ZryletTC: Hey Cristina, yeah, you're definitely on the right path, and the yes those functions will be good if you just add a return.

As for the other implementation details, I was just throwing out ideas. I think it'd be cool if we could have a single acromag device and make the questionnaire borrow signals from it but if that's too much work, this way works just as well.

One note if we do leave it this way with channel devices, we might want to use those inside the Acromag class, instead of the plain signals.

pcdsdevices/analog_signals.py
```diff
@@ -61,6 +61,56 @@ class Acromag(BaseInterface, Device):
...
+                   kind='normal', doc='Analog Input [Volts]')
+
+    tab_component_names = True
+
+    def __init__(self, prefix, channel, **kwargs):
+        self._channel = channel
+        self._prefix = prefix
+        self._name = ''.join(['ai_', self._channel])
+        super().__init__(prefix, name=self._name)
```
ZryletTC, line 86:
```suggestion
    def __init__(self, prefix, channel, name=None, **kwargs):
        self._channel = channel
        self._prefix = prefix
        self._name = name or ''.join(['ai_', self._channel])
        super().__init__(prefix, name=self._name)
```

pcdsdevices/analog_signals.py
ZryletTC: Yeah that should be good. I've just added a new suggestion for your consideration, adding the possibility to specify a name instead of taking the default, allowing the user to name it according to the channel's purpose.

pcdsdevices/analog_signals.py
cristinasewell: Thanks @ZryletTC. 
One thing here is that the way I create the Acromag Channels from the questionnaire differs a bit from the actual Acromag class. 
For the Acromag Channels from the questionnaire I have the prefix as: `XPP:USR:ao1` and then I just add the channel... something like this: `AcromagChannelOutput(prefix='XPP:USR:ao1', channel='6')`.
And currently when we create an Acromag device we would do something like this: `Acromag(prefix='XPP:USR')` - so different `prefix` expected.
Using these guys inside the `Acromag` class might require some changes to either the way we create an Acromag device or Acromag Channels - I don't think we want to change the way we create the Acromag device, correct?

ZryletTC: You're right that we don't want to change the way we create Acromag devices, but the way an Acromag defines its channel components should be a pretty simple change. I believe it should just be the following:
from
```ao1_0 = Cpt(EpicsSignal, ":ao1:0", kind='normal')```
to
```ao1_0 = Cpt(AcromagChannelOutput, ":ao1", channel=0, kind='normal')```

Comment (cristinasewell) [reply pcdsdevices/analog_signals.py:77]: Ohh I see :D - thank you

Comment (ZryletTC) [reply pcdsdevices/analog_signals.py:77]: One downside to note is that if we do this, the API for the class will be slightly different. That can be curtailed by overriding the get and put methods but still kinda eh 😕 
@klauer or @ZLLentz, what do you think?

Comment (klauer) [reply pcdsdevices/analog_signals.py:77]: Given that this thread hasn't closed yet and we're still seemingly stuck on this being a `Device` - my take is that a factory function is really what's desired here, such that we get a simple `EpicsSignal` out of it instead of a `Device`. I personally think that's just fine.

Comment (cristinasewell) [reply pcdsdevices/analog_signals.py:77]: Ok, you guys know better, I'll go with whatever you think is best. 

So, should I have a factory function that will basically determine if it is an Input or Output base on the `prefix` - `XPP:USR:ai1` vs `XPP:USR:ao1`?

Comment (klauer) [reply pcdsdevices/analog_signals.py:77]: How about use the prefix-based assumption, but allow it to be overridden by an argument?
Perhaps

```python
def factory_func(prefix, *, is_input=None, ...):
    if is_input is None:
        is_input = ':ai' in prefix
```

or

```python
def factory_func(prefix, *, signal_class=None, ...):
    if signal_class is None:
        signal_class = EpicsSignalRO if ':ai' in prefix else EpicsSignal
```

cristinasewell: ok sounds good, thank you @klauer

pcdsdevices/analog_signals.py
```diff
@@ -61,6 +61,33 @@ class Acromag(BaseInterface, Device):
...
+    name : str, optional
+        A name to refer to the Acromag Channel, if `None` it will use the
+        prefix and channel to generate a name. E.g.: `ai_1` or `ao_1`.
+    """
+    if signal_class is None:
+        signal_class = EpicsSignalRO if ':ai' in prefix else EpicsSignal
+    name_prefix = 'ai_' if ':ai' in prefix else 'ao_'
+    name = name or ''.join([name_prefix, str(channel)])
+    prefix = ''.join([prefix, ':', str(channel)])
+    return signal_class(prefix, name=name, kind='normal')
```
cristinasewell, line 88: Is this kind of what we want?

klauer: Looks good, I think! You might consider f-strings for the `join` calls, but you don't need to.

pcdsdevices/analog_signals.py
```diff
@@ -61,6 +61,33 @@ class Acromag(BaseInterface, Device):
     tab_component_names = True
 
 
+def acromag_factory_func(prefix, channel, *, signal_class=None, name=None,
```
ZLLentz, line 64: I think the code looks great. If Travis was ready I'd probably just hit merge.

I'm singling out this line to discuss opinions on our naming for factory functions. All of other other factory functions are pretending to be classes (e.g. `def Attenuator():`, `def Motor():`, this one is very honest about what it is doing. I like an honest function, I'm wondering if we want to take the lead from this kind of naming scheme and apply it elsewhere in the library in another (low priority) PR? Or do people like the magical classnamed factory functions? To be clear, I also think there is a lot to like about a factory function named, say, `AcromagChannel`

cristinasewell: Yeah great point, I went back and forth with the naming on this one, and I also took it out from the `device_types` as well - but maybe I should add it back in? (in happi qs_db for `device_class` for acromag entry creation, I just give directly the `pcdsdevices.analog_signal.acromag_factory_func` - is this ok?).
I can go back to using the `AcromagChannel` for this function too :) - might make it clearer?

ZryletTC: I personally like the "pretend to be class" format instead because, to the user, it essentially is a class. To be more honest, the docstring should be very clear about what it actually is like this factory function's is. I vote for `AcromagChannel`

pcdsdevices/analog_signals.py
ZryletTC: Resolved in 9290102

pcdsdevices/analog_signals.py
cristinasewell: Ok should I switch it back to `AcromagChannel` - and should I also import it to the `device_types`?

ZryletTC: Since I expect it'll be used pretty commonly, yes on the `device_types`. Please also add a line in `docs/source/epics_types.rst` so that it gets included on that docs page.

As for the name change, please don't do it just because I said so. Let's get at least @klauer's opinion first, maybe @silkenelson's too

silkenelson: I would expect an "Acromag" to be the whole system that can contain an ao card, an ai card, ...with the respective channels. If we are talking a single channel, I'd like to have "Channel" in the name (or "ch", "chn" ...)

ZryletTC: That's a good point. If we do go with the function format, this should be changed (although that is so long ☹️):
```suggestion
def acromag_channel_factory_func(prefix, channel, *, signal_class=None, name=None,
```
@silkenelson, what is your opinion on function versus class format for naming factory functions? `acromag_channel_factory_func` vs `AcromagChannel`

klauer: The factory function name is obvious in its intent, a big point in its favor.
The class-like behavior of `AcromagChannel` is what a user might expect. This breaks down when you do `isinstance(obj, AcromagChannel)` (or expect any other class-like behavior).

For the above reasons, I'd tend toward the style-guide-approved, lowercase, underscored name. There are alternatives, as @ZLLentz did for `PytmcSignal` - but I don't believe adding that complexity is justified here.

cristinasewell: seems like we have some mixed opinions here...

ZLLentz: I agree that we seem to be in disagreement!
Let's double-dip for now using an `AcromagChannel = acromag_factory_function` alias maybe?

ZryletTC: I'm fine with going function format. I'm also happy to have the alias in there at least for now.

ZLLentz:
I'm going to edit your description because Github is pretty sure that this closes the linked issue, and will close it for us automatically even though there are a few PRs involved. "Trying to tackle and close" is being interpreted as "will close".

ZLLentz (approved):

ZLLentz:
I'm going to close and re-open this PR to re-seed Travis

ZLLentz:
This has passed and I think it is ready to merge

PR #698, opened 2020-11-10
ENH: custom motor status print

### LINKED ISSUE #670: Custom Status Print for Motors
## Expected Behavior
```
In[1]: ipm2.target.motor
out[1]: 

IPM2 Target Y:     XPP:SB2:MMS:03
current position (user, dial): 30.23, -30.23 [mm]
user limits (low, high): -20.00, 300.00 [mm]
preset position: xxxxxx
```
- Also add a limits hit indicator

## Context

cristinasewell:
- Added a custom status print for motors by overriding the status info handler.
- Added a new component for `dial_position`
- Added a new method `check_limit_switches` to check for the limit switches.

## Motivation and Context

Trying to implement and close #670 
## How Has This Been Tested?

```
In [2]: ipm = Motor(prefix='XPP:SB2:MMS:03', name='name')

In [3]: ipm
Out[3]:
IPM2 Target Y: XPP:SB2:MMS:03
Current position (user, dial): 0.0, -0.0 [mm]
User limits (low, high): -200.0, 200.0 [mm]
Preset position: Unknown
Limit Switch: Low [] High []

```

Docstring

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -56,14 +56,41 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
     disabled = Cpt(EpicsSignal, ".DISP", kind='omitted')
     # Description is valuable
     description = Cpt(EpicsSignal, '.DESC', kind='normal')
+    # Current Dial position
+    dial_position = Cpt(EpicsSignalRO, '.DRBV', kind='normal')
 
     tab_whitelist = ["set_current_position", "home", "velocity",
-                     "enable", "disable"]
+                     "enable", "disable", "check_limit_switches"]
 
     def __init__(self, *args, **kwargs):
         # Locally defined limit overrides, note can only make limits narrower
         self._limits = (-math.inf, math.inf)
         super().__init__(*args, **kwargs)
+
+    def format_status_info(self, status_info):
+        """Override status info handler to render the motor status."""
+        lines = []
+        # in case a .DESC value is missing, use the prefix as name
+        name = ' '.join(self.prefix.split(':'))
+        name = f'{name}: {self.prefix}'
+        description = status_info['description']['value']
+        if description:
+            name = f'{description}: {self.prefix}'
+        units = status_info['user_setpoint']['units']
+        dial = status_info['dial_position']['value']
+        user = status_info['position']
+        low, high = self.limits
+        switch_limits = self.check_limit_switches()[0]
+        position = f'current position (user, dial): {user}, {dial} [{units}]'
+        limits = f'user limits (low, hight): {low}, {high} [{units}]'
+        preset = f'preset position: {self.presets.name}'
+        switch = f'Limit Switch: {switch_limits}'
+        lines.append(name)
+        lines.append(position)
+        lines.append(limits)
+        lines.append(preset)
+        lines.append(switch)
+        return '\n'.join(lines)
 
     @property
     def low_limit(self):
@@ -158,6 +185,23 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
         if self.disabled.get() == 1:
             raise MotorDisabledError("Motor is not enabled. Motion requests "
                                      "ignored")
+
+    def check_limit_switches(self):
+        """Check the limits switches."""
+        if self.low_limit_switch.get():
+            return (
+                "low",
+                "low limit switch for motor %s (pv %s) activated"
+                % (self.name, self.pvname),
+            )
+        elif self.high_limit_switch.get():
+            return (
+                "high",
+                "high limit switch for motor %s (pv %s) activated"
+                % (self.name, self.pvname),
+            )
+        else:
+            return ("ok", "")
 
 
 class PCDSMotorBase(EpicsMotorInterface):
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -22,6 +22,7 @@ from .interface import FltMvInterface
 from .pseudopos import DelayBase
 from .signal import PytmcSignal
 from .variety import set_metadata
+from .utils import get_status_value
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


cristinasewell:
This is no way near to the finished result, but I wanted to get some feedback if possible about a couple things:

- [ ] is it in the right place
- [ ] is it the right approach
- [ ] what exactly is `limit hit indicator` - I brought back an old function of `check_limit_switches` - is this kind of what it means? Also, do we eventually want to print that out as well as part of the status?

ZryletTC:
I think this is looking pretty good. Approach-wise, I think that's probably right, but @ZLLentz would know better since he wrote it.

I think the limit indicator could be a bit more concise, maybe along the lines of the following?
```
Limit Switches: High [x]  Low [ ]
```
It should probably be a bit louder somehow though... I really wish they actually gave us an example of how they wanted it :/

pcdsdevices/epics_motor.py
```diff
@@ -159,6 +204,23 @@ def check_value(self, value):
             raise MotorDisabledError("Motor is not enabled. Motion requests "
                                      "ignored")
 
+    def check_limit_switches(self):
```
ZLLentz, line 207: This function is missing the case where both limits are hit. This is nonphysical but if the IOC reports this we should know.

pcdsdevices/epics_motor.py
```diff
@@ -159,6 +204,23 @@ def check_value(self, value):
...
+                % (self.name, self.pvname),
+            )
+        elif self.high_limit_switch.get():
+            return (
+                "Low [] High [x]",
+                "high limit switch for motor %s (pv %s) activated"
+                % (self.name, self.pvname),
+            )
+        else:
+            return ("Low [] High []", "")
```
ZLLentz, line 222: This is a tuple rather than a string

pcdsdevices/epics_motor.py
```diff
@@ -159,6 +204,23 @@ def check_value(self, value):
             raise MotorDisabledError("Motor is not enabled. Motion requests "
                                      "ignored")
 
+    def check_limit_switches(self):
+        """Check the limits switches."""
```
ZLLentz, line 208: This could use a quick doc on what the return type is

pcdsdevices/epics_motor.py
```diff
@@ -56,15 +56,60 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
...
+            name = f'{description}: {self.prefix}'
+        except KeyError:
+            name = ' '.join(self.prefix.split(':'))
+            name = f'{name}: {self.prefix}'
+        try:
+            units = status_info['user_setpoint']['units']
+            dial = status_info['dial_position']['value']
+            user = status_info['position']
+        except KeyError as err:
+            logger.error('Key error %s', err)
```
ZLLentz, line 101: We probably need to do something even if these fail- as written, the printout will just be the error message, rather than a best-effort partial printout (and then we get a new bug report)

pcdsdevices/epics_motor.py
```diff
@@ -56,15 +56,60 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
...
+        else:
+            low, high = self.limits
+            switch_limits = self.check_limit_switches()[0]
+            position = (f'Current position (user, dial): {user}, {dial}'
+                        f' [{units}]')
+            limits = f'User limits (low, hight): {low}, {high} [{units}]'
```
ZLLentz, line 107:
```suggestion
            limits = f'User limits (low, high): {low}, {high} [{units}]'
```

ZLLentz:
I think the approach is right and it looks good overall. We may find we want to refactor a bit when we get to the `PseudoPositioner` printout (since they will be similar and should share some code, not sure how much).

pcdsdevices/epics_motor.py
cristinasewell: Yeah i see, maybe i should just use `.get()` here and will just print `None`  if can't get it?

cristinasewell: This might be misinforming though - we probably want to know when we could get the info and it is actually `None` vs. when we could not get the info...

ZLLentz: I think `'N/A'` is a good call as a placeholder- as you say, there is a difference between the result being `None` and the result failing.

cristinasewell: ok i'll go for that then, hopefully the users will not really see the `N/A` much

ZLLentz: I agree, it seems like it should be a really rare codepath- only if the internals totally break or if a subclass reimplements something in a strange way?

pcdsdevices/epics_motor.py
```diff
@@ -56,15 +56,58 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
     disabled = Cpt(EpicsSignal, ".DISP", kind='omitted')
     # Description is valuable
     description = Cpt(EpicsSignal, '.DESC', kind='normal')
+    # Current Dial position
+    dial_position = Cpt(EpicsSignalRO, '.DRBV', kind='normal')
```
cristinasewell, line 60: I think this is the `dial` position that scientists wanted, if anybody could confirm - that would be great.

silkenelson: Yes, that is what the scientists would call dial positions.

cristinasewell: great, thank you @silkenelson

cristinasewell:
> I think the approach is right and it looks good overall. We may find we want to refactor a bit when we get to the `PseudoPositioner` printout (since they will be similar and should share some code, not sure how much).

Ok i'll keep that in mind, i'll get to that one soon - thank you

pcdsdevices/epics_motor.py
```diff
@@ -56,15 +56,57 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
...
+        name = ' '.join(self.prefix.split(':'))
+        name = f'{name}: {self.prefix}'
+        if description:
+            name = f'{description}: {self.prefix}'
+
+        return f"""\
+{name}
+Current position (user, dial): {user}, {dial} [{units}]
+User limits (low, high): {low}, {high} [{units}]
+Preset position: {self.presets.name}
```
ZLLentz, line 106: I think this isn't the right string for the preset position- I think it's `self.presets.state()`, referring to https://github.com/pcdshub/pcdsdevices/blob/5c1154159f3034a112b92044b1cbd44ba77c0e33/pcdsdevices/interface.py#L1187

cristinasewell: thank you @ZLLentz , this would make more sense than the `name` - I need to go back to all the PRs regarding status prints and make sure I'm printing the correct data

ZLLentz: I'm trying to check it closely during review: more eyes is better

ZLLentz (approved):
I think this is good to go- will merge shortly.

PR #701, opened 2020-11-11
ENH: PIM custom status

### LINKED ISSUE #674: Custom Status Print for PIM
## Expected Behavior
```
In[1]: yag2
Out[1]:
XPP SB3 PIM: OUT      
Y position: -50.00 [mm]
Navitar Zoom: 10.0 [%]   Focus: 0.02 [mm]
```

## Context

cristinasewell:
Override the status info handler for pim.
 - [x] need to check about the focus...

## Motivation and Context

Trying to implement and close #674 

## How Has This Been Tested?

Not sure if I'm testing this correctly:
```
In [3]: pim_f = PIMWithFocus(prefix='CXI:DG1:PIM', name='yag')

In [4]: pim = PIM(prefix='CXI:DG1:PIM', name='yag')

In [5]: pim_f
Out[5]:
CXI DG1 PIM: OUT
Y Position: -51.9995 [mm]
Navitar Zoom: 0.6180 [%]  Focus: ERROR [None]

In [6]: pim
Out[6]:
CXI DG1 PIM: OUT
Y Position: -51.9995 [mm]
Navitar Zoom: 0.6180 [%]
```

Docstring 

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/674-PIM_Custom_Status_Print.rst
+++ b/docs/source/upcoming_release_notes/674-PIM_Custom_Status_Print.rst
@@ -1,0 +1,30 @@
+674 PIM Custom Status Print
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- Added a custom status print for PIM by overriding the status info handler.
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- cristinasewell
--- a/pcdsdevices/pim.py
+++ b/pcdsdevices/pim.py
@@ -109,33 +109,31 @@ class PIM(BaseInterface, Device):
             Formatted string with all relevant status information.
         """
         lines = []
-        name = ' '.join(self.prefix.split(':'))
-        focus_str = ''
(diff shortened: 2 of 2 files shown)


ZLLentz:
This looks reasonable to me. What's the sticking point about the focus?

cristinasewell:
> This looks reasonable to me. What's the sticking point about the focus?

I assumed that we only get the focus and display it for the Pim with Focus - because the other ones did not have `focus` in their `status_info` so I though they don't have focus - but wanted to double check if that's the case, or if I should somehow get a `focus` from somewhere else.

ZLLentz:
Yes, that is correct! Some of these do not have focus. We have an unfortunate situation where the LCLS-I PIMs have a wide variety of flavors (with or without zoom, with or without focus...), so we don't need to display at all for the features that do not exist.

cristinasewell:
> Yes, that is correct! Some of these do not have focus. We have an unfortunate situation where the LCLS-I PIMs have a wide variety of flavors (with or without zoom, with or without focus...), so we don't need to display at all for the features that do not exist.

I should check if I have a `Zoom` then too, i'm currently assuming not checking that...

ZLLentz:
I think this is just missing the pre-release docs before merging, it looks good to me

ZLLentz (approved):

Issue #675, opened 2020-11-05
Add 3rd harmonic to Attenuator status print


ZLLentz:
## Expected Behavior

Include 3rd harmonic:
```
In [1]: att
Out[1]:
XCS Beamline Attenuator on SB2
filter # |0|1|2|3|4|5|6|7|8|9|
 OUT     |X|X|X|X|X|X|X|X|X|X|
 IN      | | | | | | | | | | |
Transmission for 1st harmonic (E=7.92 keV): 1.0000E+00
Transmission for 3rd harmonic (E=23.76 keV): 1.000e+00
```
This might require some reworking of the Att classes so that we don't have separate 1st and 3rd harmonic devices

## Context

https://docs.google.com/document/d/1dmalnT4xLPWoJESsMHTxzIyzlGCBgTQjGNYwDhU-pzo/edit#heading=h.ylcn31lb55xr

klauer:
I was looking to close this issue as the PR was merged, but I noticed this when double-checking:

https://github.com/pcdshub/pcdsdevices/blob/e5cf6f66c9db5d5d32bb4a654f76a74e17f3e5ab/pcdsdevices/attenuator.py#L320-L321

The same variable is used for both transmissions - that can't be right, can it?

cristinasewell:
I don't think this is right, thanks for catching this. 
I think I should be getting the `readback_3rd` value instead?

klauer:
That seems like it could be it, @cristinasewell... Does `caget (that_pv).DESC` tell us anything useful? Failing that, I'd imagine the old EDM screen has a spot for 3rd harmonic transmission that you could check.

cristinasewell:
Ok i'll check this later today and put in a fix for it, thanks @klauer

cristinasewell:
```
caget MFX:ATT:COM:R3_CUR.DESC
MFX:ATT:COM:R3_CUR.DESC        Floor Ratio
```

I think it is this one. Also based on an EDM screen: 
![att](https://user-images.githubusercontent.com/62306310/103740513-352c5880-4fac-11eb-9aaf-7d003539eb0a.PNG)

The third harmonic has this pv:
```
XPP:ATT:COM:R3_CUR
```
which is similar to the one above.

klauer:
Looks like the description is wrong (likely a copy-paste bug when making the EPICS database), but if the screen shows `R3_CUR` that'd be the one 👍

PR #702, opened 2020-11-13
ENH: attenuator 3rd harmonic print

cristinasewell:
Refactored the code a bit to allow creating devices with both fundamental frequency and 3rd harmonic frequency instead of creating separated classes for them as it was previously.
Add custom status printout for the Attenuator by overriding the `format_status_info` handler.

## Motivation and Context

Trying to implement #675
## How Has This Been Tested?

```
In [1]: mfx_attenuator
Out[1]:
filter # |0|1|2|3|4|5|6|7|8|9|
 OUT     | |X| | | |X|X|X|X|X|
 IN      |X| |X|X|X| | | | | |
Transmission for 1st harmonic (E=8.23e+06 keV): 2.3446E-02
Transmission for 3rd harmonic (E=2.47e+07 keV): 2.3446E-02
```

```
In [1]: at1l0
Out[1]:
filter # |0|1|2|3|4|5|6|7|8|
 OUT     |X|X|X|X|X|X|X|X|X|
 IN      | | | | | | | | | |
Transmission for 1st harmonic (E=8.23e+06 keV): 1.0000E+00
```
## Pre-merge checklist

### DIFF
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -19,6 +19,7 @@ from .inout import InOutPositioner, TwinCATInOutPositioner
 from .interface import BaseInterface, FltMvInterface, LightpathInOutMixin
 from .signal import InternalSignal
 from .variety import set_metadata
+from .utils import get_status_value
 
 logger = logging.getLogger(__name__)
 MAX_FILTERS = 12
@@ -105,25 +106,38 @@ class AttBase(FltMvInterface, PVPositioner):
     vary. You should not instantiate this class directly, but instead use the
     :func:`Attenuator` factory function.
     """
-
+    # fundamental frequency components
+    # ================================
     # Positioner Signals
     setpoint = Cpt(EpicsSignal, ':COM:R_DES', auto_monitor=True,
                    kind='normal')
     readback = Cpt(EpicsSignalRO, ':COM:R_CUR', auto_monitor=True,
                    kind='hinted')
-    actuate = Cpt(EpicsSignal, ':COM:GO', kind='omitted')
+    actuate = Cpt(EpicsSignal, ':COM:GO', kind='normal')
     done = Cpt(EpicsSignalRO, ':COM:STATUS', auto_monitor=True,
                kind='omitted')
 
     # Attenuator Signals
     energy = Cpt(EpicsSignalRO, ':COM:T_CALC.VALE', kind='normal')
-    trans_ceil = Cpt(EpicsSignalRO, ':COM:R_CEIL', kind='omitted')
-    trans_floor = Cpt(EpicsSignalRO, ':COM:R_FLOOR', kind='omitted')
-    user_energy = Cpt(EpicsSignal, ':COM:EDES', kind='omitted')
-    eget_cmd = Cpt(EpicsSignal, ':COM:EACT.SCAN', kind='omitted')
+    trans_ceil = Cpt(EpicsSignalRO, ':COM:R_CEIL', kind='normal')
+    trans_floor = Cpt(EpicsSignalRO, ':COM:R_FLOOR', kind='normal')
+    user_energy = Cpt(EpicsSignal, ':COM:EDES', kind='normal')
+    eget_cmd = Cpt(EpicsSignal, ':COM:EACT.SCAN', kind='normal')
 
     # Aux Signals
-    calcpend = Cpt(EpicsSignalRO, ':COM:CALCP', kind='omitted')
+    calcpend = Cpt(EpicsSignalRO, ':COM:CALCP', kind='normal')
+
+    # 3rd harmonic frequency components
+    # ==================================
+    # Positioner Signals
+    setpoint_3rd = Cpt(EpicsSignal, ':COM:R3_DES', kind='normal')
+    readback_3rd = Cpt(EpicsSignalRO, ':COM:R3_CUR', kind='hinted')
+
+    # Attenuator Signals
+    energy_3rd = Cpt(EpicsSignalRO, ':COM:T_CALC.VALH', kind='normal')
+    trans_ceil_3rd = Cpt(EpicsSignalRO, ':COM:R3_CEIL', kind='omitted')
+    trans_floor_3rd = Cpt(EpicsSignalRO, ':COM:R3_FLOOR', kind='omitted')
+    user_energy_3rd = Cpt(EpicsSignal, ':COM:E3DES', kind='omitted')
 
     egu = ''  # Transmission is a unitless ratio
     done_value = 0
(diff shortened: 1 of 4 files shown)


pcdsdevices/attenuator.py
```diff
@@ -105,25 +106,38 @@ class AttBase(FltMvInterface, PVPositioner):
...
     """
-
+    # fundamental frequency components
+    # ================================
     # Positioner Signals
     setpoint = Cpt(EpicsSignal, ':COM:R_DES', auto_monitor=True,
                    kind='normal')
     readback = Cpt(EpicsSignalRO, ':COM:R_CUR', auto_monitor=True,
                    kind='hinted')
-    actuate = Cpt(EpicsSignal, ':COM:GO', kind='omitted')
```
cristinasewell, line 114: I'm going to change these back to `omitted` - just noticed that I left them here, but this was just for testing

pcdsdevices/attenuator.py
```diff
@@ -125,6 +127,18 @@ class AttBase(FltMvInterface, PVPositioner):
     # Aux Signals
     calcpend = Cpt(EpicsSignalRO, ':COM:CALCP', kind='omitted')
 
+    # 3rd harmonic frequency components
```
cristinasewell, line 130: Added the 3rd harmonic frequency components to the `AttBase` class  and gotten rid of AttBase3rd- is this kind of what you meant @ZLLentz  (***This might require some reworking of the Att classes so that we don't have separate 1st and 3rd harmonic devices***)

Or did I go a bit too far here with this approach?

ZLLentz: I think this approach makes a lot of sense, but you've actually added PVs to FeeAtt that don't exist- it doesn't have a 3rd harmonic calculator.

cristinasewell: Ohh noo - i see, thanks for pointing this out. I'm going to do some thinking here...

cristinasewell: I think I'm gonna have to go back to the original implementation of having two base classes here?, one for 3rd harmonics and one for the fundamental one?  
Or I can try to have `FeeAtt` not inherit from `AttBase` and implement its own thing... 
Or maybe other better suggestions?

ZLLentz: Maybe just add an `AttBaseWith3rd` intermediate class?

ZLLentz:
Updated PR name because when I saw this in my notifications I didn't know what it was

pcdsdevices/attenuator.py
cristinasewell: And have everybody inherit this class except `FeeAtt`? is that what you mean?

ZLLentz: Yes, that's what I'm saying.

The other approach would be to orphan the `FeeAtt` class onto its own island which is also fine- we're not revisiting that device ever (and in ops you're not "supposed" to use it)

pcdsdevices/attenuator.py
```diff
@@ -369,24 +391,22 @@ def _make_att_classes(max_filters, base, name):
...
+        cls = type(cls_name, (base, base_3rd_harmonic,), att_filters)
         cls.num_att = i
         att_classes[i] = cls
     return att_classes
 
 
-_att_classes = _make_att_classes(MAX_FILTERS, AttBase, 'Attenuator')
-_att3_classes = _make_att_classes(MAX_FILTERS, AttBase3rd, 'Attenuator3rd')
+_att_classes = _make_att_classes(
+    MAX_FILTERS, AttBase, AttBaseWith3rdHarmonic, 'Attenuator')
```
cristinasewell, line 401: I'm still not sure if this is safe - this is assuming that users will not create a Fee attenuator using this function, but I think that's the case anyway?

ZLLentz: The FEE attenuator is a one-off device.

This factory function is used because the LusiAtt devices can come in any number of flavors based on the number of silicon filters in the device. E.g. it's me being lazy and not wanting to hand-make a bunch of classes.

There shouldn't be any danger in doing this.

tests/test_attenuator.py
```diff
@@ -156,7 +153,7 @@ def test_attenuator_staging(fake_att):
 
 def test_attenuator_third_harmonic():
     logger.debug('test_attenuator_third_harmonic')
-    att = Attenuator('TRD:ATT', MAX_FILTERS-1, name='third', use_3rd=True)
+    att = Attenuator('TRD:ATT', MAX_FILTERS-1, name='third')
     att.wait_for_connection()
```
ZLLentz, line 157: I think this last test can just be removed- it no longer serves any purpose with the 3rd harmonic attenuator positioners gone

pcdsdevices/attenuator.py
```diff
@@ -291,37 +304,46 @@ def format_status_info(self, status_info):
...
+"""
+        else:
+            return f"""\
+{states}
+Transmission for 1st harmonic (E={energy:.3} keV): {trans:.4E}
+"""
 
-    _harmonic = '3rd'
 
+class AttBaseWith3rdHarmonic(FltMvInterface, PVPositioner):
```
ZLLentz, line 330: I don't think it causes a problem (so it doesn't need to be fixed), but this class doesn't really function as a `PVPositioner` on its own. I'd have preferred to either subclass `AttBase` only (e.g. extend the base class) or subclass `ophyd.Device` only (e.g. just define some PVs to mix in later).

ZLLentz (approved):
Can't argue with results, looks good

pcdsdevices/attenuator.py
cristinasewell: Ohh so what I can do is basically:
```
AttBaseAndWith3rdHarmonic(AttBase):
...
```
and in the `_make_att_class()` I would need to only import use this class instead of using both 

.. i think this is what you're saying, i'm going to change that then cause i like that much better, thank you

tests/test_attenuator.py
cristinasewell: Did not remove this but updated it to just test the Attenuator in general (0bb3c0b )

pcdsdevices/attenuator.py
cristinasewell: Changed this in b56352c - I know you said it does not need to be fixed, so let me know if I should go back, but I thought it was a good idea...

tests/test_attenuator.py
ZLLentz: fair enough!

pcdsdevices/attenuator.py
ZLLentz: looks :+1: to me

ZLLentz:
I think this is ready to be merged

Issue #597, opened 2020-09-30
motor soft limits consistency between python and epics


mseaberg:
There seem to be some inconsistencies in how hutch-python handles soft limits between python and epics. Changing the attributes "low_limit" or "high_limit" does not change any epics pv's, meaning that if the motors are run from epics these limits would have no effect. In contrast, putting or setting low_limit_travel for instance changes the epics soft limit pv, which in my opinion is the desired behavior. I have found that it is easy to end up in an inconsistent state if one changes the "low_limit" or "high_limit" attributes directly and it seems to me that these should be protected perhaps with an underscore.

## Expected Behavior

Soft limits should be consistent between epics and python.

## Current Behavior

It is possible to end up in an inconsistent state by setting python attributes which have no effect on epics pvs.

## Possible Solution

Implement a python wrapper for "low_limit_travel.put(lim)" similar to old hutch-python "set_lowlim" and "set_hilim" methods, that simply put to the soft limit epics pv's.

## Your Environment

xpp3 from xppopr@xpp-control

ZLLentz:
It's set up this way because it introduces performance penalties when an attribute access is directly connected to a PV, since most python programs assume that setattr/getattr are fast operations and interacting with a PV can time out.

I think we definitely add `set_lowlim` and `set_hilim` shortcuts here.

ZLLentz:
If you think the `high_limit` and `low_limit` attributes are traps, we can change their functionality as well.

mseaberg:
I have no problem with the python attributes, just that the implementation for setting soft limits in python needs some work. When I asked about setting soft limits in python, someone suggested that I just set the high_limit and low_limit attributes directly, which I found was causing contradictions with epics soft limits. I never had any issues with the low_limit and high_limit attributes after I only made changes using low_limit_travel, etc. This is why I think these should be more "protected" since I don't think the user should be setting them from a hutch-python session.

ZLLentz:
Understood. If you think we shouldn't see them in hutch python then we can hide them accordingly.

silkenelson:
I think I need a slow-people explanation here: what I think I should see in the motor python object is if you have hit a physical limit switch. Nothing to set here. Then we have the soft limits, that I would have expected to only be defined in the IOC as the motor record already has soft limits. If there is another set of soft limits, I'd like to understand better why we have those in addition to the soft limits in the motor record. How these are settable I have no opinion on.

ZLLentz:
[...]

PR #707, opened 2020-11-19
MNT: add set_low_lim and set_hi_lim to EpicsMotor

cristinasewell:
## Description
- Added two methods to the `epics_motors`: `set_high_limit()` and `set_low_limit()`, as well as `get_low_limit()` and `get_high_limit()`.
- Added a little method to clear limits: `clear_limits` - this sets both limits to 0.
- Removed the other limit setters and getters `low_limit`, `high_limit` `limits`
- Overrode the `limits` property

## Motivation and Context

Trying to solve the problem described in #597
## How Has This Been Tested?

Locally only.

Docstring

## Pre-merge checklist


pcdsdevices/epics_motor.py
```diff
@@ -221,6 +221,48 @@ def check_limit_switches(self):
...
+    def set_low_lim(self, value):
+        """
+        Set the low limit.
+
+        Parameters
+        ----------
+        value : float
+            Limit of travel in the negative direction.
+        """
+        if not self.moving:
```
cristinasewell, line 237: Not sure if I need to check if if it is moving or not and if `<=` current position - this is just something the `ophyd` guys are using in `set_lim()`

ZLLentz: Ah that's interesting- I wonder what kinds of problems can happen from mid-move limit sets...
I'm a bit surprised that ophyd fails silently instead of raising an exception

ZLLentz:
I think adding these two functions was a good idea.

As for the limit attributes... I'm not really sure what to do with them. I'm not 100% sure we want the properties at all- I'd like more opinions on this.

My point of view is that properties that simply get EPICS values slows down attribute inspection tools and can cause things to fail in unexpected ways without adding much utility.

I think if we just removed them entirely, the parent class's `limits` attribute is fully sufficient.

pcdsdevices/epics_motor.py
```diff
@@ -221,6 +221,48 @@ def check_limit_switches(self):
...
+            # update EPICS
+            if value <= self.position:
+                self.low_limit_travel.put(value)
+            else:
+                logger.debug('Could not set motor low limit to %f at '
```
ZLLentz, line 242: ophyd devices typically use `self.log` for this sort of thing- I think we still have a bunch of dangling `logger` references elsewhere in the code in pcdsdevices

cristinasewell: ohh thank you, i'll change this

cristinasewell: fixed this in 3290db5

cristinasewell:
> I think adding these two functions was a good idea.
> 
> As for the limit attributes... I'm not really sure what to do with them. I'm not 100% sure we want the properties at all- I'd like more opinions on this.
> 
> My point of view is that properties that simply get EPICS values slows down attribute inspection tools and can cause things to fail in unexpected ways without adding much utility.
> 
> I think if we just removed them entirely, the parent class's `limits` attribute is fully sufficient.

I removed them in 3290db5 - but I am not sure if did not break something else by doing that.

ZLLentz:
I still want more opinions on this- but I think the current code is a very good scenario:
- There is one canonical way to check the limits via getattr, the `limits` attribute, and it runs without any EPICS actions
- There are the following ways to set the limits:
  - ophyd's built-in `set_lim` (sets both, but you need to have both values ready)
  - the new `set_low_lim` (does checks and puts to the signal)
  - the new `set_hi_lim` (does checks and puts to the signal)
- There are additional `get_low_lim` and `get_hi_lim` functions that, while not strictly necessary, are easy to find if you know about the set variants.

If we have non-EPICS motors we can follow the same API with little issue.

My only suggestions are:
- Failing to set the limits for any reason should be an exception raised or a log message, not a silent failure. I think ophyd does it wrong here in `set_lim`.
- There might need to be some handling for if the user puts a high limit that's lower than the low limit, or a low limit that's higher than the high limit
- Do we need a shortcut for removing the soft limits entirely? (Setting high and low to zero at the same time)

pcdsdevices/epics_motor.py
klauer: It's a recent addition - not entirely sure why it's there. I'm positive I've set limits when a motor was moving without issue. Perhaps just a race if you set a limit very close to the current position and the readback changes beyond that value? 🤷‍♂️

klauer:
Some thoughts:
* I guess the `_lim` naming is following ophyd's `set_lim` - I don't particularly like the naming but that's OK.
* Completely agreed with Zach: this needs to raise in all conditions in which this fails silently (moving or position outside of limit).
* Limits do need special handling when `requested_low_limit > current_high_limit` (and similar scenarios). Try it with a motor record to see for yourself.
* `clear_limits` seems a nice shortcut as Zach indicated

cristinasewell:
> Some thoughts:
> 
> * I guess the `_lim` naming is following ophyd's `set_lim` - I don't particularly like the naming but that's OK.
> * Completely agreed with Zach: this needs to raise in all conditions in which this fails silently (moving or position outside of limit).
> * Limits do need special handling when `requested_low_limit > current_high_limit` (and similar scenarios). Try it with a motor record to see for yourself.
> * `clear_limits` seems a nice shortcut as Zach indicated

Do you think I should change it to something like this: `set_low_limit` and `set_high_limit` instead @klauer ? or what exactly you don't like about the naming?

pcdsdevices/epics_motor.py
```diff
@@ -191,7 +155,8 @@ def check_value(self, value):
         # Find the soft limit values from EPICS records and check that this
         # command will be accepted by the motor
         if any(self.limits):
-            if not (self.low_limit <= value <= self.high_limit):
+            if not (self._get_epics_limits()[0] <= value
```
cristinasewell, line 158: I was not sure here if I should have changed this. The `low_limit` and `high_limit` work as well (at least in some situations), because these are inherited from the ophyd Component (I think?)

pcdsdevices/epics_motor.py
```diff
@@ -221,6 +186,85 @@ def check_limit_switches(self):
...
+            When motor in motion or position outside of limit.
+        """
+        if not self.moving:
+            # update EPICS
+            _current_high_limit = self._get_epics_limits()[1]
+            if value <= self.position:
+                if value < _current_high_limit:
+                    self.low_limit_travel.put(value)
+                else:
+                    raise ValueError('Could not set motor low limit to %f. '
```
cristinasewell, line 214: This might be a bit too much, I probably could have combined the if statements here but I though raising more specific errors is not a bad idea?

pcdsdevices/epics_motor.py
```diff
@@ -221,6 +186,85 @@ def check_limit_switches(self):
...
+            High limit value to be set.
+
+        Raises
+        ------
+        ValueError
+            When motor in motion or position outside of limit.
+        """
+        if not self.moving:
+            # update EPICS
+            _current_low_limit = self._get_epics_limits()[0]
```
cristinasewell, line 246: I'm using the `_get_epics_limits` here, because it will return `0` if not set - is this fine?

pcdsdevices/epics_motor.py
```diff
@@ -221,6 +186,85 @@ def check_limit_switches(self):
         else:
             return "Low [] High []"
 
+    def get_low_limit(self):
```
klauer, line 189: I prefer the new names for sure - but you didn't _have_ to change them :)
Please also update the tab whitelist with the new names.

pcdsdevices/epics_motor.py
```diff
@@ -221,6 +186,85 @@ def check_limit_switches(self):
...
+                    raise ValueError('Could not set motor low limit to %f. '
+                                     'Low limit must be lower than the '
+                                     'current high limit: %f',
+                                     value, _current_high_limit)
+            else:
+                raise ValueError('Could not set motor low limit to %f at '
+                                 'position %f. Low limit must be lower than '
+                                 'the current position.', value, self.position)
+        else:
+            raise ValueError('Could not set the low limit, the motor is in'
```
klauer, line 223: The control flow would be much more obvious if you inverted the `if not self.moving:` and bailed early.
That is, 
```python
if self.moving:
   raise ValueError('...')
# otherwise, do this:
```

pcdsdevices/epics_motor.py
```diff
@@ -221,6 +186,85 @@ def check_limit_switches(self):
...
+            if value <= self.position:
+                if value < _current_high_limit:
+                    self.low_limit_travel.put(value)
+                else:
+                    raise ValueError('Could not set motor low limit to %f. '
+                                     'Low limit must be lower than the '
+                                     'current high limit: %f',
+                                     value, _current_high_limit)
+            else:
+                raise ValueError('Could not set motor low limit to %f at '
```
klauer, line 219: Careful - uninterpolated `%f` in your `ValueError`. Should use f-strings here.

cristinasewell: Thank you, changed it to using f-strings in my ValueErrors.... b3a7c93

pcdsdevices/epics_motor.py
cristinasewell: Thanks - i have changed this in b3a7c93

pcdsdevices/epics_motor.py
klauer: I missed this when looking earlier. Descriptive errors are very helpful 👍

pcdsdevices/epics_motor.py
cristinasewell: updated them in b3a7c93 - thanks for pointing this out

pcdsdevices/epics_motor.py
klauer: Nice - so much cleaner, don't you think? 👍

cristinasewell: Yes of course, thank you 😊

pcdsdevices/epics_motor.py
ZLLentz: For reference on inherited `low_limit` and `high_limit` attributes: https://github.com/bluesky/ophyd/blob/c91317fb21506d6acc2900e392e667aeb2684abc/ophyd/positioner.py#L142-L148

Is the concern here just the case where limits are uninitialized or invalid? It may be easiest to just wrap the `_get_epics_limits` check into an override of the `limits` attribute in that case. Then all instances of `self.limits`, `self.low_limit`, `self.high_limit` work as expected.

pcdsdevices/epics_motor.py
```diff
@@ -221,6 +186,80 @@ def check_limit_switches(self):
...
+            raise ValueError(f'Could not set motor high limit to {value}. '
+                             'High limit must be higher than the current low '
+                             f'limit: {_current_low_limit}')
+        # update EPICS limits
+        self.high_limit_travel.put(value)
+
+    def clear_limits(self):
+        """Set both low and high limits to 0."""
+        self.set_low_limit(0)
+        self.set_high_limit(0)
```
ZLLentz, line 261: This could fail if the limits are something like (-100, -50) before the function call, since the low limit can't be zero if the high limit is -50 due to the checks above. Funnily enough, `self.set_lim(0, 0)` also fails for similar reasons.

Might be easiest just to do straight puts to the PVs.

cristinasewell: Ok i changed it in 175f3e1 using the ` self.high_limit_travel.put(0)` and `self.low_limit_travel.put(0)` - i think that's what you meant. 

Thank you

pcdsdevices/epics_motor.py
cristinasewell: Overrode the `limits` attribute in 175f3e1 - and changed to using `limits` in 44f9852

pcdsdevices/epics_motor.py
```diff
@@ -52,20 +51,18 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
...
     # Current Dial position
     dial_position = Cpt(EpicsSignalRO, '.DRBV', kind='normal')
 
-    tab_whitelist = ["set_current_position", "home", "velocity",
-                     "enable", "disable", "check_limit_switches"]
+    tab_whitelist = ["set_current_position", "home", "velocity", "enable",
+                     "disable", "check_limit_switches", "get_low_limit",
+                     "set_low_limit", "get_high_limit", "set_high_limit"]
 
     def __init__(self, *args, **kwargs):
```
klauer, line 65: Looks like `__init__` can be removed here?

klauer (approved):
This looks good and ready to go to me. 

User experience when testing this may require some changes down the line, but we'll deal with that when we get there.

pcdsdevices/epics_motor.py
cristinasewell: Did we only have this because of the `_limits`?

docs/source/upcoming_release_notes/597-Motor_Soft_Limits_Consistency.rst
```diff
@@ -0,0 +1,33 @@
...
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- Added two methods to the `epics_motors`: `set_high_limit()` and `set_low_limit()`, as well as `get_low_limit()` and `get_high_limit()`.
+- Added a little method to clear limits: `clear_limits` - this sets both limits to 0.
+- Removed the other limit setters and getters `low_limit`, `high_limit` `limits`
```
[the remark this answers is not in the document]
cristinasewell: Ok thank you, i'll add this change specification to the `API Changes` section then and remove this last line from here

pcdsdevices/epics_motor.py
cristinasewell: took it out in 54b6349

docs/source/upcoming_release_notes/597-Motor_Soft_Limits_Consistency.rst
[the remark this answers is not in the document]
cristinasewell: updated this a bit in 54b6349

ZLLentz:
I think this looks clean, I'm checking the merge conflict

PR #708, opened 2020-11-19
ENH: Pseudo custom status

### LINKED ISSUE #671: Custom Status Prints for PsuedoPositioners
## Expected Behavior
```
In[1]: ccm.E
Out[1]:
Virtual Motor CCM Photon Energy   XCS:MON …(user pV?)
current position (user, dial): 7.002, 7.002n [eV]
user limits (low, high): 5.000, 15.000  [eV]
preset position: xxxxxx


In[2]: m.lxt
Out[2]:
Virtual Motor Laser X-ray Timing   [LXT USER PV?]
current position (user, dial) 2.341 [ps], 4.321452 [ns]
```

## Context

cristinasewell:
Overrode the status handler for the `PseudoSingleInterface` and for the `LaserTiming` 
Added a `verbose_name` attribute for the `PseudoSingleInterface` component.
Added `verbose_name` property to `LaserTiming` 

## Motivation and Context

Trying to tackle and close #671

## How Has This Been Tested?

```
In [1]: xpp_ccm.calc.energy
Out[1]:
Virtual Motor CCM Photon Energy XPP:MON:MPZ:07A:energy
Current position (user, dial): 9.870e+00, 9.870e+00 []
User limits (low, high): 4, 25 []
Preset position: Unknown

```
```
In [1]: xcs_lxt_fast.delay
Out[1]:
Virtual Motor XCS:LAS:MMN:03:delay
Current position (user, dial): 3.262e-12, -3.263e-12 [None]
User limits (low, high): 0, 0 [None]
Preset position: Unknown

```

```
In [3]: lxt
Out[3]:
Virtual Motor Laser X-ray Timing LAS:FS11
Current position (user, dial): -5.597e-06 [s] 5.597e-06 [s]
```

Docstring

## Pre-merge checklist


pcdsdevices/ccm.py
```diff
@@ -74,7 +74,7 @@ class CCMCalc(FltMvInterface, PseudoPositioner):
     """
 
     energy = Cpt(PseudoSingleInterface, egu='keV', kind='hinted',
-                 limits=(4, 25))
+                 limits=(4, 25), verbose_name='CCM Photon Energy')
```
cristinasewell, line 77: Added a new verbose name attribute here - not sure if this is ok for now, if not, I can just use the doc (although I had some trouble using the `__doc__` because I was getting all the docstring for the class.)

pcdsdevices/pseudopos.py
```diff
@@ -40,6 +43,41 @@ def __init__(self, prefix='', parent=None, **kwargs):
...
+
+        low, high = self.limits
+        name = self.prefix
+        if self._verbose_name:
+            name = f'{self._verbose_name} {self.prefix}'
+
+        return f"""\
+Virtual Motor {name}
+Current position (user, dial): {position}, {dial} [{units}]
+User limits (low, high): {low}, {high} [{units}]
```
cristinasewell, line 78: Are the `units` the same for limits as for the current position? I assumed they are the same here

cristinasewell:
I think I have to override the status handler in multiple places here but I am not sure where. 
Probably  for `OMMotor` and `BadSlitPositionerBase`  too - they are also inheriting from `PVPositioner` - but that does not necessarily mean they are virtual, or does it?

pcdsdevices/ccm.py
ZLLentz: I think adding a new name attribute is very appropriate

pcdsdevices/pseudopos.py
ZLLentz: Yes, the limits should always have the same units as the current position

ZLLentz:
`VirtualMotor` is a name for something that exists in only the super old code- they are all called `PsuedoPositioner`s now instead, following `ophyd`'s naming.

pcdsdevices/lxe.py
```diff
@@ -272,6 +274,8 @@ class LaserTiming(FltMvInterface, PVPositioner):
     user_offset = Cpt(AttributeSignal, attr='_user_offset',
                       kind='normal',
                       doc='A Python-level user offset.')
+    # Current Dial position
+    dial_position = Cpt(EpicsSignalRO, '.DRBV', kind='normal')
```
ZLLentz, line 278: I'm trying to figure out what this should be- I don't think there's actually a `.DRBV` field on the LXT prefix.

pcdsdevices/pseudopos.py
```diff
@@ -31,7 +31,10 @@ class PseudoSingleInterface(FltMvInterface, PseudoSingle):
         notepad_metadata={'record': 'ai', 'default_value': 0.0},
     )
 
-    def __init__(self, prefix='', parent=None, **kwargs):
+    # Current Dial position
+    dial_position = Cpt(EpicsSignalRO, '.DRBV', kind='normal')
```
ZLLentz, line 35: I'm not positive that this one exists either

pcdsdevices/lxe.py
cristinasewell: That would make sense since I got `None` for that - thanks for pointing this out, i'm going to look at the old code and see what they used

pcdsdevices/pseudopos.py
cristinasewell: yeah, same as above probably - thank you

pcdsdevices/lxe.py
cristinasewell: I can't figure out what they're using, it seems like in the old code the had a function `wm_dial()` that they kept using for the dial position, but that one was still using the `.DRBV` field.... I am still looking into this one

pcdsdevices/pseudopos.py
cristinasewell: I can try using this but not sure if works for all caseses:
```
# Current Dial position
dial_position = FCpt(EpicsSignalRO, '{self.parent.prefix}.DRBV', kind='normal')
```

ZLLentz: Does this work for any cases?

ZLLentz:
Dial position probably means "the calculated value if we used the DRBV instead of the RBV"

So any solution that doesn't involve a calculation step is probably not correct, since the reason these PsuedoPositioner classes exist is to transform some measured RBV into some calculated RBV

ZLLentz:
What is probably needed is to send the dial position of the "real" motor in the pseudopositioner through the calculation

ZLLentz:
Based on the ophyd docstrings, `PseudoPositioner.inverse` is the function that transforms the real position into the calculated one

pcdsdevices/pseudopos.py
cristinasewell: Works for this guy: xcs_lxt_fast.delay - but i am not sure if this is a good device to test against

ZLLentz: I don't think this is correct for that example- the value you're getting is in `mm`

ZLLentz:
And then `PseudoPositioner.real_positioners` can give you all the real motors in use- there's usually only one

pcdsdevices/pseudopos.py
ZLLentz: I see that it is the correct pre-calculation value- but I don't think we can assume that this is true in all cases

cristinasewell: Ohh gosh, my brain stops working on Fridays I think - sorry, this line should have been taken out, and only this one left:
```
# Current Dial position
    dial_position = FCpt(EpicsSignalRO, '{self.parent.prefix}.DRBV',
                         kind='normal')
```

cristinasewell: but regardless, i need to try your suggestions below

cristinasewell:
Would this work the same for the `PseudoSingleInterface` as well?

ZLLentz:
`PseudoSingleInterface` would have refer to `self.parent` to find these methods/attributes- but everything still applies

cristinasewell:
I am still a bit confused here, not sure what to do. 

The `position` seems to be the calculated one already I think:
https://github.com/bluesky/ophyd/blob/c91317fb21506d6acc2900e392e667aeb2684abc/ophyd/pseudopos.py#L654-L662

 - does this mean that this is actually the `dial position` then, and we can calculate the user position based on the dial maybe (`userVAL = DialVAL * DIR + OFFset`) ?

cristinasewell:
Also one thing that I noticed when testing with `LaserTimingCompensation` is that even if the names are changed to `lxt`, `txt` when calling those they don't work in an ipython shell, maybe I have an older version though - or maybe we did that on purpose for some reason I don't know, but i thought i'd mention it here.

```
In [3]: from pcdsdevices.lxe import LaserTimingCompensation as ltc
...
In [9]: m = ltc(prefix='XPP:USER:LXTTC', name='lxttc', delay_prefix='XPP:LAS:MMN:04', laser_prefix='XPP:USER')
...
In [17]: m
Out[17]:

lxttc
-----
position: -2.999968731701713e-12

pseudo
------
  position: -2.999968731701713e-12
  notepad_setpoint: 0.0
  notepad_readback: 0.0

txt
---
  position: -2.999968731701713e-12

  lxttc_delay_delay
  -----------------
    position: -2.999968731701713e-12
    notepad_setpoint: 3e-12
    notepad_readback: 3.0000020881112317e-12
  lxttc_delay_user_offset: 0.0

  lxttc_delay_motor
  -----------------
    position: 0.44968399999999953
    user_setpoint: 0.44968368700000017
    description: las_delay_new
    dial_position: 6.587563

lxt
---
  position: None
  lxttc_laser_setpoint: None [s]
  lxttc_laser_pos: 4.4683153320076e-06
  lxttc_laser_offset: 0.0

In [18]: m.txt
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-18-d3164503ea17> in <module>
----> 1 m.txt

/reg/g/pcds/pyps/apps/dev/pythonpath/ophyd/device.py in __getattr__(self, name)
   1195         # Components will be instantiated through the descriptor mechanism in
   1196         # the Component class, so anything reaching this point is an error.
-> 1197         raise AttributeError(name)
   1198
   1199     def _instantiate_component(self, attr):

AttributeError: txt

In [19]: m.delay
Out[19]:

txt
---
position: -2.999968731701713e-12

lxttc_delay_delay
-----------------
  position: -2.999968731701713e-12
  notepad_setpoint: 3e-12
  notepad_readback: 3.0000020881112317e-12
lxttc_delay_user_offset: 0.0

lxttc_delay_motor
-----------------
  position: 0.44968399999999953
  user_setpoint: 0.44968368700000017
  description: las_delay_new
  dial_position: 6.587563
```

cristinasewell:
One other thing here is that I think in the old  `LaserTiming` motor it seemed like we used two additional PVs and a direction: 
```
        direction=-1,
        pvoff="XPP:USER:VIT:T0",
        pvpos="XPP:USER:VIT:TD",
```
are these pvs not used anymore?

ZLLentz:
Sorry for the delays on returning to this- 12 days is unacceptably long, even with vacation + PAMM.

>I am still a bit confused here, not sure what to do.
>
>The position seems to be the calculated one already I think:
>https://github.com/bluesky/ophyd/blob/c91317fb21506d6acc2900e392e667aeb2684abc/ophyd/pseudopos.py#L654-L662
>
>does this mean that this is actually the dial position then, and we can calculate the user position based on the dial maybe (userVAL = DialVAL * DIR + OFFset) ?

The calculated `position` will always be calculated from the user position, not the dial position.

What a `dial_position` means for these calculated motors is very confusing because the functions don't necessarily have any nice properties (e.g. linear, monotonic), meaning that once you undo the offset/dir the "calculated dial position" could end up being completely meaningless...

I think the most charitable interpretation of "calculated dial position" follows from:
`self.position == self.inverse(self.real_position)`
substituting dial for real:
`calculated_dial_position == self.inverse(real_dial_position)`

With the understanding that, for many pseudo motors, this may be completely meaningless... but for something like the time delay stage this is actually pretty nice.

> Also one thing that I noticed when testing with LaserTimingCompensation is that even if the names are changed to lxt, txt when calling those they don't work in an ipython shell, maybe I have an older version though - or maybe we did that on purpose for some reason I don't know, but i thought i'd mention it here.

Attributes and the `ophyd` name field are completely unrelated, we'll need to do the shortcut access names somewhere else than in the class here.

> One other thing here is that I think in the old LaserTiming motor it seemed like we used two additional PVs and a direction:

I think this was an old PV notepad. If we're using Ken's generated PV notepads then we can disregard these old PVs.

pcdsdevices/pseudopos.py
```diff
@@ -39,6 +39,53 @@ def __init__(self, prefix='', parent=None, **kwargs):
             prefix = f'{parent.prefix}:{attr_name}'
 
         super().__init__(prefix=prefix, parent=parent, **kwargs)
+        self._verbose_name = verbose_name
+
+    # TODO: this implementation might be totally crazy....
+    @property
+    def calculated_dial_pos(self):
```
cristinasewell, line 46: This  implementation here does not really work for all scenarios... so probably I took the wrong approach here. 
Works fine for `xcs_lxt_fast.delay` for example, but not for `xcs_ccm.calc.energy` :
`xcs_ccm.calc.energy` - has two `real_positioners`, and does real_positioners don't have a `dial_position`.

ZLLentz: If there is no dial position I think it is okay to assume that the dial position is the same as the normal position. This is actually the case for the ccm, it doesn't have any sort of user offsets that I'm aware of.

I think it may be possible to generalize to arbitrary numbers of real positioners by passing them all into inverse as a RealPositionTuple or whatever it's called

cristinasewell: Ok that's good then, I'll assume `dial_pos` = normal position if some objects don't have that - Thank you

As for the second one - I need to do more exploration here, I think the challenge is that I need the `dial_position` of those `real_positioners`  in the `inverse` method not the actual `normal position`

cristinasewell: Ohh I think I get it now, (maybe). I would just pass into `inverse` multiple dial_positions in my case:
```
self.parent.RealPosition(*dial_pos))
```
where `dial_pos` could be for example `[13.84, 0.0]`

This will now give me 2 calculated dial position - how would I go about choosing which one to use in the `format_status_info`? would I display both of them?

ZLLentz: I think what happens here is that you'll get out some number of "dial pseudo positions" or whatever we want to call them, in the psuedo tuple, with reasonable names. And then you can pick the correct one for the pseudo single hopefully.

pcdsdevices/pseudopos.py
```diff
@@ -39,6 +39,64 @@ def __init__(self, prefix='', parent=None, **kwargs):
...
+        self._verbose_name = verbose_name
+
+    @property
+    def calculated_dial_pos(self):
+        """
+        Calculate the dial position of the real motor dial position.
+        """
+        dial_pos = []
+        # assume the name will be something like: parent_name_self_name
+        name = self.name.split('_')[-1]
```
cristinasewell, line 51: Is it safe here to assume that I will always have something like this:
```
 xcs_lxt_fast.delay.name
Out[1]: 'xcs_lxt_fast_delay'
```
where the name of the pseudo single will be at the end after a `_` (in this case `delay`)

pcdsdevices/pseudopos.py
```diff
@@ -39,6 +39,64 @@ def __init__(self, prefix='', parent=None, **kwargs):
...
+        try:
+            for real_pos in self.parent.real_positioners:
+                dial_pos.append(real_pos.dial_position.get())
+            if dial_pos:
+                calc_dial = self.parent.inverse(
+                    self.parent.RealPosition(*dial_pos))
+            # try to get the correct pseudo position
+            return f'{calc_dial.__getattribute__(name):.3e}'
```
cristinasewell, line 59: I think i am going to change this to use `._fields` probably

pcdsdevices/pseudopos.py
ZLLentz: I think `attr_name` is the correct one to use. The `_` delimiter in the name will be correct 99% of the time, but there may be situations where the underscore is ambiguous or where someone has edited the `name` attribute for some unknown reason.

pcdsdevices/interface.py
```diff
@@ -383,7 +383,7 @@ def get_value(signal):
 
 
 def get_units(signal):
-    attrs = ('units', 'egu', 'derived_units')
+    attrs = ('derived_units', 'units', 'egu')
```
cristinasewell, line 386: This is just a workaround of an issue I was having with the `units` ... i hope it does not break anything else... but I can switch this back.

The problem is that when I get the `lxt.status_info()` - the `units` are in `ns` - which is wrong - what we want is the `derived_units` (`s` in my case).

```
{'name': 'lxt',
 'kind': <Kind.config|normal: 3>,
 'is_device': True,
 'position': -5.597463492824768e-06,
 'setpoint': {'name': 'lxt_setpoint',
  'kind': <Kind.hinted: 5>,
  'is_device': False,
  'value': -5.597463492824768e-06,
  'units': 'ns'},
 'notepad_setpoint': {'name': 'lxt_notepad_setpoint',
  '.......
```

By changing the `attrs = (derived_units, ...)` to have the `derived_units` first element in the tuple - it seems like in the `status_info()` it picks the `derived_units` first and returns that:

```
{'name': 'lxt',
 'kind': <Kind.config|normal: 3>,
 'is_device': True,
 'position': -5.597463492824768e-06,
 'setpoint': {'name': 'lxt_setpoint',
  'kind': <Kind.hinted: 5>,
  'is_device': False,
  'value': -5.597463492824768e-06,
  'units': 's'},
 '.......
```

but as @ZryletTC pointed out to me in the slack chat - the  `UnitConversionDerivedSignal` most likely needs to be fixed, since the metadata for `lxt.setpoint.metadata` still has `units` as `ns` (like the original units)

I think we need to open a new PR for this issue rather than fixing it in this one - i personally don't know yet what needs fixing...

pcdsdevices/lxe.py
cristinasewell: I used the `_fs_tgt_time` as the actual `dial` position here. The `user` position gets calculated in the `_ScaledUnitConversionDerivedSignal` from what I understand, so the `setpoint` is the actual `user` position

pcdsdevices/interface.py
cristinasewell: A new issue has been added for this https://github.com/pcdshub/pcdsdevices/issues/767

ZLLentz: bizarre workaround, but if it works, it works!

ZLLentz (approved):
I don't have anything else to add here, I think it's good

pcdsdevices/lxe.py
```diff
@@ -359,6 +361,69 @@ def set_current_position(self, position):
...
+    def verbose_name(self, name):
+        """
+        Set a verbose name for the LXT motor.
+
+        Parameters
+        ----------
+        name : str
+            Verbose name for the LXT motor.
+        """
+        self._lxt_verbose_name = name
```
klauer, line 379: If you have a property that just gets/sets an internal variable with no additional logic, wrapping it with a property is unnecessary. Would suggest just renaming `_lxt_verbose_name` to `verbose_name` and leaving it at that.

klauer (approved):
Minor (optional) suggestion - looks good to me!

pcdsdevices/lxe.py
cristinasewell: Yeah that make sense, thanks Ken, removed these guys in 75a6f41

Also wrapped getting the `dial_pos` in a try/except - maybe it is not the best place to wrap the in a try/except, but if we only use it to format the status for it I think it should be fine.

klauer: No problem - and that seems a reasonable place for the try/except 👍

ZLLentz: One situation where I think it makes sense to add these sorts of flat properties is when you need to document something and it makes a lot of sense to do it at the property. This sort of documentation is easier to follow then skimming through long class docstrings or doing it inline with comments.

ZLLentz:
This has two approvals and the recent commit looks good. I will merge.

PR #711, opened 2020-11-23
ENH: add kappa pseudo motors and calculations

cristinasewell:
Add pseudomotors to Kappa object. 
Add calculations for converting kappa coordinates to spherical coordinates and vice versa. 
Add custom status print for Kappa.

Needs:
- [x] Testing
- [x] Cleanup - add missing things, remove unnecessary things (forward, inverse)

## Motivation and Context

Trying to implement and close https://github.com/pcdshub/Bug-Reports-and-Requests/issues/56
## How Has This Been Tested?

Just locally, and unit testing.

Docstring 

`check_motor_step` printing results:

Current printing example (using dummy numbers):
kappa

Custom status print:
```
In [4]: k
Out[4]:
Kappa
eta, kappa, phi: -0.20250, -0.00130, 0.51246 [deg]
e_eta, e_chi, e_phi: 0.20, -0.00, 0.51
x, y, z: 8.699, -12.786, 0.469 [mm]
```
## Pre-merge checklist

### DIFF
--- a/pcdsdevices/gon.py
+++ b/pcdsdevices/gon.py
@@ -14,7 +14,7 @@ from .interface import BaseInterface
 from .pseudopos import (PseudoPositioner, PseudoSingleInterface,
                         pseudo_position_argument, real_position_argument)
 from .sim import FastMotor
-from .utils import get_status_value
+from .utils import get_status_value, get_status_float
 
 logger = logging.getLogger(__name__)
 
@@ -550,26 +550,26 @@ class Kappa(BaseInterface, PseudoPositioner, Device):
 
     def format_status_info(self, status_info):
         """Override status info handler to render the Kappa object."""
-        x = get_status_value(status_info, 'sample_stage', 'x', 'position')
-        y = get_status_value(status_info, 'sample_stage', 'y', 'position')
-        z = get_status_value(status_info, 'sample_stage', 'z', 'position')
+        x = get_status_float(status_info, 'sample_stage', 'x', 'position')
+        y = get_status_float(status_info, 'sample_stage', 'y', 'position')
+        z = get_status_float(status_info, 'sample_stage', 'z', 'position')
         units = get_status_value(status_info, 'sample_stage', 'x',
                                  'user_setpoint', 'units')
 
-        eta = get_status_value(status_info, 'eta', 'position')
-        kappa = get_status_value(status_info, 'kappa', 'position')
-        phi = get_status_value(status_info, 'phi', 'position')
+        eta = get_status_float(status_info, 'eta', 'position')
+        kappa = get_status_float(status_info, 'kappa', 'position')
+        phi = get_status_float(status_info, 'phi', 'position')
         angle_units = get_status_value(status_info, 'eta', 'user_setpoint',
                                        'units')
-        e_eta = get_status_value(status_info, 'e_eta', 'position')
-        e_chi = get_status_value(status_info, 'e_chi', 'position')
-        e_phi = get_status_value(status_info, 'e_phi', 'position')
+        e_eta = get_status_float(status_info, 'e_eta', 'position')
+        e_chi = get_status_float(status_info, 'e_chi', 'position')
+        e_phi = get_status_float(status_info, 'e_phi', 'position')
 
         return f"""\
 Kappa
-eta, kappa, phi: {eta:.4f}, {kappa:.4f}, {phi:.4f} [{angle_units}]
-e_eta, e_chi, e_phi: {e_eta:.4f}, {e_chi:.4f}, {e_phi:.4f}
-x, y, z: {x:.4f}, {y:.4f}, {z:.4f} [{units}]
+eta, kappa, phi: {eta}, {kappa}, {phi} [{angle_units}]
+e_eta, e_chi, e_phi: {e_eta}, {e_chi}, {e_phi}
+x, y, z: {x}, {y}, {z} [{units}]
 """
 
 
--- a/pcdsdevices/gon.py
+++ b/pcdsdevices/gon.py
@@ -515,9 +515,9 @@ class Kappa(BaseInterface, PseudoPositioner, Device):
             The real position output.
         """
         pseudo_pos = self.PseudoPosition(*pseudo_pos)
-        eta, chi, phi = self.e_to_k(pseudo_pos.e_eta, pseudo_pos.e_chi,
-                                    pseudo_pos.e_phi)
-        return self.RealPosition(eta=eta, kappa=chi, phi=phi,
+        eta, kappa, phi = self.e_to_k(pseudo_pos.e_eta, pseudo_pos.e_chi,
+                                      pseudo_pos.e_phi)
+        return self.RealPosition(eta=eta, kappa=kappa, phi=phi,
                                  x=self.x.position, y=self.y.position,
                                  z=self.z.position)
 
--- a/pcdsdevices/gon.py
+++ b/pcdsdevices/gon.py
@@ -1,11 +1,17 @@
 """
 Module for goniometers and sample stages used with them.
 """
+import logging
+import numpy as np
+from prettytable import PrettyTable
 from ophyd import Device
 from ophyd import FormattedComponent as FCpt
 
 from .epics_motor import IMS
 from .interface import BaseInterface
+from .pseudopos import PseudoPositioner, PseudoSingleInterface
+
+logger = logging.getLogger(__name__)
 
 
 class BaseGon(BaseInterface, Device):
@@ -210,22 +216,31 @@ class SamPhi(BaseInterface, Device):
         super().__init__('', name=name, **kwargs)
 
 
-class Kappa(BaseInterface, Device):
+class Kappa(BaseInterface, PseudoPositioner, Device):
     """
-    Kappa stage.
+    Kappa stage, control the Kappa diffractometer in spherical coordinates.
+
+    The kappa's native coordinates (eta, kappa, phi) are mechanically
+    convenient, but geometrically awkward to think about. This module
+    replaces the coordinates with (e_eta, e_chi, e_phi) like so:
+
+    The radial component is generally fixed such that the sample is at
+    the center of rotation, but you may think of `z` as the radial
+    component (inverted because the sample is pushed into the center of the
+    coordinate system, rather than out from the center.)
 
     Parameters
     ----------
     name : str
         A name to refer to the Kappa stage device.
 
-    prefix_x : str
+    prefix_x : str, optional
         The EPICS base PV of the Kappa stage's x motor.
 
-    prefix_y : str
+    prefix_y : str, optional
         The EPICS base PV of the Kappa stage's y motor.
 
-    prefix_z : str
+    prefix_z : str, optional
         The EPICS base PV of the Kappa stage's z motor.
 
     prefix_eta : str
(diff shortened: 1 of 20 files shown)


pcdsdevices/gon.py
```diff
@@ -236,23 +252,351 @@ class Kappa(BaseInterface, Device):
...
+                                  2.0) / np.tan(kappa_ang))
+        k_eta = -(pseudo_pos.e_eta * np.pi / 180 - delta)
+        k_kap = 2.0 * np.arcsin(np.sin(pseudo_pos.e_chi * np.pi /
+                                       180 / 2.0) / np.sin(kappa_ang))
+        k_phi = pseudo_pos.e_phi * np.pi / 180 - delta
+
+        k_eta = k_eta * 180 / np.pi
+        k_kap = k_kap * 180 / np.pi
+        k_phi = -k_phi * 180 / np.pi
+        # return k_eta, k_kap, k_phi
```
cristinasewell, line 453: I think i should leave these calculations alone without the `pseudo_pos` and `real_pos` and only handle that in the `forward` and `inverse` methods, by calling these functions?

pcdsdevices/gon.py
```diff
@@ -236,23 +252,344 @@ class Kappa(BaseInterface, Device):
...
+        k_kap = 2.0 * np.arcsin(np.sin(e_chi * np.pi /
+                                       180 / 2.0) / np.sin(kappa_ang))
+        k_phi = e_phi * np.pi / 180 - delta
+
+        k_eta = k_eta * 180 / np.pi
+        k_kap = k_kap * 180 / np.pi
+        k_phi = -k_phi * 180 / np.pi
+        return k_eta, k_kap, k_phi
+
+    def mv_e_eta(self, value):
```
ZLLentz, line 448: Methods like this only existed on the old class because they were required to make the virtual motors work- if they aren't being used in forward/inverse they probably aren't needed

pcdsdevices/gon.py
```diff
@@ -236,23 +252,344 @@ class Kappa(BaseInterface, Device):
...
+        self.phi.stop()
+
+    def wait(self):
+        """Block until the action completes."""
+        self.eta.wait()
+        self.kappa.wait()
+        self.phi.wait()
+
+    @property
+    def eta_position(self):
```
ZLLentz, line 338: Are these being used anywhere? They seem redundant with `self.eta.wm()` and `self.eta.position`, for example

ZLLentz:
I haven't checked the math very carefully against the original, but this looks like a pretty clean transfer

pcdsdevices/gon.py
cristinasewell: I am only using them in `check_motor_step` and in the `mv_` functions - but I can just get it directly from `wm()` - they are redundant

pcdsdevices/gon.py
ZLLentz: I think isolating the calculations is a good idea, especially since there are three "real motor" elements that are omitted.

pcdsdevices/gon.py
```diff
@@ -236,23 +252,344 @@ class Kappa(BaseInterface, Device):
...
+        is_eta_above_max = eta_step > self.eta_max_step
+        is_kappa_above_max = kappa_step > self.kappa_max_step
+        is_phi_above_max = phi_step > self.phi_max_step
+
+        if is_eta_above_max or is_kappa_above_max or is_phi_above_max:
+            d_str = '\nDo you really intend to do the following motions?\n'
+            t = PrettyTable(['Motor', 'Current position', 'to',
+                             'Target position'])
+            if is_eta_above_max:
+                t.add_row(['eta', 5, '-->', eta])
```
ZLLentz, line 582: Quick note that these have placeholder values for current position

pcdsdevices/gon.py
cristinasewell: > I haven't checked the math very carefully against the original, but this looks like a pretty clean transfer

I'm adding some tests against the old values - they seem to work fine, i'll push them soon

pcdsdevices/gon.py
cristinasewell: Ohh thank you, i was testing thigs here :)

ZLLentz:
One more suggestion: it would be easier to test this if you make a simulated Kappa device (`KappaSim`? `SimKappa`?) in the module, by subclassing and replacing all the real motors with `pcdsdevices.sim.FastMotor`. Then you could move all the simulated motors to make sure everything fits together right.

pcdsdevices/gon.py
cristinasewell: we're using the `check_motor_step` in those functions, maybe i could include that somewhere in the forward/inverse

pcdsdevices/gon.py
```diff
@@ -236,23 +255,348 @@ class Kappa(BaseInterface, Device):
...
         self._prefix_eta = prefix_eta
         self._prefix_kappa = prefix_kappa
         self._prefix_phi = prefix_phi
+        self.eta_max_step = eta_max_step
+        self.kappa_max_step = kappa_max_step
+        self.phi_max_step = phi_max_step
+        self.kappa_ang = kappa_ang
         super().__init__('', name=name, **kwargs)
+
+    def stop(self):
```
cristinasewell, line 328: I added the `stop()` and `wait()` methods directly from the old code, but we might not need them here?
At least I might not need the `stop()` method - but I am not sure

pcdsdevices/gon.py
cristinasewell: Fixed it in f3ccec1

pcdsdevices/gon.py
cristinasewell: Took those out in f3ccec1

pcdsdevices/gon.py
ZLLentz: I don't think they are strictly necessary, but I think they are really good to have!
If someone is doing `kappa.move` but then changes their mind, they might try `kappa.stop`, and stopping all the motors at once is what they expect to happen.

cristinasewell: ok that sounds good - i'll keep them here then - thank you

pcdsdevices/gon.py
cristinasewell: I think if I add this `check_motor_step` in the `forward` method, it could get pretty annoying for a user, as she/he will have to constantly right `y` for every single move if the step is bigger than max step, but it will be safer probably. 

Or we could leave the current methods of movement as a safer alternative for people - and they'll have the option to move safe or not so safe?

Also, if we end up not using the `check_motor_step` at all, we can still at least print it out for the user so they are aware about the current position vs target position?

cristinasewell:
In the old code the kappa `x, y, z` were optional, should these be optional here as well?

pcdsdevices/gon.py
```diff
@@ -236,23 +255,348 @@ class Kappa(BaseInterface, Device):
...
+        -------
+        real_position : RealPosition
+            The real position output.
+        """
+        pseudo_pos = self.PseudoPosition(*pseudo_pos)
+        eta, kappa, phi = self.e_to_k(pseudo_pos.e_eta, pseudo_pos.e_chi,
+                                      pseudo_pos.e_phi)
+        return self.RealPosition(eta=eta, kappa=kappa, phi=phi,
+                                 x=self.x.position, y=self.y.position,
+                                 z=self.z.position)
```
cristinasewell, line 522: How do we go about this situation where I have more real motors than pseudo motors and don't need to map all o them?
Do I set them to `None`? 

```
return self.RealPosition(eta=eta, kappa=kappa, phi=phi,
                                 x=None, y=None,
                                 z=None)
```
or do i just take their `position`?

pcdsdevices/gon.py
```diff
@@ -236,23 +255,348 @@ class Kappa(BaseInterface, Device):
...
+        kappa_step = abs(kappa - self.kappa.position)
+        phi_step = abs(phi - self.phi.position)
+
+        is_eta_above_max = eta_step > self.eta_max_step
+        is_kappa_above_max = kappa_step > self.kappa_max_step
+        is_phi_above_max = phi_step > self.phi_max_step
+
+        if is_eta_above_max or is_kappa_above_max or is_phi_above_max:
+            d_str = '\nDo you really intend to do the following motions?\n'
+            t = PrettyTable(['Motor', 'Current position', 'to',
```
cristinasewell, line 576: Should I keep the `PrettyTable` here or should I implement what they had in the old code?
(there are pictures in the description of this PR of the differences)

pcdsdevices/gon.py
```diff
@@ -236,23 +255,348 @@ class Kappa(BaseInterface, Device):
...
+        pseudo_pos : PseudoPosition
+            The pseudo position input.
+        Returns
+        -------
+        real_position : RealPosition
+            The real position output.
+        """
+        pseudo_pos = self.PseudoPosition(*pseudo_pos)
+        eta, kappa, phi = self.e_to_k(pseudo_pos.e_eta, pseudo_pos.e_chi,
+                                      pseudo_pos.e_phi)
```
cristinasewell, line 519: Should I have here something like this:

```
if self.check_motor_step(eta, kappa, phi):
    return self.RealPosition(......
```
for now only the `mv_e_chi`, `mv_e_eta` and `mv_e_phi` are checking for the step...

pcdsdevices/gon.py
klauer: Stepping back for a second, the intent `PseudoPositioner` is that there's always a mapping of pseudo<->real positioners. If they are not needed, why are they in the positioner in the first place?

There is a mechanism to exclude these positioners from the real positioner list, but I believe it does not work (since I wrote the thing in 2015, I only just heard of one person trying and failing to use it this year). So, just an idea:

1. Create a `KappaSample` `Device` with the XYZ stages (if it's not a positioner, it won't get picked up by the `PseudoPositioner` mechanism)
2. On `Kappa` init, set `self.x = self.sample_stage.x` for user convenience

pcdsdevices/gon.py
klauer: My 2c: PrettyTable looks nice

klauer:
Jumping in on this PR a bit late - the `stop` method is already implemented on `PseudoPositioner` and is slightly more fault-tolerant than 3 `.stop()` calls (coming at the expense of complexity): https://github.com/bluesky/ophyd/blob/e168aea65213cd39bc5266c7b26da7a1c686463d/ophyd/pseudopos.py#L538-L562

pcdsdevices/gon.py
cristinasewell: ok thanks i'll leave that as is then

cristinasewell:
> Jumping in on this PR a bit late - the `stop` method is already implemented on `PseudoPositioner` and is slightly more fault-tolerant than 3 `.stop()` calls (coming at the expense of complexity): https://github.com/bluesky/ophyd/blob/e168aea65213cd39bc5266c7b26da7a1c686463d/ophyd/pseudopos.py#L538-L562

Ok sounds good - i should be safe to remove it from here then? The idea was that it will stop all 3 motors, but I think the `stop()` from `PseudoPositioner` will stop all of them anyway?

pcdsdevices/gon.py
cristinasewell: OK, I'll try this - thanks @klauer
(in the old code they have x, y, and z as optional (`samx`, `samy`, `samz`)) - this might have solved this problem

klauer:
Yes, you can safely remove the `stop` implementation here.

pcdsdevices/gon.py
cristinasewell: I thought I understood what I need to do but I got stuck a bit. :( 

Could I just use the `XYZStage` class instead of creating a `KappaSample`? 
https://github.com/pcdshub/pcdsdevices/blob/fa9ddd03634bd5d80bddb04dd83ff8967cd26ccd/pcdsdevices/gon.py#L154-L183

Is `sample_stage` a new component I have to add besides the x, y, and z?
```
sample_stage = Cpt(XYZStage, ...)
```
and have the user give a prefix for this guy now at init?

klauer: Looks like it's already there - good catch! Definitely we should reuse devices where possible.

`FormattedComponent` is a bit faulty here, I think you're finding - you have to know at `__init__` time what the prefixes are, but you have no control over `XYZStage.__init__` in your `Kappa` class.

There may be other/better ways; my first thought is that the following minimal subclass might work:
```python
class KappaXYZStage(XYZStage):
   def __init__(self, *args, parent, **kwargs):
       self._prefix_x = parent._sample_x_prefix  # set `_sample_x_prefix` in `Kappa.__init__` before calling `super().__init__`
       # repeat for y and z prefix
       super().__init__(*args, parent=parent, **kwargs)
```

cristinasewell: Ok thank you, this `KappaXYZStage` seems to work but I can't figure out how to set the `x` to `sample_stage.x` - i'm probably doing something very stupid here...

"On Kappa init, set self.x = self.sample_stage.x for user convenience" - I don't know of `sample_stage` in the `Kappa` `__init__` so can't set it.... 

I have this class:
```
class KappaXYZStage(XYZStage):
    def __init__(self, *args, parent, **kwargs):
        self._prefix_x = parent._prefix_x
        self._prefix_y = parent._prefix_y
        self._prefix_z = parent._prefix_z
        super().__init__(*args, parent=parent, prefix_x=self._prefix_x,
                         prefix_y=self._prefix_y, prefix_z=self._prefix_z,
                         **kwargs)
```
and then in `Kappa`:
```
    x = FCpt(IMS, '{self._prefix_x}', kind='normal')
    y = FCpt(IMS, '{self._prefix_y}', kind='normal')
    z = FCpt(IMS, '{self._prefix_z}', kind='normal')

    sample_stage = Cpt(KappaXYZStage, name='xyz_stage', kind='omitted')


.....

    def __init__(self, *, name, prefix_x, prefix_y, prefix_z,
                 prefix_eta, prefix_kappa, prefix_phi, eta_max_step=2,
                 kappa_max_step=2, phi_max_step=2, kappa_ang=50, **kwargs):
        self._prefix_x = prefix_x
        self._prefix_y = prefix_y
        self._prefix_z = prefix_z
        self._prefix_eta = prefix_eta
        self._prefix_kappa = prefix_kappa
        self._prefix_phi = prefix_phi
        self.eta_max_step = eta_max_step
        self.kappa_max_step = kappa_max_step
        self.phi_max_step = phi_max_step
        self.kappa_ang = kappa_ang
        # self.x = ..
        # self.y = ...
        # self.z = ...
        super().__init__('', name=name, **kwargs)
```

klauer: The key is that `self.x` - the instantiated component, or the `IMS` device instance - will only be available after `super().__init__`. Moving those lines down should do it: `self.x = self.sample_stage.x`

cristinasewell: Ohh ok, I tried that and I was getting an error and did not think of trying it again: 

```
--> 340         self.x = self.sample_stage.x
    341         self.y = self.sample_stage.y
    342         self.z = self.sample_stage.z

~/anaconda3/envs/bluesky/lib/python3.8/site-packages/ophyd/device.py in __set__(self, instance, owner)
    226 
    227     def __set__(self, instance, owner):
--> 228         raise RuntimeError('Do not use setattr with components; use '
    229                            'cpt.put(value)')
    230 

RuntimeError: Do not use setattr with components; use cpt.put(value)
```

klauer: That's correct, but that error is coming from `x`, `y`, and `z` till being components of `Kappa` (it thinks you're trying to do `self.x.put(self.sample_stage.x)`.

cristinasewell: Ok thanks a lot Ken for spoon-feeding me on this Issue - I thought I'd get it done sooner because they need this on Friday :( but I'm still trying to figure it out

klauer: No worries - and if you need, let me know how I can help you push it over the finish line. I'm sure we can get it working by Friday.

cristinasewell: Ok thank you, i'll try to see if I can make it work until tomorrow, if not I would love some help with it - thanks a lot. 

For now if I do:
```
  super().__init__(name=name, **kwargs)
        Kappa.x = self.sample_stage.x
        Kappa.y = self.sample_stage.y
        Kappa.z = self.sample_stage.z
```
```
In [4]: from pcdsdevices.gon import Kappa

In [5]: k = Kappa(name='name', prefix_x='x', prefix_y='y', prefix_z='z', prefix_eta='eta', prefix_kappa='kappa', prefix_phi='phi')

In [6]: k
Out[6]: 

name
----
position: ERROR

sample_stage_x
--------------
  position: ERROR
  user_setpoint: None
  description: None
  dial_position: None

sample_stage_y
--------------
  position: ERROR
  user_setpoint: None
  description: None
  dial_position: None

sample_stage_z
--------------
  position: ERROR
  user_setpoint: None
  description: None
  dial_position: None

eta
---
  position: ERROR
  user_setpoint: None
  description: None
  dial_position: None

kappa
-----
  position: ERROR
  user_setpoint: None
  description: None
  dial_position: None

phi
---
  position: ERROR
  user_setpoint: None
  description: None
  dial_position: None

e_eta
-----
  position: ERROR
  notepad_setpoint: 0.0
  notepad_readback: 0.0

e_chi
-----
  position: ERROR
  notepad_setpoint: 0.0
  notepad_readback: 0.0

e_phi
-----
  position: ERROR
  notepad_setpoint: 0.0
  notepad_readback: 0.0
```
it seems to set the `x`, `y` and `z` fine (i think) but it still sees them as `positioners` this way

I'll keep trying till tomorrow and ask for help tomorrow ;)

pcdsdevices/gon.py
```diff
@@ -304,12 +317,21 @@ class Kappa(BaseInterface, PseudoPositioner, Device):
     e_eta = FCpt(PseudoSingleInterface, kind='normal', name='gon_kappa_e_eta')
     e_chi = FCpt(PseudoSingleInterface, kind='normal', name='gon_kappa_e_chi')
     e_phi = FCpt(PseudoSingleInterface, kind='normal', name='gon_kappa_e_phi')
-
     tab_component_names = True
 
     tab_whitelist = ['stop', 'wait', 'k_to_e', 'e_to_k', 'mv_e_eta',
                      'mv_e_chi', 'mv_e_phi', 'check_motor_step']
 
+    def __new__(cls, *, name, prefix_x, prefix_y, prefix_z,
```
klauer, line 325: `__new__` should be removed.

pcdsdevices/gon.py
```diff
@@ -323,13 +345,19 @@ def __init__(self, *, name, prefix_x, prefix_y, prefix_z,
         self.kappa_max_step = kappa_max_step
         self.phi_max_step = phi_max_step
         self.kappa_ang = kappa_ang
+
         super().__init__('', name=name, **kwargs)
+        Kappa.x = self.sample_stage.x
```
klauer, line 350:
```suggestion
        self.x = self.sample_stage.x
```

pcdsdevices/gon.py
```diff
@@ -323,13 +345,19 @@ def __init__(self, *, name, prefix_x, prefix_y, prefix_z,
         self.kappa_max_step = kappa_max_step
         self.phi_max_step = phi_max_step
         self.kappa_ang = kappa_ang
+
```
klauer, line 348:
```suggestion
        self._prefix_x = prefix_x
        self._prefix_y = prefix_y
        self._prefix_z = prefix_z
```

cristinasewell: I have these guys already :) - unless you purposely suggest to move them further down

pcdsdevices/gon.py
klauer: (Resolved over chat)

pcdsdevices/gon.py
cristinasewell: have added the `check_motor_step` in the `move` method

pcdsdevices/gon.py
```diff
@@ -236,23 +266,290 @@ class Kappa(BaseInterface, Device):
...
+    @pseudo_position_argument
+    def move(self, position, wait=True, timeout=None, moved_cb=None):
+        """
+        Move to a specified position, optionally waiting for motion to
+        complete.
+
+        Checks for the motor step, and ask the user for confirmation if
+        movement step is greater than default one.
+        """
+        if self.check_motor_step(position.e_eta, position.e_chi,
```
cristinasewell, line 472: I hope i'm not doing anything crazy here but this seems to work.

Also because I added this now, I removed the other mv methods: `mv_e_eta`, `mv_e_chi` and `mv_e_phi` - I think we don't need them anymore?

pcdsdevices/gon.py
cristinasewell: resolving this because I added the `check_motor_step` as part of the `move` method

pcdsdevices/gon.py
ZLLentz: I agree the the other motor mv methods are not needed and this seems like the correct approach to me

pcdsdevices/gon.py
```diff
@@ -525,6 +525,16 @@ def forward(self, pseudo_pos):
...
 
+    @pseudo_position_argument
+    def move(self, position, wait=True, timeout=None, moved_cb=None):
+        if self.check_motor_step(position.e_eta, position.e_chi,
+                                 position.e_phi):
+            return super().move(position, wait=wait, timeout=timeout,
+                                moved_cb=moved_cb)
+        else:
+            logger.warning('Aborting moving for safety.')
+            return
```
ZLLentz, line 479: This should return a failed status object, maybe like:
```suggestion
            status = DeviceStatus(self)
            status.set_exception(ValueError('Unsafe Kappa move aborted!'))
            return status
```

ZLLentz: I did not test this snippit! So I'm sorry if it doesn't quite work right

ZLLentz: The API of the `move` method is to return a status object of some kind, I should have mentioned that first in this comment chain

ZLLentz:
I'm going to resolve the merge conflicts here, I think the code looks good

ZLLentz (approved):

PR #712, opened 2020-11-24
STY: Epics_motor metadata

### LINKED ISSUE #695: EpicsMotor Signals Need Metadata
## Expected Behavior
Screens generated via Typhos have reasonable looking interfaces, with widgets that make sense visually, as well as with correct units (or no units, when applicable). 

## Current Behavior
Many useful signals in the upstream ophyd.epics_motor.EpicsMotor class are kind='omitted'. This makes them annoying to access from Typhos screens. Things like motor homing should probably at least be kind='config'. 

Also, Ophyd doesn't use our variety package, so there is no metadata helping Typhos figure out what to do with these signals. This causes you get confusing Typhos screens like this:

image

## Possible Solution
Since upstream ophyd doesn't use our variety metadata, it seems like we need to overwrite these signals in our implementation. Seems like lines like:

```
home_forward =  Cpt(EpicsSignal, '.HOMF', kind='config', doc='Home the stage forward')
set_metadata(home_forward, dict(variety='command-proc', value=1))
```

to EpicsMotorInterface would do the trick for these signals. 

Generate a Typhos screen using a device that sub-classes ophyd.epics_motor.EpicsMotor or our EpicsMotorInterface. Example: pcdsdevices.epics_motor.SmarAct. 

## Context
These signals are hidden on the Typhos laser control screens I have generated, so laser scientists weren't sure how to home motors. Once we figured out where the signals were, the screens were confusing. 

## Your Environment
dev_python

slactjohnson:
## Description
Adds some metadata and updates kind for some motor record fields. Small update to SmarActOpenLoop to clean up the open loop position control because I couldn't resist. 

## Motivation and Context
Closes #695, more details there. 

## How Has This Been Tested?
Tested on a stage in TMO. 

Just the code. 

image

image

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -62,27 +62,22 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
     tab_whitelist = ["set_current_position", "home", "velocity",
                      "enable", "disable"]
 
-    set_metadata(EpicsMotor.home_forward, dict(variety='command-proc', value=1))
+    set_metadata(EpicsMotor.home_forward, dict(variety='command-proc',
+                                               value=1))
     EpicsMotor.home_forward.kind = Kind.normal
-    set_metadata(EpicsMotor.home_reverse, dict(variety='command-proc', value=1))
+    set_metadata(EpicsMotor.home_reverse, dict(variety='command-proc',
+                                               value=1))
     EpicsMotor.home_reverse.kind = Kind.normal
     set_metadata(EpicsMotor.low_limit_switch, dict(variety='bitmask', bits=1))
     EpicsMotor.low_limit_switch.kind = Kind.normal
     set_metadata(EpicsMotor.high_limit_switch, dict(variety='bitmask', bits=1))
     EpicsMotor.high_limit_switch.kind = Kind.normal
     set_metadata(EpicsMotor.motor_done_move, dict(variety='bitmask', bits=1))
-    EpicsMotor.motor_done_move.kind = Kind.config
+    EpicsMotor.motor_done_move.kind = Kind.omitted
     set_metadata(EpicsMotor.motor_is_moving, dict(variety='bitmask', bits=1))
-    # Had no joy here or anywhere else with enums. I still get the <val> <egu> 
-    # "0 mm" readback. Maybe I'm doing it wrong?
-#    set_metadata(EpicsMotor.motor_is_moving, dict(variety='enum',
-#                                                  enum_strings=['No', 'Yes']))
     EpicsMotor.motor_is_moving.kind = Kind.normal
     set_metadata(EpicsMotor.motor_stop, dict(variety='command-proc', value=1))
     EpicsMotor.motor_stop.kind = Kind.normal
-    # Had no luck with text either, same as enums above. 
-#    set_metadata(EpicsMotor.direction_of_travel, dict(variety='text',
-#                                                      enum_strings=['Rev', 'Fwd']))
     EpicsMotor.direction_of_travel.kind = Kind.normal
 
     def __init__(self, *args, **kwargs):
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -14,6 +14,7 @@ from ophyd.signal import EpicsSignal, EpicsSignalRO, Signal
 from ophyd.status import DeviceStatus, SubscriptionStatus
 from ophyd.status import wait as status_wait
 from ophyd.utils import LimitError
+from ophyd.ophydobj import Kind
 
 from pcdsdevices.pv_positioner import PVPositionerComparator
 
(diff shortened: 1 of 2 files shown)


pcdsdevices/epics_motor.py
```diff
@@ -54,12 +55,36 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
...
+    set_metadata(EpicsMotor.low_limit_switch, dict(variety='bitmask', bits=1))
+    EpicsMotor.low_limit_switch.kind = Kind.normal
+    set_metadata(EpicsMotor.high_limit_switch, dict(variety='bitmask', bits=1))
+    EpicsMotor.high_limit_switch.kind = Kind.normal
+    set_metadata(EpicsMotor.motor_done_move, dict(variety='bitmask', bits=1))
+    EpicsMotor.motor_done_move.kind = Kind.config
+    set_metadata(EpicsMotor.motor_is_moving, dict(variety='bitmask', bits=1))
+    # Had no joy here or anywhere else with enums. I still get the <val> <egu> 
+    # "0 mm" readback. Maybe I'm doing it wrong?
+#    set_metadata(EpicsMotor.motor_is_moving, dict(variety='enum',
```
slactjohnson, line 78: @klauer I would have thought that this would have worked. However, I didn't see anyone else using enum_strings in pcdsdevices; maybe this isn't totally working yet?

ZLLentz: The code at https://github.com/pcdshub/typhos/blob/674686b5b93c42352c5c0902f5684a9fd75bac04/typhos/widgets.py suggests that the methods to use the `enum_strings` are not yet written for `enum` or `text-enum`. Only the command varieties have handling for this.

In general the variety handling is not complete- we can justify working on this now that someone wants to take full advantage of it.

ZLLentz:
I like the way the screen looks! A few things I note from looking at this:
- The limit switches have two boxes instead of one: this is probably because they are writable signals at `ophyd` for some reason: https://github.com/bluesky/ophyd/blob/c91317fb21506d6acc2900e392e667aeb2684abc/ophyd/epics_motor.py#L62-L63
- I'm not 100% sold on `motor_done_move` being `Kind.config` in the `bluesky` sense of `Kind.config`, though it does look good for the screen. We're seeing another instance of a divergence between `bluesky` kinds and `typhos` kinds.
- It'd be interesting to see what this looks like using the `PositionerBase` template

slactjohnson (commented):
> 
> 
> I like the way the screen looks! A few things I note from looking at this:
> 
>     * The limit switches have two boxes instead of one: this is probably because they are writable signals at `ophyd` for some reason: https://github.com/bluesky/ophyd/blob/c91317fb21506d6acc2900e392e667aeb2684abc/ophyd/epics_motor.py#L62-L63
> 
>     * I'm not 100% sold on `motor_done_move` being `Kind.config` in the `bluesky` sense of `Kind.config`, though it does look good for the screen. We're seeing another instance of a divergence between `bluesky` kinds and `typhos` kinds.
> 
>     * It'd be interesting to see what this looks like using the `PositionerBase` template

Yeah, I decided to make motor_done_move Kind.config to clean up the screen. It seems redundant on the screen with motor_is_moving, and it's a little more esoterically named. I can make that Kind.normal to keep with `bluesky` standards. 

RE PositionerBase, can you explain what you mean a little more?

slactjohnson:
I think the one thing that still bugs me about this screen is that the direction_of_travel signal is still weird. I don't know of a way to do this better than with enum or text support.

ZLLentz:
> direction_of_travel signal is still weird

Strangely enough the EPICS value here isn't even an enum- so yeah, without custom enum strings it seems awkward. A bitmask doesn't help anyone here.

> RE PositionerBase, can you explain what you mean a little more?

Your screenshot is of the detailed engineering screen, showing every signal (except omitted). The default screen I usually see pop up has a special graphical widget at the top. I was just wondering if everything looks just as good on the default motor screen.

> I can make that Kind.normal to keep with bluesky standards.

In `bluesky` it would actually be `Kind.omitted` for sure. Maybe that makes the conflict of purposes more clear. We're severely underusing the capabilities of `bluesky` though so I'm not worried about having screen-centric configurations.

pcdsdevices/epics_motor.py
klauer: Just confirming @ZLLentz's comment - it's just not there quite yet.

klauer:
This is looking much better than before - good work here @slactjohnson 👍 . It also helps in that it gives us more inspiration to finish up the variety handling in typhos.

slactjohnson:
Confirmed operation on a stage in TMO.

pcdsdevices/epics_motor.py
```diff
@@ -551,11 +571,9 @@ class SmarActOpenLoop(Device):
                          doc='Clear the current step count')
     set_metadata(step_clear_cmd, dict(variety='command-proc', value=1))
     # Scan move
-    scan_move_cmd = Cpt(EpicsSignal, ':SCAN_MOVE', kind='omitted',
```
klauer, line 554: This will need at least an API change line in the release notes.

If we think there's anywhere this could be in use, we might consider some back-compat support.

klauer (changes requested):
The interactive testing you've done here seems sufficient for this PR. Just one thing to tweak with regard to the component API change, and this should be good to go.

pcdsdevices/epics_motor.py
slactjohnson: Good catch. There are very few places SmarAct stages are in use; MEC is the only other place I know of other than the laser system, and I don't expect that they are using this feature yet. @sfsyunusSLAC ?

klauer (approved):

pcdsdevices/epics_motor.py
sfsyunusSLAC: can confirm they are not

ZLLentz:
Received some feedback about this in slack thread https://slac.slack.com/archives/CV0S7HSPK/p1607189599005600, prompting me to make #722

Issue #704, opened 2020-11-17
LODCM Energy Motors


ZLLentz:
## Expected Behavior

`LODCM.energy` pseudomotor
- sets to four LODCM motors using calculations, see old formulation at https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xpp/blinst/lom.py#L1197
- reads from one motor using calculations, see old formulation at https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xpp/blinst/lom.py#L1121

## Current Behavior

Does not exist in the new python

## Possible Solution

Investigate calculations, expand `pcdscalc` may be needed.

## Context

This would allow scanning the LODCM energy

cristinasewell:
Are we not using the `T2C:REF` and `T2Si:REF` on purpose here - are they usually the same as the `T1C:REF` and `T1Si:REF`?

ZLLentz:
That's really cryptic... I have no idea. It looks like an error? But maybe it never causes an issue because they are generally the same?

cristinasewell:
This PR is on my list of Issues now, and I am trying to figure out what else still needs to be done here.

I think we covered most of this PR (or at least a portion of it) in #715 and #766, as well as in pcdshub/pcdscalc#9.
@ZryletTC do you know if the scientists got a chance to test the LODMC energy stuff at all?

I also tagged Diling a while ago in this PR (https://github.com/pcdshub/pcdsdevices/issues/763#issuecomment-785360614) to ask for any other functions we might need in LODCM but so far we did not  get anything.

ZryletTC:
@cristinasewell Yeah, they've tested but it's been awhile so I'm not sure if either they or I remember much about what still needs to be done without revisiting it more recently. The only thing I really remember is that I think there were a bunch of subscription errors popping up so we ended up disabling at least the XCS LODCM. I'll see if we can test it out again this week so I can give you a better picture of the current status.

Issue #677, opened 2020-11-05
Custom status prints for LODCM/LOM


ZLLentz:
lom:
- big big tower prints from old python
- maybe more condensed?

## Steps to Reproduce (for bugs)

1.
2.
3.

ZLLentz:
The existing classes are missing... mostly everything! When I set up the LODCM class I couldn't be bothered to keep track of 20 motor prefixes...

An example from `pcdsdevices` for using `UnrelatedComponent` to make it suck a lot less to have 20 prefixes: https://github.com/pcdshub/pcdsdevices/blob/dc9237ce539da7de76bf31371652b9506ff3ef99/pcdsdevices/sample_delivery.py#L96-L150
(I didn't have this when I made the LODCM)

Some other digging to help and to document:

Full PV lists at:
`/cds/data/iocData/<iocname>/iocInfo/IOC.pvlist`

Four iocs for XPP:
`ioc-xpp-lodcm`: soft PVs traditionally used to store aux info
`ioc-xpp-mon-lodcm-ims`: the states and most of the motors
`ioc-xpp-mon-nav-ims`: just the zoom lens
`ioc-xpp-mon-yaglom`: area detector IOC for the camera

Most useful references from old hutch configs
https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xpp/xppmotors.py#L194-L226
https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xpp/beamline.py#L455-L484
https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xpp/blinst/lom.py#L28-L58

ZLLentz:
I think a big problem even given all of this is the proliferation of super short names for each axis... makes all this old code and this request super hard to navigate. Might get lucky with the DESC fields but I'm not really sure:
```
$ caget XPP:MON:MMS:04.DESC
XPP:MON:MMS:04.DESC            LOM Xtal1 Z
```

ZLLentz:
https://sites.google.com/site/xppwiki/beamline-components/lom might be helpful for background information if you're interested in that sort of thing

ZLLentz:
And a screenshot of an internal page with even more info because I forgot what diag h, etc. was:
[...]

Issue #543, opened 2020-08-27
LODCM Class state names Rework


ZLLentz:
For LODCM state classes, stop hard-coding the state names and instead use the EPICS names (by removing the states_list)

Original card:
Typer P --> Change the slit state names in LODCM classes to have the actual widths (can these be grabbed at init from the state PVs?)
Silke --> I do not know what this means, could you be more specific

PR #715, opened 2020-12-01
ENH: Lodcm custom status + energy motors

cristinasewell:
## Description
- Have added all the missing motors and the offset motors, for Tower1, Tower2 and the Diagnostics Tower.
- Have added some functions to check the `material` of the crystal's towers, as well as the `reflection`. 
- Have added a few missing `InOutRecordPositioner` for the states.
- Have added the Energy objects, one object for the `Si` material and one for the `C` material - these could have been combined but it was easier to separate them.
- Have added the `format_status_info` to print the status of this object.
- Have added the `OffsetMotorBase` PseudoPositioner with an offset - needed for all the motors offsets (`Si` offsets and `C` offsets)

## Motivation and Context

Tackling #677 and #704 (at least partially if not completely)

## How Has This Been Tested?

For now this is kind of what I have:
```
XPP LODCM Motor Status Positions
Crystal 1 state: OUT
Current Configuration: Diamond (111)
Photon Energy: 14.5452 [keV]
-----------------------------------------------------------------
                           Crystal Tower 1           Crystal Tower 2
z [mm]                -677.0404 (658.3165)       678.2323 (662.6102)
x [mm]                     0.0071 (0.8679)          -0.1819 (1.3930)
th [deg]                 11.9522 (-7.3005)         11.9590 (16.7647)
chi [deg]                  0.0270 (3.5588)           0.1150 (6.1229)
y [mm]                   0.0200 (-18.6343)        -0.0163 (-38.7516)
hn [mm]                -10.0001 (-10.5463)         0.0001 (-18.7301)
hp [mm]                    0.0236 (0.2301)                 N/A (N/A)
diode [N/A]                      N/A (N/A)        6.0000 (-366.0000)
-----------------------------------------------------------------
                          Diagnostic Tower
diag r [deg]             23.9036 (23.9036)
diag h [mm]               0.0015 (-0.0015)
diag v [mm]             40.0000 (-40.0000)
filter [deg]           -0.0141 (-269.0666)
diode [mm]                 0.0000 (0.0000)
navitar zoom [%]                   80 (80)
```
```
XCS LODCM Motor Status Positions
Crystal 1 state: OUT
Crystal 2 state: Unknown
Current Configuration: Unknown (Unknown)
Photon Energy: Unknown [keV]
-----------------------------------------------------------------
                           Crystal Tower 1           Crystal Tower 2
z [mm]                -857.2969 (839.0964)       856.5656 (847.4449)
x [mm]                     0.0008 (3.5069)          -1.2185 (0.3971)
th [deg]                   8.7524 (2.7757)           9.9112 (4.9969)
chi [deg]                -2.1795 (-2.1653)          -1.9591 (1.9591)
y [mm]                    7.5908 (14.3321)          1.3387 (18.6875)
hn [mm]                 -11.7638 (-6.0915)          0.5999 (-1.6821)
hp [mm]                    8.7550 (0.4672)                 N/A (N/A)
diode [N/A]                      N/A (N/A)           0.0000 (0.0000)
-----------------------------------------------------------------
                          Diagnostic Tower
diag r [deg]             19.2108 (19.2273)
diag h [mm]            107.3314 (107.3314)
diag v [mm]             34.3076 (-48.3076)
filter [deg]               0.0000 (0.0000)
diode [mm]               -0.1803 (23.9823)
navitar zoom [%]                   30 (30)
```

```
In [4]: lom.get_energy()
Out[4]: 14.545257384584511

In [5]: lom.get_material()
Out[5]: 'C'

In [6]: lom.get_reflection()
Out[6]: '111'
```
## Pre-merge checklist


shs8313 (approved):

pcdsdevices/lodcm.py
```diff
@@ -186,4 +407,276 @@ def remove_dia(self, moved_cb=None, timeout=None, wait=False):
...
+        yag_zoom_units = get_status_value(status_info, 'yag_zoom',
+                                          'user_setpoint', 'units')
+        yag_zoom_user = get_status_value(status_info, 'yag_zoom', 'position')
+        yag_zoom_dial = get_status_value(status_info, 'yag_zoom',
+                                         'dial_position', 'value')
+
+        def form(left_str, center_str, right_str):
+            return f'{left_str:<15}{center_str:>25}{right_str:>25}'
+
+        def ff(float_str):
```
cristinasewell, line 620: I keep having to format the numbers to fewer decimal points, maybe I should be having an option in the `get_status_value` for that?

pcdsdevices/lodcm.py
```diff
@@ -60,6 +65,98 @@ def __init__(self, prefix, *args, **kwargs):
...
+class Y2(InOutRecordPositioner):
+    states_list = ['C', 'Si']
+    in_states = ['C', 'Si']
+    out_states = []
+
+
+LODCM_MOTORS = {
+    # CRYSTAL TOWER ONE
+    'z1': {'prefix': 'XPP:MON:MMS:04', 'description': 'LOM Xtal1 Z',
+           'motor': Motor('XPP:MON:MMS:04', name='z1')},
```
klauer, line 101: The calls to the factory function `Motor()` are instantiating ophyd devices that talk to PVs on import - this will have to be removed/refactored.

cristinasewell: ohh no, i see - thanks @Ken, i'll remove it because I don't need it anyway

pcdsdevices/lodcm.py
klauer: Given that many values will be floating points, that seems reasonable. However, you might not want `get_status_value` to return a `str` instead a `float`, so maybe a new function `get_status_value_as_float` (or something less verbose) with the `precision` kwarg? 🤷‍♂️

pcdsdevices/lodcm.py
```diff
@@ -60,6 +65,74 @@ def __init__(self, prefix, *args, **kwargs):
         super().__init__(prefix, *args, **kwargs)
 
 
+class CHI1(InOutRecordPositioner):
```
cristinasewell, line 68: These guys here all have only `Unknown`, `C` and `Si` states - are they still considered `InOUtRecordPositioners` or should they be something else? Also, does this mean that they don't move in and out of the beamline? what's their role?

pcdsdevices/lodcm.py
```diff
@@ -84,21 +157,145 @@ class LODCM(BaseInterface, Device):
...
     # QIcon for UX
     _icon = 'fa.share-alt-square'
 
-    tab_whitelist = ['h1n', 'yag', 'dectris', 'diode', 'foil', 'remove_dia']
+    tab_whitelist = ['h1n', 'yag', 'dectris', 'diode', 'foil',
+                     'remove_dia']
 
     def __init__(self, prefix, *, name, main_line='MAIN', mono_line='MONO',
                  **kwargs):
+        kwargs['z1_prefix'] = kwargs.get('z1_prefix') or get_prefix('z1')
```
cristinasewell, line 273: I don't know if I should keep this way of defaulting to some pre-defined prefixes? 
Are they changing based on XPP and XCS?
Maybe storing them somewhere else not in this module might be better - if we go this way?
Also, this method does not make it very clear to the user that they are even going to be using these prefixes when creating this device.

pcdsdevices/lodcm.py
```diff
@@ -186,4 +383,276 @@ def remove_dia(self, moved_cb=None, timeout=None, wait=False):
 
         return status
 
+    def _is_tower_1_c(self):
```
cristinasewell, line 386: So i got these guys from the old code, but I am not sure if this is still the case now? (`_is_tower_1_c`, `_is.......`)
These functions will be also used in PR #704 

It seems like right now they all return `False` - if `h1n` position is `C` then `y1`'s position is `Unknown` or some type of combination like this that don't make them `True`

I guess my question here is how do we determine if the tower 1 is `C` or `Si` - is it only based on `h1n`? or are these combinations here still valid?

pcdsdevices/lodcm.py
klauer: It does seems surprising there are no "out" states for these... Someone who's used them should be able to chime in to clarify.

pcdsdevices/lodcm.py
cristinasewell: solved in #718

pcdsdevices/lodcm.py
```diff
@@ -84,13 +158,129 @@ class LODCM(BaseInterface, Device):
...
+    df_prefix : str
+        LOM Dia Filter Wheel motor prefix.
+
+    dd_prefix : str
+        LOM Dia PIPS motor prefix.
+
+    yag_zoom_prefix : str
+        LOM Zoom motor prefix.
+    """
     h1n = Cpt(H1N, ':H1N', kind='hinted')
```
cristinasewell, line 222: I think `h1n`'s name could be changed to something like `h1n_state` or positioner.... this guy is different from the other `h1n_m = UCpt(IMS, kind='normal', doc='LOM Xtal1 Hn')` - i think... i need to look into it more

pcdsdevices/lodcm.py
```diff
@@ -187,3 +403,389 @@ def remove_dia(self, moved_cb=None, timeout=None, wait=False):
         return status
 
     remove_dia.__doc__ += insert_remove
+
+    @pseudo_position_argument
+    def forward(self, pseudo_pos):
+        energy = self.get_energy()
+        # TODO this seems ridiculous, needs revising
+        return self.RealPosition(th1=energy, h1n=None, yag=None,
```
cristinasewell, line 411: I am not sure how the mapping will work here

pcdsdevices/lodcm.py
```diff
@@ -84,13 +158,129 @@ class LODCM(BaseInterface, Device):
...
     foil = Cpt(Foil, ":FOIL", kind='omitted')
+    y1_state = Cpt(Y1, ':Y1', kind='omitted')
+    chi1_state = Cpt(CHI1, ':CHI1', kind='omitted')
+    h2n_state = Cpt(H2N, ':H2N', kind='omitted')
+    y2_state = Cpt(Y2, ':Y2', kind='omitted')
+    chi2_state = Cpt(CHI2, ':CHI2', kind='omitted')
+    # TODO energy component??
+    energy = Cpt(PseudoSingleInterface, egu='eV')
+
+    # Crystal Tower 1
```
cristinasewell, line 235: I'm wondering if it would make more sense to have a separate class for each Tower... might make it cleaner but because i'm not entirely sure how this class is supposed to behave yet I don't know if there will be something ahead that would complicate things that way

klauer: Looking at @ZLLentz 's image with the descriptions, 3 tower classes seems appropriate - assuming their calculations are independent of one another. 

Instead of trying to figure that out by studying the code, asking someone who knows about it will save a bunch of time/effort here...

pcdsdevices/lodcm.py
ZLLentz: chi is a rotation axis, so having no "out" state makes sense.

I'm not 100% sure what they are rotating (surely it's the crystal) or why there are states at all if it is the crystal. Maybe its a nominally aligned position at a known energy?

pcdsdevices/lodcm.py
```diff
@@ -60,7 +97,44 @@ def __init__(self, prefix, *args, **kwargs):
...
+
+
+class CHI2(InOutRecordPositioner):
+    states_list = ['C', 'Si']
+    in_states = ['C', 'Si']
+    out_states = []
+
+
+class H2N(InOutRecordPositioner):
+    states_list = ['C', 'Si']
```
ZLLentz, line 113: There's probably a less wordy way to set all of these up, because of https://github.com/pcdshub/pcdsdevices/blob/757834326d6b95a1ff49507999502b2f81e8f3b3/pcdsdevices/state.py#L36-L37
This is related to #543
Might not be in scope for this PR

pcdsdevices/lodcm.py
ZLLentz: It's confusing because the `h1n` does include a `motor` subdevice... with hopefully the same PV as the `h1n_m`. These state devices have the motor embedded into them.

pcdsdevices/lodcm.py
ZLLentz: I think making separate tower classes is appropriate if it helps organize the code. Unfortunately the towers don't share too many commonalities with each other and there will be some redundancy in the naming, like:
lodcm.tower1.z1
lodcm.tower2.z2
but there wouldn't be a z1 on tower2

pcdsdevices/lodcm.py
```diff
@@ -84,13 +158,129 @@ class LODCM(BaseInterface, Device):
...
+    th2_c = Cpt(PseudoSingleInterface, ':TH2:OFF_C', kind='normal',
+                name='th2_diamond', doc='Th2 motor offset for C')
+    z1_si = Cpt(PseudoSingleInterface, ':Z1:OFF_Si', kind='normal',
+                name='z1_silicon', doc='Z1 motor offset for Si')
+    z1_c = Cpt(PseudoSingleInterface, ':Z1:OFF_C', kind='normal',
+               name='z1_diamond', doc='Z1 motor offset for C')
+    z2_si = Cpt(PseudoSingleInterface, ':Z2:OFF_Si', kind='normal',
+                name='z1_diamond', doc='Z2 motor offset for Si')
+    z2_c = Cpt(PseudoSingleInterface, ':Z2:OFF_C', kind='normal',
+               name='z1_diamond', doc='Z2 motor offset for C')
```
ZLLentz, line 283: oh my this tower of pseudo positioners looks terrifying

pcdsdevices/lodcm.py
ZLLentz: XCS does have a different set of prefixes, but otherwise it's the same device

I do think defaulting to the correct prefixes is good

Do we need to allow the user to change these at all? Could we just fully hard-code prefixes for the XPP and XCS LODCMs? There must be a concise way to define this... it's escaping me at the moment

pcdsdevices/lodcm.py
ZLLentz: We need to re-frame this, this sort of construct is going to be rough

Only the `PositionerBase` subclasses will be picked up in the `RealPosition` tuple, which will cut down on this quite a bit... but it's still a bit ridiculous

Maybe we can find an alternate way of  implementing the offset motors to bring these calculation areas down to just handling the energy psuedomotor?

pcdsdevices/lodcm.py
ZLLentz: h1n brings the crystal into the beam, moving horizontally (normal to the beam and parallel to the ground). It's also the only axis that can fully clear the crystal assembly from the beam.
y1 brings the crystal into the beam, moving vertically (normal to the beam and normal to the ground)
chi1 rotates the crystal assembly, presumably selecting a crystal on an axis that I don't have a geometric understanding of

So I think this function is correct insofar as:
- If they all agree, we are at that crystal
- If they disagree, we are not at any particular crystal, we're just misaligned

The confusing part is the h1n out clause. I think this is saying "we are technically out, but if we were to insert into the beam, we'd be aligned to the C crystal"

ZLLentz: Also, tower 2 is never removed so there is no need for the h1n out clause? (guessing)

pcdsdevices/lodcm.py
ZLLentz: Oh shoot I'm totally off-base, the state positioners ARE `PositionerBase`! That's awful

ZLLentz: Honestly this sounds like a terrible experience
Maybe we need to make a subdevice `PseudoPositioner` for `energy` instead of defining the whole LODCM as a `PseudoPositioner`, and just duplicate the motor components we need for the calculations

ZLLentz: Then instead of using `PseudoSingle` for the offset motors we have some sort of `OffsetIMS` class

pcdsdevices/lodcm.py
silkenelson: Now that you have spoken with Dilingm you are likely aware of the desire of the scientists to type as few characters as possible, so we will need to make aliases if we add another component that would require an extra tab or character to type in a 'user facing' device. They will want lom_z1  as this is what they had (lodcm_z1 could also work as they should be able to tab complete after lod

cristinasewell: Yes that's correct Silke, I think if I end up making separated classes I would still define all those motors in LODCM class, something I did for Kappa for the XYZStage:

```
    self.z1 = self.first_tower.z1
    self.x1 = self.first_tower.x1
    ...... and so on....
```
but I need to make sure they are available when using `tab` though, not sure if they will be this way....

pcdsdevices/lodcm.py
cristinasewell: Yeah that's something I have a hard time wrapping my head around.
When we do the calculations, depending on what material we are using, we would be mapping different pseudo positioners. 
So something like this would not really work:

```
  if material == "Si":
            return self.PseudoPosition(th1_si=th, th2_si=th, z1_si=-z, z2_si=z)
  elif material == "C":
            return self.PseudoPosition(th1_c=th, th2_c=th, z1_c=-z, z2_c=z)
```
So we probably should have two different Energy subdevices in this case?...maybe...

ZLLentz: I think we should make it easier on ourselves by doing something like https://github.com/pcdshub/pcdsdevices/pull/715#discussion_r538970273

pcdsdevices/lodcm.py
cristinasewell: In the old code we have something like this:

```
self._stateh1n = stateiocDevice("%s:H1N" % pvBase)

(full pv: XPP:LODCM:H1N)
```
and also:
```
self.h1n = h1n

(full pv: XPP:MON:MMS:09)
```

I just have `h1n_m` instead of `h1n` because it was taking, that's why I was thinking about changing the current `h1n` to something like `h1n_state`

cristinasewell: So from a programming perspective, I still have to have two components though correct?
```
h1n_state = Cpt(H1N, .....)
h1n = Cpt(IMS, ....)

h2n_state = Cpt(H2N, ....)
h2n = Cpt(IMS, ...)

... and so on...
```
They have different PVs...

pcdsdevices/lodcm.py
cristinasewell: That would be great.

Is the prefix for the XCS motors: `HFX` ? (based on old code)
https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xcs/xcsmotors.py#L527-L554

If these are the pvs from the link above, there is only  one (i think) PV that differs from XPP:
```
self.lom_df = psmotor("XPP:MON:MMS:27", "lom_df")

self.lom_df = psmotor("HFX:MON:MMS:22", "lom_df")  # calib foils
```

ZLLentz: That looks correct to me. `HFX` a prefix used for some items in the tunnel between `XPP` and `XCS`, and since the mono is in there the PV name makes sense.

It's very lucky if only one of the number indices is different. I wonder why the engineer changed exactly one... maybe the `XPP` mono motor prefixes include the CCM motors? `XPP` has two monochromators right next to each other... Maybe the foil stage was added later? I can only guess.

pcdsdevices/lodcm.py
ZLLentz: `InOutRecordPositioner` already automatically has a `motor` subdevice, but if it makes the implementation/usage easier it is good to duplicate it anyway.

I think using `_state` to designate the state devices is a good idea, it is very clear.

pcdsdevices/lodcm.py
klauer: @cristinasewell, would you mind adding a class docstring indicating the reason for the lack of the out state?

pcdsdevices/lodcm.py
klauer: (I have nothing useful to add here; just another request to add the above very useful comment in class docstrings somehow as you see fit...)

pcdsdevices/lodcm.py
klauer: It looks like we really need to fix up ophyd's support for limiting the real positioners used in calculations. The alternative you propose sounds very reasonable, though.

pcdsdevices/lodcm.py
```diff
@@ -60,6 +69,437 @@ def __init__(self, prefix, *args, **kwargs):
...
+    # def set(value):
+    #     if use_ims_preset:
+    #         Pv.put("%s_SET" % offset_pv, motor.wm() - value)
+    #     Pv.put(offset_pv, motor.wm() - value)
+
+    def __init__(self, prefix, *args, **kwargs):
+        super().__init__('', *args, **kwargs)
+
+
+class CrystalTower1State(BaseInterface, Device):
```
cristinasewell, line 212: So I separated the 'state' related functions into separated classes, because I'm using them in the calculations, and since I did not need all the motors there I did not want to include them here, but I think I should include the motors here and have only 2 classes (Tower1, Tower2) instead of 4 (Tower1motors, Tower1States, .....)

cristinasewell: In the calculations I only use `th1` motor.....

pcdsdevices/lodcm.py
```diff
@@ -60,6 +69,437 @@ def __init__(self, prefix, *args, **kwargs):
...
+
+    def __init__(self, prefix, *args, **kwargs):
+        self._prefix = prefix
+        self._df_suffix = '27'
+        if 'XCS' in self._prefix:
+            self._df_suffix = '22'
+        super().__init__(prefix, *args, **kwargs)
+
+
+class OffsetIMS(PseudoSingleInterface):
```
cristinasewell, line 194: I think I'm not supposed to inherit from `PseudoSingleInterface`? I am not sure what the `OffsetIMS` should be

ZLLentz: The driving idea for `OffsetIMS` (bad name maybe?) was to let us separate the various unrelated psuedopositioner-like things from the main class, e.g. you shouldn't have to pass 10+ unrelated values through the same two `forward` and `inverse` calculation functions. Each different calculation can be independent, making it simpler to implement and maintain.

So `OffsetIMS` would be a motor that has an offset. That's it.

Two reasonable ways to implement this I think?
- `PseudoPositioner` with an `IMS` as the real motor and a `PseudoSingleInterface` as the offset motor (e.g. the Si and C motors), there might already be a good base class for this? Maybe not though.
- A subclass of `IMS` with an offset PV built in and move/readback adjusted appropriately (actually this way sounds messy)

pcdsdevices/lodcm.py
ZLLentz: I think it's OK if the calculation class has some redundancy... this mega device would be very hard to write without it. I wish there was an easy way to have something like a "reference" component, e.g. a tool like:
```
class Energy(PseudoPositioner):
    energy = Cpt(PseudoSingleInterface)
    th1 = RefCpt('{parent.tower1.th1}')
```
Would be neat if it existed/were possible in the `ophyd` framework

cristinasewell: Yes, this would be very nice, I agree

pcdsdevices/lodcm.py
```diff
@@ -60,6 +69,452 @@ def __init__(self, prefix, *args, **kwargs):
...
+class OffsetIMS(PseudoPositioner):
+    """
+    IMS motor with an offset.
+
+    The driving idea for OffsetIMS is to let us separate the
+    various unrelated PseudoPositioner-like things from the main class.
+    Each different calculation can be independent, making it simpler to
+    implement and maintain.
+    """
+    motor = FCpt(IMS, '{self._prefix}', kind='normal')
```
cristinasewell, line 333: @ZLLentz  and @klauer  - I still don't know if I got this little guy right, can you pretty please help when you get a chance. 

I think what we eventually want to have is simulate the old code:?

```
    def _create_crystal_vm(self, name, motor, offset_pv, use_ims_preset=False):
        offset_pv = "%s:%s" % (self.pvBase, offset_pv)

        def move(pos):
            motor.mv(pos + Pv.get(offset_pv))

        def wm():
            return motor.wm() - Pv.get(offset_pv)

        def set(value):
            if use_ims_preset:
                Pv.put("%s_SET" % offset_pv, motor.wm() - value)
            Pv.put(offset_pv, motor.wm() - value)

        return virtualmotor.VirtualMotor(
            name, move, wm, set=set, motorsobj=self._motors
        )

```
I have a 'REAL' offset motors, that have valid PVs, (`th1_c`, `th1_si`... and so on)

I either don't know how to test this guy properly or I am missing something or have some stuff wrong. 

Here is how I am testing it:

```
In [3]: from pcdsdevices.lodcm import SIMOffsetIMS

In [4]: from ophyd.sim import make_fake_device

In [5]: fake_ims = make_fake_device(SIMOffsetIMS)

In [6]: ims = fake_ims('FAKE:IMS', name='fake_ims')

In [7]: ims
Out[7]: 

fake_ims
--------
position: 0.0

motor
-----
  position: 0

offset
------
  position: 0
  notepad_setpoint: 0
  notepad_readback: 0

In [8]: ims.move(2)
Out[8]: MoveStatus(done=True, pos=fake_ims, elapsed=0.0, success=True, settle_time=0.0)

In [9]: ims
Out[9]: 

fake_ims
--------
position: 0.0

motor
-----
  position: 2

offset
------
  position: 0
  notepad_setpoint: 2
  notepad_readback: 0
```
Now for a real position of 3, I would have to move the offset at 1?
```
In [10]: ims.inverse(3)
Out[10]: FakeSIMOffsetIMSPseudoPos(offset=1)

In [11]: ims.offset.move(1)
Out[11]: MoveStatus(done=True, pos=fake_ims, elapsed=0.0, success=True, settle_time=0.0)

In [12]: ims
Out[12]: 

fake_ims
--------
position: 0.0

motor
-----
  position: 3

offset
------
  position: 0
  notepad_setpoint: 1
  notepad_readback: 0
```
Inverse seems to work, maybe.... although I don't know why the position does not get updated for offset...... 

The `forward` guys is something else though, I am doing something wrong here. So for a pseudo position of 2, i would have to move the motor at 5?
```
In [13]: ims.forward(2)
Out[13]: FakeSIMOffsetIMSRealPos(motor=5)

In [14]: ims.motor.move(5)
Out[14]: MoveStatus(done=True, pos=fake_ims_motor, elapsed=0.0, success=True, settle_time=0.0)

In [15]: ims
Out[15]: 

fake_ims
--------
position: 0.0

motor
-----
  position: 5

offset
------
  position: 0
  notepad_setpoint: 1
  notepad_readback: 0
```

The offset however does not change, where exactly am I going wrong here?

cristinasewell: Ohh and i have this guy for testing:

```
class SIMOffsetIMS(OffsetIMS):
    """Simulation of OffsetIMS device for testing."""
    motor = Cpt(FastMotor, limits=(-100, 100))

```

cristinasewell: I think this works but I don't know how legal this is:

```
    @pseudo_position_argument
    def forward(self, pseudo_pos):
        pseudo_pos = self.PseudoPosition(*pseudo_pos)
        motor_value = pseudo_pos.offset + self.offset.notepad_setpoint.get()
        return self.RealPosition(motor=motor_value)

    @real_position_argument
    def inverse(self, real_pos):
        real_pos = self.RealPosition(*real_pos)
        offset = real_pos.motor - self.offset.notepad_setpoint.get()
        return self.PseudoPosition(offset=offset)
```

pcdsdevices/lodcm.py
cristinasewell: I am running into circles here with the LODCMEnergy class:
-  if I add the Offset motors after the `super().__init...` then it will not recognize them as real motors when `return self.RealPosition(...)`
 - if I define them as real components before at `init` then  `inverse` and `forward` will complain that I am missing some real motors, because I don't set all the motors, only set half of them depending of `Si` or `C`

cristinasewell: I only deal with `Offset Motors` when setting/getting energy..

pcdsdevices/lodcm.py
klauer: I think one of us ophyd experts should work on providing a generalized "`PseudoPositioner` with a user offset" - it's something that _should_ be easy but isn't just yet. `DelayBase` is probably the closest thing so far: https://github.com/pcdshub/pcdsdevices/blob/b0a85fb42effa63bf29e3c8f9b668671f6e60df3/pcdsdevices/pseudopos.py#L229

That may mean a further a delay in getting this PR in, though, if that's acceptable...

silkenelson: I think it is: we are nearly done with running in 2020 and there is now the target alignment & shooting project to work on as well.

cristinasewell: I am sorry and thank you. 🙏

pcdsdevices/lodcm.py
```diff
@@ -60,6 +71,683 @@ def __init__(self, prefix, *args, **kwargs):
...
+        if length == 0:
+            # don't bother transforming this
+            # TODO maybe catch error in common.wave.. when send 0
+            return 0
+        # TODO i might not need to devide here by 1000
+        energy = common.wavelength_to_energy(length) / 1000
+        return self.PseudoPosition(energy=energy)
+
+
+class LODCMEnergySi(PseudoPositioner):
```
cristinasewell, line 598: I split the Energy motors into `Si` and `C` motors - because the `energy` motor is dealing with 8 motors in total, and only 4 at the time depending on the material. So when placing these 8 motors into one PseudoPositioner and using only 4 out of them - it will complain that I am not using the other ones.

I think eventually what users would want to use is only one `energy` motor (as of now i have `energy_c` and `energy_si`)- so I could maybe reassign the `energy` motor at `LODCM __init__` time somehow - based on a function that will let me know if it is `C` or `Si` ....
```
if material == 'C':
    self.energy = self.energy_c
elsif material == 'Si':
    self.energy = self.energy_si

```
or maybe I could go back before I split this motor into two, and in there possibly redefine the `_real` motors to only have 4 at the time.... 

Any suggestions here?

ZLLentz: Before suggesting anything, I'd like to note a few issues:
- There is some prefix mishap in these classes: the `dr` is supposed to be `XPP:MON:MMS:19` and the `th1_c` is supposed to be `XPP:LODCM:TH1:OFF_C`
- The `OffsetIMS` here are not actually motors, they are just pointing to PVs that store a value:
```
$ caget XPP:LODCM:TH1:OFF_C.RTYP
XPP:LODCM:TH1:OFF_C.RTYP       ao
```
That would say something like `ims` or `motor` if it was a motor:
```
$ caget XPP:MON:MMS:18.RTYP
XPP:MON:MMS:18.RTYP            ims
```

Given this info, I think these two classes can be combined into one that checks whether we're at Si or C and then does the calculation appropriately, if I'm understanding correctly.

cristinasewell: > Before suggesting anything, I'd like to note a few issues:
> 

Thanks for pointing this out - that's correct, I need to change that.
> * There is some prefix mishap in these classes: the `dr` is supposed to be `XPP:MON:MMS:19` and the `th1_c` is supposed to be `XPP:LODCM:TH1:OFF_C`

Ok thank you, I need to revisit the `MotorOffsetBase` class to see what I have done there...
> * The `OffsetIMS` here are not actually motors, they are just pointing to PVs that store a value:
> 
> ```
> $ caget XPP:LODCM:TH1:OFF_C.RTYP
> XPP:LODCM:TH1:OFF_C.RTYP       ao
> ```
> 
> That would say something like `ims` or `motor` if it was a motor:
> 
> ```
> $ caget XPP:MON:MMS:18.RTYP
> XPP:MON:MMS:18.RTYP            ims
> ```
> 
So here I would have to check if is `Si` or `C` at `__init__` time - which does not sound very good, because I don't know in advance what material we're dealing with... or even if we know, we might want to change it after initialization?
The reason for that, is because the `forward` method will complain for not using all the `_real` motors.

Ex:
if `C`:
     use only 4 motors (`C` motors)
if `Si`:
    use the other 4 motors (`Si` motors)

Are we able to re-define what the `_real` motors are in our class after we initialize the class?
> Given this info, I think these two classes can be combined into one that checks whether we're at Si or C and then does the calculation appropriately, if I'm understanding correctly.

ZLLentz: You use the same 4 motors in both cases. The `OFF_C` constructs are not motors, they just store values.

cristinasewell: Ohh ok 🤦 , I'd just use the `th1`, `th2` `z1` and `z2` (maybe `dr` as well) these are the motors - I will come back for more questions then - thanks @ZLLentz

cristinasewell: Ok back to this guy, can't wait to be done with it. 😓 

Now, if I understand  correctly (probably not yet) our `OffsetMotor` will have the `real_motor` as eg: `th1` and the `pseudo_motor` as `th1_c` **OR** `th1_si` - depending on the material? 
(I have a version of this OffsetMotor that is a PsedoPositioner with an IMS - motor and PseudosingleInterface as the offsets - either `OFF_Si` or `OFF_C`)

Maybe I'm still getting the OffsetMotor architecture wrong, but based on what I understand, we'll have to define these 4 motors at init time, including their `OFF_C` or `OFF_Si` PVs correct?

Something like this:
```
    # thata movement
    th2 = FCpt(OffsetMotor, '{self._th2_offset_prefix}',
               motor_prefix='{self._m_prefix}:MON:MMS:13',
               add_prefix=('prefix', 'motor_prefix'),
               kind='normal', doc='LOM Xtal2 Theta')
```

where `self._th2_offset_prefix` is either `f'{self._prefix}:TH2:OFF_C'` **OR** this one `f'{self._prefix}:TH2:OFF_Si'` 


If I have to do this there is no way for me to know at `__init__` time if we're dealing with `Si` or `C` ... are there ways around this?

ZLLentz: That's definitely one way to do it. With the offset motor approach (which I had completely forgotten about in my previous comment) you'd need to include both sets of offset motors and decide which ones to link to the energy dynamically based on the Si/C states.

The other way is to drop the offset motors and handle the offset PVs directly in the energy calculations.

tests/test_lodcm.py
```diff
@@ -331,77 +318,61 @@ def test_get_energy_si(fake_energy_si, monkeypatch):
...
+
+def test_forward_si(fake_energy_si, monkeypatch):
+    energy = fake_energy_si
+
+    energy.th1.set_current_position(-23)
+    with patch("pcdsdevices.lodcm.LODCMEnergySi.get_reflection",
+               return_value=(1, 1, 1)):
         res = energy.get_energy()
         assert res == 5.059840436879476
+        res = energy.forward(5.059840436879476)
```
cristinasewell, line 334: The forward function is giving back what the 5 motor positions should be when the pseudo_pos is 5.0598....
It seems like it gives me back the correct stuff, however when I try to `move` the energy - it does not change the motor positions - this could be because of me not setting up correctly the `fake_lodcm` or the `fake_energy` objects.... 

Is there another easier way to check if the `move` is actually working somehow?

ZLLentz: The motor's you're using in the test aren't actually motors, they are just collections of PVs that get their `user_setupoint` set when you try to move them and do no other changes. So as long their setpoints are changing I think it works.

If you want to see them "working" we need to actually sub in a simulated motor object instead of just replacing all the PVs with fake PVs like is done currently.

cristinasewell: Got it, I'd have to use the good old `FastMotor`  for every IMS I have in my case and maybe others.... ok, thank you

pcdsdevices/lodcm.py
```diff
@@ -85,42 +853,131 @@ class LODCM(BaseInterface, Device):
...
+        self.x2_si = self.tower2.x2_si
+        self.y2_c = self.tower2.y2_c
+        self.y2_si = self.tower2.y2_si
+        self.chi2_c = self.tower2.chi2_c
+        self.chi2_si = self.tower2.chi2_si
+        self.h2n_c = self.tower2.h2n_c
+        self.h2n_si = self.tower2.h2n_si
+
+    @property
+    def energy(self):
```
cristinasewell, line 953: I think I should not have this property here like that.... might not be good practice? since it is an actual object component not a simple property...

The idea here is that I wanted the user to be able to do something like this:
```
lom.energy.move(10)
```
instead of 
```
lom.energy_si.move(10) or lom.energy_c.move(10)
```
but there's probably a better way of dealing with this

cristinasewell:
We should probably open more Issues with additional stuff for `LODCM` object - this PR is getting a bit large.

The following might be something scientist might want?:

`tweakX`:
https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xpp/blinst/lom.py#L1133

`tweakParallel`:
https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xpp/blinst/lom.py#L1165

`setE`:
https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xpp/blinst/lom.py#L1197

`others...`
..but I think we should also get some feedback from scientists about what else we need to include here, as I am not sure yet what else they use 

@ZryletTC  should I change the naming into this PR? from `th1_c` to `th1C` and all the other ones? just so we get that out of the way?

pcdsdevices/lodcm.py
```diff
@@ -60,6 +70,763 @@ def __init__(self, prefix, *args, **kwargs):
...
+
+    Assume material is 'Si' without checking if the crystal's materials are
+    aligned for both towers.
+    """
+    tower1 = FCpt(CrystalTower1, '{self._prefix}', kind='normal')
+    tower2 = FCpt(CrystalTower2, '{self._prefix}', kind='normal')
+    dr = FCpt(IMS, '{self._m_prefix}:MON:MMS:19',
+              kind='normal', doc='LOM Dia Theta')
+
+    th1 = FCpt(OffsetMotor, prefix='{self._prefix}:TH1:OFF_Si',
```
cristinasewell, line 495: I should probably change these guys just to be consistent - from `th1` to `th1_Si` or `th1Si` eventually - they are after all the offset motors - but in the same time they are under the `LODCMEnergySi` object, so obviously they'll be the `Si` offsets

pcdsdevices/lodcm.py
```diff
@@ -60,6 +70,763 @@ def __init__(self, prefix, *args, **kwargs):
...
+                    * diffraction.d_space('C', reflection))
+        if length == 0:
+            # don't bother transforming this
+            # TODO maybe catch error in common.wave.. when send 0
+            return 0
+        energy = common.wavelength_to_energy(length) / 1000
+        return self.PseudoPosition(energy=energy)
+
+
 class LODCM(BaseInterface, Device):
```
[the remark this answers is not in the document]
ZLLentz: I think it's OK to not go through and add all 50

pcdsdevices/lodcm.py
ZLLentz: I think this is a reasonable solution to the problem, the only other one I can think of is an "energy" motor object that picks whether to forward the commands to instead of a property that returns one or the other. But I'm not sure which approach is better.

Either way, the only thing that seems "wrong" here is returning `None` sometimes (and maybe having properly access wait for EPICS, that's usually not great too)

cristinasewell: Ok I can probably raise an error here instead of returning `None`... ? 
Can you please elaborate on this: ***(and maybe having properly access wait for EPICS, that's usually not great too)***

ZLLentz: I think raising an exception or just picking one to use as the default are both fine.

When you access a python property, there is some assumption that it will be "fast" (e.g., just checking a value)

In this case, accessing `lodcm.energy` can be slow if EPICS is being slow, since you have to check the material type before deciding which underlying object to return.

cristinasewell: Ohhh got it, thank you. 

I'm just going to pick one as default I think for now - but this won't solve the wait access on the property - so maybe your previous suggestion with an `energy` proxy motor object might be better, i don't know for sure how it would look like yet

ZLLentz: It's probably correct to not worry about that so much for now and consider changing it in the future if it becomes a problem. A lot of times these sorts of things never actually become a problem.

cristinasewell: Ok sounds good, pasting here for future reference - I think it will look similar to this example here: https://ruddra.com/python-proxy-object/

tests/test_lodcm.py
```diff
@@ -76,6 +194,197 @@ def test_hutch_foils():
...
+    lodcm = fake_lodcm
+    with patch("pcdsdevices.lodcm.LODCMEnergySi.get_reflection",
+               return_value=(1, 1, 1)):
+
+        lodcm.energy_si.move(10, wait=False)
+        assert lodcm.energy.th1.motor.user_setpoint.get() == 11.402710639982848
+        assert lodcm.energy.th2.motor.user_setpoint.get() == 11.402710639982848
+        assert lodcm.energy.dr.user_setpoint.get() == 11.402710639982848*2
+        assert lodcm.energy.z1.motor.user_setpoint.get() == -713.4828146545175
+        assert lodcm.energy.z2.motor.user_setpoint.get() == 713.4828146545175
```
ZryletTC, line 385: Are there any top-level motion methods that should also get tested?

ZryletTC (approved):
Sorry for the long and late review, as always.
Please disregard most of it for now. I think the only thing that should be fixed before merge is the missing test. Consider everything else as notes for the next PR, which I can help with if desired.

Thank you for doing all this, @cristinasewell!!! It looks really good 😃

cristinasewell:
@ZryletTC thank you very much for the thorough review - I appreciate it 😉 
I am going to fix the Offset IMS test, and fix some little tests on lodcm part- and all the rest will probably go in the next PR as you mentioned.

tests/test_pseudopos.py
```diff
@@ -134,3 +136,20 @@ class MyLUTPositioner(LookupTablePositioner):
 
     assert lut.real.limits == (0, 9)
     assert lut.pseudo.limits == (40, 400)
+
+
+def fake_offset_ims():
```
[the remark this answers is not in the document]
cristinasewell: Thanks for catching this - I think I will be moving the test into `test_epics_motor` instead because I define this motor there anyway

tests/test_lodcm.py
```diff
@@ -1,20 +1,35 @@
...
+                               CHI1, Y2, CHI2, H2N, SimFirstTower,
+                               SimSecondTower, LODCMEnergySi, LODCMEnergyC)
 
 logger = logging.getLogger(__name__)
 
 
+def motor_setup(mot, pos=0):
+    mot.user_readback.sim_put(0)
+    mot.user_setpoint.sim_put(0)
+    mot.user_setpoint.sim_set_limits((0, 0))
```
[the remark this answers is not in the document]
cristinasewell: 1. If we don't use a fake motor (like FastMotor) for example, we have to fake a few things in order to be able to test with a motor, these below I had to setup specifically for the IMS motor, I think they don't all require the same things:
```
user_readback
user_setpoint
limits
motor_spg # to 2 - because unable to move otherwise, 2 - Go
```

2. The limits are mostly set here so we make `check_value` from `conftest` happy, otherwise we'll get a type error when trying to get the limits:
```
 low_limit, high_limit = self.limits
```
As far as limits(0, 0) - I am not sure how that works with real motors and the limits for those motors, but when testing - yes, it does seem like (0, 0) means `NO LIMITS` - or unlimited travel like you said - I could be wrong here though

pcdsdevices/pseudopos.py
```diff
@@ -459,3 +460,97 @@ def inverse(self, real_pos: tuple) -> tuple:
...
+        pseudo_pos : PseudoPosition
+            The pseudo position input, a namedtuple.
+
+        Returns
+        -------
+        real_pos : RealPosition
+            The real position output, a namedtuple.
+        """
+        pseudo_pos = self.PseudoPosition(*pseudo_pos)
+        motor_value = pseudo_pos.pseudo_motor + self.user_offset.get()
```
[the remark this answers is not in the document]
cristinasewell: It could be that I got multiple things wrong here and they all work out together somehow.
It should probably be something like this instead:
```
def forward():
.....
motor_value = pseudo_pos.pseudo_motor - self.user_offset.get()
```

```
def inverse():
...
offset = real_pos.motor + self.user_offset.get()
```

```
def set_current_position(position):
...
new_offset = self.position[0] - position
```

I might have gotten them all reversed?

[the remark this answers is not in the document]
cristinasewell: https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xpp/blinst/lom.py#L446-L462

pcdsdevices/lodcm.py
```diff
@@ -60,6 +70,760 @@ def __init__(self, prefix, *args, **kwargs):
...
+                self.chi2_state.position == 'C')
+
+    def is_silicon(self):
+        """Check if tower 2 is with Silicon (Si) material."""
+        return ((self.h2n_state.position == 'Si' or
+                 self.h2n_state.position == 'OUT') and
+                self.y2_state.position == 'Si' and
+                self.chi2_state.position == 'Si')
+
+    def get_reflection(self, as_tuple=False, check=False):
```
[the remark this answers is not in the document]
cristinasewell: I got rid of the check from `get_material` - that one was more annoying and easier to get rid of.
I think in the next PR i should get rid of this one too, but I need to know what exactly we want to do here, since this check was a bit more involved

pcdsdevices/pseudopos.py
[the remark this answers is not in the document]
cristinasewell: @ZryletTC  do you think I should change this?

pcdsdevices/epics_motor.py
```diff
@@ -570,6 +570,14 @@ class DelayNewport(DelayBase):
     motor = Cpt(Newport, '')
 
 
+class OffsetMotor(OffsetMotorBase):
```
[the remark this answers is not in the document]
cristinasewell: Yes, I think it should be OffsetIMS instead - will change this in next PR - thank you

pcdsdevices/epics_motor.py
```diff
@@ -570,6 +570,14 @@ class DelayNewport(DelayBase):
...
+    motor = FCpt(IMS, '{self._motor_prefix}')
+
+    def __init__(self, prefix, motor_prefix, *args, **kwargs):
+        self._motor_prefix = motor_prefix
+        super().__init__(prefix, motor_prefix, *args, **kwargs)
```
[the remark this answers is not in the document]
cristinasewell: Took these guys out in 171f6be - thank you (sorry I did not commit them here - I missed that you actually had the suggestion, thought it was only a comment)

pcdsdevices/pseudopos.py
```diff
@@ -459,3 +460,97 @@ def inverse(self, real_pos: tuple) -> tuple:
             self._table_data_by_name[pseudo_field]
         )
         return self.PseudoPosition(**{pseudo_field: pseudo_value})
+
+
+class OffsetMotorBase(FltMvInterface, PseudoPositioner):
+    """
+    Motor with an offset.
+    """
```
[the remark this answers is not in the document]
cristinasewell: 👍

pcdsdevices/pseudopos.py
```diff
@@ -459,3 +460,97 @@ def inverse(self, real_pos: tuple) -> tuple:
...
+        ``position`` as the new current position.
+
+        Parameters
+        ----------
+        position
+            The new current position.
+        '''
+        self.user_offset.put(0.0)
+        new_offset = position - self.position[0]
+        self.user_offset.put(new_offset)
```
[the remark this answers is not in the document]
cristinasewell: I followed the `DelayBase` implementation - I assumed there was a reason for it that I don't understand yet, but I think your suggestion works as well - I just don't know why we have two `put` here.

This works so far, i think this is what you meant:
```
new_offset = position - self.motor.position
```

tests/test_pseudopos.py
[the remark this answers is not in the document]
cristinasewell: Fixed this in 171f6be - thanks a lot for catching this

tests/test_lodcm.py
```diff
@@ -1,20 +1,35 @@
...
+        return "Si"
+    monkeypatch.setattr(LODCM, 'get_material', get_material)
+
     lodcm = FakeLODCM('FAKE:LOM', name='fake_lom')
-    lodcm.h1n.state.sim_put(1)
-    lodcm.h1n.state.sim_set_enum_strs(['Unknown'] + H1N.states_list)
+    lodcm.h1n_state.state.sim_put(1)
+    lodcm.h1n_state.state.sim_set_enum_strs(['Unknown'] + H1N.states_list)
     lodcm.yag.state.sim_put(1)
     lodcm.yag.state.sim_set_enum_strs(['Unknown'] + YagLom.states_list)
```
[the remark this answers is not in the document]
cristinasewell: Ohh this is no dumb question, it is not very clear - and I might actually have some misleading comments there. 
Yes, it sets the `h1n_state` to `OUT`  here

The other fake sim_put for `y1_state` for `chi1_state` , `y2_state` and `chi2_state` - sets them to `C` ...

In the next PR I will probably change these a bit to be more clear...

tests/test_lodcm.py
```diff
@@ -23,9 +38,112 @@ def fake_lodcm():
...
+
+    return energy
+
+
+@pytest.fixture(scope='function')
+def fake_energy_c(monkeypatch):
+    FakeLODCMEnergy = make_fake_device(LODCMEnergyC)
+    energy = FakeLODCMEnergy('FAKE:ENERGY:C', name='fake_energy_c')
+
+    motor_setup(energy.th1.motor, pos=77)
```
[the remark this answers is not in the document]
cristinasewell: You are correct, I have not used it - I will make a SimLODCM and will be able to test things better in next PR- (i think)

pcdsdevices/lodcm.py
```diff
@@ -60,6 +70,760 @@ def __init__(self, prefix, *args, **kwargs):
...
+    h1n = FCpt(IMS, '{self._m_prefix}:MON:MMS:09',
+               kind='normal', doc='LOM Xtal1 Hn')
+    # paralell to the crystal surface movement
+    h1p = FCpt(IMS, '{self._m_prefix}:MON:MMS:20',
+               kind='normal', doc='LOM Xtal1 Hp')
+
+    # states
+    h1n_state = Cpt(H1N, ':H1N', kind='hinted')
+    y1_state = Cpt(Y1, ':Y1', kind='omitted')
+    chi1_state = Cpt(CHI1, ':CHI1', kind='omitted')
```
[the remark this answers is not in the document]
cristinasewell: No particular reason - should they be `normal`?

pcdsdevices/lodcm.py
```diff
@@ -60,6 +70,760 @@ def __init__(self, prefix, *args, **kwargs):
...
+                  motor_prefix='{self._m_prefix}:MON:MMS:20', kind='normal',
+                  add_prefix=('prefix', 'motor_prefix'),
+                  name='h1p_si', doc='H1p motor offset for Si [mm]')
+
+    tab_component_names = True
+    tab_whitelist = ['is_diamond', 'is_silicon', 'get_reflection',
+                     'get_material']
+
+    def __init__(self, prefix, *args, **kwargs):
+        self._m_prefix = ''
```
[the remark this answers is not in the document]
cristinasewell: `m` is for the `motor` - for some reason that made sense for me back then - but I think `_hutch_prefix` is better - will change this in next PR - thank you

pcdsdevices/lodcm.py
```diff
@@ -60,6 +70,760 @@ def __init__(self, prefix, *args, **kwargs):
...
+              kind='normal', doc='LOM Dia PIPS')
+    # yag screen
+    yag_zoom = FCpt(IMS, '{self._m_prefix}:MON:CLZ:01',
+                    kind='normal', doc='LOM Zoom')
+
+    tab_component_names = True
+
+    def __init__(self, prefix, *args, **kwargs):
+        self._m_prefix = ''
+        self._df_suffix = '27'
```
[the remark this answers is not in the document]
cristinasewell: The `df` component - (filters wheel i think) - has a different PV suffix in `XPP` vs `XCS` - I should add  a comment in there to explain that

```
XPP:MON:MMS:27
```
vs

```
HFX:MON:MMS:22
```

tests/test_lodcm.py
```diff
@@ -76,6 +194,197 @@ def test_hutch_foils():
...
+    with patch("pcdsdevices.lodcm.LODCMEnergySi.get_reflection",
+               return_value=(1, 1, 1)):
+        res = energy.get_energy()
+        assert res == 5.059840436879476
+        res = energy.forward(5.059840436879476)
+        received = []
+        for i in res:
+            received.append(i)
+        expected = [23, 23, 46, -289.70663244212216, 289.70663244212216]
+        assert np.allclose(sorted(received), sorted(expected))
```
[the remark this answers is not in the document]
cristinasewell: The `forward` does not necessarily give me back the real positions in the order I set them - maybe it is doing a sort itself (of just the fake object is)?

For example in `forward` i have this:
```
 return self.RealPosition(th1=th,
                                 th2=th,
                                 z1=-z,
                                 z2=z,
                                 dr=2*th)
```
but when I use it I get this:
```
FakeLODCMEnergySiRealPos(dr=46.00000000000001, th1=23.000000000000004, th2=23.000000000000004, z1=-289.70663244212216, z2=289.70663244212216)
```


But yes, now that I think about it - it is not a good idea to sort them - could be fooling myself

tests/test_lodcm.py
```diff
@@ -76,6 +194,197 @@ def test_hutch_foils():
...
+    assert tower1.is_silicon()
+    tower1.h1n_state.move('Si')
+    assert tower1.is_silicon()
+    # tower1 c: y1_state = C, chi1_state = C, h1n = C or Out
+    tower1.h1n_state.move('OUT')
+    tower1.y1_state.move('C')
+    tower1.chi1_state.move('C')
+    assert tower1.is_diamond()
+    tower1.h1n_state.move('C')
+    assert tower1.is_diamond()
```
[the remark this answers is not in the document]
cristinasewell: Sounds good - also just realized that `h2n_state` - does not have an `OUT` position - do you remember how we decided to have a similar check for Tower 2, like tower 1 - to determine if it is Diamond or Silicon - that's why we don't check the same way for tower 2 - i will switch back

tests/test_lodcm.py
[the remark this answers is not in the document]
cristinasewell: changed the way I test this in fa843ab

pcdsdevices/lodcm.py
```diff
@@ -187,3 +1043,350 @@ def remove_dia(self, moved_cb=None, timeout=None, wait=False):
...
+        material = material or self.get_material()
+        reflection = reflection or self.get_reflection(
+            as_tuple=True, check=True)
+        if material == 'Si':
+            return self.energy_si.get_energy()
+        elif material == 'C':
+            return self.energy_c.get_energy()
+        else:
+            raise ValueError('Cannot initialize energy motor because could not'
+                             ' determine the material.')
```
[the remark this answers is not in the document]
cristinasewell: Nope, thanks for pointing this out - change the wording a bit in fa843ab

pcdsdevices/lodcm.py
```diff
@@ -85,42 +850,133 @@ class LODCM(BaseInterface, Device):
     mono_line : str, optional
         Name of the mono, double-bounce beamline.
     """
-
-    h1n = Cpt(H1N, ':H1N', kind='hinted')
+    # h1n_state = Cpt(H1N, ':H1N', kind='hinted')
```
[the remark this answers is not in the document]
cristinasewell: removed it in fa843ab

pcdsdevices/lodcm.py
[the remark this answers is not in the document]
cristinasewell: I added a comment about this in fa843ab

pcdsdevices/lodcm.py
```diff
@@ -60,6 +70,760 @@ def __init__(self, prefix, *args, **kwargs):
...
+
+class Y2(InOutRecordPositioner):
+    states_list = ['C', 'Si']
+    in_states = ['C', 'Si']
+    out_states = []
+
+
+class CrystalTower1(BaseInterface, Device):
+    """
+    Crystal Tower 1.
```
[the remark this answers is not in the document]
cristinasewell: Ohh ok.... we'll need to decide how we want these too look then

cristinasewell:
Ok I think I am going to stop here with the changes in this PR, and change other suggestions in the follow up PR.

pcdsdevices/epics_motor.py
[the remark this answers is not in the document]
cristinasewell: I need to double check here if this works though - might need to put them back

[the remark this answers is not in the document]
cristinasewell: Added this back in e2feed0 - because it was not working without it - if this still needs to be taken out, we can change it in the next PR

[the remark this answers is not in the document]
ZryletTC: I understand why you need those lines but they are already in OffsetMotorBase, right? Does the inheritance not work? :/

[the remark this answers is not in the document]
cristinasewell: Ohh you are correct @ZryletTC  - what happen is I got rid of this guy too: 
```
motor = FCpt(IMS, '{self._motor_prefix}')
```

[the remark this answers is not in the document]
cristinasewell: Thank you :)

pcdsdevices/pseudopos.py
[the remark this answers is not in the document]
ZryletTC: @klauer Any reason you wrote it that way?

ZLLentz:
Ready to merge then? Or wait a short bit?

cristinasewell:
> Ready to merge then? Or wait a short bit?

I am ready to merge this, unless @ZryletTC thinks I should be doing some more changes here?

pcdsdevices/pseudopos.py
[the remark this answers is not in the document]
ZryletTC: No it's not something that you did. The old code has it like that as well. The issue is that this convention is the opposite of how we run most other offset values.

ZLLentz:
OK just standing by then for Mr. Travis to do his thing

pcdsdevices/lodcm.py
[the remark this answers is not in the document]
ZryletTC: Sorry, I wasn't very clear. I meant why is the `_m_prefix` handled differently than `_df_prefix`?
As in why is df set to `27` instead of `''` before the `if` with the `self._df_suffix = '27'` moved to inside the if condition

ZLLentz:
:+1: good work on this one, hopefully we can minimize the number of 3 month pull requests going forward!

pcdsdevices/pseudopos.py
```diff
@@ -459,3 +460,97 @@ def inverse(self, real_pos: tuple) -> tuple:
...
+    def __init__(self, prefix, motor_prefix, * args, **kwargs):
+        if self.__class__ is OffsetMotorBase:
+            raise TypeError('OffsetMotorBase must be subclassed with '
+                            'a "motor" component, the real motor to move.')
+        self._motor_prefix = motor_prefix
+        super().__init__(prefix, *args, **kwargs)
+
+    @pseudo_position_argument
+    def check_value(self, value):
+        return super().check_value(value)
```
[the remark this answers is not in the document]
klauer: Yes, this should be an ophyd-level fix I believe.

pcdsdevices/pseudopos.py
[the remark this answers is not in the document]
klauer: I recall borrowing this from another implementation

klauer:
👏 (Didn't see I was tagged on the above things, sorry!)

pcdsdevices/lodcm.py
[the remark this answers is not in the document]
cristinasewell: Ohh got it, no reason at all, one was set to default to one and the other was not - I should probably change them to have similar implementations

tests/test_lodcm.py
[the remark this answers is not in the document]
cristinasewell: Added more comments in https://github.com/pcdshub/pcdsdevices/pull/766/commits that should make it clearer now about those states.

pcdsdevices/lodcm.py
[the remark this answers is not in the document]
cristinasewell: Changed the behavior in https://github.com/pcdshub/pcdsdevices/pull/766/commits

tests/test_lodcm.py
```diff
@@ -34,20 +152,20 @@ def test_lodcm_destination(fake_lodcm):
...
+    lodcm.h1n_state.move('C')
     assert len(lodcm.destination) == 2
     # Block the mono line
     lodcm.yag.move('IN')
     assert len(lodcm.destination) == 1
-    lodcm.h1n.move('Si')
+    lodcm.h1n_state.move('Si')
     assert len(lodcm.destination) == 0
     lodcm.yag.move('OUT')
     assert len(lodcm.destination) == 1
```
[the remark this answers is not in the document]
cristinasewell: I have added some extra lines that test the `lodcm.main_line` and `lodcm.mono_line` here: https://github.com/pcdshub/pcdsdevices/pull/766/commits/7556127dd33e120a638158b57380b6bea5e5606a

I left the `len` the way it is, because that was somebody else's tests, and also because I think it might be important to test to make sure only one line is in certain situation, or no line at all or both...

PR #717, opened 2020-12-02
Lxt devices update

ZryletTC:
## Description
Primarily, this adds notepad signals to LaserTiming and DelayBase devices.
It also fixes the reversal of the ReversedTimeToolDelay class to properly handle offsets and adjusts limits to allow for negative delay values.

## Motivation and Context
Makes these devices easier to use and collect data from.

## How Has This Been Tested?

Tested interactively in xpp3 while monitoring the notepad PVs.

I think docstring have been updated appropriately. Please lmk if I missed something or wasn't very clear.

## Pre-merge checklist
- [ ] Add some paragraphs to the docs about these devices


ZryletTC:
This was initially working til @silkenelson helped me realize that the `OphydReadback` I had set up was really more of a `OphydSetpoint` so I wrote 66ee9b2 to address that, changing the signal over to setpoint and creating a new PV for the readback. This, however, did not work and the setpoint seems to update after the readback does:
```
LAS:FS11:lxt:OphydReadback     2020-12-01 21:37:13.181064 1e-12  
LAS:FS11:lxt:OphydSetpoint     2020-12-01 21:37:13.181300 1e-12  
```
Maybe I just need to look at it again in the morning but if you notice something I did wrong, please point it out. There's probably also a better way to be doing these subscriptions. If so, please also let me know and I'll try to do it the right way.
Thanks :)

pcdsdevices/lxe.py
```diff
@@ -286,18 +298,40 @@ def __init__(self, prefix='', *, egu=None, **kwargs):
...
-
-    @_user_offset.setter
-    def _user_offset(self, value):
         self.setpoint.user_offset = value
 
+    @done.sub_value
+    def	_update_position(self, value=None, **kwargs):
+        """The move was completed. Update the notepad readback."""
+        if value == self.done_value:
+            try:
```
klauer, line 312: Now that we're repeating this block in more than a couple places, it may be worth adding as a method on the `NotepadLinkedSignal` itself - what do you think? `put_safely(value)`/`put_nofail` or something?

klauer:
Subscriptions look fine. 

And as an aside, I'm glad you're taking care of this on the beamline with access to the device - it's been a very difficult device to get a correct description of. Maybe when it's finally in fully working shape, would you mind adding even a short paragraph of documentation to the class?

pcdsdevices/lxe.py
ZryletTC: Ahh, yes, that's a good idea.

ZryletTC:
Thanks for taking a look. Yeah, I'll try adding some documentation to help make it clearer what these devices do and why we need these weird classes.

- Got it, fixed in 4ebf699. Although I wish I didn't hard code the 0.
- Yeah, _setup_move definitely seems like a better spot for the setpoint. Moved in new commit.
- A PVPositionerWithNotepadSignals would be nice. I'll make an issue for it.

One of those two changes or the combination of both seemed to fix it (my money is on the _setup_move change). The setpoint now updates a whole 10ms before the readback. Thanks for the help. I'll get started on making this PR ready for merge (including the better docs) later today.

klauer:
Great! Glad it seems to be working better 👍

ZLLentz:
I see pre-rel docs, is this ready for review?

ZryletTC:
Unfortunately, no. Hoping to get it ready either tonight or tomorrow.

ZryletTC:
I was going to try and create a doc page about these devices but I'm not going to get through that before the end of the day so I'm just going to create an issue and push it until after break. Please critique at will.

tests/test_lxe.py
```diff
@@ -0,0 +1,307 @@
...
+    That signal's value is off by 10e9 for the ns to s conversion
+    and negative due to the convention that positive lxt means earlier laser.
+    """
+
+    logger.debug('test_laser_timing_delay')
+    assert lxt.wm() == 0
+    assert lxt._fs_tgt_time.get() == -0
+    lxt.mv(1e-6)
+    assert lxt.wm() == 1e-6
+    assert lxt._fs_tgt_time.get() == pytest.approx(-1000)
```
klauer, line 169: `pytest.approx` is new to me, and I'm its newest fan 👍

pcdsdevices/lxe.py
klauer: Issue created for this; will resolve it now.

klauer (approved):
Nothing further jumps out as problematic. Most importantly, it's working on the beamline in your local testing. Thanks again for fixing my mess 👍 😁

ZryletTC:
Kicking travis...

klauer:
Uncle Travis rang. He wants his money. (Could really use that stimulus, `@`usgovt, ...)

klauer:
Pennies have been scraped together; Travis is satisfied. Let's try this again...

PR #718, opened 2020-12-03
MNT: add get_status_float function

cristinasewell:
Added a new function in `utils`: `get_status_float` that is very similar to the existing one `get_status_value` except that this one is formatting the value with the precision that was requested by the user. 

## Motivation and Context

I keep having to format multiple numbers with similar precision of 4 when overriding the `format_status_info`
```
{value:.4f}
```
So this will make it simpler and has an additional check before formatting the value to make sure the value is a float in the first place.
## How Has This Been Tested?

Unit test.

Docstring

## Pre-merge checklist


klauer (approved):
Looks good! Assuming tests pass, feel free to self-merge.

ZLLentz (approved):

PR #720, opened 2020-12-05
ENH: Included Exit slits class

ghalym:
New class added to slits.py for the exit slits.
## Description
New class added to slits.py for the exit slits. The class contains all the motion pvs, RTDs, Camera and CIL pvs required for the exit slits.

## How Has This Been Tested?
This has been tested from my local directory.

image

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/issue_number-add_ExitSlits.rst
+++ b/docs/source/upcoming_release_notes/issue_number-add_ExitSlits.rst
@@ -27,4 +27,4 @@ Maintenance
 
 Contributors
 ------------
-- N/A
+- ghalym
--- a/docs/source/upcoming_release_notes/issue_number-add_ExitSlits.rst
+++ b/docs/source/upcoming_release_notes/issue_number-add_ExitSlits.rst
@@ -1,0 +1,30 @@
+issue_number add ExitSlits
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- N/A
--- a/docs/source/upcoming_release_notes/issue_number-add_ExitSlits.rst
+++ b/docs/source/upcoming_release_notes/issue_number-add_ExitSlits.rst
@@ -15,7 +15,7 @@ Device Updates
 
 New Devices
 -----------
-- N/A
+- Added `ExitSlits` device.
 
 Bugfixes
 --------
--- a/pcdsdevices/slits.py
+++ b/pcdsdevices/slits.py
@@ -24,17 +24,16 @@ from ophyd.pv_positioner import PVPositioner
 from ophyd.signal import Signal
 from ophyd.status import wait as status_wait, Status
 
-from .areadetector.detectors import (PCDSAreaDetectorEmbedded,
-                                     PCDSAreaDetectorTyphosTrigger)
+from .areadetector.detectors import (PCDSAreaDetectorTyphosTrigger)
 
-from .epics_motor import BeckhoffAxis, EpicsMotor, PCDSMotorBase
-from .interface import BaseInterface, FltMvInterface, MvInterface,LightpathMixin, LightpathInOutMixin
+from .epics_motor import BeckhoffAxis
+from .interface import (BaseInterface, FltMvInterface, MvInterface,
+                        LightpathMixin, LightpathInOutMixin)
 from .signal import PytmcSignal, NotImplementedSignal
 from .sensors import RTD, TwinCATTempSensor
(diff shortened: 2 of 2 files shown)


docs/source/upcoming_release_notes/issue_number-add_ExitSlits.rst
```diff
@@ -0,0 +1,30 @@
...
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- N/A
```
klauer, line 30:
```suggestion
- ghalym
```

docs/source/upcoming_release_notes/issue_number-add_ExitSlits.rst
```diff
@@ -0,0 +1,30 @@
...
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
```
klauer, line 18:
```suggestion
- Added `ExitSlits` device.
```

pcdsdevices/slits.py
```diff
@@ -492,3 +500,49 @@ class PowerSlits(BeckhoffSlits):
 
     rtds = DDCpt(_rtd_fields(RTD, 'rtd', range(1, 9)))
     fsw = Cpt(NotImplementedSignal, ':FSW', kind='normal')
+
```
klauer, line 503: The style check is failing as there are 3 blank lines here instead of 2 (https://pcdshub.github.io/style.html)

pcdsdevices/slits.py
```diff
@@ -492,3 +500,49 @@ class PowerSlits(BeckhoffSlits):
...
+ 
+    tab_component_names = True
+
+    lightpath_cpts = ['target']
+    _icon = 'fa.video-camera'
+
+    target = Cpt(TwinCATStatePMPS, ':YAG:STATE', kind='hinted',
+                 doc='Control of the YAG  stack via saved positions.')
+    yag_motor = Cpt(BeckhoffAxis, ':MMS:YAG', kind='normal',
+                  doc='Direct control of the Yag Stack motor.')
```
klauer, line 516: Similar to the other PR, make sure lines like this (`doc=...` here) align with the character after `(` on the previous line.
For example:
```python
    yag_motor = Cpt(BeckhoffAxis, ':MMS:YAG', kind='normal',
                    doc='Direct control of the Yag Stack motor.')
```

pcdsdevices/slits.py
```diff
@@ -492,3 +500,49 @@ class PowerSlits(BeckhoffSlits):
 
     rtds = DDCpt(_rtd_fields(RTD, 'rtd', range(1, 9)))
     fsw = Cpt(NotImplementedSignal, ':FSW', kind='normal')
+
+
+
+class ExitSlits(BaseInterface,Device,LightpathInOutMixin):
```
klauer, line 506:
```suggestion
class ExitSlits(BaseInterface, Device, LightpathInOutMixin):
```

klauer (changes requested):
Outside of style, everything looks good here.

See all of the style issues with their line numbers: https://travis-ci.com/github/pcdshub/pcdsdevices/jobs/454906950
You can also run this locally by running `flake8 pcdsdevices/slits.py`

klauer (approved):

pcdsdevices/slits.py
```diff
@@ -492,3 +499,55 @@ class PowerSlits(BeckhoffSlits):
 
     rtds = DDCpt(_rtd_fields(RTD, 'rtd', range(1, 9)))
     fsw = Cpt(NotImplementedSignal, ':FSW', kind='normal')
+
+
+class ExitSlits(BaseInterface, Device, LightpathInOutMixin):
+    tab_component_names = True
+
+    lightpath_cpts = ['target']
+    _icon = 'fa.video-camera'
```
ZLLentz, line 508: I think the only thing I would have suggested was a different lightpath icon

ZLLentz: Not that we're using lightpath or anything

silkenelson: That makes me sad....

ghalym: @ZLLentz The target is the Yag on the exit slits, and I thought that is what we use. I can change it. what icons should it be?

ZLLentz: I'm not sure, but I'm pretty sure that if someone is looking for "slits" they will glance over a video camera icon. I couldn't find a better option when I post-reviewed this (something to capture both elements- slits and camera) so I don't really have much standing here.

ghalym: I wasn't thinking of the slits. I was thinking of only the Yag in the beam path. But if we need to include the slits, then can we add two icons?

ZLLentz: That's a good point, only the yag blocks beam, the slits are more an energy selector than a beam shaper.  We should leave it for now.

PR #725, opened 2020-12-15
MNT: clean up average signal to avoid timeout messages

ZLLentz:
## Description

The logic in `AvgSignal` (used for gas detector) includes a `wait_for_connection` in a background thread that can time out in many cases with a large traceback.

This pull request replaces that instance with a straightforward `subscribe` call that does not time out, it just sits there and waits for the PV to be connected.

## Motivation and Context

I'm working though a `hutch-python` unit testing bug right now and these tracebacks are annoying.
```
Exception in thread Thread-16:
  File "/u1/zlentz/miniconda3/envs/dev/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/u1/zlentz/miniconda3/envs/dev/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/u1/zlentz/miniconda3/envs/dev/lib/python3.6/site-packages/pcdsdevices/signal.py", line 243, in _init_subs
    self.raw_sig.wait_for_connection()
  File "/u1/zlentz/miniconda3/envs/dev/lib/python3.6/site-packages/ophyd/signal.py", line 1032, in wait_for_connection
    self._ensure_connected(self._read_pv, timeout=timeout)
  File "/u1/zlentz/miniconda3/envs/dev/lib/python3.6/site-packages/ophyd/signal.py", line 1011, in _ensure_connected
    raise TimeoutError(f"{pv.pvname} could not connect within "
TimeoutError: GDET:FEE1:241:ENRC could not connect within 1.0-second timeout.
```
## How Has This Been Tested?

Hopefully the tests still pass, it looks pretty benign.

Added a pre-release docs page

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/725-AvgSignal-Spam.rst
+++ b/docs/source/upcoming_release_notes/725-AvgSignal-Spam.rst
@@ -1,0 +1,32 @@
+725 AvgSignal-Spam
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- AvgSignal will no longer spam exceptions text to the terminal when the signal
+  it is averaging is disconnected. This will primarily be noticed in the
+  BeamStats class, loaded in every hutch-python session.
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- zllentz
--- a/pcdsdevices/signal.py
+++ b/pcdsdevices/signal.py
@@ -12,7 +12,7 @@ if __name__ != 'pcdsdevices.signal':
 import logging
 import numbers
 import typing
-from threading import RLock, Thread
+from threading import RLock
 
 import numpy as np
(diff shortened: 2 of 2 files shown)


klauer:
Seems fine to require it to be connected at the start.

If for whatever reason we need to keep the old functionality, adding a `SUB_META` subscription, checking for connectivity, and then calling `subscribe()` could be a (complicated) improvement.

ZLLentz:
In the current formulation (pre this PR) it initializes just fine but dumps that traceback into the terminal if it cannot connect to the PV.

It is my understanding that, since I really just care about the values coming in, the connection wait is redundant. When (if) we connect, the subscribe call will ensure the values start coming into the callback.

ZLLentz:
This worked very well interactively in both the non-connecting case and the connecting case.
I think this also fixes a bug where the `AvgSignal` never starts working ever if the PV begins as disonnected.

klauer (approved):
Looks good to me.

I'm not sure where my recollection that `signal.subscribe()` blocked when disconnected came from - faulty memory...

ZLLentz:
I think it used to block! Which is why this code block existed in its original state.

PR #727, opened 2020-12-15
MNT: add support to save presets to XYTargetGrid and map points

cristinasewell:
## Description
A new object has been added to the targets file, the `XYGridStage` - not a "true" `ophyd` object - this will probably get refactor in the future.
This object is supporting mapping points from a Sample wafer mounted on an XY Grid Stage. It also supports saving those mapped points into a `yaml` file along with some additional information about the targets.

A summary (steps) of how somebody would use this object:
[Testing the fixed target scan in xppopr.pdf](https://github.com/pcdshub/pcdsdevices/files/5890048/Testing.the.fixed.target.scan.in.xppopr.pdf)

## Testing locally:
```
import matplotlib.pyplot as plt
from pcdsdevices.targets import XYGridStage as xy
from pcdsdevices.sim import FastMotor
import pcdsdevices
from bluesky import RunEngine
from bluesky.utils import install_kicker
RE = RunEngine({})
install_kicker()
from bluesky.callbacks.best_effort import BestEffortCallback
bec = BestEffortCallback()
RE.subscribe(bec)
from ophyd.sim import det4

x = FastMotor()
y = FastMotor()
s = FastMotor()

pcdsdevices.interface.setup_preset_paths(hutch='presets')
path='/home/cristina/workspace/pcdsdevices/samples.yml'
stage = xy(x_motor=x, y_motor=y, m_points=101, n_points=4, path=path)

xx, yy = stage.map_points()

stage.save_grid('test4')
# Now you should have a test sample to work with - you can use this `test4` sample and its path 
# to do scanning with the `fixed_target_scan` function from `nabs.plans`
```

The sample file can contain multiple samples, and this is how a sample would look like:
```
test3:
  time_created: '2021-01-28 11:32:52.007296'
  top_left:
  - -6.199999999999994
  - 13.599999999999968
  top_right:
  - -3.000000000000001
  - 13.599999999999968
  bottom_right:
  - -3.300000000000001
  - 23.600000000000065
  bottom_left:
  - -6.199999999999994
  - 23.700000000000067
  M: 101
  N: 4
  coefficients:
  - -6.199999999999994
  - 3.199999999999993
  - 0.0
  - -0.30000000000000027
  - 13.599999999999968
  - 0.0
  - 10.1000000000001
  - -0.10000000000000142
  xx:
  - pos: -6.199999999999994
    status: true
  - pos: -5.133333333333329
    status: true
  - pos: -4.0666666666666655
    status: true
  - pos: -3.000000000000001
  - ...
   yy:
  - pos: 13.599999999999968
    status: true
  - pos: 13.599999999999968
    status: true
  - pos: 13.599999999999968
  - ...
```

Docstrings in the code.
Also have provided operators with simple steps/files for how to use this code.

## Pre-merge checklist


pcdsdevices/targets.py
```diff
@@ -260,3 +264,122 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+        try:
+            # create presets
+            self.x.presets.add_hutch(value=pos[0], name="x_bottom_left")
+            self.x.presets.add_hutch(value=pos[2], name="x_top_left")
+            self.x.presets.add_hutch(value=pos[4], name="x_top_right")
+            self.y.presets.add_hutch(value=pos[1], name="y_bottom_left")
+            self.y.presets.add_hutch(value=pos[3], name="y_top_left")
+            self.y.presets.add_hutch(value=pos[5], name="y_top_right")
+        except AttributeError:
+            logger.warning('No folder setup for motor presets. '
```
klauer, line 304: It'd be unfortunate to get to this point, after picking 3 points for it to say "sorry - it isn't setup correctly!"

pcdsdevices/targets.py
```diff
@@ -260,3 +264,122 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+        # The meshgrid function returns
+        # two 2-dimensional arrays
+        xx, yy = np.meshgrid(x_grid_points, y_grid_points,
+                             sparse=False, indexing='xy')
+
+        plt.plot(xx, yy, marker='.', color='k', linestyle='none')
+        plt.show()
+
+        # flat out the arrays of points
+        flat_xx = list(chain.from_iterable(xx))
```
cristinasewell, line 381: This flattening process is not placing the points in the list the way we want..... (snake-style)

So I think I should not be flattening the 2d arrays here at all, and return a 2d arrays for x points, and a 2d array for y points? and deal with the rest in the scan function. (`nabs`)

klauer: Mind pasting the plot here? Haven't gotten around to running this yet

pcdsdevices/targets.py
cristinasewell: yes this would be very unfortunate indeed.... thank you

pcdsdevices/targets.py
cristinasewell: This is the plot that I get from here:
```
    xx, yy = np.meshgrid(x_grid_points, y_grid_points,
                             sparse=False, indexing='xy')

        plt.plot(xx, yy, marker='.', color='k', linestyle='none')
        plt.show()
```
My presets are these:
`bottom_left`: `(-100, 1316)`
`top_left`: `(-100, -100)`
`top_right`: `(2036, -100)`

This plot is only showing the grid points:
point_plot

And this one is with this guy:
```
plan = list_scan([det], motor1, xx, motor2, yy)
plot_raster_path(plan, 'motor1', 'motor2', probe_size=.001)
```
scan_points

not looking too good

pcdsdevices/targets.py
cristinasewell: Is there a neat way of checking if this is setup or do I just try to set it right after I asked for the first point and see if this works?

pcdsdevices/targets.py
cristinasewell: With the addition of `get_snake_grid_list` in ba1e39f I get something like in this picture below, this function is not too neat, i'm thinking that python must have something better for such thigs...
snake_like

pcdsdevices/targets.py
klauer: `if not hasattr(self.y, 'presets')` or something, perhaps?

pcdsdevices/targets.py
klauer: There may be something that could serve as a good example in bluesky itself for cleaning up the snaking code. 

Perhaps you could use bluesky's snaking code calculations, then fix up the resulting points based on the transformation matrix determined by the user-selected points? Easier said than done, of course...

pcdsdevices/targets.py
cristinasewell: Ok thanks

pcdsdevices/targets.py
```diff
@@ -260,3 +264,210 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+        if not snake_like:
+            flat_xx = list(chain.from_iterable(xx))
+            flat_yy = list(chain.from_iterable(yy))
+            return flat_xx, flat_yy
+
+        xx = self.snake_grid_list(xx)
+        yy = self.snake_grid_list(yy)
+        return xx, yy
+
+    def map_skewed(self, height=2500, width=2500, snake_like=True,
```
cristinasewell, line 384: This is definitely not how I should handle the affine and projective transformation, but I just have this here for now for testing ... not even sure if it will help at all or if it is correct
This is just a little skew adjustment, that's all.... 
From a grid like this:
skewed_grid

skewed_plan

cristinasewell:
Testing in `xxpopr` (pasting here for future reference):
- This is due to change of course and it is not completed yet, but I'm adding here the info to know what I have so far
- Also, I am putting everything in this class for now, but some calculations will probably not belong here, will move them somewhere else later, for now it is easier to test this way.

0. open the camera from a terminal
```
camViewer -c 18
```
1. create the xy stage object from an xpp session
```
from pcdsdevices.targets import XYGridStage as xy 
path = '/cds/home/opr/xppopr/experiments/xppx40318/samples.yml' 
stage = xy(name='name', x_motor='XPP:USR:PRT:MMN:06', y_motor='XPP:USR:PRT:MMN:07', m_points=3, n_points=10, path=path) 
```
2. set presets here for a grid of 3 by 10 (3 in x direction and 10 in y direction), this will ask you to tweak the motors using the arrow keyboard keys. 
- `Up&Down` keys for y and `Lef&Right` for x. To change the scale `Shift+Up&Down` key.
- Press `q` when satisfied with the position
- preset order: 00, M0, MN, 0N (top_left, top_right, bottom_right, bottom_left) corners
```
stage.set_presets()  
```
-  you can do a `stage.get_presets()` to see the saved presets
```
In [43]: stage.get_presets()
Out[43]: ((-11.461, 34.555), (-10.949, 34.542), (-10.905, 36.774), (-11.418, 36.799))
```

3. map points `def map_points(snake_like=True, show_grid=False)`
```
xx, yy = stage.map_points(show_grid=True) 
```
4.  do a simple list scan (points are saved snake_like pattern)
```
RE(bp.list_scan([], x.epix_x, xx, x.epix_y, yy)) 
```

**Additional Stuff**
* To see the current m and n points:
```
In [40]: stage.m_n_points
Out[40]: (3, 10)
```
* To change the m and n point
```
stage.m_n_points = 3, 10
```
* To see the current coefficients if any: (you can also change them, but i might need to take the setter out from this one because it is based on calculations..)
```
In [41]: stage.coefficients
Out[41]:
[0.7653300975628099,
 0.02719240783434529,
 -2.075445824255439,
 0.2900208033688614,
 0.8302966107817156,
 4.50353181205719,
 0.009031594049928593,
 -0.0009276606223413975]
```
* To save the current mapped grid: (if the same name is provided for a sample that is already saved, it should override it)
```
stage.save_grid('sample3by10')
```
* To see the current mapped samples:
```
In [44]: stage.mapped_samples()
Out[44]: ['sample3by10']
```
* To get the info for a specific sample saved:
```
In [45]: stage.get_sample('sample3by10') - this will return a dictionary so you should be able to get specific data out of it
Out[45]:
{'time_created': '2021-01-07 12:25:25.744436',
 'top_left': [-11.461, 34.555],
 'top_right': [-10.949, 34.542],
 'bottom_right': [-10.905, 36.774],
 'bottom_left': [-11.418, 36.799],
 'M': 3,
 'N': 10,
 'coefficients': [0.7653300975628099,
  0.02719240783434529,
  -2.075445824255439,
  0.2900208033688614,
  0.8302966107817156,
  4.50353181205719,
  0.009031594049928593,
  -0.0009276606223413975]}
```

cristinasewell:
I need to find a nice and user friendly way of being able to ask for any  m and n points based on the data saved in the `.yml` file for that specific grid, and get the actual physical x y location on the grid - as I am not saving the mapped grid points anywhere - only the metadata needed to map those points.
```
{'time_created': '2021-01-07 12:25:25.744436',
 'top_left': [-11.461, 34.555],
 'top_right': [-10.949, 34.542],
 'bottom_right': [-10.905, 36.774],
 'bottom_left': [-11.418, 36.799],
 'M': 3,
 'N': 10,
 'coefficients': [0.7653300975628099,
  0.02719240783434529,
  -2.075445824255439,
  0.2900208033688614,
  0.8302966107817156,
  4.50353181205719,
  0.009031594049928593,
  -0.0009276606223413975]}
```
The user can ask for only one point, or for multiple of them.... or in a certain pattern (like snake_like pattern for example)... for now i just have the option to get all the xx and yy mapped points.
This should not be too complicated but there are multiple ways we can go about

cristinasewell:
Updated instructions on how to test here:
[Testing the fixed target scan in xppopr.1.13.21.pdf](https://github.com/pcdshub/pcdsdevices/files/5810912/Testing.the.fixed.target.scan.in.xppopr.1.13.21.pdf)

pcdsdevices/targets.py
```diff
@@ -260,3 +267,786 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+
+        Returns
+        -------
+        coord : tuple
+            Four coordinate positions.
+            (top_left, top_right, bottom_right, bottom_left)
+        """
+        try:
+            # corner (0, 0)
+            top_left = (round(self.x.presets.positions.x_top_left.pos, 5),
```
cristinasewell, line 370: I probably should not round this guys here... but I am not getting exactly the same points back.

```
In [28]: stage.set_presets()

Setting coordinates for (0, 0) top left corner:

: 0.0000, : 0.0000, scale: 0.1

Setting coordinates for (0, M) top right corner:

: 0.0000, : 2.0000, scale: 0.1

Setting coordinates for (N, M) bottom right corner:

: 2.0000, : 2.0000, scale: 0.1

Setting coordinates for (N, 0) bottom left corner:

: 2.0000, : 1.0000, scale: 0.1
```

```
In [29]: stage.get_presets()
Out[29]:
((0, 0),
 (0, 2.0000000000000004),
 (2.0000000000000004, 2.0000000000000004),
 (2.0000000000000004, 0.9999999999999997))
```

klauer: Rounding for display purposes is just fine (encouraged, even), but rounding prior to doing calculations should be avoided.

cristinasewell: ok thanks, it'll take those off then...

pcdsdevices/targets.py
```diff
@@ -260,3 +267,795 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+        Parameters
+        ----------
+        path : string, optional
+            Path of .yml file with samples.
+        """
+        path = path or self._path
+        with open(path, 'w') as sample_file:
+            yaml.safe_dump(None, sample_file)
+
+    def mesh_interpolation(self, top_left, top_right, bottom_right,
```
cristinasewell, line 739: I have a few functions here that probably can be moved to the `pcdscalc` - 
- `mesh_interpolation`
- `get_unit_meshgrid`
- `convert_to_physical`
- maybe others....

They are however pretty specific to this `XYGridStage` object, should I still move them?

pcdsdevices/targets.py
```diff
@@ -125,9 +132,9 @@ class XYTargetGrid():
                           x_spacing=1.0, y_spacing=1.0, x_comp=0.05,
                           y_comp=0.01)
     """
-    def __init__(self, x=None, y=None, x_init=None, x_spacing=None,
-                 x_comp=0.0, y_init=None, y_spacing=None, y_comp=0.0,
-                 name=None):
+
+    def __init__(self, name, x=None, y=None, x_init=None, x_spacing=None,
+                 x_comp=0.0, y_init=None, y_spacing=None, y_comp=0.0):
```
ZLLentz, line 137: Just being a little cautious- any worries about changing the signature here?
Previously, `name` was an optional kwarg
Now, `name` is a required arg (positional  or keyword okay)
Usually in ophyd, `name` is a required keyword-only kwarg (which is neither the previous nor the updated situation here)

ZLLentz: @slactjohnson This `XYTargetGrid` is being used in `mec`, correct? Should double-check that a signature update isn't immediately breaking

pcdsdevices/targets.py
```diff
@@ -260,3 +267,795 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+    """
+    Class that helps support multiple samples on a mount for an XY Grid setup.
+
+    We could have multiple samples mounted in a setup. This class helps
+    figuring out the samples xy coordinates for each target on the grid,
+    maps points accordingly, and saves those records in a file.
+
+    Parameters
+    ----------
+    name
```
ZLLentz, line 395:
```suggestion
    name : str
```

pcdsdevices/targets.py
ZLLentz: I think it's correct to keep things that are very specific to this class here- after all, if it's only this class that wants to use them, then we have to track down a central library if they need an update.

When we find ourselves wanting to re-use the code for something else, that's when it's time to send it to `pcdscalc`.

pcdsdevices/targets.py
```diff
@@ -260,3 +267,795 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+        Use the Down arrow to move y motor down.
+        Use the Up arrow to move y motor up.
+        Use Shift & Up arrow to scale*2.
+        Use Shift & Down arrow to scale/2.
+        Press q to quit.
+        """
+        tweak_base(self.x, self.y)
+
+    # TODO: maybe add an option here for setting parameters directly
+    def set_presets(self):
```
ZLLentz, line 286: These presets are only used in the subclass, I'm not sure if these methods belong in the parent

pcdsdevices/targets.py
```diff
@@ -260,3 +267,795 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+                            self.y.presets.positions.y_bottom_right.pos)
+            # corner (N, 0)
+            bottom_left = (self.x.presets.positions.x_bottom_left.pos,
+                           self.y.presets.positions.y_bottom_left.pos)
+            return top_left, top_right, bottom_right, bottom_left
+        except Exception:
+            logger.warning('Could not get presets, try to set_presets.')
+
+
+class XYGridStage(XYTargetGrid):
```
ZLLentz, line 385: Which features of `XYTargetGrid` are you using in `XYGridStage`? To me it looks like you're defining new `move_to` functions, for example, and not using the parent's resources. The grid is computed differently too, the parent using this comp/skew parameter, for example, that is being ignored here.

Am I missing a reason that this needs to be a subclass?

pcdsdevices/targets.py
```diff
@@ -260,3 +267,795 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+            The sample name that we want the grid for. To see current
+            available samples call `mapped_grids`
+        path : str, optional
+            Path to the `.yml` file. Defaults to the path defined when
+            creating this object.
+
+        Returns
+        -------
+        data : dictionary
+            Dictionary of all the information for a saved sample.
```
ZLLentz, line 600: This can also return `None` if there are errors. You should do one of:
- Document that it can return `None`
- Make it raise errors
- Make it always return a `dict`

pcdsdevices/targets.py
```diff
@@ -260,3 +267,795 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+                m_points = sample['M']
+                n_points = sample['N']
+            except Exception as ex:
+                logger.error('Something went wrong when getting the '
+                             'information for sample %s. %s', sample_name, ex)
+                return
+        else:
+            logger.error('This sample probably does not exist. Please call'
+                         ' mapped_samples() to see which ones are available.')
+            return
```
ZLLentz, line 667: Another one where the ambiguous return type should be documented or made non-ambiguous

ZLLentz:
I thought everything else looked good

pcdsdevices/targets.py
cristinasewell: I think I was initially using the `x_spacing` and `y_spacing` - and it made more sense to inherit from this class, but as of now I am not using those, I am only using the `StageStack` portion of it. So I actually don't need to inherit from this class. I think I should move the twek method with the set and get presets to my object instead.

The only thing that I am using now is the `StageStack` from this class, which allows the x_motor and y-motor to either be created from a PV or a motor object, I can inherit from `StageStack` only if I need to keep this functionality

cristinasewell: So you are correct, I am not really using the parent's resources - thank you for pointing this out

ZLLentz: Okay, that makes sense- stuff evolves over time as we code/test!

pcdsdevices/targets.py
cristinasewell: I went back to the original version in 393ee51 - I am still using `StackStage` - just so we can give the option to the user to insert either a motor object or a PV, but I am wondering if I should not use `StackStage` at all?

pcdsdevices/targets.py
cristinasewell: thanks - fixed this in 393ee51

pcdsdevices/targets.py
cristinasewell: I placed them in the `XYGridTarget` object in 393ee51

pcdsdevices/targets.py
cristinasewell: Took the inheritance out of this object (393ee51)

pcdsdevices/targets.py
cristinasewell: Ok thanks, I think I covered most of (hopefully all but will double check) the ambiguous returns in 393ee51

pcdsdevices/targets.py
cristinasewell: thanks - fixed this in 393ee51

pcdsdevices/targets.py
slactjohnson: @ZLLentz I believe we have `XYTargetGrid` instantiated for MEC, but I don't know how much they use it (even though they asked for it). Based on the comments below, it looks like the change in signature may have been avoided anyway?

ZLLentz: Yes, it turns out we didn't need to do this, which is great. Changing API can be a hassle when it needs to be done.

pcdsdevices/targets.py
cristinasewell: Ok that sounds good. If I go about calculating each point  during a scan - then some of these functions might be useful in the `nabs` as well - I am currently trying this in my scan function in `nabs`

klauer (commented):
I think this is OK overall, if everyone is happy with the functionality and how it works.

pcdsdevices/targets.py
```diff
@@ -260,3 +267,802 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+        xx, yy : tuple
+            Tuple of two lists with all mapped points for x and y positions in
+            the grid.
+        """
+        top_left = top_left or self.get_presets()[0]
+        top_right = top_right or self.get_presets()[1]
+        bottom_right = bottom_right or self.get_presets()[2]
+        bottom_left = bottom_left or self.get_presets()[3]
+
+        if all([top_left, top_right, bottom_right, bottom_left]) is None:
```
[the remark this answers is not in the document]
cristinasewell: thanks for pointing this out - changed this in 270f55b

pcdsdevices/targets.py
```diff
@@ -260,3 +267,802 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+                              "bottom_right": bottom_right,
+                              "bottom_left": bottom_left,
+                              "M": m_points,  # number of rows
+                              "N": n_points,  # number of columns
+                              "coefficients": coefficients}}
+        try:
+            temp_data = json.loads(json.dumps(data[sample_name]))
+            jsonschema.validate(temp_data, self.sample_schema)
+        except jsonschema.exceptions.ValidationError as err:
+            logger.warning('Invalid input: %s', err)
```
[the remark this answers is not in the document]
cristinasewell: re-raising the error now in 270f55b - thanks

pcdsdevices/targets.py
```diff
@@ -260,3 +267,802 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+        bottom_right = bottom_right or self.get_presets()[2]
+        bottom_left = bottom_left or self.get_presets()[3]
+
+        if all([top_left, top_right, bottom_right, bottom_left]) is None:
+            raise ValueError('Could not get presets, make sure you set presets'
+                             ' first using the `set_presets` method.')
+        rows = m_rows or self.m_n_points[0]
+        columns = n_columns or self.m_n_points[1]
+
+        a_coeffs, b_coeffs = self.mesh_interpolation(top_left, top_right,
```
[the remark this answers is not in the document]
cristinasewell: I wrote a few methods here with the intention that I'll be moving them to a different package (like in `pcdscalc`) - it was easier to make them part of the class so I can test them... but it makes sense to move them outside the class now as independent functions if we decided to keep them here.

I will try to pay attention to what you pointed out in the feature 😉, thanks -  but it seems like this whole object does not truly belong in `pcdsdevices` - nor somewhere else

I pulled out of this class a few general functions like you suggested (in 270f55b )- I hope that's sensible enough - but let me know if that's not what you meant.

pcdsdevices/targets.py
```diff
@@ -260,3 +267,802 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+        data = {sample_name: {"time_created": now,
+                              "top_left": top_left,
+                              "top_right": top_right,
+                              "bottom_right": bottom_right,
+                              "bottom_left": bottom_left,
+                              "M": m_points,  # number of rows
+                              "N": n_points,  # number of columns
+                              "coefficients": coefficients}}
+        try:
+            temp_data = json.loads(json.dumps(data[sample_name]))
```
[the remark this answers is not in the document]
cristinasewell: Yes the initial `dump/load` is unnecessary - thanks - I removed it in 270f55b

This class does not need to do JSON schema validation - it was useful for testing and making sure I write things correctly to the file, but I can take this out of here if it is too much - please let me know what you think.
The move functions were requested by the scientists

pcdsdevices/targets.py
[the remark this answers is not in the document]
ZLLentz: I think this object definitely belongs in pcdsdevices. It represents real but generic physical hardware and provides device-specific functionality for using that hardware in Python.

klauer:
Thanks, @cristinasewell. This is looking better! 

And as usual, I hope you're taking my comments as they're intended, as constructive criticism. I stumbled around alone for ages trying to figure the same things out that I'm suggesting to you - and although I know not everyone's the same - I wish someone had been there to help critique/fix my coding style long ago.

cristinasewell:
> Thanks, @cristinasewell. This is looking better!
> 
> And as usual, I hope you're taking my comments as they're intended, as constructive criticism. I stumbled around alone for ages trying to figure the same things out that I'm suggesting to you - and although I know not everyone's the same - I wish someone had been there to help critique/fix my coding style long ago.

Yes - that's what I do - that's why I requested you to review my code - I know it takes effort and time to review somebody else's code so I appreciate it very much. I find that constructive criticism is one of the best ways to learn ;) - It will always be welcomed and appreciated 😉

cristinasewell:
One thing that concerns me now is that the sample file can get pretty large since I am also saving the actual target points - because of time constrains I think this will be fine for now, but in the future I should probably add a folder with sample files, where every sample is saved in different files - instead of multiple samples in one file?

ZLLentz:
Yes, we may need to return to this later and make a change here. Perhaps we can just cull the accumulated target points? There may be some way to know that they are no longer needed. But I think it will be okay for now.

ZLLentz:
The only thing that bugs me about this PR is the class name because it is extremely similar to the existing to the point of being a potential source of confusion:
Existing class: `XYTargetGrid`
New class defined here: `XYGridStage`

I have not brought this up yet because I can't figure out any better name to use. So I will resign myself to let it go.

cristinasewell:
> Yes, we may need to return to this later and make a change here. Perhaps we can just cull the accumulated target points? There may be some way to know that they are no longer needed. But I think it will be okay for now.

Yeah, I am sure there are better ways to do this, but something simple would probably be to save them as tuples with a bit indicator, something like this: xx = [(1, 0), (2, 0), (3, 0), (4, 1), (5, 1)] - to indicate that 1's have been shot and 0's are available (.. or True and False instead of 1 and 0.. ) and in the scan besides/(instead of) saving the last index shot, we can set the bits for every target position.

cristinasewell:
> The only thing that bugs me about this PR is the class name because it is extremely similar to the existing to the point of being a potential source of confusion:
> Existing class: `XYTargetGrid`
> New class defined here: `XYGridStage`
> 
> I have not brought this up yet because I can't figure out any better name to use. So I will resign myself to let it go.

Same here, I can't think of a better name, but open for changing it to something better for sure

klauer:
I think the name is OK and can't think of anything better. Should we move forward with this?

cristinasewell:
I am doing some final testing right now - there is one more thing I just realized I need to add and should be good to go after that.
I need to allow for re-calibration of points without losing the `last_index_shot` value in the sample files.

cristinasewell:
One last question for now - should I have this object inherit from ophyd `Device`? - or is this fine for now the way is?

ZLLentz:
Making this object inherit from `Device` would require some refactoring. The advantage would be that it would behave more similarly to all the other devices. The disadvantage would be that you can no longer pass arbitrary stages in, it'd have to be prefixes only with classes determined ahead of time.

klauer:
If this goes the ophyd Device route, I'd picture it to be something like a `PVPositioner`, where you're required to make a subclass with a certain set of components (`x` and `y` here).

On one hand, I'm not sure this really needs it. But then again a device-looking object that doesn't behave as a device may confuse users in the end. In the interest of moving forward, I'd think using it as-is should be fine for now.

cristinasewell:
@klauer, @ZLLentz , @ZryletTC 
Here are some things I would like you to take a look at, they are mostly the changes I made in this commit (e381d22):

I am now saving the statuses of those targets along with their points. The tricky part here was to be able to save one sample file, and if we need to override it (save it again under the same name) we would save new information about everything except the statuses (we need to keep the same statuses as before)- and I'd like you to give me some feedback about how I did that, and if you could spot potential edge cases here that I have not considered.

pcdsdevices/targets.py
```diff
@@ -260,3 +268,907 @@ def move(self, nxspaces, nyspaces, wait=False):
...
+                x_points.append(i)
+                y_points.append(j)
+        if snake_like:
+            x_points = snake_grid_list(
+                np.array(x_points).reshape(rows, columns))
+        self.positions_x = x_points
+        self.positions_y = y_points
+        return x_points, y_points
+
+    def compute_mapped_point(self, sample_name, m_row, n_column,
```
cristinasewell, line 907: Just realized that this function also needs to know if it was saved in snake-like pattern .... maybe it was not a good idea to save the points in a snake-like pattern.
😭 

I should probably add one more little field in the file to indicate if snake-like

cristinasewell: I think this is not True actually.... i will come back to this.

ZLLentz:
I actually can't follow this at all- I've tried a few times throughout the day. I'm not sure what you're asking or if there are any issues in the code.

ZLLentz:
Okay, after our talk today I think we should merge both PRs if you're satisfied with today's testing. It is totally good to make more changes later and PR/merge them separately.

cristinasewell:
Sounds good @ZLLentz  - we will be testing this on `xxp` today, sometimes after 5 - I guess we can wait until tomorrow?

ZryletTC (approved):
I can't say I've read through it all yet but we've seen it work and I'm in favor of merging now and making further edits in a new PR.

cristinasewell:
> I can't say I've read through it all yet but we've seen it work and I'm in favor of merging now and making further edits in a new PR.

I do have a little function I want to add that will reset the statuses for the targets in a sample- the scientists want a 'rehearsal' type scanning where the targets don't get marked as `shot`, this is not the same but in the time frame this is something that will get the job done (or help with it) - I will go ahead and push that function in - I should not need to push other changes in after that.

ZLLentz:
Ok I'm going to merge everything here and in https://github.com/pcdshub/nabs/pull/31 since data was taken!

We can revise in follow-ups. Good work on this one!

PR #740, opened 2021-01-15
TST: try to get status printouts from all fake devices

klauer:
## Description
Attempt to instantiate all fake devices and print status information.
Non-instantiable devices are skipped (about 13 classes).

## Motivation and Context
* Status information has close to 0 test coverage
* Generically performing these tests may catch problems in the future

## How Has This Been Tested?
This is testing things.

In this PR text.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -311,21 +311,24 @@ class AttBase(FltMvInterface, PVPositioner):
 
         states = '\n'.join(render_ascii_att(blade_states))
 
-        energy = get_status_value(status_info, 'energy', 'value') * 1e3
-        energy_3rd = get_status_value(status_info, 'energy_3rd', 'value')
-        trans = get_status_value(status_info, 'position')
+        energy = get_status_float(
+            status_info, 'energy', 'value', scale=1e3, precision=3)
+        energy_3rd = get_status_float(
+            status_info, 'energy_3rd', 'value', scale=1e3, precision=3)
+        trans = get_status_float(
+            status_info, 'position', precision=4, format='E')
 
         if energy_3rd != 'N/A':
-            energy_3rd = energy_3rd * 1e3
-            return f"""\
-{states}
-Transmission for 1st harmonic (E={energy:.3} keV): {trans:.4E}
-Transmission for 3rd harmonic (E={energy_3rd:.3} keV): {trans:.4E}
-"""
+            status_3rd = (
+                f'Transmission for 3rd harmonic (E={energy_3rd} keV): {trans}'
+            )
         else:
-            return f"""\
+            status_3rd = ''
+
+        return f"""\
 {states}
-Transmission for 1st harmonic (E={energy:.3} keV): {trans:.4E}
+Transmission for 1st harmonic (E={energy} keV): {trans}
+{status_3rd}
 """
 
 
@@ -1054,8 +1057,9 @@ class AT2L0(FltMvInterface, PVPositionerPC, LightpathInOutMixin):
         self.calculator.run_calculation.put(1, wait=True)
         return super()._setup_move(position)
 
-    def __init__(self, *args, limits=None, **kwargs):
-        UCpt.collect_prefixes(self, dict(calculator_prefix='AT2L0:CALC'))
+    def __init__(self, *args, limits=None, calculator_prefix='AT2L0:CALC',
+                 **kwargs):
+        UCpt.collect_prefixes(self, dict(calculator_prefix=calculator_prefix))
         limits = limits or (0.0, 1.0)
         super().__init__(*args, limits=limits, **kwargs)
 
--- a/pcdsdevices/lens.py
+++ b/pcdsdevices/lens.py
@@ -1,10 +1,11 @@
 """
 Module for Beryllium Lens positioners.
 """
+import logging
 import time
 from collections import defaultdict
 from datetime import date
-import logging
+
 import numpy as np
 from ophyd.device import Component as Cpt
 from ophyd.device import FormattedComponent as FCpt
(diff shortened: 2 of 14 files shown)


ZLLentz:
I'm a big fan of generic library-wide tests for behemoths like this library that are essentially just many similar classes

klauer:
This one will be draft until I get all the tests passing with appropriate per-class kwargs. Would like to lower the number of skips, if possible, too.

klauer:
OK, now that the attenuator one is in, this should be good to go. 🤞 

_All_ `BaseInterface`-mixed in classes now instantiate with the best-effort code (13 classes are intentionally skipped as they are base classes). There are no status print failures locally, at least.

pcdsdevices/attenuator.py
```diff
@@ -311,21 +311,24 @@ def format_status_info(self, status_info):
...
 
-        energy = get_status_value(status_info, 'energy', 'value') * 1e3
-        energy_3rd = get_status_value(status_info, 'energy_3rd', 'value')
-        trans = get_status_value(status_info, 'position')
+        energy = get_status_float(
+            status_info, 'energy', 'value', scale=1e3, precision=3)
+        energy_3rd = get_status_float(
+            status_info, 'energy_3rd', 'value', scale=1e3, precision=3)
+        trans = get_status_float(
+            status_info, 'position', precision=4, format='E')
```
ZLLentz, line 319: Random thought, once we propagate variety information more thoroughly through the library then these sorts of calls can just check the display information from the variety metadata

pcdsdevices/lens.py
```diff
@@ -520,14 +521,21 @@ class LensStack(LensStackBase):
 
     """
 
-    def __init__(self, *args, path, **kwargs):
+    def __init__(self, x_prefix, y_prefix, z_prefix, z_offset, z_dir, E,
```
ZLLentz, line 524: I realize that making these more explicit in many places allows for instrospectability, but I think we intentionally left this out in many cases to leverage UCpt to avoid writing all of this out. Do we need to always write out the full extended prefix set from now on in all classes to let the test suite do its thing?

tests/test_interface.py
```diff
@@ -233,6 +234,46 @@ def test_tab_completion(cls):
...
+    # Otherwise, include any Devices that inherit from BaseInterface.
+    return BaseInterface in cls.mro()
+
+
+@pytest.mark.parametrize(
+    'cls',
+    [pytest.param(cls, id=f'{cls.__module__}.{cls.__name__}')
+     for cls in conftest.find_all_device_classes()
+     if _should_check_status_prints(cls)
+     ]
```
ZLLentz, line 269: Can we sort this list (and the one for tab completion) so that the test readout is easier to parse? It looks like this now:
```
...
tests/test_interface.py::test_smoke_status_prints[pcdsdevices.sample_delivery.M3BasePLCDevice] PASSED [ 52%]
tests/test_interface.py::test_smoke_status_prints[pcdsdevices.sample_delivery.PCM] PASSED [ 53%]
tests/test_interface.py::test_smoke_status_prints[pcdsdevices.spectrometer.VonHamosCrystal] PASSED [ 53%]
tests/test_interface.py::test_smoke_status_prints[pcdsdevices.gon.Kappa] PASSED [ 53%]
tests/test_interface.py::test_smoke_status_prints[pcdsdevices.state.StateRecordPositionerBase] PASSED [ 53%]
tests/test_interface.py::test_smoke_status_prints[pcdsdevices.lens.LensStackBase] PASSED [ 53%]
tests/test_interface.py::test_smoke_status_prints[pcdsdevices.lodcm.H1N] PASSED [ 53%]
tests/test_interface.py::test_smoke_status_prints[pcdsdevices.timetool.TimetoolWithNav] PASSED [ 53%]
tests/test_interface.py::test_smoke_status_prints[pcdsdevices.lens.LensStack] PASSED [ 53%]
tests/test_interface.py::test_smoke_status_prints[pcdsdevices.pim.PPM] PASSED [ 54%]
tests/test_interface.py::test_smoke_status_prints[pcdsdevices.state.CombinedStateRecordPositioner] PASSED [ 54%]
tests/test_interface.py::test_smoke_status_prints[pcdsdevices.lxe._ReversedTimeToolDelay] PASSED [ 54%]
tests/test_interface.py::test_smoke_status_prints[pcdsdevices.attenuator.AttenuatorCalculator_AT2L0] PASSED [ 54%]
...
```

tests/conftest.py
```diff
@@ -99,14 +105,83 @@ def find_all_device_classes() -> list:
...
+    pcdsdevices.lens.LensStack: dict(
+        path=str(MODULE_PATH / 'test_lens_sets' / 'test'),
+    ),
+    pcdsdevices.lens.SimLensStack: dict(
+        path=str(MODULE_PATH / 'test_lens_sets' / 'test'),
+    ),
+    pcdsdevices.lxe.LaserEnergyPositioner: dict(
+        calibration_file=MODULE_PATH / 'xcslt8717_wpcalib_opa',
+    ),
+}
```
ZLLentz, line 135: I'm surprised there aren't more of these!

ZLLentz (approved):

pcdsdevices/lens.py
klauer: `UnrelatedComponent` is supported directly in the best-effort instantiation code. It _does_ assume you don't do something funny in `__init__` with the prefix (see the diff on how I changed AT2L0 - I was previously setting `calc_prefix` explicitly and not allowing the kwarg, which meant that `UCpt` wouldn't remove the argument from the kwargs dict).

See `pcdsdevices.lxe` or `pcdsdevices.jet` for passing `UCpt` examples.

The explicitness here is required for `FormattedComponent`, primarily. I personally think that requiring those in `__init__` kwarg lists explicitly is a reasonable thing.

pcdsdevices/attenuator.py
klauer: Very good point!

pcdsdevices/lens.py
ZLLentz: Ahh! Okay that makes a lot of sense. Seems :+1:

klauer: That said, hopefully I didn't screw up any of the `__init__`s here inadvertently ;)

tests/test_interface.py
klauer: This annoyed me and I (apparently) made an attempt to fix it, but didn't realize it was doing the equivalent of `return list(set(sorted(modules)))` before. Fixed for good in ce3512d .

ZLLentz: Looks :+1: :+1: to me!

klauer:
_🚲ing travis_

klauer:
Tests passing; moving forward based on previous approval. Will have a follow-up PR that better accounts for devices in submodules - which will be even more important once the library is reorganized.

PR #752, opened 2021-02-09
MNT: edge case for timeouts on lxt moves to the same location.

ZLLentz:
## Description

Skip waiting for the `done` signal in cases where a move of `LaserTiming` is less than 20fs. Instead, simply sleep for 0.01s and move on.

## Motivation and Context

If you move the `LaserTiming` (`lxt`) device in XPP to the same location as the previous move, the `done` signal does not transition at all, it stays at `1`. Since there is never a transition from `0` to `1`, there is never any real `done` status. This is a workaround to allow scans to proceed if they are triggering tiny moves.

## How Has This Been Tested?

Live on the real deal in xpp3

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/lxt-wait-752.rst
+++ b/docs/source/upcoming_release_notes/lxt-wait-752.rst
@@ -1,0 +1,31 @@
+lxt-wait 752
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- `LaserTiming` will now report done after 0.01s for infintesimal moves,
+  rather than the previous freeze-and-never-report-done.
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- ZLLentz
--- a/docs/source/upcoming_release_notes/lxt-wait-752.rst
+++ b/docs/source/upcoming_release_notes/lxt-wait-752.rst
@@ -19,7 +19,7 @@ New Devices
 
 Bugfixes
 --------
-- `LaserTiming` will now report done after 0.01s for infintesimal moves,
+- `LaserTiming` will now report done after 0.01s for infinitesimal moves,
   rather than the previous freeze-and-never-report-done.
 
 Maintenance
--- a/pcdsdevices/lxe.py
+++ b/pcdsdevices/lxe.py
@@ -29,6 +29,7 @@ LXT::
 """
 
 import pathlib
+import time
 import types
 import typing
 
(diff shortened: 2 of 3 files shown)


ZLLentz:
Actually I screwed this up a bit, whoops

docs/source/upcoming_release_notes/lxt-wait-752.rst
```diff
@@ -0,0 +1,31 @@
...
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- `LaserTiming` will now report done after 0.01s for infintesimal moves,
```
klauer, line 22:
```suggestion
- `LaserTiming` will now report done after 0.01s for infinitesimal moves,
```

pcdsdevices/lxe.py
```diff
@@ -315,8 +316,20 @@ def _setup_move(self, position):
         except Exception as ex:
             self.log.debug('Failed to update notepad setpoint to position %s',
                            position, exc_info=ex)
+
+        # Trigger dmov for tiny moves since done PV doesn't work here
+        if np.isclose(position, self.setpoint.get(), atol=20e-15, rtol=0):
```
klauer, line 321:
```suggestion
        short_move = np.isclose(position, self.setpoint.get(), atol=20e-15, rtol=0)
```

(Should be a multi-line suggestion; don't accept this one directly - even if you like it)

klauer (approved):
Minor, optional suggestions - and of course, thanks for fixing up my mess!

docs/source/upcoming_release_notes/lxt-wait-752.rst
ZLLentz: 🥔

pcdsdevices/lxe.py
ZLLentz: ahh yes ok will do

ZLLentz:
There might be more to this. I got a call from Silke that certain moves were stalling the scan- same deal as last night where moves above 1-e6 in distance were causing freezes, even though this did not happen during the day! Something is wrong with the motor maybe?

I'll drop this PR into draft purgatory for now.

ZLLentz:
11th hour edit to the live directory with:
```
+        # Something is wrong with done signal, just sleep and pretend
+        time.sleep(1)
+        self._move_changed(value=1-self.done_value)
+        self._move_changed(value=self.done_value)
```

ZLLentz:
Hypothesis: the underlying setpoint PV can do some number of things that don't necessarily involve the signal we've picked for done! And this behavior can vary between shift to shift depending on what we're doing with the laser!

klauer:
That's not a good sign... It does highlight how difficult it is to make a proper device here without having a full understanding of _what lies beneath_ on the EPICS level.

ZLLentz:
I absolutely agree. We either need to:
- learn what's inside the black box and define `done` more appropriately
- or set the `done` to automatically finish after a set period of time given no `done` signal (eww gross my error handling)

ZLLentz:
Maybe the prudent solution is to instead:
- Remove the `done` signal altogether (it doesn't really work)
- Configure a settling time (the correct way to add a "wait" at the end of a move)

klauer:
Is the settle time reasonably repeatable that with an added buffer you could be certain about it status? Or rather, with no reliable done status, how can we... be sure?

ZLLentz:
I have no idea Ken! We're in black box land! Maybe it's time to do some research on how the IOC works.

ZLLentz:
Closing this PR in favor of either:
- Learning what's in the black box
- Making the janky patch used in XPP/XCS an official PR until we have time to investigate the black box

PR #761, opened 2021-02-18
Update gon.py

dlzhu:
check_motor_step currently takes 'real motor positions', rather than the Euler angles. This is one of the issues that led to the print out giving wrong target position values even though the motion largely did the right thing. Also, phi motor was moving right amount, but status of kappa.e_phi shows the wrong value. Thus when moving, e.g. kappa.e_chi(5), from (0,0,0), at the end of motion, it looks like the motors moved the right mount/direction, but kappa.e_phi show 2X value as kappa.phi, rather than 0.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/gon.py
+++ b/pcdsdevices/gon.py
@@ -521,8 +521,9 @@ class Kappa(BaseInterface, PseudoPositioner, Device):
         Checks for the motor step, and ask the user for confirmation if
         movement step is greater than default one.
         """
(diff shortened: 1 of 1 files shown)


ZLLentz:
Good catch, though the proposed diff here raises an AttributeError in the automatic test since position_k is just a normal tuple

ZLLentz:
I'll be working on the related items in #762

Issue #763, opened 2021-02-18
LODCM - additional functions


cristinasewell:
## Expected Behavior
The following functions might be something scientist might want for the `LODCM`:

`tweakX`:
https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xpp/blinst/lom.py#L1133

`tweakParallel`:
https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xpp/blinst/lom.py#L1165

`setE`:
https://github.com/pcdshub/old-hutch-configs/blob/18c7769d66a647ab558ad7ebc34d019fc98fbf48/xpp/blinst/lom.py#L1197

`others...`

We might need some more feedback from scientists about other things we should include in the `LODCM` object after #715 is merged in.

## Current Behavior

We don't have those functions.

## Possible Solution

I have started looking into these so I'll paste some code here of how approximately they should look (needs cleanup)

```
    def set_energy(self, energy, material=None, reflection=None):
        """
        Set the current positions of th1, th2, z1 and z2 offset motors.

        Parameters
        ----------
        energy : number
            Energy in keV.
        material : str, optional
            Crystal material. E.g.: `Si`
        reflection : tuple, optional
            Crystal's reflections. E.g.: `(1,1,1)`
        """
        # try to determine the material and reflection:
        material = material or self.get_material()
        reflection = reflection or self.get_reflection(
            as_tuple=True, check=True)
        th, z = diffraction.get_lom_geometry(energy, material, reflection)
        if material == 'Si':
            self.th1Si.set_current_position(th)
            self.th2Si.set_current_position(th)
            self.z1Si.set_current_position(-z)
            self.z2Si.set_current_position(z)
        elif material == 'C':
            self.th1C.set_current_position(th)
            self.th2C.set_current_position(th)
            self.z1C.set_current_position(-z)
            self.z2C.set_current_position(z)
        else:
            raise ValueError("Invalid material: %s", material)

    def tweak_x(self, x, material=None):
        """
        Tweak some motors to center the beam on YAG2 horizontally based on
        the calibrated s4 position.

        Parameters
        ----------
        x : number
            The value that we want to move `x2` by from current position.
        material : str, optional
            Crystal material. E.g.: `Si`
        """
        # try to determine the material:
        material = material or self.get_material()
        if material == 'Si':
            th = self.th2Si.wm()
[...]

PR #766, opened 2021-02-24
Lodcm additional functions

cristinasewell:
## Description

- Added 4 new methods:
`tweak_x`, `tweak_parallel`, `set_energy`, `wait_energy`
- Overrode the `custom_status_info` for the 3 towers as well as the energy classes
- Implemented some suggestions from here:  https://github.com/pcdshub/pcdsdevices/pull/715#pullrequestreview-594626588 

## Motivation and Context

Trying to tackle some things from #763

## How Has This Been Tested?

Has only been tested locally
The methods that require movement have not been tested with the real motors.

Docstring

## Pre-merge checklist


cristinasewell:
@ZryletTC  I think I am going to keep tagging you in this PR to get some help. Also I don't think I am going to touch the OffsetIMS in this PR, I think it is better to open another PR for that.

For `get_reflection` - do you think we could safely remove both the `check` and the `as_tuple`?
I think we don't really use the string version of reflection except for the status display - but maybe the scientists use it for something else that I am not aware of?

https://github.com/pcdshub/pcdsdevices/blob/5b59352d9243fd1c258194644147bd26866dbbcb/pcdsdevices/lodcm.py#L228-L255

pcdsdevices/lodcm.py
```diff
@@ -1139,7 +1148,124 @@ def calc_energy(self, energy, material=None, reflection=None):
...
+        # try to determine the material and reflection:
+        material = material or self.get_material()
+        reflection = reflection or self.get_reflection(
+            as_tuple=True, check=True)
+        th, z = diffraction.get_lom_geometry(energy * 1e3, material,
+                                             reflection)
+        # TODO: probably not the correct approach here to use
+        # set_current_position
+        if material == 'Si':
+            self.th1Si.set_current_position(th)
```
cristinasewell, line 1175: I am not sure if I am supposed to use `set_current_position` directly here or if I should be using something else instead?

In the old code they have their own `set` method for it:

```
        def set(value):
            if use_ims_preset:
                Pv.put("%s_SET" % offset_pv, motor.wm() - value)
            Pv.put(offset_pv, motor.wm() - value)
```

And use it here:
```
        if ID == "Si":
            self.th1Si.set(th)
            self.z1Si.set(-z)
        elif ID == "C":
            self.th1C.set(th)
            self.z1C.set(-z)
```

pcdsdevices/lodcm.py
```diff
@@ -1137,9 +1131,124 @@ def calc_energy(self, energy, material=None, reflection=None):
...
+            self.z2Si.set_current_position(z)
+        elif material == 'C':
+            self.th1C.set_current_position(th)
+            self.th2C.set_current_position(th)
+            self.z1C.set_current_position(-z)
+            self.z2C.set_current_position(z)
+        else:
+            raise ValueError("Invalid material: %s", material)
+
+    def tweak_x(self, x_value, material=None, wait=False):
```
cristinasewell, line 1170: Should I be changing the naming here as well to reflect exactly what we have in the old code?
instead of `tweak_x` -> `tweakX`?

and the other ones?
`tweeak_parallel` - `tweakParallel`
`set_energy` - `setE`
.....
others?

I think it would not be too hard to get used to the new style, but it is easy for me to say, I am not the one using these....

cristinasewell:
A few more questions here that I need to address in this PR:

- [ ] https://github.com/pcdshub/pcdsdevices/pull/715#discussion_r580914627
In what sense this should be improved @ZryletTC  - is the wording confusing in the docstring?

- [ ] https://github.com/pcdshub/pcdsdevices/pull/715#discussion_r580928118
How do we want them too look like? I thought these classes `CrystalTower1`, `CrystalTower2`, `DiagnosticsTower` ,`LODCMEnergySi` and `LODCMEnergyC` - are kind of hidden from the end users, in the sense that users won't really use `lom.tower1` for example to see what's going on there, instead they only use `lom` to see the entire status printout.... 

Should I add the entire status for example:
`CrystalTower1`:
```
Crystal Tower 1
Crystal 1 state: OUT
Current Configuration: Diamond (111)
z [mm]                -400.5795 (381.8556)
x [mm]                     0.0087 (0.8663)
th [deg]                 18.4118 (-0.8408)
chi [deg]                  0.0270 (3.5588)
y [mm]                   0.0203 (-18.6346)
hn [mm]                -10.0001 (-10.5463)
hp [mm]                    0.0242 (0.2295)
diode [N/A]                      N/A (N/A)
```

`Crystal Tower2`:
```
Crystal Tower 2
Crystal 2 state: OUT
Current Configuration: Diamond (111)
z [mm]                 401.7717 (386.1496)
x [mm]                -0.1880 (1.3869)
th [deg]                 18.4168 (23.2225)
chi [deg]              0.0750 (6.1629)
y [mm]                 -0.0165 (-38.7514)
hn [mm]             -0.0003 (-18.7297)
hp [mm]              N/A (N/A)
diode [N/A]        6.0000 (-366.0000)
```

... and so on

`LODCMEnergySi`
what would we want here? just energy or more/

cristinasewell:
@ZryletTC  I think I implemented most of your suggestions from the other PR here (except the ones regarding the `OffsetMotor`), please let me know if I missed anything important.

ZLLentz:
This has been sitting for a while, I'll give it a review and try to answer the questions

pcdsdevices/lodcm.py
ZLLentz: I don't have any insight here

pcdsdevices/lodcm.py
```diff
@@ -105,109 +106,111 @@ class Y2(InOutRecordPositioner):
...
-    `chi1_state` rotates the crystal assembly.
+    The `y1_state` brings the crystal into the beam, moving vertically
+    (normal to the beam and normal to the ground).
+    The `chi1_state` rotates the crystal assembly.
 
     Parameters
     ----------
     prefix : str
         Epics base Pv prefix.
+    name : str
```
ZLLentz, line 125:
```suggestion
    name : str, keyword-only
```

pcdsdevices/lodcm.py
```diff
@@ -1137,9 +1472,124 @@ def calc_energy(self, energy, material=None, reflection=None):
...
+        """
+        # try to determine the material:
+        material = material or self.get_material()
+        if material == 'Si':
+            th = self.th2Si.wm()
+        elif material == 'C':
+            th = self.th2C.wm()
+        else:
+            raise ValueError("Invalid material: %s", material)
+        z_value = x_value / np.tan(2 * th)
```
ZLLentz, line 1533: Are there issues here with angular units? `np.tan` is in radians, I thought all the motors were in degrees.

ZLLentz:
I think the rest is :+1:

pcdsdevices/lodcm.py
cristinasewell: Oh thanks for catching this, yeah I think i need to convert that to degrees first

pcdsdevices/lodcm.py
cristinasewell: I completely overlooked this, thank you for pointing that out. I added a new `offset_set_pv` to the `OffsetMotorBase` in 4d32fc3 but I am not sure if that was the right approach - I think we want to only add it for certain motors not for all of them, even if we won't set those PVs for the other motors, and I am not entirely sure how I'd do that (i am going to give it more more though tomorrow):

https://github.com/pcdshub/pcdsdevices/blob/4a3e4f061557a5925c23496a7fdfedb8a3f54366/pcdsdevices/pseudopos.py#L1010-L1012

cristinasewell:
Locally I have a couple of recent added tests that are failing (after merging master), do I need to update some other packages?

pcdsdevices/lodcm.py
cristinasewell: Fixed this in 4d32fc3 - I had my test wrong too because I manually calculated the values of what they should be - I thought it was simple math 🙅

pcdsdevices/lodcm.py
ZLLentz: I think we need to have a subclass of the offset motor that has this additional `{self._prefix}_SET` PV and also puts to it during `set_current_position`, then carefully pick which ones to grant this subclass to... I still don't know exactly why the old classes do this but I'm scared to upset the balance of "years of legacy code"

cristinasewell: Ohh that makes more sense, I will try that instead, thank you.

ZLLentz:
Yeah, it's possible that the requirements have updated... maybe check your environment against: https://github.com/pcdshub/pcdsdevices/blob/4c757eb143a1fbc0e6e518d95ec4bbd1ff9b6586/conda-recipe/meta.yaml#L19-L33

pcdsdevices/lodcm.py
cristinasewell: I added a new subclass in f11f91e - I think this is what you meant, please let me know if I missed your suggestion.

As of now, I am not using any motors with the old `use_ims_preset=True` in any of the calculations, but the old `lodcm` object is so big - I am sure we'll have to add more stuff into it later on... and probably scientists are using these motors independently too... 

I also have no idea what is the `_SET` for, @ZryletTC  do you know?

pcdsdevices/epics_motor.py
```diff
@@ -625,6 +625,35 @@ class OffsetMotor(OffsetMotorBase):
...
 
 
+class OffsetIMSWithPreset(OffsetMotorBase):
+    """
+    Offset IMS with an additiona Offset _SET PV.
+
+    This motor puts to an additional PV (_SET) during `set_current_postion`.
+    """
+    motor = FCpt(IMS, '{self._motor_prefix}')
+    offset_set_pv = FCpt(EpicsSignal, '{self._prefix}:_SET', kind='normal')
```
ZLLentz, line 635:
```suggestion
    offset_set_pv = FCpt(EpicsSignal, '{self._prefix}_SET', kind='normal')
```

cristinasewell: ohh thank you @ZLLentz - I think my brain is just so tired of this LODCM object that does not want to function well hahah

ZLLentz: Honestly, same, I want to merge and forget

pcdsdevices/lodcm.py
ZryletTC: The `_SET` PVs are just the setpoints for certain states. In this case, I think this would be the diamond or silicon states setpoints. And so when you update the offset for a motor, the states should get changed too.
@ZLLentz does that match your understanding?

ZLLentz:
@ZryletTC @cristinasewell 
Should I merge this for today's tagging? Or leave it?

cristinasewell:
> @ZryletTC @cristinasewell
> Should I merge this for today's tagging? Or leave it?

@ZryletTC I would like you to decide on this if that's ok, as you work more closely with the scientists, what do you think? 
Do you think the scientists would like to test this before it gets merged? I have not tried to test these additional functions with real devices as they would move things.

ZryletTC:
Yes, I'd say go ahead and merge. We have another week until experiment in XPP to test with.

ZryletTC (approved):

klauer:
Just missed the tag. Oh well, I'll merge and help with making another minor one while Zach fixes gateway problems.

PR #771, opened 2021-03-01
MNT: Lock timetool compensation to static offsets

### LINKED ISSUE #770: Issues with lxt, txt, lxt_ttc in XPP and XCS
**Bug Report**

Some sequences of move, set_current_position, move results in strange and unpredictable behavior for lxt_ttc.

Example issue paraphrased from meeting notes:
```
set lxt to zero
set txt to zero
move ttc to 10ps
txt move correctly
lxt move to 2ns
```

**Expected Behavior**

lxt_ttc should always move each motor to the correct position

**Context**

Makes it impossible to use these in timing scans, despite having worked previously!

ZLLentz:
## Description

Allow subclasses of `SyncAxesBase` to define static offsets. Use this in the timetool compensation construct to keep the delay and laser motors always the same value.

## Motivation and Context

closes #770 
`SyncAxesBase` sets its offsets on the first move, which is great if you've already zeroed them, otherwise this fails. It was designed to take the arbitrary set of four motors on a large stand and keep them at the same offsets as when you first started moving.

In this case, we want very similar behavior, but we always want to move the two motors to the same position, so just hardcode the offsets in the class definition.

## How Has This Been Tested?

Interactively with test motors

WIP

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/pseudopos.py
+++ b/pcdsdevices/pseudopos.py
@@ -4,11 +4,11 @@ import typing
 import numpy as np
 import ophyd
 import ophyd.pseudopos
-from ophyd.signal import EpicsSignal
 from ophyd.device import Component as Cpt
 from ophyd.device import FormattedComponent as FCpt
 from ophyd.pseudopos import (PseudoSingle, pseudo_position_argument,
                              real_position_argument)
+from ophyd.signal import EpicsSignal
 from scipy.constants import speed_of_light
 
 from .interface import FltMvInterface
@@ -167,6 +167,7 @@ class SyncAxesBase(FltMvInterface, PseudoPositioner):
     """
 
     pseudo = Cpt(PseudoSingleInterface)
+    _offsets = None
 
     def __init__(self, *args, **kwargs):
         if self.__class__ is SyncAxesBase:
@@ -174,7 +175,6 @@ class SyncAxesBase(FltMvInterface, PseudoPositioner):
                              'the axes to synchronize included as '
                              'components'))
         super().__init__(*args, **kwargs)
-        self._offsets = None
 
     def calc_combined(self, real_position):
         """
@@ -222,7 +222,7 @@ class SyncAxesBase(FltMvInterface, PseudoPositioner):
 
     @real_position_argument
     def inverse(self, real_pos):
-        """Combined axis readback is the mean of the composite axes."""
+        """Combined axis readback is defined in calc_combined."""
         return self.PseudoPosition(pseudo=float(self.calc_combined(real_pos)))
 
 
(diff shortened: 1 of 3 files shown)


ZLLentz:
Hmm, maybe this should be a normal option to SyncAxesBase or a subclass in there that keeps everything truly synchronized with no offsets

klauer:
Seems like that kind of flexibility would be reasonable.

pcdsdevices/sim.py
```diff
@@ -41,6 +41,12 @@ def __init__(self, *args, init_pos=0, **kwargs):
             kwargs.pop(kw, None)
         super().__init__(init_pos=init_pos, **kwargs)
 
+    def set_current_position(self, position):
+        """
+        Adjust offset until the current position is the input position.
```
ZryletTC, line 46: FastMotors don't have offsets, right?

ZryletTC: Also, was this put here as a start for adding `set_current_position` to `SyncAxesBase` as well?

pcdsdevices/pseudopos.py
```diff
@@ -167,14 +167,14 @@ class Parallel(SyncAxesBase):
     """
 
     pseudo = Cpt(PseudoSingleInterface)
+    _offsets = None
```
ZryletTC, line 170: What are your thoughts on this later becoming a notepad signal in some form?

ZryletTC (commented):
Looks like this'll work 👍 Thanks for working through all that with Diling!

ZLLentz:
I actually am planning to scrap this in favor of #775 which has more of Diling's checklist

pcdsdevices/sim.py
ZLLentz: I don't remember why I put this here

pcdsdevices/pseudopos.py
ZLLentz: It's plausible to make this a notepad signal, but it's even better to statically define it for the class

ZryletTC: Right, it should be defined per class, but for classes where they could be variable, it might be good to make their offsets recorded as PVs, right?

ZLLentz: I'm not sure how to do this generically for a parent class that has no idea how many real axes there will be in the subclass

ZryletTC: Sorry I'm not being very clear. I'm not requesting this be added here. I'm just thinking about whether it would be beneficial for the child classes to implement something like that in the future.

ZLLentz: This would be useful in the cases where your desired offsets change during an experiment

PR #773, opened 2021-03-01
ENH: add set_current_position to PseudoPositioner subclasses

ZLLentz:
## Description

Add a generic set_current_position for PseudoPositioners
Might be a candidate for something to upstream to Ophyd?

## Motivation and Context

closes #772 

## How Has This Been Tested?

WIP

## Where Has This Been Documented?

WIP

## Pre-merge checklist
- [ ] Code works interactively
- [ ] Code contains descriptive docstrings, including context and API
- [ ] New/changed functions and methods are covered in the test suite where possible
- [ ] Test suite passes locally
- [ ] Test suite passes on travis
- [ ] Ran docs/pre-release-notes.sh and created a pre-release documentation page
- [ ] Pre-release docs include context, functional descriptions, and contributors as appropriate

PR #775, opened 2021-03-04
ENH: Sync features for lxt_ttc

### LINKED ISSUE #770: Issues with lxt, txt, lxt_ttc in XPP and XCS

### LINKED ISSUE #772: lxt_ttc.set_current_position (XPP, XCS)
**Feature Request**

The lxt_ttc motor in XPP/XCS should have a `set_current_position` method that correctly sets the current position of the component lxt and txt motors.

**Existing Alternatives**

Need to call `set_current_position` independently on each of the component motors.

**Context**

This could probably be handled generically for pseudopositioners

### LINKED ISSUE #776: Add warning for inconsistent lxt_ttc
**Feature Request**

When lxt_ttc is in an inconsistent state prior to a move (delay and timing are not equal/opposite), show a warning and ask if you want Python to move txt to fix this.

**Existing Alternatives**

Currently it just "fixes" itself immediately, which may not be what you want.

**Context**

Request from Diling

### LINKED ISSUE #777: lxt_ttc should be defined specially per instrument
**Feature Request**

Currently, the lxt_ttc class is defined very specifically in pcdsdevices (newport delay with 2 bounces and vitara timing)
But this class isn't very clear and shouldn't even be included in pcdsdevices, rather each hutch should have something like this in their beamline.py file:
```
class MyDelayClass(DelayBase):
    motor = Cpt(IMS, '')

class TTC(SyncAxesBase):
    txt = Cpt(LaserTiming, MY:XPP:PV, <some kwargs>)
    lxt = Cpt(MyDelayClass, MY:XPP:PV2, n_bounces=14, invert=True, <some kwargs>)

lxt_ttc = TTC('', name='lxt_ttc')
```
This allows the hutch to have better agency over how to define the scripting for their endstation or experiment-specific delay scanning

**Existing Alternatives**

Modify random properties on `pcdsdevices.lxe.LaserTimingCompensation` post-instantiation

**Context**

Based on conversation with Diling

### LINKED ISSUE #778: Add an invert/direction parameter to DelayBase
**Feature Request**

Allow us to define whether increasing the motor value increases or decreases the delay

**Existing Alternatives**

Could reverse the underlying EPICS motor?
Could make special subclasses? https://github.com/pcdshub/pcdsdevices/blob/3b3041da26bcf89ae050ead6c9d06bda6b3e6b8a/pcdsdevices/lxe.py#L376

**Context**

Requested by Diling in zoom

ZLLentz:
## Description

This PR handles many of the requests for `lxt_ttc` by expanding the sync motor handling. To go through them one by one:

- Fix issues where `lxt_ttc` could have inconsistent offset behavior by allowing the `SyncAxis` to define static offsets (which is going to be the default). Since I'm changing a default, I've decided to deprecate the previous `SyncAxesBase`. (closes #770)
- Add `set_current_position` to all pcdsdevices `PseudoPositioner`s (closes #772)
- Add a warning when any `SyncAxis` subclass has motor positions that do not match each other properly, and provide a method (`fix_sync`) to re-synchronize them (closes #776) 
- Provide an example of a clean way to implement `lxt_ttc` per instrument so that the scientists can reconfigure it as needed (closes #777)
- Add direction inversion directly to `DelayBase` instead of requiring a special inversion class. (closes #778)
- Add `set_position` as an alias to `set_current_position` (no issue number, but I remember this from the meeting)

Obsoletes the following PRs:
- closes #771
- closes #773

## Motivation and Context

`lxt_ttc` had issues recently that made the code unusable, we want to:
- fix the issues
- use a more stable and configurable base
- allow the scientists to configure `lxt_ttc` themselves, as it will change between experiments
- add requested features

## How Has This Been Tested?

Added some unit tests

Added pre-release docs

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/pseudopos.py
+++ b/pcdsdevices/pseudopos.py
@@ -750,7 +750,7 @@ def delay_class_factory(motor_class):
 
 
 def delay_instance_factory(
-        motor_class, prefix, egu='s', n_bounces=2, invert=False, **kwargs
+        prefix, motor_class, egu='s', n_bounces=2, invert=False, **kwargs
         ):
     cls = delay_class_factory(motor_class)
     return cls(prefix, egu=egu, n_bounces=n_bounces, invert=invert, **kwargs)
--- a/pcdsdevices/pseudopos.py
+++ b/pcdsdevices/pseudopos.py
@@ -485,11 +485,13 @@ class DelayBase(FltMvInterface, PseudoPositioner):
                       notepad_metadata={'record': 'ao', 'default_value': 0.0})
     motor = None
 
-    def __init__(self, *args, egu='s', n_bounces=2, **kwargs):
+    def __init__(self, *args, egu='s', n_bounces=2, invert=False, **kwargs):
         if self.__class__ is DelayBase:
             raise TypeError(('DelayBase must be subclassed with '
                              'a "motor" component, the real motor to move.'))
         self.n_bounces = n_bounces
+        if invert:
+            self.n_bounces *= -1
         super().__init__(*args, egu=egu, **kwargs)
 
     @pseudo_position_argument   # TODO: upstream this fix
--- a/pcdsdevices/pseudopos.py
+++ b/pcdsdevices/pseudopos.py
@@ -539,25 +539,12 @@ class SyncAxis(FltMvInterface, PseudoPositioner):
         The calculation is:
         1. Subtract the offset
         2. Divide by the scale
-
-        If warn_inconsistent is True, and any of the other real motors has a
-        position inconsistent with the first, this will warn the user about the
-        inconsistency if the calculated position differs by more than the
-        warn_deadband.
         """
         self._setup_offsets()
-        pseudo_calcs = []
-        for attr, pos in real_pos._asdict().items():
-            calc = self.inverse_single(attr, pos)
-            pseudo_calcs.append(calc)
-        pick_answer = pseudo_calcs[0]
-        if self.warn_inconsistent:
-            for calc in pseudo_calcs:
-                if not np.isclose(pick_answer, calc,
-                                  atol=self.warn_deadband, rtol=0):
-                    self.consistency_warning()
-                    break
-        return self.PseudoPosition(sync=pick_answer)
+        attr = real_pos._fields[0]
+        pos = real_pos[0]
+        calc = self.inverse_single(attr, pos)
+        return self.PseudoPosition(sync=calc)
 
     def inverse_single(self, attr, pos):
         """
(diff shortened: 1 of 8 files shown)


ZLLentz:
This is more or less where I want to be with this PR, it has all the things needed for the `lxt_ttc` stuff and more, but generalized into `SyncAxis`. I'm going to add tests and maybe some more docs before I call it done.

I'd like feedback on:
- Was this even a good idea at all?
- Implementation quality/issues

pcdsdevices/pseudopos.py
```diff
@@ -282,6 +309,314 @@ def inverse(self, real_pos):
...
+                raise ValueError(
+                    'Invalid fix_sync_keep_still. Must match a motor.'
+                    ) from None
+
+    def _check_info_dict(self, setting, info_kind):
+        """Helper function to check that all keys in the dict are Cpts"""
+        if setting is not None:
+            try:
+                for attr, _ in setting.items():
+                    getattr(self.__class__, attr)
```
klauer, line 469: Perhaps `if attr not in self.component_names:` instead?

pcdsdevices/pseudopos.py
```diff
@@ -282,6 +309,314 @@ def inverse(self, real_pos):
...
+                self.offset_mode = SyncAxisOffsetMode[self.offset_mode]
+            except KeyError:
+                raise ValueError(
+                    f'Invalid offset_mode: {self.offset_mode}'
+                    ) from None
+        self._check_info_dict(self.offsets, 'offsets')
+        self._check_info_dict(self.scales, 'scales')
+        if self.fix_sync_keep_still is not None:
+            try:
+                getattr(self.__class__, self.fix_sync_keep_still)
```
klauer, line 458: Perhaps `self.fix_sync_keep_still in self.component_names`?

It's set at class definition time, so it should be available here.

pcdsdevices/pseudopos.py
```diff
@@ -282,6 +309,314 @@ def inverse(self, real_pos):
...
+        """Helper function to fill default values into the dict"""
+        if setting is None:
+            setting = {}
+        elif isinstance(setting, dict):
+            setting = copy.copy(setting)
+        else:
+            raise ValueError(
+                f'Invalid {info_kind}: {setting}, must be dict or None')
+        for attr in self._real_attrs:
+            if attr not in setting:
```
klauer, line 485: `setting.setdefault(attr, default)` is kind of nice shorthand for this:

```
In [2]: a = {}

In [3]: a.setdefault('a', 4)
Out[3]: 4

In [4]: a
Out[4]: {'a': 4}

In [5]: a['b'] = 5

In [6]: a.setdefault('b', 7)
Out[6]: 5

In [7]: a
Out[7]: {'a': 4, 'b': 5}
```

pcdsdevices/pseudopos.py
```diff
@@ -282,6 +309,314 @@ def inverse(self, real_pos):
...
+        This gives us the sync readback position.
+        """
+        return (pos - self.offsets[attr]) / self.scales[attr]
+
+    def consistency_warning(self):
+        """
+        Let the user know that the axes are desync'd
+        """
+        if self.warn_inconsistent:
+            logger.warning(
```
klauer:
```suggestion
            self.log.warning(
```

klauer (commented):
I really like this implementation. It's got a lot of flexibility and it's easily configurable.

Some optional style/code suggestions, but none of it has any bearing on the actual implementation.

klauer:
And just noticed a commit came in while I was reviewing - including that info in the status repr makes a lot of sense. 👍

pcdsdevices/pseudopos.py
ZLLentz: I always forget about the self.log

pcdsdevices/pseudopos.py
ZLLentz: I'll do this, avoiding the dunder variables is preferred when possible

pcdsdevices/pseudopos.py
ZLLentz: I'll also do this

pcdsdevices/pseudopos.py
ZLLentz: Never knew about this, will revise accordingly

ZLLentz:
That's the last thing I wanted to do here, should probably see if we can take `LxtTtcExample` for a test drive in XPP and copy that definition to XPP's beamline file

ZryletTC:
Yes, I'll try to test in XPP later today. They might want to run woth the old way for this experiment though.

ZLLentz:
Cool, whatever they want to do is fine

ZLLentz:
This should be done

ZLLentz:
If there are no objections, I'll merge before I stop for the day

klauer:
I take it @ZryletTC 's tests were successful? If so, I have nothing further to add here.

ZryletTC (approved):
They most basic tests are good. It doesn't break everything and an object can be created without errors. Any further implementation stuff can be handled in a new PR if necessary.

klauer (approved):

ZLLentz:
Resolving merge conflict and confirming no CI mishap before merging

PR #787, opened 2021-03-10
ENH: ObjectComponent and InterfaceDevice

### LINKED ISSUE #780: InstanceComponent
## Expected Behavior

Random idea that I had last night.
For one-off devices in beamline setup files, it might be useful to have a construct like:

```
from ophyd import Device, EpicsSignal
from pcdsdevices.component import InstanceComponent as ICpt

my_signal = EpicsSignal('MY:PV')

class MyDevice(Device):
    sig = ICpt(my_signal)
```

So that we can build things like one-off pseudo positioners using existing motor objects constructively rather than duplicatively.

This would never be acceptable in `happi` or even useful at all in applications outside of a single hutch's operations, but I think it could be helpful where users don't really want to know implementation details like `'MY:PV'` but do want to assemble interesting devices.

ZLLentz:
## Description

Create two new kinds of components and one new kind of device:
- `ObjectComponent` (`OCpt`): a component that is simply an inclusion of an already-instantiated device, to be used in a beamline setup file as a shorthand for making one-off combination classes constructively from e.g. `happi` definitions of individual components rather than needing to completely recreate them via duplicating the constructors and `__init__` args.
- `InterfaceComponent` (`ICpt`): a component to assist in creating `InterfaceDevice`s. `InterfaceDevice` lets us do something similar to the above `ObjectComponent` but in a way that keeps class structures re-usable and inspectable. We define an interface where the class will have e.g. a `Signal` named `"sig"` and an `EpicsMotor` named `"mot"`, then at class instantiation time it allows us to pass these pre-assembled objects to the class via the `__init__` `**kwargs`, e.g. `MyClass('PREFIX', name='name', sig=my_sig, mot=my_mot)`

Adds the following helpers:
- `to_interface`: takes a normal class and replaces all elements with `InterfaceComponent` definitions.

## Motivation and Context

- Help us create more intuitive beamline startup files where we want to re-use motors in different contexts as elements of larger devices e.g. `PseudoPositioner`s.
- Potential tie-in with `happi` for users who want to include `happi` entries inside of other `happi` entries via `to_interface` (provided the interface side of this PR got upstreamed to `ophyd`)
- closes #780

## How Has This Been Tested?

Add pytest tests

Docstrings and included in full api docs

## Pre-merge checklist


ZLLentz:
I think there is more to handle here, especially with component kwargs and with documentation, but it's far enough along to warrant a quick look and opinions about the direction/concept

ZLLentz:
I'm also considering deprecating `pcdsdevices.component` and moving it all into `pcdsdevices.device` because the two may need to reference each other more closely and because I'd like to mimic `ophyd`'s structure

klauer:
This is clean and _way_ simpler than I was expecting. I like it and the direction you indicate you're going with it 👍

ZLLentz:
Remaining things to think about and/or do here, plus my initial thoughts:

- Which kwargs of component should be handled, and how?
    - `suffix`: ignore (device already made)
    - `lazy`: ignore (device already made)
    - `trigger_value`: ignore (barely even used in ophyd)
    - `add_prefix`: ignore (device already made)
    - `doc`: pass to `super().__init__` and let it do its thing (not used after component instantiation)
    - `kind`: I have no idea what to do here. Suppose the device is re-used in multiple devices, we don't want to keep changing it, right? I need to look into how ophyd tracks kind on a device vs a subdevice vs a component.
    - `**kwargs`: ignore (device already made)

- Need to look through the library and decide if any existing components are better handled via `InterfaceDevice`
    - obvious candidate is a `DelayMotor(real_motor)` class
    - other `PseudoPositioner`s?

ZLLentz:
Remaining things to think about and/or do here, plus my initial thoughts:

- Which kwargs of component should be handled, and how?
    - `suffix`: ignore (device already made)
    - `lazy`: ignore (device already made)
    - `trigger_value`: ignore (barely even used in ophyd)
    - `add_prefix`: ignore (device already made)
    - `doc`: pass to `super().__init__` and let it do its thing (not used after component instantiation)
    - `kind`: I have no idea what to do here. Suppose the device is re-used in multiple devices, we don't want to keep changing it, right? I need to look into how ophyd tracks kind on a device vs a subdevice vs a component.
    - `**kwargs`: ignore (device already made)

- Need to look through the library and decide if any existing components are better handled via `InterfaceDevice`
    - obvious candidate is a `DelayMotor(real_motor)` class
    - other `PseudoPositioner`s?

klauer:
> kind: I have no idea what to do here. Suppose the device is re-used in multiple devices, we don't want to keep changing it, right? I need to look into how ophyd tracks kind on a device vs a subdevice vs a component.

The "reused in multiple devices" concern is valid, though hopefully not such a huge deal for users.
If we had the device overlay thing we chatted about ages ago, the overlay could wrap the device and you could change kinds as you pleased... I think. (If my recollection of that idea is still accurate)

> Need to look through the library and decide if any existing components are better handled via InterfaceDevice

Pretty well any device we have in the library could be considered a template/interface and be customized... but I think I'm getting ahead of things here.

klauer (approved):
I'm completely sold on this concept and implementation. Really nice work!

Diff looks good style-wise as well.

ZLLentz:
I'll figure out where the bad import is coming from and add a real docstring to `DelayMotor` before merging.

PR #791, opened 2021-03-16
HACK: include janky LaserTiming fix from xpp/xcs live deployment

ZLLentz:
## Description

Sleep a second whenever we move LaserTiming

## Motivation and Context

Things are bad, this works in prod, I have heard no complaints, and I have other things I'd prefer to work on.

### DIFF
--- a/docs/source/upcoming_release_notes/791-LaserTiming-hack.rst
+++ b/docs/source/upcoming_release_notes/791-LaserTiming-hack.rst
@@ -1,0 +1,32 @@
+791 LaserTiming-hack
+####################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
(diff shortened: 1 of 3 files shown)


ZLLentz:
See #752 for a previous cleaner but very failed attempt

klauer (approved):
Works in prod ▶️ approve 👍

klauer:
Oh, wait - `tests/test_lxe.py::test_lasertiming_dmov_pass FAILED                     [ 77%]` - test failure wasn't random, it appears.

ZLLentz:
It's not so surprising that a test about dmov fails when we intentionally break dmov. I'll put a little bit of time into that.

ZryletTC (approved):
I don't like janky hack but better to merge than leave it in dev checkouts so 👍

PR #797, opened 2021-03-23
Tmo lamp magnetic bottle configuration

aegger13:
## Motivation and Context

We have a second configuration for LAMP motors with magnetic bottle

## How Has This Been Tested?

I've tested in hutch python and by changing in Happi
Wrote some other tests

doesn't really

Has some description

## Pre-merge checklist


aegger13:
Added all three of you to see who approves first!

tests/test_lamp_motion.py
```diff
@@ -0,0 +1,40 @@
...
+MOTOR6 = 'TMO:LAMP:MMS:06'
+MOTOR7 = 'TMO:LAMP:MMS:07'
+MOTOR8 = 'TMO:LAMP:MMS:08'
+MOTOR9 = 'TMO:LAMP:MMS:09'
+MOTOR10 = 'TMO:LAMP:MMS:10'
+
+def test_lamp():
+    """Make sure we have assigned correct motors pvs we expect"""
+    lamp = LAMP(prefix='TMO:LAMP', name='test')
+    assert getattr(lamp, 'gas_jet_x').prefix == MOTOR1
```
klauer, line 20: Nitpick
```suggestion
    assert lamp.gas_jet_x.prefix == MOTOR1
```

klauer (commented):
Seems fine, would approve after you do the "Ran docs/pre-release-notes.sh and created a pre-release documentation page" step 👍 

(As far as the tests, they seem a bit superfluous to double-check every prefix - that should be something you as the device-writer do. There's no real problem in principle there, I suppose.)

ZLLentz:
This would fall into the "it's not possible to write a useful test" category

tests/test_lamp_motion.py
ZLLentz: It's probably best to just remove the `test_lamp_motion` file. When a test suite is just "is my code exactly what I wrote in my code" it becomes a maintenance hazard more than a self-check.

ZLLentz: Also note that this is instantiating an object with real PV connections to real devices inside of a test suite! This could potentially be hazardous if something is written wrong and this is run on the tmo subnet.

klauer: For reference - see others that use `make_fake_device()`

aegger13: True, mocks would be better here.  It was mostly a self check.  I'll remove the test file, the lamp_motion devices don't really fit into the framework I see here.

aegger13:
Ok, I think this is ready to merge.  Let me know if you see something else I should add/remove.

docs/source/upcoming_release_notes/800-New_lamp_configuration_for_magnetic_bottle.rst
```diff
@@ -0,0 +1,30 @@
+800 New lamp configuration for magnetic bottle
+#################
+
+API Changes
+-----------
+- new LAMPMangeticBottle device class
```
ZLLentz, line 6: This isn't an API change- or if it is, I need to rename this category to "Backwards-incompatible API breaks". This should not slow down the merge.

ZLLentz (approved):

PR #806, opened 2021-04-01
MNT: fix set position bug

### LINKED ISSUE #805: motor object .set_current_position leaves motor in 'set' state
Currently for some motors (XPP and XCS for sure, no idea about other hutches), in particular dumb motors, the .set_current_position function hangs in the terminal. When Ctr+C out of it, it leaves the motor in the 'set' state.

ZLLentz:
## Description

- Fix an issue where calling ``set_current_position`` on certain motors would cause the ipython session to freeze, leaving the motor in the ``set`` state instead of bringing it back to the ``use`` state.
- Hacky workaround for IMS motor part number strings being unable to be read through pyepics when they contain invalid utf-8 characters. (Is there a better way to do this? Perhaps by picking a different character set for the pyepics string interpretation? There's something I half remember about this)

## Motivation and Context

closes #805 

## How Has This Been Tested?

Locally on `XPP:USR:PRT:MMS:25`

pre-release notes

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/__init__.py
+++ b/pcdsdevices/__init__.py
@@ -20,7 +20,7 @@ def make_new_bts(old_bts):
         try:
             return old_bts(st1)
         except UnicodeDecodeError:
-            return new_bts(st1[1:])
+            return st1.decode("utf-8", "ignore")
     return new_bts
 
 
--- a/pcdsdevices/__init__.py
+++ b/pcdsdevices/__init__.py
@@ -1,4 +1,5 @@
-# Hacky ophyd hotfix
+# Hacky ophyd and pyepics hotfixes
+import epics.ca
 from ophyd.device import Device
 
 from ._version import get_versions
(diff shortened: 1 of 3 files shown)


pcdsdevices/__init__.py
```diff
@@ -12,5 +13,25 @@ def __contains__(self, value):
...
 del __contains__
 
+
+# Fix handling of often corrupt IMS PN fields
+def make_new_bts(old_bts):
+    def new_bts(st1):
+        try:
+            return old_bts(st1)
+        except UnicodeDecodeError:
+            return new_bts(st1[1:])
```
klauer, line 23: Can we try instead `.decode(st1, "ignore")`? https://docs.python.org/3/howto/unicode.html

klauer (approved):
Except for the unicode stuff, looks good to me. And nice find 👍

pcdsdevices/__init__.py
ZLLentz: Yes please, taking all suggestions here

ZLLentz: Here is the string we are dealing with btw:
```
$ caget XPP:USR:PRT:MMS:25.PN
XPP:USR:PRT:MMS:25.PN          \377\375\003\377\373\003\377\373\001MFI3CRL17N4-EE
```

klauer:
```python
In [6]: b'\377\375\003\377\373\003\377\373\001MFI3CRL17N4-EE'.decode("ascii", "ignore")
Out[6]: '\x03\x03\x01MFI3CRL17N4-EE'

In [7]: b'\377\375\003\377\373\003\377\373\001MFI3CRL17N4-EE'.decode("latin-1", "ignore")
Out[7]: 'ÿý\x03ÿû\x03ÿû\x01MFI3CRL17N4-EE'

In [8]: b'\377\375\003\377\373\003\377\373\001MFI3CRL17N4-EE'.decode("utf-8", "ignore")
Out[8]: '\x03\x03\x01MFI3CRL17N4-EE'
```

I think stripping out those bad characters then should be pushed to the component, but at least the above won't raise

ZLLentz: Yep, it definitely works with ignore. The other way we could approach this is to set the pyepics decoding to use the latin-1 character set via changing the environment variable:
https://github.com/pyepics/pyepics/blob/e2370d277f9a1ca200740446fefa2cac234462b9/epics/utils.py#L7

PR #811, opened 2021-04-05
ENH: Slit interface changes and precision in status printout

ZryletTC:
## Description
These are two simple changes requested by scientists in XPP/XCS that have been living in a dev checkout for a little while now. Thought it might be a good time to PR them.

First is slit interface changes. Primarily, this change allows users to do things like `s3(4,5)` and adds the aliases, `vo, vg, ho, hg`, for access to the components.

The other thing is specification of precision for the values used in motor status printouts. For now, I've hardcoded a precision of 3 decimal places, but am definitely open to other suggestions. There's probably a good way to use PREC or smth, right?

It would also be nice if we could update precision globally within an hutch-python session. `%precision` doesn't seem to be work for these printouts but might be worth further investigation.

## Motivation and Context
Scientists like their old-python conventions and unnecessarily long numbers are annoying.

## How Has This Been Tested?
Been running in dev for at least a month I think. Did some basic interactive tests when I wrote it back then. Passes Travis.

...oops

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -22,7 +22,7 @@ from .doc_stubs import basic_positioner_init
 from .interface import FltMvInterface
 from .pseudopos import OffsetMotorBase, delay_class_factory
 from .signal import EpicsSignalEditMD, EpicsSignalROEditMD, PytmcSignal
-from .utils import get_status_float
+from .utils import get_status_float, get_status_value
 from .variety import set_metadata
 
 logger = logging.getLogger(__name__)
@@ -103,10 +103,8 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
         status: str
             Formatted string with all relevant status information.
         """
-        description = get_status_float(status_info, 'description', 'value',
-                                       precision=3)
-        units = get_status_float(status_info, 'user_setpoint', 'units',
-                                 precision=3)
+        description = get_status_value(status_info, 'description', 'value')
+        units = get_status_value(status_info, 'user_setpoint', 'units')
         dial = get_status_float(status_info, 'dial_position', 'value',
                                 precision=3)
         user = get_status_float(status_info, 'position', precision=3)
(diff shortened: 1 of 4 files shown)


pcdsdevices/epics_motor.py
```diff
@@ -103,10 +103,13 @@ def format_status_info(self, status_info):
         status: str
             Formatted string with all relevant status information.
         """
-        description = get_status_value(status_info, 'description', 'value')
-        units = get_status_value(status_info, 'user_setpoint', 'units')
-        dial = get_status_value(status_info, 'dial_position', 'value')
-        user = get_status_value(status_info, 'position')
+        description = get_status_float(status_info, 'description', 'value',
+                                       precision=3)
```
ZLLentz, line 107: Is the description a float?

pcdsdevices/epics_motor.py
```diff
@@ -103,10 +103,13 @@ def format_status_info(self, status_info):
...
+        units = get_status_float(status_info, 'user_setpoint', 'units',
+                                 precision=3)
```
ZLLentz, line 109: Are the units a float?

ZLLentz:
I think setting the precision on these to 3 is fine, these are for the real EPICS motors (not the pseudos, which can have arbitrarily small values)

You can get the precision if you want it via `self.user_readback.metadata['precision']`

ZLLentz:
Also I think the slits changes are :+1:

pcdsdevices/epics_motor.py
ZryletTC: Lol I guess I got carried away. The way `get_status_float` is written though, it functions exactly the same as `get_status_value` for non-floats.

ZryletTC: Would it be a bad idea to condense them to one function?

ZLLentz: I'm not motivated to refactor get_status things because I want to refactor the entire printout setup to use jinja, so if this doesn't cause a problem we can leave it

klauer:
You can do recursive f-string formatting?! This changes things...

klauer (approved):
Looks good to me 👍

ZLLentz (approved):

PR #816, opened 2021-04-08
ENH: Updates for lightpath

ZLLentz:
## Description

Accumulated updates to make devices work in the KFE lightpath

## Motivation and Context

They asked for a lightpath, and it will be more useful if all the elements work

## How Has This Been Tested?

Interactively

Pre-release notes

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/slits.py
+++ b/pcdsdevices/slits.py
@@ -438,7 +438,7 @@ class BeckhoffSlits(SlitsBase):
 
     def __init__(self, prefix, *, name, **kwargs):
         self._started_move = False
-        super().__init__(prefix, name=name, **kwargs)
+        super().__init__(prefix, name=name, nominal_aperture=3, **kwargs)
 
     @exec_queue.sub_value
     def _exec_handler(self, *args, value, old_value, **kwargs):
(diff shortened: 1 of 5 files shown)


ZLLentz:
I'd like to merge this bit (after cleaning up docs) because I haven't made a change to this in a day or so

pcdsdevices/slits.py
```diff
@@ -438,7 +438,7 @@ class BeckhoffSlits(SlitsBase):
 
     def __init__(self, prefix, *, name, **kwargs):
         self._started_move = False
-        super().__init__(prefix, name=name, **kwargs)
+        super().__init__(prefix, name=name, nominal_aperture=3, **kwargs)
```
klauer, line 441: Related nice explanation of this parameter is in the `LusiSlits`:

```
    ... we set a
    `nominal_aperture` for each unique device along the beamline. This is
    value is considered the smallest the slit aperture can become without
    blocking the beamline. Both the `xwidth` and the `ywidth`(height) need to
    exceed this `nominal_aperture` for the slits to be considered removed.
```

and the check is pretty simple in `SlitsBase`, where widths for the case of `BeckhoffSlits` is `xwidth` and `ywidth`:

```python
        self._inserted = (min(widths) < self.nominal_aperture.get())
```
```

klauer (approved):
I think this looks good. Comment does not require any action, just a note as I read through what this PR was doing.

pcdsdevices/slits.py
ZLLentz: Might want to move this parameter to `happi` at some point- not particularly in a hurry to do so.

PR #819, opened 2021-04-13
MNT: pass the object's class to the cpt init so pseudopos work correctly

ZLLentz:
## Description

In `ObjectComponent`, instead of just passing `object` as the `cls`, pass the class of the object we're using.

## Motivation and Context

Some classes like `PseudoPositioner` check the `cls` fields of their components as part of the initialization process. This aims to make the `cls` field for an `ObjectComponent` more similar to a normal component.

## How Has This Been Tested?

Interactively in xpp's beamline file

will add pre-release docs

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/819-objcpt-cls.rst
+++ b/docs/source/upcoming_release_notes/819-objcpt-cls.rst
@@ -1,0 +1,31 @@
+819 objcpt-cls
+##############
+
+API Changes
+-----------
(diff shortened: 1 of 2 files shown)


klauer (approved):
Makes sense, for introspection as well as these pseudopositioner things (which I had forgotten about).

Looks good to me. 👍

PR #829, opened 2021-05-12
ENH: UpdateComponent, lock down mirror offsets and others

ZLLentz:
## Description

- Add `UpdateComponent`, a `Component` subclass that allows you to inherit and update a super class's components.
- Add `BeckhoffAxisNoOffset`, a variation on `BeckhoffAxis` that uses `UpdateComponent` to set the user offset related signals to read-only.
- Apply `BeckhoffAxisNoOffset` to all static beamline motors.

## Motivation and Context

For the beamline Beckhoff axes, we don't want to be changing their directions or offsets accidentally. In fact these should always always always be 0/POS in 99% of cases, and some other static value in the other 1% of cases. An easy/clean way to handle this is to prevent the Python from writing to them by making them `EpicsSignalRO`. This has to be specifically targeted at specific instances, both for backcompat and for the end stations that do want to make use of the EPICS-level direction and offset controls.

Matt Seaberg originally pointed this out for the mirrors in https://jira.slac.stanford.edu/browse/LCLSPC-131, I have extended this to all beamline axes that I deemed appropriate.

To implement this, I decided to finally create the thing I've wanted for a while: an easy way to inherit/edit components inside a device definition, the `UpdateComponent`. I think this implementation minimizes the duplication of typing and feels natural to use.

## How Has This Been Tested?

Only interactively, so it is possible that I have broken something- hopefully the tests will catch it.
Before merging I need to add some tests for `UpdateComponent` (which I plan to upstream to `ophyd`).

Not enough yet

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/device.py
+++ b/pcdsdevices/device.py
@@ -272,17 +272,13 @@ class UpdateComponent(Component):
     def __init__(self, **kwargs):
         # Store the original kwargs for use later
         self.update_kwargs = kwargs
+        # Begin with "something" as the copy_cpt to avoid issues on edge cases.
+        self.copy_cpt = None
 
     def __set_name__(self, owner, attr_name):
-        # Find the parent cpt of the same name
+        # Find the parent cpt of the same name and copy it
         parent_cpt = getattr(owner.mro()[1], attr_name)
-
-        if isinstance(parent_cpt, UpdateComponent):
-            # If the parent is also an updater, copy the copy
-            self.copy_cpt = copy.copy(parent_cpt.copy_cpt)
-        else:
-            # Otherwise, copy the original
-            self.copy_cpt = copy.copy(parent_cpt)
+        self.copy_cpt = copy.copy(parent_cpt)
 
         # Edit this object as needed
         for key, value in self.update_kwargs.items():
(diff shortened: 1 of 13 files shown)


ZLLentz:
I won't merge this until next week when I'll have finished the tests and Ken has had a chance to review after his vacation, but I might put this into dev pcdsdevices to get the typhos screen update out if the test suite passes.

pcdsdevices/device.py
```diff
@@ -261,3 +263,70 @@ def to_interface(device_class):
...
+    """
+    def __init__(self, **kwargs):
+        # Store the original kwargs for use later
+        self.update_kwargs = kwargs
+        # Begin with "something" as the copy_cpt to avoid issues on edge cases.
+        self.copy_cpt = None
+
+    def __set_name__(self, owner, attr_name):
+        # Find the parent cpt of the same name and copy it
+        parent_cpt = getattr(owner.mro()[1], attr_name)
```
klauer, line 303: I think this is good enough for the majority of cases, but you might search the `mro` to deal with multiple inheritance.

klauer (commented):
This is pretty neat! I think it's implemented cleanly and just had one minor suggestion.
I'm assuming that since this is already in use, it's working well.

Aside from the necessary documentation steps, I can't see anything else that might need to be done here. Will wait for that to approve.

pcdsdevices/device.py
ZLLentz: Ahh, good point, multiple inheritance breaks the simple grab here.

ZLLentz: Fixed this now by iterating through `mro()[1:]` (skip own class)

klauer (approved):
Nothing to add from previous review, as you resolved the one thing I noticed. Looks good 👍

PR #846, opened 2021-07-09
MNT: pmgr CI fixes

ZLLentz:
## Description

- Create the pmgr singleton only on first instance of IMS class, we don't need to do it if we aren't using IMS
- Allow us to move on if the pmgr singleton fails to be created
- flake8 edits for style consistency (line length, import sorting)

## Motivation and Context

- Fails travis on import because we are running a script that only exists on NFS when we load the IMS class
- Fails the flake8 check in travis

## How Has This Been Tested?

Locally and hopefully travis passes now

I'll make a release snippit

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/846-pmgr-ci.rst
+++ b/docs/source/upcoming_release_notes/846-pmgr-ci.rst
@@ -1,0 +1,33 @@
+846 pmgr
+########
+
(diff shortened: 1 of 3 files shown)


mcb64 (approved):
Looks good to me.

PR #851, opened 2021-07-14
BLD: some pypi fixes

### LINKED ISSUE #852: Error in .gitattributes breaks version packaging
https://github.com/pcdshub/pcdsdevices/blob/24a61eb480609c0b6fc48dc47edd60f8342bb05e/.gitattributes#L1

This tells github to replace the template strings in this file with version information when we make a tag, but since the path is wrong, it does not work, so when built from github archive tar we are missing the version information.

klauer:
Closes #852

(For the record, @ZLLentz and I agreed offline that no real changelog is necessary here; this is mainly a package building bugfix thing. While going through these fixes, we also ran into #852 after considerable confusion)


ZLLentz:
closes #852

ZLLentz (approved):

PR #859, opened 2021-08-17
ENH: GroupDevice

### LINKED ISSUE #617: Idea: OrphanerDevice (for large, system-level device hierarchies)
Large hierarchies of devices can be convenient to design, but cause problems when scanning.

ophyd (really bluesky) make the broad assumption that *all* devices in a hierarchy must be functioning to properly stage any device in the tree. This is erring on the side of caution and overall makes sense.

The split and delay system may be the largest example we have (?): https://github.com/pcdshub/hxrsnd/blob/9b05312df491903ec9cc9e7bc14a11b7c4984112/hxrsnd/sndsystem.py#L20

At the top level, all devices are related mainly as a convenient grouping into a "system" - but if a camera is failing to stage in a distant relative's tree, scanning a leaf-node motor should not be an issue as it is currently.

"OrphanerDevice"
-------------------

```python
class System(OrphanerDevice):
    subsystem1 = Cpt(...)
    subsystem2 = Cpt(...)
    subsystem3 = Cpt(...)

system = System()
RE(scan([system.subsystem1], ...))   # only stages subsystem1 and below

# It could do this by effectively orphaning its children at init time:
assert system.subsystem1.parent is None  # no going back up the hierarchy, sorry

# system.stage() -> RuntimeError
```

This allows you to still group the system together using the convenient component notation, scan unrelated subsystems, and so on.

I'm sure there are downsides I'm not thinking of here, but throwing it out there for now.

ZLLentz:
#### ENH
- Add GroupDevice: A device that is a group of components that will act independently.

This has the following implications:
- Components will have no references to this parent device. If
      accessed and used out of context, the components will be as if
      they were instantited completely separately.
- If a component is staged in a bluesky plan, it will not stage
      the ``GroupDevice``, and therefore will not stage the entire
      device tree.
- ``GroupDevice`` instances by default do nothing when staged.
      You can add specific subdevices to the stage list by setting the
      ``stage_group`` class attribute to a list of components.
- Following from the previous point, note that ``GroupDevice``
      instances cannot process top-level ``stage_sigs``. If you need
      top-level ``stage_sigs``, you should instead contain them in a
      subdevice that is not a ``GroupDevice``.
- ``GroupDevice`` instances that implement ``set`` are required
      to specify a ``stage_group`` to help remind you that these classes
      really do need to stage "something" before scanning. If your
      movable device really does not need this, you can set ``stage_group``
      to an empty list.
- When represented in typhos, we'll see the ``GroupDevice`` screen
      instead of the default device screens.
      (Note: at time of writing, this hypothetical ``GroupDevice``
      ui template does not yet exist).
- Certain devices will completely break if we remove their subdevice
      references: for example, consider the ``PsuedoPositioner`` class.
      For classes like these, we'll need to keep the parent references
      for the ``PseudoSingle`` instances. For the full list of classes that
      need to retain their ``parent`` attribute, see
      ``GroupDevice.needs_parent``.

#### Apply GroupDevice to:
- Acromag
- ArrivalTimeMonitor
- BaseGon
- BeckhoffJet
- BeckhoffJetManipulator
- BeckhoffJetSlits
- CCM
- CrystalTower1
- CrystalTower2
- CVMI
- DiagnosticTower
- ExitSlits
- FFMirror
- FlowIntegrator
- GasManifold
- ICT
- Injector
- IPIMB
- IPMDiode
- IPMMotion
- Kappa
- KBOMirror
- KMono
- KTOF
- LAMP
- LAMPMagneticBottle
- LaserInCoupling
- LCLS2ImagerBase
- LODCM
- LODCMEnergyC
- LODCMEnergySi
- Mono
- MPODApalisModule
- MRCO
- OffsetMirror
- PCM
- PIM
- PulsePickerInOut
- ReflaserL2SI
- RTDSBase
- SamPhi
- Selector
- SlitsBase
- StateRecordPositionerBase
- VonHamosCrystal
- VonHamosFE
- Wave8
- WaveFrontSensorTarget
- XOffsetMirror
- XYZStage

#### BUG
- Fix some race conditions in FuncPositioner
- Fix a race condition in schedule_task that could cause a task to never be run

#### MNT
- Clean up pmgr loading a bit (my previous edit to Mike's setup was a bit rough)
- Edit stage/unstage on PIMY to be compatible with GroupDevice
- Edit stage/unstage and the class definition on SlitsBase to be compatible with GroupDevice
- Move PVStateSignal from state.py to signal.py to avoid a circular import
- Make the tests importable and runnable on Windows

#### BLD
- Require Python 3.9 for type annotations
- Make pmgr optional, but if installed make sure it has a compatible version.

#### CI
- Update to 3.9-only version
- Fix the PIP test build

## Motivation and Context

closes #617 
A better approach to staging than in #857 

## How Has This Been Tested?

Added `GroupDevice` tests

Added pre-release notes

## Pre-merge checklist

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -13,10 +13,10 @@ build:
 
 requirements:
   build:
-    - python >=3.6
+    - python >=3.9
     - setuptools
   run:
-    - python >=3.6
+    - python >=3.9
     - bluesky >=1.6.4
     - happi
     - jsonschema
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -32,6 +32,8 @@ requirements:
     - pyyaml
     - schema
     - scipy
+  run_constrained:
+    - pmgr >=2.0.2
 
 test:
   imports:
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -26,7 +26,6 @@ requirements:
     - pcdscalc >=0.2.0
     - pcdsutils >=0.4.0
     - pint
-    - pmgr >=2.0.2
     - prettytable
     - pyepics >=3.4.2
     - pytmc >=2.7.0
--- a/docs/source/upcoming_release_notes/859-group_device.rst
+++ b/docs/source/upcoming_release_notes/859-group_device.rst
@@ -15,54 +15,54 @@ Device Updates
 
   - Acromag
   - ArrivalTimeMonitor
-  - CCM
-  - CVMI
-  - KTOF
-  - ICT
   - BaseGon
-  - XYZStage
-  - SamPhi
-  - Kappa
-  - IPMDiode
-  - IPMMotion
-  - IPIMB
-  - Wave8
-  - Injector
+  - BeckhoffJet
   - BeckhoffJetManipulator
   - BeckhoffJetSlits
-  - BeckhoffJet
+  - CCM
+  - CrystalTower1
+  - CrystalTower2
+  - CVMI
+  - DiagnosticTower
+  - ExitSlits
+  - FFMirror
+  - FlowIntegrator
+  - GasManifold
+  - ICT
+  - Injector
+  - IPIMB
+  - IPMDiode
+  - IPMMotion
+  - Kappa
+  - KBOMirror
+  - KMono
+  - KTOF
   - LAMP
   - LAMPMagneticBottle
   - LaserInCoupling
-  - CrystalTower1
-  - CrystalTower2
-  - DiagnosticTower
-  - LODCMEnergySi
-  - LODCMEnergyC
+  - LCLS2ImagerBase
   - LODCM
-  - OffsetMirror
-  - XOffsetMirror
-  - KBOMirror
-  - FFMirror
+  - LODCMEnergyC
+  - LODCMEnergySi
+  - Mono
   - MPODApalisModule
   - MRCO
+  - OffsetMirror
+  - PCM
   - PIM
-  - LCLS2ImagerBase
   - PulsePickerInOut
   - ReflaserL2SI
   - RTDSBase
+  - SamPhi
   - Selector
-  - PCM
-  - FlowIntegrator
-  - GasManifold
   - SlitsBase
-  - ExitSlits
-  - KMono
+  - StateRecordPositionerBase
   - VonHamosCrystal
   - VonHamosFE
-  - Mono
-  - StateRecordPositionerBase
+  - Wave8
   - WaveFrontSensorTarget
+  - XOffsetMirror
+  - XYZStage
 
 - Clean up pmgr loading a bit on IMS (my previous edit to Mike's setup was a bit rough)
 - Edit stage/unstage on PIMY to be compatible with GroupDevice
--- a/docs/source/upcoming_release_notes/859-group_device.rst
+++ b/docs/source/upcoming_release_notes/859-group_device.rst
@@ -79,7 +79,7 @@ Bugfixes
 
 Maintenance
 -----------
-- Move PVstateSignal from state.py to signal.py to avoid a circular import
+- Move PVStateSignal from state.py to signal.py to avoid a circular import
 - Make the tests importable and runnable on Windows
 - Require Python 3.9 for type annotations
 - Make pmgr optional, but if installed make sure it has a compatible version.
--- a/docs/source/upcoming_release_notes/859-group_device.rst
+++ b/docs/source/upcoming_release_notes/859-group_device.rst
@@ -1,0 +1,91 @@
+859 group_device
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- Add GroupDevice: A device that is a group of components that will act independently.
+
+Device Updates
+--------------
+- The following devices have become group devices:
+
+  - Acromag
+  - ArrivalTimeMonitor
+  - CCM
+  - CVMI
+  - KTOF
+  - ICT
+  - BaseGon
+  - XYZStage
+  - SamPhi
+  - Kappa
+  - IPMDiode
+  - IPMMotion
+  - IPIMB
+  - Wave8
+  - Injector
+  - BeckhoffJetManipulator
+  - BeckhoffJetSlits
+  - BeckhoffJet
+  - LAMP
+  - LAMPMagneticBottle
+  - LaserInCoupling
+  - CrystalTower1
+  - CrystalTower2
+  - DiagnosticTower
+  - LODCMEnergySi
+  - LODCMEnergyC
+  - LODCM
+  - OffsetMirror
+  - XOffsetMirror
+  - KBOMirror
+  - FFMirror
+  - MPODApalisModule
+  - MRCO
+  - PIM
+  - LCLS2ImagerBase
+  - PulsePickerInOut
+  - ReflaserL2SI
+  - RTDSBase
+  - Selector
+  - PCM
+  - FlowIntegrator
+  - GasManifold
+  - SlitsBase
+  - ExitSlits
+  - KMono
+  - VonHamosCrystal
+  - VonHamosFE
+  - Mono
+  - StateRecordPositionerBase
+  - WaveFrontSensorTarget
+
+- Clean up pmgr loading a bit on IMS (my previous edit to Mike's setup was a bit rough)
+- Edit stage/unstage on PIMY to be compatible with GroupDevice
+- Edit stage/unstage and the class definition on SlitsBase to be compatible with GroupDevice
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- Fix some race conditions in FuncPositioner
+- Fix a race condition in schedule_task that could cause a task to never be run
+
+Maintenance
+-----------
+- Move PVstateSignal from state.py to signal.py to avoid a circular import
+- Make the tests importable and runnable on Windows
+- Require Python 3.9 for type annotations
+- Make pmgr optional, but if installed make sure it has a compatible version.
+- Update to 3.9-only CI
+- Fix the CI PIP test build
+
+Contributors
+------------
+- ZLLentz
--- a/pcdsdevices/beam_stats.py
+++ b/pcdsdevices/beam_stats.py
@@ -1,9 +1,9 @@
 import logging
 
 from ophyd.device import Component as Cpt
+from ophyd.device import Device
 from ophyd.signal import AttributeSignal, EpicsSignal, EpicsSignalRO
 
-from .device import GroupDevice
 from .interface import BaseInterface
 from .pv_positioner import PVPositionerDone
 from .signal import AvgSignal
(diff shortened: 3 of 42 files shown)


ZLLentz:
I'm running into a few very specific edge cases here, which collide in the CCM and LODCM constructs:
1. A `PseudoPositioner`, whose `PseudoSingle`s need `parent` set correctly
2. A `Device` that contains a `GroupDevice` can not be staged
3. `Positioner`s in general that become `GroupDevice` instances cannot be used in scans

When we strip the parental structure from a `PseudoSingle`, it breaks the `PseudoPositioner`, but some `PseudoPositioner`s really are group devices outside of one or two real motors that should be connected and staged.

When a `Device` contains a `GroupDevice`, the top-level device can no longer be staged because it causes a cascade where the `GroupDevice` must be staged. But what if the top level device is a `Positioner` or a `PseudoPositioner`, and therefore can not be a `GroupDevice`? Everything falls apart.

I think the `GroupDevice` needs the following
1. A formalism for partial staging: only some subdevices are staged. This is for things where your `set` method on the top-level devices interacts with specific subdevices.
2. An automatic blacklist of subdevices that are not allowed to be orphaned (so far, just `PseudoSingle`). These retain the `parent` link.
3. No exceptions raised on group device staging (I can't figure out what the issue was here)

ZLLentz:
This is essentially ready for review. Before marking it as such I'm going to clean up the description a bit.

pcdsdevices/pim.py
```diff
@@ -45,14 +47,22 @@ class PIMY(InOutRecordPositioner, BaseInterface):
...
-        self._original_vals[self.state] = self.state.get()
+        self._pre_stage_state = self.state.get()
+        return super().stage()
+
+    def unstage(self) -> list[OphydObject]:
+        """Load the original position that was saved on `stage`."""
+        if self._pre_stage_state is not None:
+            self.state.put(self._pre_stage_state)
+        self._pre_stage_state = None
         return super().stage()
```
klauer, line 62:
```suggestion
        return super().unstage()
```

Makes me think we might need a "generically try to stage/unstage all fake device classes" test

pcdsdevices/slits.py
```diff
@@ -58,10 +58,13 @@ class SlitsBase(MvInterface, Device, LightpathMixin):
...
-    ywidth = None
-    xcenter = None
-    ycenter = None
+    xwidth = Cpt(SignalRO)
+    ywidth = Cpt(SignalRO)
+    xcenter = Cpt(SignalRO)
+    ycenter = Cpt(SignalRO)
+
+    # The gap opens/closes when we move the slits device
+    stage_group = [xwidth, ywidth]
```
klauer, line 67: I see, because this has to refer to a `Component` instance. Ophyd worked around that by accepting either a string name or an instance (if I recall correctly). I think allowing one type as you've done here is probably cleaner.

Semi-related - should there be an `__init_subclass__` for this to ensure that `xwidth` etc are set appropriately? I realize as I'm typing this it's probably overkill, but I'm already near the end of my sentence so 'submit'.

pcdsdevices/slits.py
```diff
@@ -243,10 +246,16 @@ def stage(self):
...
-        self._original_vals[self.xwidth.setpoint] = self.xwidth.readback.get()
-        self._original_vals[self.ywidth.setpoint] = self.ywidth.readback.get()
+        self._pre_stage_gap = self.position
         return super().stage()
 
+    def unstage(self):
+        """
+        Restore the initial values of the aperture position.
+        """
+        self.move(self._pre_stage_gap[0], self._pre_stage_gap[1], wait=True)
```
klauer, line 256: This seems like the right approach as compared to the old one 👍 

Not-important-but-a-notable-in-my-mind change: `_pre_stage_gap` doesn't seem to be set in `__init__` above, meaning this will raise a different exception if `unstage` is called before `stage` than before.

pcdsdevices/utils.py
```diff
@@ -198,13 +198,12 @@ def schedule_task(func, args=None, kwargs=None, delay=None):
             break
 
     def schedule():
-        if matched_thread is None:
+        if matched_thread is not None and context.event_thread is not None:
```
klauer, line 201: Pesky `if` statements - this seems an important fix! Main source of the race condition you ran into, perhaps?

klauer (commented):
This all looks good to me. Nothing I saw required any real changes.

Fine print: I did not give as much thought as you have to what should be a `GroupDevice`, nor did I consider every case of it. If you would like me to do that double-check, just let me know.

conda-recipe/meta.yaml
```diff
@@ -26,7 +26,6 @@ requirements:
     - pcdscalc >=0.2.0
     - pcdsutils >=0.4.0
     - pint
-    - pmgr >=2.0.2
```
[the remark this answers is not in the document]
ZLLentz: Thank you, I wanted to do this but couldn't find the keyword in the conda docs

pcdsdevices/beam_stats.py
```diff
@@ -1,18 +1,17 @@
...
+from .device import GroupDevice
 from .interface import BaseInterface
 from .pv_positioner import PVPositionerDone
 from .signal import AvgSignal
 
-
 logger = logging.getLogger(__name__)
 
 
-class BeamStats(BaseInterface, Device):
```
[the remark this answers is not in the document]
ZLLentz: I went back and forth on this one and a few others multiple times. I ended up making it a GroupDevice because of the differing IOC sources, but in practice nothing is done to an EpicsSignal on stage... so the decision is pretty moot. This doesn't even affect connection checking. It doesn't do anything at all I'm pretty sure. Maybe if it doesn't do anything at all it's better to reduce complexity and I should change singnal-only devices back.

pcdsdevices/device.py
```diff
@@ -336,3 +341,82 @@ def __copy__(self):
...
+        PVStateSignal,
+        ]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Remove references to parent (in this case, self)
+        for cpt_name in self.component_names:
+            cpt = getattr(self, cpt_name)
+            # The following types break without parents
+            if not any(isinstance(cpt, cls) for cls in self.needs_parent):
```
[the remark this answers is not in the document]
ZLLentz: Ahh, thanks for the simplification. I had a memory of this syntax working, but then I went to check the docs and misread them, deciding that it wouldn't work. But this definitely does.

pcdsdevices/device.py
```diff
@@ -336,3 +341,82 @@ def __copy__(self):
...
+                cpt._parent = None
+        if self.stage_group is None:
+            self.stage_group = []
```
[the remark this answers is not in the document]
ZLLentz: I got a bit confused reading this because of the line this is attached to, which I think is not related to your comment.

But what you're saying is: in most cases, there is a class-defined `stage_group: list[Component]` and the list is the same object referenced from every instance. And while this does not currently cause a problem, it is good practice to replace the class-shared definition with a new list of the same contents to avoid issues with future code modifying a shared resource.

Is that correct?

pcdsdevices/pim.py
ZLLentz: Oops, this one even passes the PIM stage/unstage test though!

pcdsdevices/device.py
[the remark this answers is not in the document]
klauer: Yes, you have it exactly. Sorry for the confusion there; it may have been better to comment on the class body list instead.

pcdsdevices/beam_stats.py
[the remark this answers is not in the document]
klauer: I reflected on this a bit more since looking at the code - and I agree that it shouldn't really change anything either way at the device level.

But let's go one level up: it _may_ influence what happens in typhos, when that future GroupDevice template comes about. In which case, you could either design that template around this case being OK, or just not allowing it to start with. Hard to say until you get there, but the latter option may be slightly easier.

pcdsdevices/slits.py
ZLLentz: Right, I locked it as `Component`-only to encourage lintable code, with the assumption that we wanted to easily catch typos in component names that are harder to catch when you pass in a string.

I think it would make sense to take it a step further, something like making sure they refer to valid components?

pcdsdevices/slits.py
ZLLentz: Thanks, the old way was confusing, but perhaps more importantly, did not work as a `GroupDevice`

I think it's worth initializing the storage variable here. I'll revise appropriately.

pcdsdevices/utils.py
ZLLentz: Yes! The critical bug here was that it was possible for `schedule` to exit without running any code at all!

pcdsdevices/beam_stats.py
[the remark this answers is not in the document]
ZLLentz: Oh that's a really good point. With that logic, these should definitely not be `GroupDevice`s, the template will assume that we have 2+ subdevices and deprioritize the top-level signals.

pcdsdevices/slits.py
klauer: I like that - the additional check of 'is this a valid component'. 👍

pcdsdevices/beam_stats.py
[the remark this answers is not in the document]
ZLLentz: I've identified 4 devices to change back:
- BeamStats (all sigs)
- LCLS (all sigs)
- MPODApalisModule (all sigs)
- IonPumpWithController (mostly sigs, one subdevice)

ZLLentz:
To-do:
- Writeup about the changes
- Direct unit tests (currently, tested indirectly through existing tests)
- Grep around for `stage`, `self.parent`, `self._parent`

pcdsdevices/beam_stats.py
```diff
@@ -8,7 +8,6 @@
 from .pv_positioner import PVPositionerDone
 from .signal import AvgSignal
 
-
```
ZLLentz, line 11: Hilariously, after changing here and changing back, all the remains is a flake8 fix.

docs/source/upcoming_release_notes/859-group_device.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Features
+--------
+- Add GroupDevice: A device that is a group of components that will act independently.
+
+Device Updates
+--------------
+- The following devices have become group devices:
+
+  - Acromag
```
klauer, line 16: Nitpick: sort by name?

docs/source/upcoming_release_notes/859-group_device.rst
```diff
@@ -0,0 +1,91 @@
...
+- N/A
+
+Bugfixes
+--------
+- Fix some race conditions in FuncPositioner
+- Fix a race condition in schedule_task that could cause a task to never be run
+
+Maintenance
+-----------
+- Move PVstateSignal from state.py to signal.py to avoid a circular import
```
klauer, line 82:
```suggestion
- Move PVStateSignal from state.py to signal.py to avoid a circular import
```

klauer (approved):
Not much more to add - this looks great! 👍

docs/source/upcoming_release_notes/859-group_device.rst
ZLLentz: You're right. I'm just being a bit lazy at the end of a day, should have sorted to begin with.

ZLLentz:
Waiting on travis to re-run docs and then pressing merge and moving on to the more direct CCM updates

PR #868, opened 2021-09-08
ENH: filter motor status spam from other sessions

ZLLentz:
## Description

Make the post-move error or warning about alarm levels only appear if we requested the move in the current session.
This is achieved by adding a context-sensitive log filter to the motor's `self.log`.

Note that this has a limitation where it will have unpredictable behavior if we have multiple objects with the same name. This is because the name is the piece of metadata that identifies the log messages.

## Motivation and Context

Related `ophyd` code:
https://github.com/bluesky/ophyd/blob/dd4b3e389a0202ecacce39fc3965d703c616b0d4/ophyd/epics_motor.py#L283-L291
Note how these are run in a callback based on the done moving PV, which applies to all open sessions.

ref https://jira.slac.stanford.edu/browse/LCLSECSD-168

## How Has This Been Tested?

Added some targetted tests that snoop on the logging outputs.

Docstrings and release notes.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -4,7 +4,7 @@ Module for LCLS's special motor records.
 import logging
 import os
 import shutil
-import typing
+from types import SimpleNamespace
 
 from ophyd.device import Component as Cpt
 from ophyd.device import Device
@@ -15,8 +15,7 @@ from ophyd.signal import EpicsSignal, EpicsSignalRO, Signal
 from ophyd.status import DeviceStatus, MoveStatus, SubscriptionStatus
 from ophyd.status import wait as status_wait
 from ophyd.utils import LimitError
-from ophyd.utils.epics_pvs import (AlarmSeverity, raise_if_disconnected,
-                                   set_and_wait)
+from ophyd.utils.epics_pvs import raise_if_disconnected, set_and_wait
 from pcdsutils.ext_scripts import get_hutch_name
 
 from pcdsdevices.pv_positioner import PVPositionerComparator
@@ -58,8 +57,8 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
            the beamline.
         5. The post-move error logs only appear after a move in that session,
            not after a move in another user's session. This is achieved by
-           modulating self.tolerated_alarm appropriately to tolerate alarms
-           except when we ask for a move in this session.
+           keeping track of whether or not a move was caused by this session
+           and filtering self.log appropriately.
     """
     # Enable/Disable puts
     disabled = Cpt(EpicsSignal, ".DISP", kind='omitted')
@@ -95,16 +94,14 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
     EpicsMotor.low_limit_travel.kind = Kind.config
     EpicsMotor.direction_of_travel.kind = Kind.normal
 
-    # Alarm modulation
-    _tolerated_alarm: AlarmSeverity = AlarmSeverity.NO_ALARM
-    _moved_in_session: bool = False
-
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.subscribe(
             self._reset_moved_in_session,
             event_type=self.SUB_DONE,
+            run=True,
         )
+        self._install_motion_error_filter()
 
     def move(self, position, wait=True, **kwargs):
         self._moved_in_session = True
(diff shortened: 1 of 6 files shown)


ZLLentz:
`LoggerAdapter` objects don't have `addFilter`... sigh

ZLLentz:
Needs one more adjustment, I don't like my solution because it applies one filter per device to the generic filter. I need this to only apply one filter or it could get slow.

ZLLentz:
OK I have tested this with fake devices and it seems to be behaving as I intend. I'm going to click "Ready for review" but not request reviewers until I've cleaned up the description and completed the checklist.

ZLLentz:
There's no obvious real motor I can test this on right now- so I'm leaving the first checkbox blank. I wish we weren't having outages and I wish I still had my motion test PLC.

pcdsdevices/epics_motor.py
```diff
@@ -89,12 +95,20 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
     EpicsMotor.low_limit_travel.kind = Kind.config
     EpicsMotor.direction_of_travel.kind = Kind.normal
 
+    # Registry of log filters by device name
+    _motion_error_log_filters: dict[str, filter] = {}
```
klauer, line 99:
```suggestion
    _motion_error_log_filters: typing.ClassVar[dict[str, typing.Callable]] = {}
```

Maybe? Lazy callable annotation above

pcdsdevices/epics_motor.py
```diff
@@ -89,12 +95,20 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
...
+    _moved_in_session: bool
     _egu: str
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._egu = ''
         self.motor_egu.subscribe(self._cache_egu)
+        self._install_motion_error_filter()
```
klauer, line 107: I think the hard reference in`_motion_error_log_filters` will stop `EpicsMotorInterface` instances from being garbage collected. Not a big deal in production where our objects aren't created for temporary usage, but... may affect the test suite a bit, perhaps? 🤷‍♂️

pcdsdevices/epics_motor.py
```diff
@@ -89,12 +95,20 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
...
+
+    def move(self, position: float, wait: bool = True, **kwargs) -> MoveStatus:
+        self._moved_in_session = True
```
klauer, line 110: Thinking of bluesky, `set` will call `move`, so this is a good spot for the toggle.

klauer (approved):
I like the approach overall; it's a sensible one. 

Add it to the list of things that would be nice to have in ophyd, ...

pcdsdevices/epics_motor.py
```diff
@@ -323,12 +337,75 @@ def egu(self) -> str:
...
+        """
+        try:
+            name = record.ophyd_object_name
+        except AttributeError:
+            logger.debug('logging filter applied to wrong logger')
+            return True
+        try:
+            filt = cls._motion_error_log_filters[name]
+        except KeyError:
+            logger.debug(f'logging filter not found for {name}')
```
ZryletTC, line 363: Aren't we supposed to use %s syntax instead for logger messages?

ZryletTC (approved):
I would like to try out some interactive testing when our systems are back online, but otherwise I think this looks good 😃

pcdsdevices/epics_motor.py
ZLLentz: You're correct, I used the wrong type hinting here.

pcdsdevices/epics_motor.py
ZLLentz: I think you're correct, I need to use the weakref stuff

pcdsdevices/epics_motor.py
ZLLentz: %s in logger messages has two advantages that I know of:
- Allows programs to filter out repeated messages better, since they can be grouped by the original text, provided we intend to use such filters.
- Is slightly faster when a log message is not logged, since the substitution operation can be skipped.

And in my head, one disadvantage:
- The code is harder to read since we are being encouraged to use different substitution strategies in different places.

I'm not interested in putting %s back into my code for any reason and I wish python would support more formatting options in log messages.

klauer: I lean toward `%`-logger-formatting as good practice for the reasons Zach mentioned, plus the fact that formatting bugs can't cause unexpected exceptions. I'd consider that the primary advantage, since a `logger.debug` causing a crash is a facepalm moment for me (I frequently mess them up, if I'm being honest).

I second the wish for better formatting options, and also wouldn't reject what is proposed here. It would be ideal to get coverage on such lines.

pcdsdevices/epics_motor.py
ZLLentz: Hmm, it turns out that a weakref to a method is always immediately dropped, or at least it is in this context. I either don't need to do this or I need to switch it to a weakref mapping of objects instead of callables.

ZLLentz: If we do need a weakref mapping, I'm just going to add an ophyd instantiation callback during the pcdsdevices startup

klauer: https://docs.python.org/3/library/weakref.html#weakref.WeakMethod

ZLLentz: Heading down this WeakMethod path is starting to make the code more complicated than it needs to be for a log filter. I think I'm going to back up and use the ophyd registry stuff because I think it will make everything cleaner (plus we may be interested in the registry itself for other cases)

pcdsdevices/epics_motor.py
ZLLentz: I'll probably either add a test or remove these debug statements outright because it's a bit much to bother. I added these during a testing loop where I couldn't figure out why the filter wasn't filtering.

ZLLentz:
I think the code is much cleaner now, and we're no longer holding permanent object references. `isort` has mangled my comment, but other than that I'm happy with this.

ZLLentz:
I'd like some quick thoughts on the mini `registry` module I added before merging

PR #871, opened 2021-09-10
ENH: CCM calc constant PVs and set_current_position(energy)

ZLLentz:
### CCM calc constant PVs
CCM class currently gets calculation constants from three places:
- defaults in ccm.py
- definitions in beamline.py
- definitions in happi

This replaces all of those options with:
- from PVs defined in the XPP and XCS hutches

This is accomplished using a mix-in class that is applied in several places, along with some light aliasing, so that all of these are valid ways to read/write to the constants:
- ccm.E.theta0.put(15)
- ccm.theta0.put(15)
- ccm.energy.theta0.put(15)
- ccm.E_Vernier.theta0.put(15)

And these will all be synchronized with each other, even between different hutch python sessions.

Note: since this removes configuration options, it will likely require changes to beamline.py and the happi database.

### CCM psuedomotor set_current_position(energy)
Adjusting the CCM offset manually is a pain. This function helps us do it efficiently.
If the scientist has knowledge that the current energy passing through the CCM is a specific value, they can now call `set_current_position` to enforce the value, which will automatically adjust the `theta0` constant appropriately to make it so.
With some light aliasing, this is accessible in multiple locations, such as:
- ccm.E.set_current_position(8)
- ccm.energy.set_current_position(8)
- ccm.energy.energy.set_current_position(8)

### Misc
Adjust the handling of the `egu` property to prevent property access from timing out while assembling device status.

## Motivation and Context

The constants for these calculations are not synchronized between sessions, meaning we can get different results in different terminals and this can mess with both the data collection and the PVs that get written to with the calculated value. This also makes it so we can save values between sessions that are separated in time.

The offset for the CCM is not an intuitive constant to be setting, but the current photon energy is.

## How Has This Been Tested?

Added one new test, verified that the real PVs connect, fixed some things I noticed interactively.

Added pre-release docs.

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/871-ccm-pvs.rst
+++ b/docs/source/upcoming_release_notes/871-ccm-pvs.rst
@@ -1,0 +1,37 @@
+871 ccm-pvs
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- Switch the CCM energy devices to use user PVs as the canonical source
+  of calculation constants. This allows the constants to be consistent
+  between sessions and keeps different sessions in sync with each other.
+- Add ``CCM.energy.set_current_position`` utility for adjusting the CCM
+  theta0 offset in order to synchronize the calculation with a known
+  photon energy values.
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- Fix an issue where epics motors could time out on the getting of
+  the ``egu`` property, which was causing issues with the displaying
+  of device status.
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- zllentz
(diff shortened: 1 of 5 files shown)


ZLLentz:
Note to self: revisit XCS and figure out why the PVs aren't there and set the defaults + switch their theta0 to 0.2311318654103 * 180/np.pi

ZLLentz:
The xcs user pv host is offline- but this helped me test the warning text a bit which was nice. Still need to revisit to drop their theta0 in, but I think the code here is sound. I'll request reviews once I do the release notes writeup.

ZLLentz:
Adding one more test before I'm done here

ZLLentz:
Notes on comments from Ken's zoom review:

- [ ] blocks of string - wrap in parens to normalize the intent level
- [ ] type hints on instance attributes: hint them in the class definition body instead of in the __init__, do not have to assign default values
- [ ] overlap with OptionalEpicsSignal - consider if usage is appropriate
- [ ] consider wrapping the calc value defaults into a re-usable utility
- [ ] consider type aliases for the more complex re-used types
- [ ] egu potential race condition in get

ZLLentz:
Merging with Ken's zoom approval. I'd rather move on to other things than do the last few optional cleanups here.

PR #879, opened 2021-10-01
MNT: Various changes from XPP

ZryletTC:
## Description
This is a collection of changes that were being used in the XPP's dev version of pcdsdevices.
This includes:
- updating the default theta0 for CCM objects
- using a hacky version of BeckhoffAxis in the Liquid Jet classes to work with the old version currently used by IOC
- adding the acceleration field to the IMS class
- changing component names, aliases, and tab whitelists

## Motivation and Context

Some parts were done to make things work (better).
Some parts were done because scientists requested them.

## How Has This Been Tested?

Interactively.

Docstrings have (mostly) been updated accordingly, and everything will be covered in the pre-release notes.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/ccm.py
+++ b/pcdsdevices/ccm.py
@@ -30,7 +30,7 @@ si_111_dspacing = 3.1356011499587773
 si_511_dspacing = 1.0452003833195924
 
 # Defaults
-default_theta0_deg = 14.9792
+default_theta0_deg = 15.1027
 default_theta0 = default_theta0_deg * np.pi/180
 default_dspacing = si_111_dspacing
 default_gr = 3.175
--- a/pcdsdevices/ccm.py
+++ b/pcdsdevices/ccm.py
@@ -918,11 +918,9 @@ class CCM(BaseInterface, GroupDevice, LightpathMixin, CCMConstantsMixin):
              doc='Combined motion of the CCM Y motors.')
 
     lightpath_cpts = ['x']
-    tab_component_names = True
-    tab_whitelist = ['x1', 'x2', 'y1', 'y2', 'y3', 'E', 'E_vernier',
-                     'th2coarse', 'th2fine', 'alio2E', 'E2alio',
-                     'home', 'kill', 'status',
-                     'insert', 'remove', 'inserted', 'removed']
+    tab_whitelist = ['x1', 'x2', 'y1', 'y2', 'y3', 'E', 'E_Vernier',
+                     'th2fine', 'alio2E', 'E2alio', 'alio', 'home',
+                     'kill', 'insert', 'remove', 'inserted', 'removed']
 
     _in_pos: float
     _out_pos: float
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -1009,7 +1009,7 @@ class BeckhoffAxis(EpicsMotorInterface):
         return status
 
 
-class BeckhoffAxis_Pre140PLC(BeckhoffAxisPLC):
+class BeckhoffAxisPLC_Pre140(BeckhoffAxisPLC):
     """
     Disable some newly introduced signals.
     """
@@ -1026,7 +1026,7 @@ class BeckhoffAxis_Pre140(BeckhoffAxis):
     prior to v1.4.0, which is when the homing routines were
     introduced.
     """
-    plc = Cpt(BeckhoffAxis_Pre140PLC, ':PLC:', kind='normal',
+    plc = Cpt(BeckhoffAxisPLC_Pre140, ':PLC:', kind='normal',
               doc='PLC error handling.')
 
     def home(self, *args, **kwargs):
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -577,7 +577,6 @@ class IMS(PCDSMotorBase):
     velocity = Cpt(EpicsSignal, '.VELO', limits=True, kind='config')
     velocity_base = Cpt(EpicsSignal, '.VBAS', kind='omitted')
     velocity_max = Cpt(EpicsSignal, '.VMAX', kind='config')
-    acceleration = Cpt(EpicsSignal, '.ACCL', kind='config')
 
     tab_whitelist = ['reinitialize', 'acceleration', 'clear_.*']
 
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -579,6 +579,7 @@ class IMS(PCDSMotorBase):
     velocity = Cpt(EpicsSignal, '.VELO', limits=True, kind='config')
     velocity_base = Cpt(EpicsSignal, '.VBAS', kind='omitted')
     velocity_max = Cpt(EpicsSignal, '.VMAX', kind='config')
+    acceleration = Cpt(EpicsSignal, '.ACCL', kind='config')
 
     tab_whitelist = ['auto_setup', 'reinitialize', 'clear_.*']
 
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -69,8 +69,8 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
     # Current Dial position
     dial_position = Cpt(EpicsSignalRO, '.DRBV', kind='normal')
 
-    tab_whitelist = ["set_current_position", "home", "velocity", "enable",
-                     "disable", "check_limit_switches", "get_low_limit",
+    tab_whitelist = ["set_current_position", "home", "velocity",
+                     "check_limit_switches", "get_low_limit",
                      "set_low_limit", "get_high_limit", "set_high_limit"]
 
     set_metadata(EpicsMotor.home_forward, dict(variety='command-proc',
@@ -447,8 +447,6 @@ class PCDSMotorBase(EpicsMotorInterface):
     # paused and ready to resume on Go 'Paused', and to resume a move 'Go'.
     motor_spg = Cpt(EpicsSignal, '.SPG', kind='omitted')
 
-    tab_whitelist = ["spg_stop", "spg_pause", "spg_go"]
-
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.stage_sigs[self.motor_spg] = 2
@@ -581,7 +579,7 @@ class IMS(PCDSMotorBase):
     velocity_max = Cpt(EpicsSignal, '.VMAX', kind='config')
     acceleration = Cpt(EpicsSignal, '.ACCL', kind='config')
 
-    tab_whitelist = ['auto_setup', 'reinitialize', 'clear_.*']
+    tab_whitelist = ['reinitialize', 'acceleration', 'clear_.*']
 
     # The singleton parameter manager object.
     _pm = None
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -1014,6 +1014,7 @@ class BeckhoffAxis_Pre140PLC(BeckhoffAxisPLC):
     Disable some newly introduced signals.
     """
     cmd_home = None
+    home_pos = None
     user_enable = None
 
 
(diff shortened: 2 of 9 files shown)


ZryletTC:
Some of these things are a little controversial. Please let me know what you think @klauer and @ZLLentz.

In particular, if you'd rather not have the OldBeckhoffAxis, I'll take it out, I was very on the fence about adding it here. If it is fine, then I'll add appropriate docstrings for it.

pcdsdevices/epics_motor.py
```diff
@@ -579,8 +577,9 @@ class IMS(PCDSMotorBase):
     velocity = Cpt(EpicsSignal, '.VELO', limits=True, kind='config')
     velocity_base = Cpt(EpicsSignal, '.VBAS', kind='omitted')
     velocity_max = Cpt(EpicsSignal, '.VMAX', kind='config')
+    acceleration = Cpt(EpicsSignal, '.ACCL', kind='config')
```
ZLLentz, line 580: This is already defined at https://github.com/bluesky/ophyd/blob/dd4b3e389a0202ecacce39fc3965d703c616b0d4/ophyd/epics_motor.py#L56, does something we do remove it?

pcdsdevices/interface.py
```diff
@@ -34,15 +34,12 @@
...
-                    "wait_for_connection", "stop", "get", "configure"]
-Signal_whitelist = ["value", "put"]
+OphydObject_whitelist = []
+BlueskyInterface_whitelist = []
+Device_whitelist = ["stop"]
+Signal_whitelist = ["value", "put", "get"]
 Positioner_whitelist = ["settle_time", "timeout", "egu", "limits", "move",
-                        "position", "moving", "set_position",
-                        "set_current_position"]
+                        "position", "moving", "set_current_position"]
```
ZLLentz, line 42: I understand that there are requests to cut down on tabs here, but the following removals need a few words of justification:
- `name` is a pretty critical attribute for understanding the bluesky scans
- `connected` is immensely useful for understanding if your object's IOC is running properly
- `set_position` is an alias we were asked to add- only to be asked to remove it??

pcdsdevices/ccm.py
```diff
@@ -918,11 +918,9 @@ class CCM(BaseInterface, GroupDevice, LightpathMixin, CCMConstantsMixin):
...
 
     lightpath_cpts = ['x']
-    tab_component_names = True
-    tab_whitelist = ['x1', 'x2', 'y1', 'y2', 'y3', 'E', 'E_vernier',
-                     'th2coarse', 'th2fine', 'alio2E', 'E2alio',
-                     'home', 'kill', 'status',
-                     'insert', 'remove', 'inserted', 'removed']
+    tab_whitelist = ['x1', 'x2', 'y1', 'y2', 'y3', 'E', 'E_Vernier',
+                     'th2fine', 'alio2E', 'E2alio', 'alio', 'home',
+                     'kill', 'insert', 'remove', 'inserted', 'removed']
```
ZLLentz, line 923: `status` is a scientist-request function- why is it removed from the tab list?

ZLLentz: well, maybe it's added somewhere else in a parent class? I'll read on

pcdsdevices/epics_motor.py
```diff
@@ -69,8 +69,8 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
     # Current Dial position
     dial_position = Cpt(EpicsSignalRO, '.DRBV', kind='normal')
 
-    tab_whitelist = ["set_current_position", "home", "velocity", "enable",
-                     "disable", "check_limit_switches", "get_low_limit",
+    tab_whitelist = ["set_current_position", "home", "velocity",
+                     "check_limit_switches", "get_low_limit",
```
ZLLentz, line 73: enable and disable were requested functions by someone in CXI a few years ago

pcdsdevices/epics_motor.py
```diff
@@ -447,8 +447,6 @@ class PCDSMotorBase(EpicsMotorInterface):
     # paused and ready to resume on Go 'Paused', and to resume a move 'Go'.
     motor_spg = Cpt(EpicsSignal, '.SPG', kind='omitted')
 
-    tab_whitelist = ["spg_stop", "spg_pause", "spg_go"]
```
ZLLentz, line 450: I don't remember the origin of these- probably ok to remove without much more consideration

pcdsdevices/epics_motor.py
```diff
@@ -579,8 +577,9 @@ class IMS(PCDSMotorBase):
     velocity = Cpt(EpicsSignal, '.VELO', limits=True, kind='config')
     velocity_base = Cpt(EpicsSignal, '.VBAS', kind='omitted')
     velocity_max = Cpt(EpicsSignal, '.VMAX', kind='config')
+    acceleration = Cpt(EpicsSignal, '.ACCL', kind='config')
 
-    tab_whitelist = ['auto_setup', 'reinitialize', 'clear_.*']
```
ZLLentz, line 583: the removal of auto_setup confuses me but I think its ok

pcdsdevices/epics_motor.py
```diff
@@ -1011,6 +1010,15 @@ def home(self, direction=None, wait=True, **kwargs):
         return status
 
 
+class OldBeckhoffAxisPLC(BeckhoffAxisPLC):
+    cmd_home = None
+
+
+class OldBeckhoffAxis(BeckhoffAxis):
+    plc = Cpt(OldBeckhoffAxisPLC, ':PLC:', kind='normal',
+              doc='PLC error handling.')
```
ZLLentz, line 1019: I don't mind having these, just add docstrings that note that they are for lcls-twincat-motion prior to v1.4.0 (pre-Nov2020) and we should also override the `home` method appropriately

pcdsdevices/ipm.py
```diff
@@ -323,9 +329,9 @@ class IPIMB(BaseInterface, GroupDevice):
         Trigger component.
     """
 
-    tab_whitelist = ['isum', 'xpos', 'ypos']
+    tab_whitelist = ['sum', 'xpos', 'ypos']
 
-    isum = Cpt(EpicsSignalRO, ':SUM', kind='hinted')
+    sum = Cpt(EpicsSignalRO, ':SUM', kind='hinted')
```
ZLLentz, line 334: I don't even remember why this was isum

pcdsdevices/pim.py
```diff
@@ -88,11 +88,11 @@ class PIM(BaseInterface, GroupDevice):
     _prefix_start = ''
 
     state = Cpt(PIMY, '', kind='normal')
-    zoom_motor = FCpt(IMS, '{self._prefix_zoom}', kind='normal')
+    zoom = FCpt(IMS, '{self._prefix_zoom}', kind='normal')
```
ZLLentz, line 91: I don't mind these shortenings. We likely put the longer versions in out of an abundance of caution not to collide with other things, but it seems like that isn't an issue.

pcdsdevices/epics_motor.py
klauer: I think few scientists would want to use these - agreed on the removal.

pcdsdevices/epics_motor.py
klauer: This is explicitly removing `auto_monitor=True` for some reason as well.

pcdsdevices/epics_motor.py
klauer: Consider a name like `BeckhoffAxis_Pre140` or something to be more explicit.

Ophyd also supports device class versioning for scenarios like this, but it's probably overkill.

pcdsdevices/epics_motor.py
ZLLentz:
```
$ ipython
Python 3.9.6 | packaged by conda-forge | (default, Jul 11 2021, 03:39:48)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.26.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import pcdsdevices

In [2]: pcdsdevices.__version__
Out[2]: '4.7.1'

In [3]: from pcdsdevices.epics_motor import IMS

In [4]: mot = IMS('XPP:USR:MMS:01', name='mot')
Using production server.

In [5]: mot.acceleration
Out[5]: EpicsSignal(read_pv='XPP:USR:MMS:01.ACCL', name='mot_acceleration', parent='mot', value=1.0, timestamp=1632762543.584904, tolerance=0.0001, auto_monitor=True, string=False, write_pv='XPP:USR:MMS:01.ACCL', limits=False, put_complete=False)
```

pcdsdevices/interface.py
klauer: The removal of attrs and all those that Zach mentioned surprised me a bit - scientists don't tweak attrs for bluesky scanning?

We may have been including too much to start with, but overall this strikes me as now including too little. Where the balance is in the middle, I don't really know, to be fair.

pcdsdevices/interface.py
```diff
@@ -552,7 +549,7 @@ class MvInterface(BaseInterface):
     applications.
     """
 
-    tab_whitelist = ["mv", "wm", "camonitor", "wm_update"]
```
klauer, line 555: I personally never use it, but my understanding was `camonitor` was an "old hutch Python thing that was really desirable". No longer the case?

pcdsdevices/ipm.py
klauer: Would recommend a `@property isum -> warn and return self.sum` for back-compat, unless you can say for certain there is no code using this.

klauer (commented):
Just my 2c

pcdsdevices/interface.py
ZLLentz: I think wm_update does exactly the same thing

pcdsdevices/interface.py
ZryletTC: @dlzhu Can you comment on all the tab whitelist changes?

ZLLentz:
@ZryletTC do you mind if I make some small tweaks to this so we can merge?

ZryletTC:
Yes, please do 👍

ZLLentz:
I've implemented the non-whitelist-related suggestions
I'd appreciate a spot-check on my diff so we can get this merged and do the big tag

ZLLentz (approved):

pcdsdevices/ipm.py
```diff
@@ -443,9 +461,9 @@ class IPM_Det(BaseInterface, Device):
     """Base class for IPM_IPIMB and IPM_Wave8. Not meant to be instantiated."""
     tab_component_names = True
 
-    def isum(self):
-        """Returns the detector's isum value."""
-        return self.det.isum.get()
+    def sum(self):
```
klauer, line 464: Seems a bit odd to have accessor methods like these which alias component names of `det`, but it appears to be a common feature here, so 🤷‍♂️

pcdsdevices/jet.py
```diff
@@ -74,19 +74,19 @@ class BeckhoffJetManipulator(BaseInterface, GroupDevice):
 
     tab_component_names = True
 
-    x = Cpt(BeckhoffAxis, ':X', kind='normal')
-    y = Cpt(BeckhoffAxis, ':Y', kind='normal')
-    z = Cpt(BeckhoffAxis, ':Z', kind='normal')
+    x = Cpt(OldBeckhoffAxis, ':X', kind='normal')
```
klauer, line 77: Would recommend using the versioned name here since it's in the PR.
Is the `Old` alias even used outside of this?

klauer (approved):
Looks good overall. 👍

pcdsdevices/jet.py
ZLLentz: I think it might be used in a beamline config somewhere. I left it in out of an abundance of caution.

pcdsdevices/ipm.py
ZLLentz: I largely agree that this pattern can be very confusing, as helpful as it would be sometimes.

PR #897, opened 2021-10-15
Add vrc cls ls

jyin999:
## Pre-merge checklist

### DIFF
--- a/pcdsdevices/valve.py
+++ b/pcdsdevices/valve.py
@@ -128,6 +128,16 @@ class VRC(VVC, LightpathMixin):
         self._removed = lightpath_values[self.open_limit]['value']
 
 
+class VRC_ClsLS(VVC):
+    """
+    Class for Gate Valves with Control and only close limit switch readback.
+    """
+
+    state = Cpt(EpicsSignalRO, ':STATE_RBV', kind='normal', doc='Valve state')
+    closed_limit = Cpt(EpicsSignalRO, ':CLS_DI_RBV', kind='hinted',
+                       doc='Closed limit switch digital input')
+
+
 class VGC(VRC):
     """Class for Controlled Gate Valves."""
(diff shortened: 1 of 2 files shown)


pcdsdevices/valve.py
```diff
@@ -128,6 +128,14 @@ def _set_lightpath_states(self, lightpath_values):
         self._removed = lightpath_values[self.open_limit]['value']
 
 
+class VRCClsLS(VVC):
+    """Class for Gate Valves with Control and Closed Limit Switch Readback."""
```
ZryletTC, line 132: Should LightpathMixin be used here?

pcdsdevices/valve.py
```diff
@@ -128,6 +128,14 @@ def _set_lightpath_states(self, lightpath_values):
         self._removed = lightpath_values[self.open_limit]['value']
 
 
+class VRCClsLS(VVC):
```
ZryletTC, line 131: Looks like a lot of classes in this file aren't covered by the test suite :/

ZLLentz: We should have some overall tests that try to make every device
There's not much meaningful testing that we can do if the classes are just PV definitions

pcdsdevices/valve.py
ZLLentz: I think so

ZLLentz: Probably OK to do in a follow-up PR

pcdsdevices/valve.py
klauer: We kinda do: https://github.com/pcdshub/pcdsdevices/blob/39a9dca3adaf8b46a2f5958084aee87948e0f99f/pcdsdevices/tests/test_interface.py#L271-L281

PR #899, opened 2021-10-19
ENH: allow enum_strs to be overridden with EpicsSignalEditMD

### LINKED ISSUE #867: Alternative source of enum strings for a given signal
Based on https://github.com/pcdshub/typhos/issues/453
I (personally) think it's more appropriate to solve at this level, so I'm duplicating and suggesting something similar here.

## Current Behavior

State positioners (at least semi-PCDS specific) use the enum strings associated with the PV to determine combo box drop down items:

image

## Expected Behavior

However, the underlying device may have more specific - user-friendly, detailed - information pertaining to those states.
For example, in the above we might have this name from the PLC:

`fbStage1.stFilter6.sName                   := '(6) 6.18 um Al';`

It'd be ideal to display these strings instead, which are available through a separate ophyd signal.

## Possible solution

A custom signal type here that allows for an override such as: `PreferredEnumEpicsSignal` or something along those lines with a `preferred_enum_attr_names = ["state1.name", "state2.name", ...]` could solve this at the ophyd level rather than at the typhos level.

Then we'd be free to use either the original enum strings or the preferred ones at the ophyd/bluesky level.

klauer:
## Description

In short - this updates the `pcdsdevices.signal.EpicsSignalEditMD` class variant to allows enum information to be sourced from different PVs (signals) than the original PV. Alternatively, they may be hard-coded.

State positioners (at least semi-PCDS specific) use the enum strings associated with the PV to determine combo box drop down items. Sometimes our enum strings are less accurate than is desirable (`Filter 1` instead of `5 µm Si` for example), with the detailed information stored in another PV.

A few things to consider:
* I think this deserves revisiting on the pytmc end of things, but perhaps this also has general use for non-pytmc signals.
* This is strictly for `EpicsSignal` (but if you have another Signal type, you probably have control over it directly, right?)
* ✅  ~Possible refactor: could merge directly with `EpicsSignalBaseEditMD`. Downside is a bit more code complexity, upside is less inheritance complexity~ 

## Motivation and Context
Closes #867 
 
## How Has This Been Tested?
Typhos-based testing, paired with https://github.com/pcdshub/typhos/pull/459

I'm not sure of a good way to unit test this. Ideally we'd have some EPICS IOC running in Travis like we do in some other repos - but I know that's somewhat controversial here.

Release notes, so far

image

caproto:
```python
class MyEnum(enum.IntEnum):
    off = 0
    on = 1
    unknown = 2


class EnumIOC(PVGroup):
    ""
    primary = pvproperty(value=MyEnum.on, record="mbbi")
    alt_0 = pvproperty(value="Alternate Zero")
    alt_1 = pvproperty(value="Alternate One")
    alt_2 = pvproperty(value="Alternate Two")
```

ophyd:
```python
class EnumTest(pcdsdevices.interface.BaseInterface, ophyd.Device):
    enum0 = ophyd.Component(ophyd.EpicsSignal, "alt_0", string=True)
    enum1 = ophyd.Component(ophyd.EpicsSignal, "alt_1", string=True)
    enum2 = ophyd.Component(ophyd.EpicsSignal, "alt_2", string=True)
    enum_normal = ophyd.Component(pcdsdevices.signal.EpicsSignal, "primary")
    enum_with_attrs = ophyd.Component(
        pcdsdevices.signal.EpicsSignalEditMD, "primary",
        enum_attrs=["enum0", "enum1", "enum2"],
    )
    enum_with_strs = ophyd.Component(
        pcdsdevices.signal.EpicsSignalEditMD, "primary",
        enum_strs=["enum0", "enum1", "enum2"],
    )
```

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/signal.py
+++ b/pcdsdevices/signal.py
@@ -1037,6 +1037,25 @@ class FakeEpicsSignalEditMD(FakeEpicsSignal):
     API stand-in for EpicsSignalEditMD
     Add to this if you need it to actually work for your test.
     """
+    def __init__(
+        self,
+        *args,
+        enum_attrs: Optional[list[Optional[str]]] = None,
+        enum_strs: Optional[list[str]] = None,
+        **kwargs
+    ):
+        super().__init__(*args, **kwargs)
+        self._enum_attrs = enum_attrs
+        self._enum_strs = enum_strs
+
+    @property
+    def enum_attrs(self):
+        return self._enum_attrs
+
+    @property
+    def enum_strs(self):
+        return self._enum_attrs or self._enum_strs or None
+
     def _override_metadata(self, **kwargs):
         ...
 
@@ -1046,6 +1065,25 @@ class FakeEpicsSignalROEditMD(FakeEpicsSignalRO):
     API stand-in for EpicsSignalROEditMD
     Add to this if you need it to actually work for your test.
     """
+    def __init__(
+        self,
+        *args,
+        enum_attrs: Optional[list[Optional[str]]] = None,
+        enum_strs: Optional[list[str]] = None,
+        **kwargs
+    ):
+        super().__init__(*args, **kwargs)
+        self._enum_attrs = enum_attrs
+        self._enum_strs = enum_strs
+
+    @property
+    def enum_attrs(self):
+        return self._enum_attrs
+
+    @property
+    def enum_strs(self):
+        return self._enum_attrs or self._enum_strs or None
+
     def _override_metadata(self, **kwargs):
         ...
 


pcdsdevices/signal.py
```diff
@@ -802,7 +805,7 @@ class FakeEpicsSignalEditMD(FakeEpicsSignal):
     Add to this if you need it to actually work for your test.
     """
     def _override_metadata(self, **kwargs):
-        pass
+        ...
```
ZLLentz, line 808: Curious: is there a functional difference here? Or is it a stylistic thing?

pcdsdevices/signal.py
```diff
@@ -811,8 +814,214 @@ class FakeEpicsSignalROEditMD(FakeEpicsSignalRO):
...
+
+
+class EpicsSignalBaseEditEnum(EpicsSignalBaseEditMD):
+    """
+    EpicsSignal variant which allows for user correction of enum strings.
+
+    Parameters
+    ----------
+    enum_attrs : list of str
+        List of attribute names, relative to the parent device.  That is to
```
ZLLentz, line 831: Some clarification that these are not general attribute strings, these are signals that hold string values
```suggestion
        List of signal names, relative to the parent device.  That is to
```

PR #900, opened 2021-10-20
MNT: lodcm calc should not raise in background thread

ZLLentz:
## Description

If our two crystal orientations do not match, make the calculation NaN instead of throwing an exception.

## Motivation and Context

`inverse` is called in a callback, so if it throws an exception it ~barfs all over the screen~
It no longer spams the screen due to other changes since making this PR, but it still spams the logfile and it means we'll either have no value stored for the pseudoposition, or a stale value in the case of this error state.

## How Has This Been Tested?

Interactively only

Added pre-release notes

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/900-lodcm-nan.rst
+++ b/docs/source/upcoming_release_notes/900-lodcm-nan.rst
@@ -1,5 +1,5 @@
 900 lodcm-nan
-#################
+#############
 
 API Changes
 -----------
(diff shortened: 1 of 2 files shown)


ZLLentz:
I did not write a test for this, but it works interactively and this module needs a full rework anyway.

klauer (approved):

PR #924, opened 2022-01-07
Automated Ophyd Device Generation

### LINKED ISSUE #923: Scripted Ophyd Device Generation
## Expected Behavior
It would be nice to have a script for automated generation of an Ophyd device from a DB file. Many times Ophyd objects are just a bunch of signals, which would be straightforward to generate from a DB file. More complicated devices could be modified as needed by the user after the rote signal components are added. If there are more things in the DB file than are needed in the device, it is usually easier to remove things than to add them. 

## Current Behavior
Currently people code devices by hand. 

## Context
I recently wrote a very large device class for the Wave8 that took a small piece of my soul. After completing that device I would like to avoid that kind of drudgery in the future.

slactjohnson:
## Description
Adds in a script to help automatically generate an ophyd device from an IOC db file. 

## Motivation and Context
This helps reduce the amount of rote work that happens when writing a device composed of signals. 

Closes #923 

## How Has This Been Tested?
Tested on the pgpWave8 IOC, with lots of structure to the DB file, as well as the qadc IOC, which is a pretty flat db file that is representative of many IOCs that we have. 

Docstrings and comments. The argparser provides some CLI help. 

Output for qadc:
image

Fraction of the output for the wave8: 
image

*snip*

image

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/923-Add_script_for_generating_ophyd_classes.rst
+++ b/docs/source/upcoming_release_notes/923-Add_script_for_generating_ophyd_classes.rst
@@ -8,7 +8,7 @@ API Changes
 Features
 --------
 - Adds a new script, make_ophyd_device.py, that helps with autogeneration of
-  an ophyd device class from an IOC db file. 
+  an ophyd device class from an IOC db file. Includes a helper script. 
 
 Device Updates
 --------------
--- a/docs/source/upcoming_release_notes/923-Add_script_for_generating_ophyd_classes.rst
+++ b/docs/source/upcoming_release_notes/923-Add_script_for_generating_ophyd_classes.rst
@@ -1,0 +1,31 @@
+923 Add script for generating ophyd classes
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- Adds a new script, make_ophyd_device.py, that helps with autogeneration of
+  an ophyd device class from an IOC db file. 
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- tjohnson
--- a/make_ophyd_device
+++ b/make_ophyd_device
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+# Source keeps stealing my args
+sourceWrapper() {
+ source /reg/g/pcds/pyps/conda/py36env.sh
+}
+
+sourceWrapper
+
+python -m pcdsdevices.make_ophyd_device "$@"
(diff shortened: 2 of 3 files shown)


slactjohnson:
Looking for some feedback on what's here so far. Currently the script appears to capture the structure of the Wave8 properly, coming up with something surprisingly similar to what I wrote. 

A few thoughts on weaknesses/missing features:

- Repeated units, e.g. adcconfig0, adcconfig1, etc are not captured as a single repeated class.
- All signals that are not RBV signals are setup as EpicsSignals. I could do some smarter things with the metadata from whatRecord to decide whether or not to make a signal EpicsSignalRO. 
- The CLI could be cleaner.

klauer:
Looking good so far! A few random thoughts you shouldn't feel obligated to implement or even respond to:
* RE: `EpicsSignal` vs `EpicsSignalRO` - for a first pass you could also just check that the record type is one of ai, bi, stringin, etc.
* Fortunately we don't really have to worry about motor records popping up in these database files as this would be used with primitive types only
* It'd be good to automatically set up `string=True` for `stringin`/`stringout`/`lsi`/`lso` records (and `waveform` records with `FTVL`=`CHAR`)
* The script assumes you've appropriately broken up your database files sensibly and didn't lump everything into one file - overall a good assumption
* You might consider taking a look at the `inflection` library for converting from snake/camel-case and back. (I _think_ it may already be in our environment, but if not, feel free to ignore this)
* Is the `Class` suffix necessary for the newly-created devices?
* Maybe you could add docstrings indicating the database file the class was generated from?
* How about setting the component`doc=` to the record `.DESC` field?

ZLLentz:
I think this looks really good! Some additional thoughts here.

Small thoughts about kinds:
- Is there any way at all to intuit which signals should be which kind? My guess is no. I think `normal` is a sensible default and we can expect the developer to carefully consider which signals should be hinted/configure/omitted

Small thoughts about packaging:
- This imports from `whatrecord`. Do we need to add add `pcdsdevices` dependencies for scripts like this?

Smaller thoughts about formatting:
- It looks like this uses tabs, whereas `pcdsdevices` is typically 4-space delimited
- Is it worth running the result through an auto-formatter like `black`?  This will take care of a lot of the minutia related to formatting for us without needing to bake it in to the script or manually edit it later.

slactjohnson:
@ZLLentz @klauer Thanks for all the feedback. I implemented some of it. I would like to do the rest, but I'm not sure when I'll get around to it, and I want to stop sitting on this PR. I think it's still useful as is.

ZLLentz (approved):
Given your last comment, I think this is good to merge.

ZLLentz:
travis failure unrelated

klauer (approved):
Maybe just consider making a TODO issue as a note 👍

slactjohnson:
I don't appear to have the ability to merge without the build passing. 

image

klauer:
Going to restart that Travis job and see if it doesn't complete this time. Unexpected failure...

ZLLentz:
Travis says let's go!

PR #929, opened 2022-01-28
ENH: Hdf5 Gige Saving

### LINKED ISSUE #928: [LCLSECSD-571] Add Filestore and trigger mixins for lcls hdf5 devices
## Expected Behavior
We would like to: 
- Save GigE hdf5 files with bluesky style scans
- Provide a method for saving files interactively

## Current Behavior
Current work-arounds like [this](https://github.com/pcdshub/pcdsdevices/blob/216a0283a9c455d8d7f04b870b736351d26327fa/pcdsdevices/areadetector/detectors.py#L65-L85,), which are useful outside of scans but not inside them.  

## Possible Solution
Implement ophyd trigger mixin methods to pcdsdevices area detector mixins.
Adding to pcdsdevices would allow use across different hutches

## Your Environment
Requested by XCS, workarounds exist in RIX

ZLLentz:
## Description

Add an area detector variant class, `PCDSHDF5BlueskyTriggerable`, to satisfy beamline needs for camera image saving outside of the the DAQ in a scan and outside of scans altogether.

## Motivation and Context

closes #928 

Existing implementations are functional, but cannot be used seamlessly in bluesky scans.

## How Has This Been Tested?

Interactively in XCS and by using IM1L1
I'm not sure what kind of unit test to make here and am open to suggestions.

Pre-release notes added
Lots of docstrings added

## Pre-merge checklist


ZLLentz:
I think this worked in practice. I'll clean up the documentation here with the intention of merging.

ZLLentz:
The error is:
```
E   ModuleNotFoundError: No module named 'hutch_python'
```
Since we made a pcdsutils tag with the function we stole from hutch_python, I'll adjust this PR to use the pcdsutils import instead.

pcdsdevices/areadetector/detectors.py
```diff
@@ -62,6 +64,147 @@ def get_plugin_graph_edges(self, *, use_names=True, include_cam=False):
...
+    You can decide how many images we'll take at each point by setting
+    the `num_images_per_point` attribute.
+
+    There are two modes provided: "always aquire" and "aquire at points"
+    with key differences in the behavior. You can select which one
+    to use using the ``always_acquire`` keyword argument, which
+    defaults to True.
+
+    With always_aquire=True (default):
+    - Viewers will remain updated althroughout the scan, even between
```
klauer, line 85:
```suggestion
    - Viewers will remain updated throughout the scan, even between
```

tangkong (approved):

pcdsdevices/areadetector/plugins.py
```diff
@@ -162,3 +168,42 @@ class HDF5Plugin(ophyd.plugins.HDF5Plugin, FilePlugin):
 
 class MagickPlugin(ophyd.plugins.MagickPlugin, FilePlugin):
     pass
+
+
+class HDF5FileStore(FileStoreHDF5IterativeWrite, HDF5Plugin_V31):
```
tangkong, line 173: It's good to use the most updated plugin, but do we actually make use of the signals added in the most recent version?  File writing doesn't need it, if I'm not mistaken.

klauer: Hmm, good point - we should probably use the minimum version class we can...

ZLLentz:
Final thoughts here:
- I don't want to spend time testing this again, so I'd like to keep the class structure as is and merge
- I never figured out a good way to make a unit test for this sort of thing

PR #935, opened 2022-02-01
FIX: EpicsSignalEditMD and state positioner-related fixes

klauer:
## Description

This fixes some issues related to a flurry of recent bug reports. The following changes have been made:
- The MD update signal will be more lenient for cases where specific unset values ("Invalid") persist from the TwinCAT PLC. Devices like TwinCAT state positioners will stop throwing large confusing errors related to invalid enum strings.
- The MD update signal will not send metadata updates until all composite signals are ready and have contributed their values.
- State names are no longer case-sensitive: states like "Out", "out", and "OUT" will be understood to be the same.
- HelpfulIntEnum has been temporarily vendored from pcdsutils/pcdsdaq. This will change to an import in a future release.

## Motivation and Context

- Several state devices have spammy errors when created that interfere with our ability to adjust their enum strings dynamically as viewed from Python.
- Several state devices have variations on the "OUT" string that all mean effectively the same thing, but were previously being interpreted as distinct entities. This lead to some devices being both "IN" and "OUT" because of a misunderstanding of the "Out" string as being "Not "OUT"", for example.
- [Related Jira Ticket](https://jira.slac.stanford.edu/browse/LCLSECSD-550)

## How Has This Been Tested?

The unit tests for pcdsdevices, which covers this section extensively, still pass.
Interactive testing was done to verify that affected real devices now have their behavior corrected, and that related devices have not had their behaviors broken.

Here in this PR and in release notes. Some docstrings are included where appropriate and necessary.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/state.py
+++ b/pcdsdevices/state.py
@@ -252,18 +252,26 @@ class StatePositioner(MvInterface, Device, PositionerBase):
         # Check for a malformed string digit
         if isinstance(value, str) and value.isdigit():
             value = int(value)
-        try:
-            return self.states_enum[value]
-        except KeyError:
+
+        if isinstance(value, int):
             try:
                 return self.states_enum(value)
             except ValueError:
-                err = ('{0} is not a valid state for {1}. Valid state names '
-                       'are: {2}, and their corresponding values are {3}.')
-                enum_names = [state.name for state in self.states_enum]
-                enum_values = [state.value for state in self.states_enum]
-                raise ValueError(err.format(value, self.name, enum_names,
-                                            enum_values))
+                ...
+        elif isinstance(value, self.states_enum):
+            return value
+        else:
+            for state_enum in self.states_enum:
+                if state_enum.name.lower() == value.lower():
+                    return state_enum
+
+        enum_names = [state.name for state in self.states_enum]
+        enum_values = [state.value for state in self.states_enum]
+        raise ValueError(
+            f"{value} is not a valid state for {self.name}."
+            f"Valid state names are: {enum_names}, "
+            f"and their corresponding values are {enum_values}."
+        )
 
     def _do_move(self, state):
         """
--- a/pcdsdevices/state.py
+++ b/pcdsdevices/state.py
@@ -7,7 +7,7 @@ import copy
 import functools
 import logging
 from enum import Enum
-from typing import ClassVar, Optional
+from typing import ClassVar, List, Optional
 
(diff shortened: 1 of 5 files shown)


tangkong:
So I don't forget, maybe we should solve case sensitivity issues via `_missing_` or `_missing_value_` methods, similar to how `aenum` does

https://stackoverflow.com/questions/42658609/how-to-construct-a-case-insensitive-enum

klauer:
Vendored helpfulintenum, then pushed it into pcdsutils as well - https://github.com/pcdshub/pcdsutils/pull/44

pcdsdevices/state.py
```diff
@@ -108,7 +108,13 @@ def _state_init(self):
                 self.states_enum = self._create_states_enum()
             self._state_initialized = True
 
-    def _late_state_init(self, *args, enum_strs=None, **kwargs):
+    def _late_state_init(
+        self,
+        *args,
+        connected: bool = False,
```
ZLLentz, line 114: Is this unused now?

pcdsdevices/state.py
```diff
@@ -304,7 +309,7 @@ def _create_states_enum(self):
                 for alias in aliases:
                     state_def[alias] = i
         enum_name = self.__class__.__name__ + 'States'
-        enum = Enum(enum_name, state_def, start=0)
+        enum = HelpfulIntEnum(enum_name, state_def, start=0, module=__name__)
```
ZLLentz, line 312: TIL the enum type constructor has a `module` arg

ZLLentz (approved):
Small comments, but I think this is actually 100% good if the tests still pass

pcdsdevices/state.py
klauer: Same for me - the ["Warning!"](https://docs.python.org/3.11/howto/enum.html#functional-api) brought it to my attention - unpickleable without it:
image

pcdsdevices/state.py
klauer: Yes, removing it now

pcdsdevices/state.py
ZLLentz: Huh, interesting

tangkong (approved):
I have no issues with this.  If this clears the angry state error issues then I think our original goal has been met

klauer:
I can't approve my own PR, but the release notes look good @ZLLentz :+1:

PR #942, opened 2022-02-03
WIP/ENH: proposal for alternative states customization + mr1l0_homs fix

klauer:
## Pre-merge checklist

### DIFF
--- a/pcdsdevices/mirror.py
+++ b/pcdsdevices/mirror.py
@@ -790,8 +790,12 @@ class CoatingState(Device):
 
     This makes it appear at the top of the screen in typhos.
     """
-    coating = Cpt(TwinCATMirrorStripe, ':COATING:STATE', kind='hinted',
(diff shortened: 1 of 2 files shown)


klauer:
@ZLLentz an idea here - not sure if better or worse than the status quo, but the excessive explicit subclassing was getting to me.

ZLLentz:
This is a really interesting framing- the pattern (classmethod to create a subclass, pass into a component) might be useful beyond the scope here. Explicit and compact, I like it.

klauer:
This could potentially be a generic pattern applied to the base interface; `.customized(component_name=dict(**update_args))`

I'm trying to imagine how it could be abused. Perhaps the lack of a subclass existing for inheritance purposes means that it "dies" at a single level of ugliness and can't be abused further. Not sure.

klauer:
I think this could have potential, but we already have a solution for now. Can revisit later; closing

Issue #945, opened 2022-02-07
sim.slow_motor fails in scans if starting scan point is current position


tangkong:
## Expected Behavior
`sim.slow_motor` should move if used in a scan regardless of initial position

## Current Behavior
If any of the `sim.slow_motor` are used in a scan where the starting position is their current position, the scan fails with a failed MoveStatus.   

```python
---------------------------------------------------------------------------
FailedStatus                              Traceback (most recent call last)
<ipython-input-1-25d2bc539cbf> in <module>
----> 1 RE(bp.scan([], sim.slow_motor2, 0, 1, num=10))

/cds/group/pcds/pyps/conda/py39/envs/pcds-5.1.1/lib/python3.9/site-packages/bluesky/run_engine.py in __call__(self, *args, **metadata_kw)
    880             self._task_fut.add_done_callback(set_blocking_event)
    881 
--> 882         plan_return = self._resume_task(init_func=_build_task)
    883 
    884         if self._interrupted:

/cds/group/pcds/pyps/conda/py39/envs/pcds-5.1.1/lib/python3.9/site-packages/bluesky/run_engine.py in _resume_task(self, init_func)
   1018                     if (exc is not None
   1019                             and not isinstance(exc, _RunEnginePanic)):
-> 1020                         raise exc
   1021                     # Only try to get a result if there wasn't an error,
   1022                     # (other than a cancelled error)

/cds/group/pcds/pyps/conda/py39/envs/pcds-5.1.1/lib/python3.9/site-packages/bluesky/run_engine.py in _run(self)
   1641             exit_reason = str(err)
   1642             self.log.exception("Run aborted")
-> 1643             raise err
   1644         finally:
   1645             if not exit_reason:

/cds/group/pcds/pyps/conda/py39/envs/pcds-5.1.1/lib/python3.9/site-packages/bluesky/run_engine.py in _run(self)
   1501                     else:
   1502                         try:
-> 1503                             msg = self._plan_stack[-1].send(resp)
   1504                         # We have exhausted the top generator
   1505                         except StopIteration:

/cds/group/pcds/pyps/conda/py39/envs/pcds-5.1.1/lib/python3.9/site-packages/bluesky/plans.py in scan(detectors, num, per_step, md, *args)
   1103     full_cycler = plan_patterns.inner_product(num=num, args=args)
   1104 
-> 1105     return (yield from scan_nd(detectors, full_cycler,
   1106                                per_step=per_step, md=_md))
   1107 

[...]

PR #947, opened 2022-02-08
BUG: mark success if already at goal

tangkong:
## Description
Simulated Slow motors fail to move in a scan if the first position is the current position. 
Broadly, this issue appears any time sim.slow_motor was moved to its current position with the `wait` argument enabled

## Motivation and Context
This is a functionality that should be included.  
Addresses #945 

## How Has This Been Tested?
Tested interactively, test added to `test_interface.py`

Code has been commented, related issue, this PR.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/sim.py
+++ b/pcdsdevices/sim.py
@@ -64,6 +64,12 @@ class SlowMotor(FastMotor):
 
         def update_thread(positioner, goal):
             positioner._moving = True
+            if positioner.position == goal:
+                # if already at the requested position, mark success
+                positioner._set_position(goal)  # needed for LiveTable format
+                positioner._done_moving(success=True)
+                return
+
             while positioner.position != goal and not self._stop:
                 if goal - positioner.position > 1:
                     positioner._set_position(positioner.position + 1)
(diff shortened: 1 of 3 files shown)


pcdsdevices/sim.py
```diff
@@ -64,6 +64,12 @@ def _setup_move(self, position, status):
 
         def update_thread(positioner, goal):
             positioner._moving = True
+            if positioner.position == goal:
+                # if already at the requested position, mark success
+                positioner._set_position(goal)  # needed for LiveTable format
```
tangkong, line 69: I'm not sure why this line is needed.  Without it, LiveTable fails to write new rows after the first.

klauer: Does `_set_position` do a subscription callback by chance?

tangkong:
It's possible to fix the issue by removing the `positioner.position != goal` condition from the while loop, but that scares me so I'm going to leave this as is

klauer (approved):
Looks good to me 👍

ZLLentz (approved):
Looks good to me, somehow I thought this would be much longer

PR #952, opened 2022-02-17
Add SRS DG645 support class

vespos:
## Description
Add a class to handle and scan SRS DG645 channels

## Motivation and Context
Needed for the upcoming xpp exp. General enough that it can be useful in the future again.

## How Has This Been Tested?
Tested by scanning a channel delay with the DAQ at XPP, changing reference.

Doc strings

## Pre-merge checklist


slactjohnson:
This looks useful for sure. Is there a reason not to include other configurable parameters like polarity? There are a lot of features supported by this IOC.

vespos:
That's a good point. At the moment I have only implemented the channel specific stuff. There are indeed many more parameters that can be set from the screen that I could / want to add. By polarity, do you mean the trigger source setting?

ZLLentz (commented):
Left various suggestions relating to conventions, not relating to the feature set or overall approach which looks good

pcdsdevices/delay_generator.py
```diff
@@ -0,0 +1,141 @@
...
+    Parameters
+    ----------
+    prefix: str
+        Base PV for the delay generator channel
+
+    name: str
+        Alias of the channel
+    """
+    delay = Cpt(EpicsSignal, 'DelayAO', kind='hinted')
+    delay_rbk = Cpt(EpicsSignalRO, 'DelaySI', kind='normal')
```
slactjohnson, line 50: You can also combine PVs that form a read-write pair into a single component, e.g.:
`delay = Cpt(EpicsSignal, 'DelaySI', write_pv='DelayAO', kind=normal)`

pcdsdevices/delay_generator.py
```diff
@@ -0,0 +1,141 @@
...
+
+    Parameters
+    ----------
+    prefix: str
+        Base PV for the delay generator channel
+
+    name: str
+        Alias of the channel
+    """
+    delay = Cpt(EpicsSignal, 'DelayAO', kind='hinted')
```
slactjohnson, line 49: Usually we include the colon in the component suffix to make things a little cleaner when instantiating the device, e.g. `delay = Cpt(EpicsSignal, ':DelayAO', kind='hinted')`

slactjohnson: Nevermind, I see you handle this below. Don't mind me....

pcdsdevices/delay_generator.py
```diff
@@ -0,0 +1,141 @@
...
+        return
+
+
+channel_cpts = {}
+for channel in CHANNELS[:-1]:
+    channel_cpts[f'ch{channel}'] = Cpt(
+        Dg_channel, f":{channel.lower()}", name=f"ch{channel}"
+        )
+
+Dg = type('Dg', (Delay_generator, ), channel_cpts)
```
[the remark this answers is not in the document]
klauer: There's also:
```
In [2]: ophyd.device.create_device_from_components?
Signature:
ophyd.device.create_device_from_components(
    name,
    *,
    docstring=None,
    default_read_attrs=None,
    default_configuration_attrs=None,
    base_class=<class 'ophyd.device.Device'>,
    class_kwargs=None,
    **components,
)
Docstring:
Factory function to make a Device from Components

Parameters
----------
name : str
    Class name to create
docstring : str, optional
    Docstring to attach to the class
default_read_attrs : list, optional
    Outside of Kind, control the default read_attrs list.
    Defaults to all `component_names'
default_configuration_attrs : list, optional
    Outside of Kind, control the default configuration_attrs list.
    Defaults to []
base_class : Device or sub-class, optional
    Class to inherit from, defaults to Device
**components : dict
    Keyword arguments are used to map component attribute names to
    Components.

Returns
-------
cls : Device
    Newly generated Device class
```

pcdsdevices/delay_generator.py
```diff
@@ -0,0 +1,141 @@
...
+
+    tab_component_names = True
+    tab_whitelist = ['set_reference', 'get_str']
+
+    def get(self):
+        """
+        The readback is a string formatted as"<REF> + <DELAY>".
+        Returns the <DELAY> as a float
+        """
+        return float(self.delay_rbk.get().split("+")[1])
```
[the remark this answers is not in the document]
vespos: Yes, I was really surprised by that too.
Not sure it ever is different. I'm sure what is best between using the delayAO for both the read and write, or overwrite the get method ( i wasnt too comfortable with it either)
Suggestions here?

pcdsdevices/delay_generator.py
vespos: Need to see how the string PV ends being handled when scanning this component with the DAQ etc. I would not be surprised that such a string would break something down the line here.
(This mostly a note to myself for when I can take the xpp daq)

pcdsdevices/delay_generator.py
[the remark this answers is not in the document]
ZLLentz: If it's never different then we can ignore the string PV altogether, but if the string PV is the ultimate holder of truth then any suggestion will be at least a little distasteful.

Another approach to this general problem (that I'm not sure maps well here) is to make a derived signal subclass that processes the string and exposes a float (https://github.com/bluesky/ophyd/blob/df60483867e521dcda83648756cab8fc7b80dd17/ophyd/signal.py#L521)

pcdsdevices/delay_generator.py
ZLLentz: Hmm yeah mixing two data types for the read/write will almost certainly have consequences

pcdsdevices/delay_generator.py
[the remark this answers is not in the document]
vespos: Thanks, I'll look into the derived signal.
I am also trying to see really how the DelayAO and DelaySI differ in the IOC. I need to understand a bit more how asyn works, but it in the IOC deb file, the AO and SI are defined as:

```
record(stringin,"$(DG):aDelaySI")
{
    field(PINI, "YES")
    field(SDIS, "$(DG):disableScanBI.VAL NPP NMS")
    field(DISS, "INVALID")
    field(DESC, "Chan A")
    field(DTYP, "asynOctetRead")
    field(INP,  "@asyn($(PORT),101) NONE")
}
record(ao,"$(DG):aDelayAO")
{
    field(SDIS, "$(DG):disableScanBI.VAL NPP NMS")
    field(DISS, "INVALID")
    field(DESC, "Chan A Delay")
    field(PREC, "12")
    field(EGU,  "Sec")
    field(DTYP, "asynFloat64")
    field(OUT,  "@asyn($(PORT),101) NONE")
    field(FLNK, "$(DG):aDelayFO")
}
record(fanout,"$(DG):aDelayFO")
{
    field(SDIS, "$(DG):disableScanBI.VAL NPP NMS")
    field(DISS, "INVALID")
    field(LNK1, "$(DG):statusSI")
    field(LNK2, "$(DG):aDelaySI")
}
```

To my almost inexistant asyn knowledge, the exact same command is used in both the AO and SI PVs. Does that mean we can safely forget about the SI?

pcdsdevices/delay_generator.py
```diff
@@ -0,0 +1,133 @@
...
+
+    tab_component_names = True
+    tab_whitelist = ['set', 'set_reference', 'get_str']
+
+    def get_str(self):
+        """ Returns the full <REF> + <DELAY> string. """
+        return self.delay_str.get()
+
+    def set(self, new_delay):
+        return set_and_wait(self.delay, new_delay)
```
ZLLentz, line 61: This will break `bluesky` scans of `DgChannel` devices, since `set` is no longer behaving correctly. `set` is supposed to return a status object immediately without waiting.
If you want `set` to pass through to the `delay` signal so `DgChannel` can be used as a settable/movable in a scan, something like this would work better:

```suggestion
    def set(self, new_delay, **kwargs):
        return self.delay.set(delay, **kwargs)
```
If you don't need to pass a full `DgChannel` into a `bluesky` scan you probably don't need to override the `set` method.
If you're intending to have a "wait for the value to change" interface it needs to be named something else than `set`.

vespos: Thanks for your comment. I will go with what you suggest, as this is a variable we will want to scan. I've played a bit with the set, it is almost instantaneous, so the set_and_wait functionality does not really adds anything useful here.

ZLLentz (approved):
This looks pretty clean to me, I'd be happy to merge if you think it's done

ZLLentz:
Oh, backing up a few steps, I'll check on the travis failure

pcdsdevices/delay_generator.py
```diff
@@ -0,0 +1,133 @@
...
+
+    tab_component_names = True
+    tab_whitelist = ['set', 'set_reference', 'get_str']
+
+    def get_str(self):
+        """ Returns the full <REF> + <DELAY> string. """
+        return self.delay_str.get()
+
+    def set(self, new_delay, **kwargs):
+        return self.delay.set(delay, **kwargs)
```
ZLLentz, line 61: This function has a typo, the arg is `new_delay` and the passed value is `delay`

vespos, line 61:
```suggestion
        return self.delay.set(new_delay, **kwargs)
```

vespos: That'll teach me accepting random suggestions from strangers without double checking them :-p

vespos:
Merging after approval by @ZLLentz.
Usage will tell whether we want additional features.

PR #962, opened 2022-03-07
ENH: add post_elog_status to BaseInterface

tangkong:
## Description
With newfound experience posting to the ELog, we can give every device the ability to post its status to the elog.  

The `elog` package comes with a method for registering the elog so we leverage that and use `elog` as an optional dependency.  [(elog is registered on init)](https://github.com/pcdshub/elog/blob/dbe280a04d7f9d0209087b24ad938d25bcb98e5e/elog/elog.py#L146)

## Motivation and Context
[Jira ticket](https://jira.slac.stanford.edu/browse/LCLSPC-124)

We're hoping to provide scientists an easy and accessible way to log their beamline state. 

## How Has This Been Tested?
Interactively.  Tests for this seem redundant with existing status print tests and elog's own tests.  I've chosen to not write any new tests, but this time not out of sloth.

I also accidentally posted `at2l0` status to the xcsp23820 elog, so...

Docstrings, though there's not much to describe

image

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -368,16 +368,19 @@ class BaseInterface:
         """
         Post device status to the primary elog, if possible.
         """
-        if has_elog:
-            try:
-                elog = get_primary_elog()
-                final_post = f'<pre>{self.status()}</pre>'
-                elog.post(final_post, tags=['ophyd_status'],
-                          title=f'{self.name} status report')
-            except ValueError:
-                logger.info('elog exists but has not been registered')
-        else:
+        if not has_elog:
             logger.info('No primary elog found, cannot post status.')
+            return
+
+        try:
+            elog = get_primary_elog()
+        except ValueError:
+            logger.info('elog exists but has not been registered')
+            return
+
+        final_post = f'<pre>{self.status()}</pre>'
+        elog.post(final_post, tags=['ophyd_status'],
+                  title=f'{self.name} status report')
 
 
 def get_name(obj, default):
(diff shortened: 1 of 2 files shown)


pcdsdevices/interface.py
```diff
@@ -358,6 +364,21 @@ def subdevice_filter(info):
 
         return ophydobj_info(self, subdevice_filter=subdevice_filter)
 
+    def post_elog_status(self):
+        """
+        Post device status to the primary elog, if possible.
+        """
+        if has_elog:
```
klauer, line 371: Some nitpicks:
* In general, it's better to exit early for bad conditions, at least so you can keep better indentation
```python
if not has_elog: 
    logger.info(...)
    return
```

* The `try/except ValueError` is wrapping 3 statements below; which does it apply to? 
* At least the `final_post = ...` assignment can be moved out of the try/except block, but not knowing about `get_primary_elog` or `elog.post()` I don't know where `ValueError` is expected to be raised

klauer (approved):
Style suggestion / nitpick, but functionally this looks good to me 👍

ZLLentz (approved):
Simple and clean, I like this. I can absolutely imagine someone calling this on the beamline.

ZLLentz:
Was most of the html/formatting complexity from the "post many devices" PR from the aggregation? I have to admit I didn't have enough experience to quickly dig into all the details.

tangkong:
Yep, it was mostly iterating/recursing through a namespace and organizing things into collapsible blocks.  If we just want to post a single status, we can do away with most of that.

PR #970, opened 2022-03-15
ENH: multi derived signal

### LINKED ISSUE #969: Feature: DerivedSignal with multiple sources
## Current / Expected Behavior
* `DerivedSignal` allows us to take in a single signal and perform some calculations on it, and make a new signal
* It should be possible to take in multiple input signals and, whenever one changes, perform some calculations on it and publish a new signal

## Possible Solution
It could be called `MultiDerivedSignal` or something like that.
`Cpt(MultiDerivedSignal, attrs=["attr1", "attr2", "attr3"])`

## Context
Talking with Mike E. about how to approach summarizing error codes from 19 x 3 locations (for each attenuator blade: motor, state, PLC error codes) in a single signal that can be displayed readily in typhos or used in hutch-python.

klauer:
## Description
- [x] Move reusable logic from `PVStateSignal` to `AggregateSignal`
- [x] Keep track of subscriptions and unsubscribe when the device is destroyed
- [x] Add test conditions for the above
- [x] Add MultiDerivedSignal

## Motivation and Context
* Closes #969
* Attempts to address the issue of working with many signals at once, both reading and writing

## How Has This Been Tested?
* Test suite with contrived examples and fake epics signals
* Mocked up how AT2L0 could potentially use it

AT2L0 potentially could have something like:

<details>

```python
def _check_errors(signals: SignalToValue) -> str:
    result = []
    for sig, value in signals.items():
        if value not in (0, ""):
            result.append(f"{sig.name}: {value}")
    return "\n".join(result)

error_summary = Cpt(
    MultiDerivedSignal,
    calculate=_check_errors,
    attrs=sum(
        (
            [
                f"blade_{_blade:02d}.state.error",
                f"blade_{_blade:02d}.state.error_id",
                f"blade_{_blade:02d}.state.error_message",
                f"blade_{_blade:02d}.motor.plc.err_code",
            ]
            for _blade in range(1, 19)
        ),
        [],
    ),
)

def _reset_errors(self, value: OphydDataType) -> SignalToValue:
    return {sig: 1 for sig in self.parent.reset_errors.signals}

reset_errors = Cpt(
    MultiDerivedSignal,
    calculate=lambda values: 0,
    calculate_on_put=_reset_errors,
    attrs=[f"blade_{_blade:02d}.motor.plc.cmd_err_reset" for _blade in range(1, 19)],
)
```
</details>

* Very open to input here; clunky / reasonable / etc?

* Docstrings

## Pre-merge checklist
- [x] Tested with real `EpicsSignal` instances
- [x] Tested the real example in typhos
- [x] Test real state positioners to ensure the refactored code was not broken


klauer:
This is ready for at least a conceptual agreement that this is a reasonable approach / API / etc. 
Test suite covers almost every line of the diff.

pcdsdevices/signal.py
```diff
@@ -116,30 +122,61 @@ class FakePytmcSignalRO(FakePytmcSignal, FakeEpicsSignalRO):
...
+    Can you expect the ophyd object to ever give subscription callbacks?
+
+    If not, pcdsdevices considers that object to be "meek" or quiet.  This can
+    be useful to determine whether or not you should wait on a subscription
+    callback to happen before continuing.
+    """
+    if not isinstance(obj, ophyd.ophydobj.OphydObject):
+        raise ValueError(f"{obj} is not an ophyd object")
+
+    return any(isinstance(obj, cls) for cls in fake_device_cache.values())
```
ZLLentz, line 136: I don't quite understand this implementation. An ophyd device is quiet if and only if it is fake?

pcdsdevices/signal.py
```diff
@@ -157,55 +194,222 @@ def _calc_readback(self):
...
+                return
             if value != old_value or not self._update_only_on_change:
                 self._run_subs(sub_type=self.SUB_VALUE, obj=self, value=value,
                                old_value=old_value)
 
+    @property
+    def connected(self) -> bool:
+        """Are all relevant signals connected?"""
+        # Subscriptions are used to see if the device is connected:
+        self._setup_subscriptions()
```
ZLLentz, line 303: I'm not sure if its appropriate to set up EPICS anything inside of a property due to network latency. Is this necessary/are my concerns unfounded?
Are our subscriptions set up on an as-needed basis?

pcdsdevices/signal.py
```diff
@@ -116,30 +122,61 @@ class FakePytmcSignalRO(FakePytmcSignal, FakeEpicsSignalRO):
...
+    connected: bool = False
+    #: The last value retrieved from a value callback, or a direct get request.
+    value: Optional[OphydDataType] = None
+    #: The value subscription callback ID (None if not yet subscribed)
+    value_cbid: Optional[int] = None
+    #: The meta subscription callback ID (None if not yet subscribed)
+    meta_cbid: Optional[int] = None
+
+
 class AggregateSignal(Signal):
```
ZLLentz, line 160: It seems like the edits to AggregateSignal are primarily aimed at cleaning up the rough edges of the class and adding proper support for metadata. Is my understanding correct, or did I miss/misunderstand something?

pcdsdevices/signal.py
```diff
@@ -157,55 +194,222 @@ def _calc_readback(self):
...
+                info.value_cbid = None
+
+            if info.meta_cbid is not None:
+                signal.unsubscribe(info.meta_cbid)
+                info.meta_cbid = None
+
+        return super().destroy()
+
 
 class PVStateSignal(AggregateSignal):
```
ZLLentz, line 414: I think the refactor here maintains the original behavior wrt values, I might opt to find and test a device or two though.

pcdsdevices/signal.py
```diff
@@ -266,6 +466,157 @@ def put(self, value, **kwargs):
         self.parent.move(value, **kwargs)
 
 
+class MultiDerivedSignal(AggregateSignal):
```
ZLLentz, line 469: I'm a big fan the API for this class and how you've implemented it.

pcdsdevices/tests/test_state.py
```diff
@@ -96,6 +96,7 @@ def test_pvstate_positioner_logic():
     lim_obj.states_enum['Unknown']
     with pytest.raises(KeyError):
         lim_obj.states_enum['defer']
+    lim_obj.destroy()
```
ZLLentz, line 99: What exactly is the new dangling thing to worry about here? Do we need to worry about this in the live env?

pcdsdevices/type_hints.py
```diff
@@ -0,0 +1,41 @@
+from __future__ import annotations
```
ZLLentz, line 1: big :+1: on type_hints.py

pcdsdevices/utils.py
```diff
@@ -461,6 +464,92 @@ def exclude(
...
+        The owner object, to be used for logging / Status object attribution.
+
+    timeout : float, optional
+        Per-signal timeout to configure during set.
+
+    settle_time : float, optional
+        Per-signal settle time to configure during set.
+
+    raise_on_set_failure : bool, optional
+        Raise if any of the ``set`` calls fail.
```
ZLLentz, line 495: This docstring is malformed, the parameters section has lost its header and dropped below the returns

pcdsdevices/utils.py
```diff
@@ -461,6 +464,92 @@ def exclude(
         return set(cls.__members__.values()) - cls.include(identifiers)
 
 
+def set_many(
```
ZLLentz, line 467: This is a really good convenience method

pcdsdevices/utils.py
```diff
@@ -461,6 +464,92 @@ def exclude(
...
+        st.set_finished()
+        return st
+
+    status = statuses[0]
+    for st in statuses[1:]:
+        status = ophyd.status.AndStatus(status, st)
+    return status
+
+
+def maybe_make_method(
```
ZLLentz, line 524: This one is pretty nifty too

ZLLentz (commented):
Lots of words
Mostly this is very good, I like the approach a lot
some questions/quibbles

pcdsdevices/signal.py
klauer: I'm open to suggestions on this one, but the goal is to find a way of determining whether or not a given signal (or object) will _ever_ give a value subscription without being poked at.

In the above case, I used a shortcut of "if it's a FakeEpicsSignal or a FakePytmcSignal or ..." it has no control layer underneath it so it will not be getting any value callbacks without a specific `put()` in the right spot in the test suite.

pcdsdevices/signal.py
klauer: I agree, but this was due to being painted in a corner:

* `.wait_for_connection()` on a device says "ensure that all components are connected by checking `.connected`"
* A call to `.connected` on a `MultiDerivedSignal` as-implemented returns the connectivity status based on whether or not subscriptions have been configured
* Without subscriptions configured, `.connected` will always return `False` and as such `wait_for_connection()` will raise a `TimeoutError`

Alternatively, perhaps, an entirely separate code path could be added for being used solely with `.get()` and `.put()` and never using subscriptions. I feel that this will likely make things more complicated logic-wise, but it could be doable if we determine the concerns are strong enough for it.

pcdsdevices/signal.py
klauer: That is correct. The PR started off as a light change to the class but gradually became more thorough, as these things tend to go.

pcdsdevices/signal.py
klauer: Given how we use these in nearly everything, full testing would be strongly recommended...

pcdsdevices/tests/test_state.py
klauer: Live environment this is not really a concern as we create signals and use them once.

For testing and ephemeral usage (typhos?) we absolutely should be more concerned about correctly destroying devices and signals.

pcdsdevices/utils.py
klauer: Good catch, and weird

pcdsdevices/signal.py
klauer: I took the alternative approach from my above comment in [7f6c62c](https://github.com/pcdshub/pcdsdevices/pull/970/commits/7f6c62c2c363e7fad836e47231c441372418e88d); what do you think?

klauer:
* Uploaded pcdsutils latest version to pypi to get those builds working again.
* `.get()` to fill the cache on subscription was causing `disconnected` tests to fail.  The logic of where this came in is not so clear to me just yet, but I'm not sure the cache-filling step is necessary at this point so I removed it. An alternate could be `.get() only if .connected`, but that's flimsy too.

klauer:
<details>
<summary>Updated at2l0 sample based on this</summary>

```python
    def _check_errors(signals: SignalToValue) -> str:
        result = ["Error summary:", "Test"]
        for sig, value in signals.items():
            if value not in (0, ""):
                result.append(f"{sig.name}: {value}")
        return "\n".join(result)

    error_summary = Cpt(
        MultiDerivedSignalRO,
        calculate=_check_errors,
        attrs=sum(
            (
                [
                    f"blade_{_blade:02d}.state.error",
                    f"blade_{_blade:02d}.state.error_id",
                    f"blade_{_blade:02d}.state.error_message",
                    f"blade_{_blade:02d}.motor.plc.err_code",
                ]
                for _blade in range(1, 20)
            ),
            [],
        ),
    )

    def _reset_errors(self, value: OphydDataType) -> SignalToValue:
        return {sig: 1 for sig in self.parent.reset_errors.signals}

    reset_errors = Cpt(
        MultiDerivedSignal,
        calculate=lambda values: 0,
        calculate_on_put=_reset_errors,
        attrs=[
            f"blade_{_blade:02d}.motor.plc.cmd_err_reset"
            for _blade in range(1, 20)
        ],
    )
    set_metadata(reset_errors, dict(variety='command-proc', value=1))
```
</details>

<details><summary>Typhos representation</summary>
image
</details>

pcdsdevices/signal.py
ZLLentz: It's not clear to me if any of these possible approaches is better than the others.

I guess the gap in my head that prompted this question was the understanding that "normal" epics signals start initiating the connection process in a background thread after init, whereas this multi derived signal seemed to be doing it on first request instead (because it needs to do a bit more than a normal epics signal?), making stuff like this kind of wonky.

pcdsdevices/signal.py
```diff
@@ -152,60 +175,251 @@ def _calc_readback(self):
...
+        RuntimError
+            If called after .subscribe() or used without a parent Device.
+        """
+        if self._has_subscribed:
+            raise RuntimeError(
+                "Cannot add signals to an AggregateSignal after it has been "
+                "subscribed to."
+            )
+
+        sig = self.parent
```
tangkong, line 396: (More musings, less of a suggestion/request).  

I suppose there's no case where we would want to walk further up the component chain to get information from the parent's parent?  

Say we have something with the structure:
```
TopLevelDevice
    component1
        sub1_cpt(MultiDerivedSignal)
        ...
    component2
        sub2_cpt(MultiDerivedSignal)
        ...
```

Then `sub2_cpt` wouldn't be able to reach information about `sub1_cpt`.  It could, however, reach anything within `component2`

I don't necessarily think we should make changes to support this; it might just be easier to flatten the components if we expect them to talk.

tangkong (commented):
Phew, lots of good stuff in here.  I'm not entirely through it, but I really like the functionality it provided.  Going through and properly type annotating the library is a herculean task, but I think you've set out a nice foundation for doing that.   It might even be something educational for a newer contributor to do 👀

pcdsdevices/signal.py
tangkong: I suppose you could just specify an attr as `'.parent.component1.sub1_cpt...'`, though that seems subversive

klauer: I can't yet imagine a scenario where I (personally) would choose to implement a device/component hierarchy like that. That doesn't exclude it from being a possibility, of course. If this pcdsdevices feature sees any use at all outside of just me, I'd expect the question to pop up at some point.

The resolution is - as you reasoned - use `.parent.cpt` to reach your biological great-uncle + aunt components.

pcdsdevices/signal.py
```diff
@@ -266,6 +476,169 @@ def put(self, value, **kwargs):
...
+        settle_time: Optional[float] = None
+    ) -> ophyd.status.StatusBase:
+        if self.calculate_on_put is None:
+            raise ReadOnlyError(
+                f"{self.name} is read-only. "
+                f"`calculate_on_put` function not defined for {self.name} "
+                f"({type(self).__name__})."
+            )
+        to_write = self.calculate_on_put(value) or {}
+        if not isinstance(to_write, Mapping):
```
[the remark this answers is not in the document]
klauer: I think technically the only acceptable result would be duck-typed bluesky `Movable` - https://github.com/bluesky/bluesky/blob/47e33af7f6d2c836ace4e54e2cf039a93c8b8beb/bluesky/protocols.py#L119 - but we can be a bit more restrictive here because we're working within the bounds of ophyd/pcdsdevices and not throwing other libraries into the mix.

Why not `isinstance(to_write, SignalToValue)`? Limitations of the type hinting system and being lazy about validation:
```python
In [1]: from pcdsdevices.type_hints import SignalToValue

In [2]: isinstance({}, SignalToValue)
...
TypeError: Subscripted generics cannot be used with class and instance checks
```

klauer:
I have nothing further to add to this PR based on the above feedback. It's in the hands of the reviewers to decide if any changes need to be made prior to merging.

tangkong (approved):
I don't think I have any further suggestions / thoughts w.r.t. design.  The only other thing I think might be helpful is an example in the documentation of how this might be implemented.  The `at2l0` example you provided helped immensely in reviewing the code, and I imagine it'd be helpful for people trying to use this in the future.

klauer:
Good call, @tangkong - I think I'll make a follow-up PR documenting/giving examples of the various signal types we provide, since the docs are pretty severely lacking in that regard.

ZLLentz:
I'm inclined to merge this- are we ok to go? I didn't actually have time to hunt down all the devices that previously used the PVStateSignal, which is the only reason for my apprehension

ZLLentz:
Starting some last minute testing because I don't want to stall this out
The following classes use PVStateSignal:
https://github.com/pcdshub/pcdsdevices/blob/9fb4faf0e68fd3b1696704d2e1d7b515329d004b/pcdsdevices/stopper.py#L17
https://github.com/pcdshub/pcdsdevices/blob/bec2fbf0c29d99f6090340a691922d4e34515b06/pcdsdevices/valve.py#L27
https://github.com/pcdshub/pcdsdevices/blob/4c757eb143a1fbc0e6e518d95ec4bbd1ff9b6586/pcdsdevices/movablestand.py#L10
https://github.com/pcdshub/pcdsdevices/blob/e6c0d4e3ea835605f0923c3673c84cde84c24bbf/pcdsdevices/pulsepicker.py#L19
https://github.com/pcdshub/pcdsdevices/blob/e6c0d4e3ea835605f0923c3673c84cde84c24bbf/pcdsdevices/pulsepicker.py#L170

And now I'll check up on all instances of each of these in happi... WIP

ZLLentz:
I created a short script to do the test and it has the same output before and after applying this PR.
<details>
  <summary>The Script</summary>
from happi.client import Client

try:
    from pcdsdevices.signal import MultiDerivedSignal
except ImportError:
    print('Running without multi derived signal edits')
else:
    print('Running with multi derived signal edits')

client = Client.from_config()

classes = (
    'Stopper',
    'GateValve',
    'MovableStand',
    'PulsePicker',
    'PulsePickerInOut',
)

search_results = []
for cls in classes:
    search_results.extend(client.search_regex(device_class=f'.*{cls}'))

devices = []
for result in search_results:
    devices.append(result.get())

for device in devices:
    try:
        device.wait_for_connection(1.0)
    except Exception:
        print(f'Device {device.name}, connected=False')
    print(f'Device {device.name}, connected={device.connected}')
    if device.connected:
        try:
            print(f'State {device.state.get()}')
        except Exception:
            print(f'Error reading device state')
</details>
<details>
  <summary>Output before this PR</summary>
Running without multi derived signal edits
Device hxd_dg2_stopper, connected=True
State IN
Device sh1, connected=False
Device sh1, connected=False
Device sh2, connected=True
State IN
Device sh4, connected=True
State IN
Device sh45, connected=True
State IN
Device sh5, connected=True
State IN
Device sh6, connected=True
State IN
Device st1k2_pps, connected=True
Error reading device state
Device st1l0_pps, connected=True
State IN
Device st3k4_pps, connected=True
Error reading device state
Device um6_stopper, connected=True
State OUT
Device xcs_sb1_stopper, connected=False
Device xcs_sb1_stopper, connected=False
Device cxi_dg1_vgc_01, connected=True
State IN
Device cxi_dg1_vgc_02, connected=True
State IN
Device cxi_dg2_vgc_01, connected=True
State OUT
Device cxi_dg2_vgc_02, connected=True
State OUT
Device cxi_dia_valve, connected=True
State OUT
Device cxi_dsc_vgc_01, connected=True
State IN
Device cxi_kb1_vgc_01, connected=True
State IN
Device cxi_kb1_vgc_02, connected=True
State IN
Device cxi_kb2_vgc_01, connected=True
State IN
Device cxi_sc1_vgc_01, connected=True
State IN
Device cxi_sc1_vgc_02, connected=True
State IN
Device cxi_sc3_vgc_01, connected=True
State IN
Device cxi_snd_valve, connected=True
State OUT
Device hfx_mon_valve_1, connected=True
State OUT
Device hfx_mon_valve_2, connected=True
State OUT
Device hfx_mon_valve_3, connected=True
State OUT
Device hfx_mon_valve_4, connected=True
State OUT
Device hx2_valve, connected=True
State Unknown
Device hxd_dg2_valve, connected=True
State OUT
Device hxd_mon_valve_1, connected=True
State OUT
Device hxd_mon_valve_2, connected=True
State OUT
Device mec_bxl_valve, connected=True
State OUT
Device mec_hxm_valve, connected=True
State OUT
Device mec_mxt_valve, connected=True
State OUT
Device mfx_dg1_valve_1, connected=True
State OUT
Device mfx_dg1_valve_2, connected=True
State OUT
Device mfx_dia_valve_01, connected=True
State OUT
Device mfx_dia_valve_02, connected=True
State OUT
Device mfx_dvd_valve, connected=True
State OUT
Device mfx_mxt_valve, connected=True
State OUT
Device mxt_valve_1, connected=True
State IN
Device mxt_valve_2, connected=True
State OUT
Device uvd_valve, connected=True
State Unknown
Device xcs_dg3_valve_2, connected=True
State OUT
Device xcs_lam_valve_1, connected=True
State OUT
Device xcs_lam_valve_2, connected=True
State IN
Device xcs_pbt_valve_1, connected=True
State OUT
Device xcs_pbt_valve_2, connected=True
State OUT
Device xcs_pbt_valve_3, connected=True
State OUT
Device xcs_pbt_valve_4, connected=True
State OUT
Device xcs_sb1_valve, connected=True
State OUT
Device xcs_sb2_vgc_01, connected=True
State OUT
Device xpp_sb2_valve, connected=True
State OUT
Device xrt_dvd_valve, connected=True
State OUT
Device xrt_mxt_valve, connected=True
State OUT
Device cxi_ds1_pulsepicker, connected=True
State CLOSED
Device cxi_pulsepicker, connected=True
State CLOSED
Device mec_pulsepicker, connected=True
State CLOSED
Device mfx_pulsepicker, connected=True
State CLOSED
Device xcs_pulsepicker, connected=True
State OPEN
Device xpp_pulsepicker, connected=True
State OPEN

</details>
<details>
  <summary>Output after this PR</summary>
Running with multi derived signal edits
Device hxd_dg2_stopper, connected=True
State IN
Device sh1, connected=False
Device sh1, connected=False
Device sh2, connected=True
State IN
Device sh4, connected=True
State IN
Device sh45, connected=True
State IN
Device sh5, connected=True
State IN
Device sh6, connected=True
State IN
Device st1k2_pps, connected=True
Error reading device state
Device st1l0_pps, connected=True
State IN
Device st3k4_pps, connected=True
Error reading device state
Device um6_stopper, connected=True
State OUT
Device xcs_sb1_stopper, connected=False
Device xcs_sb1_stopper, connected=False
Device cxi_dg1_vgc_01, connected=True
State IN
Device cxi_dg1_vgc_02, connected=True
State IN
Device cxi_dg2_vgc_01, connected=True
State OUT
Device cxi_dg2_vgc_02, connected=True
State OUT
Device cxi_dia_valve, connected=True
State OUT
Device cxi_dsc_vgc_01, connected=True
State IN
Device cxi_kb1_vgc_01, connected=True
State IN
Device cxi_kb1_vgc_02, connected=True
State IN
Device cxi_kb2_vgc_01, connected=True
State IN
Device cxi_sc1_vgc_01, connected=True
State IN
Device cxi_sc1_vgc_02, connected=True
State IN
Device cxi_sc3_vgc_01, connected=True
State IN
Device cxi_snd_valve, connected=True
State OUT
Device hfx_mon_valve_1, connected=True
State OUT
Device hfx_mon_valve_2, connected=True
State OUT
Device hfx_mon_valve_3, connected=True
State OUT
Device hfx_mon_valve_4, connected=True
State OUT
Device hx2_valve, connected=True
State Unknown
Device hxd_dg2_valve, connected=True
State OUT
Device hxd_mon_valve_1, connected=True
State OUT
Device hxd_mon_valve_2, connected=True
State OUT
Device mec_bxl_valve, connected=True
State OUT
Device mec_hxm_valve, connected=True
State OUT
Device mec_mxt_valve, connected=True
State OUT
Device mfx_dg1_valve_1, connected=True
State OUT
Device mfx_dg1_valve_2, connected=True
State OUT
Device mfx_dia_valve_01, connected=True
State OUT
Device mfx_dia_valve_02, connected=True
State OUT
Device mfx_dvd_valve, connected=True
State OUT
Device mfx_mxt_valve, connected=True
State OUT
Device mxt_valve_1, connected=True
State IN
Device mxt_valve_2, connected=True
State OUT
Device uvd_valve, connected=True
State Unknown
Device xcs_dg3_valve_2, connected=True
State OUT
Device xcs_lam_valve_1, connected=True
State OUT
Device xcs_lam_valve_2, connected=True
State IN
Device xcs_pbt_valve_1, connected=True
State OUT
Device xcs_pbt_valve_2, connected=True
State OUT
Device xcs_pbt_valve_3, connected=True
State OUT
Device xcs_pbt_valve_4, connected=True
State OUT
Device xcs_sb1_valve, connected=True
State OUT
Device xcs_sb2_vgc_01, connected=True
State OUT
Device xpp_sb2_valve, connected=True
State OUT
Device xrt_dvd_valve, connected=True
State OUT
Device xrt_mxt_valve, connected=True
State OUT
Device cxi_ds1_pulsepicker, connected=True
State CLOSED
Device cxi_pulsepicker, connected=True
State CLOSED
Device mec_pulsepicker, connected=True
State CLOSED
Device mfx_pulsepicker, connected=True
State CLOSED
Device xcs_pulsepicker, connected=True
State OPEN
Device xpp_pulsepicker, connected=True
State OPEN
</details>

And a quick small diff:
```
$ diff before_pvstate_edit.txt after_pvstate_edit.txt
1c1
< Running without multi derived signal edits
---
> Running with multi derived signal edits
```

ZLLentz:
Ok my little pastes seem to have butchered the line breaks but the tl;dr is I think we can merge

PR #983, opened 2022-03-30
Do not merge: patch used in xcs for hxr spec

ZLLentz:
## Description

As a record and a point for discussion, the IMS motors on the HXR spectrometer have some wonky behavior where their control limits are never updated, and therefore the python session doesn't normally know where they are correctly. Unless we override them with the explicit values from the PV fields, they often do not know the correct soft limit locations.

In this case, I opted to set `limits=False` in the setpoint signal and I opted to override the `limits` property to use the field PV values directly.

## Motivation and Context

Do other IMS motors on the beamline also have this issue? We need to check and decide if this patch should be part of the IMS class, or if it just applies to this IOC.

## How Has This Been Tested?

This device works properly in XCS

Not yet

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/spectrometer.py
+++ b/pcdsdevices/spectrometer.py
@@ -3,9 +3,10 @@ Module for the various spectrometers.
 """
 from ophyd.device import Component as Cpt
 from ophyd.device import FormattedComponent as FCpt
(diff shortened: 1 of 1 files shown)


ZLLentz:
This is superseded by #1045

PR #984, opened 2022-03-30
ENH: add scale argument to tweak function

tangkong:
## Description
Simple addition of `scale` optional keyword argument to the tweak function.  

## Motivation and Context
[Jira](https://jira.slac.stanford.edu/browse/LCLSECSD-937)

## How Has This Been Tested?
interactively with simulated motors. 

Added relevant bits to docstrings

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -844,7 +844,7 @@ class FltMvInterface(MvInterface):
         except ophyd.utils.LimitError:
             return
 
-    def tweak(self, **kwargs):
+    def tweak(self, scale=0.1):
         """
         Control this motor using the arrow keys.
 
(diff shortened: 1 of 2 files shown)


pcdsdevices/interface.py
```diff
@@ -844,16 +844,21 @@ def mv_ginput(self, timeout=None):
         except ophyd.utils.LimitError:
             return
 
-    def tweak(self):
+    def tweak(self, **kwargs):
```
ZLLentz, line 847: The only suggestion I have is to switch this to the much simpler `scale=0.1` in the signature here so that the user can more easily see the arguments when they inspect the function. I realize that this is a duplication of the default value, but I think the API clarity is more important.

ZLLentz:
This is exactly what was needed for James's request here

ZLLentz:
@tangkong can you make a quick pre-release doc for this? We should merge and include in the tag

ZLLentz (approved):

PR #964, opened 2022-03-08
FIX: call super() after setting up enum attrs

klauer:
## Description
Callbacks *can* happen prior to the rest of `EpicsSignalROEditMD.__init__` being configured, so `.connected` raises AttributeError.

Re-orders initialization of affected classes.

## Motivation and Context
Closes #963 

## How Has This Been Tested?
Test suite and reloading xcspython

## Pre-merge checklist
- [x] Code works interactively
- [x] Code contains descriptive docstrings, including context and API
- [x] New/changed functions and methods are covered in the test suite where possible
- [x] Test suite passes locally
- [x] Test suite passes on travis
- [x] Ran docs/pre-release-notes.sh and created a pre-release documentation page
- [x] Pre-release docs include context, functional descriptions, and contributors as appropriate

Issue #963, opened 2022-03-08
EpicsSignalROEditMD and others can fail at initialization


klauer:
## Current Behavior
Callbacks happen prior to the rest of `EpicsSignalROEditMD.__init__` being configured, so `.connected` raises AttributeError.

```python
ERROR    run - Exception occurred during callback <function PyepicsShimPV.get_all_metadata_callback.<locals>.get_metadata_thread at 0x7f7ea90ec5e0> (pvname='XCS:LAS:MMN:04.RBV')
Traceback (most recent call last):
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.3.0/lib/python3.9/site-packages/ophyd/_dispatch.py", line 45, in run
    callback(*args, **kwargs)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.3.0/lib/python3.9/site-packages/ophyd/_pyepics_shim.py", line 106, in get_metadata_th  read
    callback(pvname, md)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.3.0/lib/python3.9/site-packages/ophyd/signal.py", line 931, in _initial_metadata_call  back
    self._set_event_if_ready()
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.3.0/lib/python3.9/site-packages/ophyd/signal.py", line 989, in _set_event_if_ready
       self._run_metadata_callbacks()
  File "/cds/home/k/klauer/dev_path/pcdsdevices/signal.py", line 1038, in _run_metadata_callbacks
    self._metadata_override["connected"] = self.connected
  File "/cds/home/k/klauer/dev_path/pcdsdevices/signal.py", line 991, in connected
    and not len(self._pending_signals)
AttributeError: 'EpicsSignalROEditMD' object has no attribute '_pending_signals'
```

## Possible Solution
Fix ordering of `super()` call.

## Steps to Reproduce (for bugs)
Load `xcspython` and look for exceptions.

## Your Environment
pcds-5.3.0

PR #985, opened 2022-03-30
FIX: follow-up to #964; subscribe-after-init

### LINKED ISSUE #978: EpicsSignalEditMD still has initialization (?) issues (wrt `.log`)
## Behavior
`EpicsSignalEditMD` can throw these errors in typhos:

```python
  File "/cds/group/pcds/pyps/apps/dev/pythonpath/ophyd/ophydobj.py", line 462, in inner
    cb(*args, **kwargs)
  File "/cds/group/pcds/pyps/apps/dev/pythonpath/pcdsdevices/signal.py", line 979, in _enum_string_updated
    self.log.debug(
AttributeError: 'EpicsSignalEditMD' object has no attribute 'log'
```

Related to my changes in #963 assuredly

Ripple effect from the full traceback here:
<details>

```python
[2022-03-24 09:23:22,347] [ERROR   ] - Subscription value callback exception (PytmcSignalRO(read_pv='PF1K0:WFS:MMS:STATE:06:NAME_RBV', name='pf1k0_target_config_state06_state_name', parent='pf1k0_target_config_state06', value='TARGET5', timestamp=1648139001.923994, auto_monitor=False, string=True))
  File "/cds/group/pcds/pyps/apps/dev/pythonpath/ophyd/device.py", line 223, in __get__
    return instance._signals[self.attr]
KeyError: 'target'
 
During handling of the above exception, another exception occurred:
 
  File "/cds/group/pcds/pyps/apps/dev/pythonpath/ophyd/device.py", line 223, in __get__
    return instance._signals[self.attr]
KeyError: 'state'
 
During handling of the above exception, another exception occurred:
 
  File "/cds/group/pcds/pyps/apps/dev/pythonpath/ophyd/ophydobj.py", line 462, in inner
    cb(*args, **kwargs)
  File "/cds/group/pcds/pyps/apps/dev/pythonpath/pcdsdevices/signal.py", line 979, in _enum_string_updated
    self.log.debug(
AttributeError: 'EpicsSignalEditMD' object has no attribute 'log'
```
</details>

## Possible Solution
Ensure `.log` and other vital attributes are appropriately setup

Load `kfe` screens

## Context
Mike E showed these tracebacks in his terminal while debugging a typhos issue

## Your Environment
pcds-dev

klauer:
## Description
`EpicsSignalBaseEditMD` and related classes can have initialization problems as in #978 and #963.

## Motivation and Context
* Intending to close #978 
* Supersedes my failed fix in #964 
* I'm hoping to come up with a good fix the 2nd time around so a 3rd time around isn't necessary.

## How Has This Been Tested?

* Test suite
* Loaded up LFE screen as in the issue and didn't notice a traceback
* Current `lfe` still shows log-related error messages

* PR and issue text

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/963-init_ordering.rst
+++ b/docs/source/upcoming_release_notes/963-init_ordering.rst
@@ -20,7 +20,7 @@ New Devices
 Bugfixes
 --------
 - Fix race condition on initialization of new ``EpicsSignalEditMD`` and
-  ``EpicsSignalROEditMD``. (#963)
+  ``EpicsSignalROEditMD``. (#963, #978)
 
 Maintenance
 -----------
--- a/pcdsdevices/signal.py
+++ b/pcdsdevices/signal.py
@@ -1182,11 +1182,6 @@ class EpicsSignalBaseEditMD(EpicsSignalBase, SignalEditMD):
         name: Optional[str] = None,
         **kwargs
     ):
-        # Duplicate superclass handling of ``parent`` and ``name`` so we can
-        # initialize everything prior to calling super().__init__.
-        self._parent = parent
-        self._name = name or ""
-
         self._enum_attrs = list(enum_attrs or [])
         self._pending_signals = set()
         self._original_enum_strings = []
@@ -1213,13 +1208,11 @@ class EpicsSignalBaseEditMD(EpicsSignalBase, SignalEditMD):
             # The end result is:
             # -> self.metadata["enum_strs"] => self._enum_strings
             self._metadata_override["enum_strs"] = self._enum_strings
-            if self.parent is None:
+            if parent is None:
                 raise RuntimeError(
-                    "This signal {self.name!r} must be used in a "
+                    "This signal {name!r} must be used in a "
                     "Device/Component hierarchy."
                 )
-
-            self._subscribe_enum_attrs()
 
         elif enum_strs:
             # Override with strings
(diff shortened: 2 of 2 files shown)


klauer:
I think this is arguably a _better_ fix than the previous one. It would benefit from a second set of eyes on it, for sure.

tangkong:
I was poking around with this earlier and actually saw the traceback regardless of of whether or not I reverted recent changes.  I still seem to see the `.log` traceback if I load `kfe` though.  If I try `lfe` and it hangs for a while, it'll also show the traceback.  

I dunno if it's just a me-and-my-machine thing, but the timing is truly finicky

klauer:
I think the timing shouldn't be too difficult to understand when taking into account the following:
* A `.subscribe()` call can immediately result in a callback
* If a `.subscribe()` call is used in `__init__`, it must be done _after_ everything is setup
    * This includes the `super().__init__()` call, specifically for setting up `OphydObj` stuff like `.log`
    * This includes the subclass-specific implementation details (such as the enum value/metadata override stuff here)

The original fix worked for one scenario but not all. I think this could (or should) work for all if my assumptions are correct.

ZLLentz:
Gotcha, I understand at a surface-level now. If you observe this reducing the number of errors I say we go for it and merge.

In general, should we advise against creating EPICS subscriptions on init? Does this sort of thing have startup costs when e.g. the PVs are disconnected? Those are the main things I'm worried about for other devices.

klauer:
> In general, should we advise against creating EPICS subscriptions on init? Does this sort of thing have startup costs when e.g. the PVs are disconnected? Those are the main things I'm worried about for other devices.

I don't think there is a blocking effect any longer when the PV is disconnected, though that may have been the case at some point.
As to race conditions at `__init__` time, we should be extremely careful and follow the "at the true tail-end of `__init__`" thing I mentioned above

ZLLentz (approved):
Let's merge this then, if the observed failures were just test environment interference

PR #989, opened 2022-03-31
DOC: add release notes from v5.2.0

ZLLentz:
## Description

- Add release notes using the release notes script, bringing all the pre-release notes together.
- Rebuild the api docs

Requested a review from every contributor as a notification that your contribution is being tagged. If you don't have time, it's ok to not give a review- I won't wait for all of them as these can be revised later.

## Motivation and Context

I want to make a tag


klauer (approved):
Looks good to me! 🍾 

We sure made our lives easier with those pre-release scripts...

tangkong (approved):
Look all the hard work!  The mixture of line lengths in the `releases.rst` kinda tweaks me out, but it's not really important.

ZLLentz:
The line lengths in `releases.rst` doesn't matter for the sphinx docs, but it does end up looking terrible in the github releases page! I'll tweak them when I paste them over there.

Issue #988, opened 2022-03-31
Reconsider user function signature for MultiDerivedSignal


klauer:
One possibility might be:

```python
class MyClass(Device):
    def _get_state_velo(state_velo: MultiDerivedSignal, items: SignalToValue) -> float:
        """For state_velo, calculate the velocity to show."""
        return max(value for value in items.values())

    def _set_state_velo(state_velo: MultiDerivedSignal, value: float) -> SignalToValue:
        """For state_velo, distribute the puts to all fields."""
        return {sig: value for sig in state_velo.signals}

    state_velo = Cpt(
        MultiDerivedSignal,
        calculate=_get_state_velo,
        calculate_on_put=_set_state_velo,
    )
```

_Originally posted by @ZLLentz in https://github.com/pcdshub/pcdsdevices/pull/987#discussion_r839915521_

Thoughts from that PR:
1. `self` shouldn't be special, as this adds to confusion
2. Flexible approach (1 OR 2 arguments) vs strict approach (always 2 arguments)?
3. What are "best practices" with respect to defining these methods? 
    a. We all appear to not be a fan of double-underscore methods (with name mangling)
    b. `_method` is probably sufficient, and preferable compared to lambdas

Additional things to ponder:
1. Should `calculate` be `calculate_on_get` for clarity?
2. Anything else?

PR #990, opened 2022-04-01
REF/API: change MDS API based on discussions

klauer:
## Description
Per #988

* Renames `calculate` to `calculate_on_get`.
* Adjusts signatures as noted below.

### General required signature
```python
calculate_on_get(mds: MultiDerivedSignal, items: SignalToValue) -> OphydDataType
calculate_on_put(mds: MultiDerivedSignal, value: OphydDataType) -> SignalToValue
```

### Signatures when defined as a method in a `Device` class
When used in a subclass of `MultiDerivedSignal` this becomes:
```python
calculate_on_get(self: Device, mds: MultiDerivedSignal, items: SignalToValue) -> OphydDataType
calculate_on_put(self: Device, mds: MultiDerivedSignal, value: OphydDataType) -> SignalToValue
```
`self` refers to the `Device` as you'd expect here.

### Signatures when used as a subclass of `MultiDerivedSignal`
```python
calculate_on_get(self: MultiDerivedSignal, mds: MultiDerivedSignal, items: SignalToValue) -> OphydDataType
calculate_on_put(self: MultiDerivedSignal, mds: MultiDerivedSignal, value: OphydDataType) -> SignalToValue
```
`self` refers to the `MultiDerivedSignal` as you'd expect here. Note that this duplicates the `mds` parameter.

## Motivation and Context
To close https://github.com/pcdshub/pcdsdevices/issues/988

## Pre-merge checklist


ZLLentz:
I'm :+1: on this in general, I need to look closer at the code though. Friday brain isn't reviewing well.

klauer:
Bump - 

I think we should either merge this or close it before too much more time gets on.  Or we might get more code that needs updating, if we are to merge it.

Signal documentation will need updating

ZLLentz:
I feel like there's some implementation minutia that I'm just not grasping no matter how many times I read the code. I'm :+1: on the API change, I don't fully understand the signature checking details, and I think the CI fails because of the other elements in the code that use the previous `calculate` kwarg.

klauer:
> I'm 👍 on the API change,

I think this is the most important question to answer.

@tangkong or @ZryletTC, any input? See the PR description for API information

> I don't fully understand the signature checking details,

Hmm, can the code be made simpler or more easy-to-read? If we can narrow down as to what makes it difficult, I'm happy to try and clean it up...

CI/other devices/etc will be fixed up if we agree on the API

tangkong:
I'm most happy with the `calculate` --> `calculate_on_get` change, I think that's much more clear.  I don't have any major issues with the API as it stands.  

I went back and forth on whether I'd advocate for always requiring `self` on the `calculate_on_*` functions, but I suppose if we want to reuse a function across several devices, we want to avoid duplicating code in each of the devices.

So in the end: LGTM

pcdsdevices/type_hints.py
```diff
@@ -29,13 +29,19 @@ def __call__(**kwargs) -> None:
 SignalToValue = Dict[ophyd.Signal, OphydDataType]
 
 
-class MdsCalculateFunction(Protocol):
+@runtime_checkable
+class MdsOnGetFunction(Protocol):
```
ZLLentz, line 33: I guess I'm just skeptical/inexperienced: is this as magical as it looks? Define a `Protocol` subclass, slap on a decorator, and now it can be used in an `isinstance` check for an arbitrary callable? Let python take the wheel?

What exactly is the `isinstance` check looking for when we compare a function to these protocols? Call signature? Argument count? Annotations?

klauer: You were right to be skeptical. The removal of the signature check is wrong and should be covered by the test suite.

https://docs.python.org/3/library/typing.html#typing.runtime_checkable

I would either re-add the original checks or a utility function for "check signature against Protocol".

ZryletTC:
I have nothing to add 😶
The API change is good in my books.

ZLLentz:
Any interest in finalizing this prior to the tag? Or is it ok to leave it?

pcdsdevices/type_hints.py
klauer: Resolved in c1b8b97db1

docs/source/upcoming_release_notes/988-api-mds-clarify.rst
```diff
@@ -0,0 +1,42 @@
+988 api-mds-clarify
+###################
+
+API Changes
```
ZLLentz, line 4: Looks like we'll be triggering pcdsdevices v6.0.0!

klauer:
It's worth noting that I've only tested this with the test suite and not on actual devices just yet. If this PR passes the smell check, we should ensure that it's good with the attenuators/state positioners prior to bringing it to the dev environment.

ZLLentz (approved):
As-is this looks good

I'm a bit nervous that the attenuator error checking stuff probably doesn't have a test suite entry for it, so I'll at least double check that right now

ZLLentz:
the imager and attenuator calculation signals are passing the smell test at least

ZLLentz:
timing is bad for actually causing real errors for testing, unfortunately

ZLLentz:
I'd be comfortable merging this though

klauer:
OK, nothing further to add on my end. If we're in agreement, let's get this merged.

tangkong:
Good by me.

PR #801, opened 2021-03-25
Coating state

ghalym:
## Description
Implementation of Coating State mover for the MR1L0 and MR2L0 mirrors

Since this hasn't been implemented yet to all offset mirrors additional classes have been added to e used by those two mirrors.

## Motivation and Context

Implementation of Coating State mover for the MR1L0 and MR2L0 mirrors
## How Has This Been Tested?
This has been tested on my local environment

## Where Has This Been Documented?

image
image

## Pre-merge checklist
- [x] Code works interactively
- [x] Code contains descriptive docstrings, including context and API
- [x] New/changed functions and methods are covered in the test suite where possible
- [x] Test suite passes locally
- [ ] Test suite passes on travis
- [x] Ran docs/pre-release-notes.sh and created a pre-release documentation page
- [x] Pre-release docs include context, functional descriptions, and contributors as appropriate

PR #994, opened 2022-04-12
ENH: add util for rearranging the component order for typhos screens.

ZLLentz:
## Description

Functional proof of concept for a "device rearranger" for typhos screens without any modifications to typhos. This relies on some ophyd internal implementation details, but in the future it could be swapped out for some explicit typhos order-signaling feature.

Usage
```
reorder_components(
    XPIM,
    start_with=['zoom_lock', 'focus_lock'],
    end_with=['target'],
)
```
or, equivalently
```
@reorder_components(start_with=['zoom_lock', 'focus_lock'], end_with=['target'])
class XPIM(Device):
    ....
```
Result: the `zoom_lock` and `focus_lock` are at the top of the `typhos` screen, the `target` has been shipped from the top to the bottom.
image

This has been expanded to include the following helpers as well:
```
move_subdevices_to_start
sort_components_by_name
sort_components_by_kind
set_standard_ordering 
```

## Motivation and Context

#993 identified that there will be cases where we want to use inheritance but we also want to put the signals into specific orders. #801 uses a workaround to achieve the same thing, but this is more general and requires less code (and less thinking) per device.

## How Has This Been Tested?

Unit tests added + interactive tests

Pre-release notes and docstrings

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/utils.py
+++ b/pcdsdevices/utils.py
@@ -13,7 +13,7 @@ import time
 from collections.abc import Iterable
 from functools import reduce
 from types import MethodType
-from typing import Callable, Dict, Iterator, Optional, Union
+from typing import Callable, Dict, Iterator, List, Optional, Union
 
 import ophyd
 import pint
@@ -687,3 +687,31 @@ def post_ophyds_to_elog(objs, allow_child=False, hutch_elog=None):
 
     hutch_elog.post(final_post, tags=['ophyd_status'],
                     title='ophyd status report')
+
+
+def reorder_components(
+    cls: type,
+    start_with: Optional[List[str]] = None,
+    end_with: Optional[List[str]] = None,
+) -> None:
+    """
+    Rearrange the components in cls for typhos displays.
+
+    Internally, this works by switching around the keys in the _sig_attrs
+    OrderedDict.
+
+    Parameters
+    ----------
+    cls : type
+        The Device subclass that we'd like to rearrange the order of.
+    start_with : list of str, optional
+        The component names to bring to the top of the screen.
+    end_with : list of str, optional
+        The component names to bring to the bottom of the screen.
+    """
+    start_with = start_with or []
+    end_with = end_with or []
+    for cpt_name in reversed(start_with):
+        cls._sig_attrs.move_to_end(cpt_name, last=False)
+    for cpt_name in end_with:
+        cls._sig_attrs.move_to_end(cpt_name, last=True)
--- a/pcdsdevices/utils.py
+++ b/pcdsdevices/utils.py
@@ -754,7 +754,7 @@ def _normalize_reorder_list(
                 output.append(reverse_map[obj])
             except KeyError as exc:
                 raise ValueError(
-                    f'Recieved component {obj}, which is not from the device '
+                    f'Received component {obj}, which is not from the device '
                     f'class {cls}. We have components with the following '
                     f'names: {", ".join(cls._sig_attrs)}'
                 ) from exc
@@ -762,7 +762,7 @@ def _normalize_reorder_list(
             output.append(obj)
         else:
             raise TypeError(
-                f'Recieved object {obj}, which is not a str or Component.'
+                f'Received object {obj}, which is not a str or Component.'
             )
     return output
 
@@ -884,7 +884,6 @@ def sort_components_by_kind(cls: type[Device]) -> type[Device]:
     omitted = []
     for name, cpt in cls._sig_attrs.items():
         if check_kind_flag(cpt.kind, Kind.hinted):
-            print(cpt.kind)
             hinted.append(name)
         elif check_kind_flag(cpt.kind, Kind.normal):
             normal.append(name)
--- a/pcdsdevices/utils.py
+++ b/pcdsdevices/utils.py
@@ -720,15 +720,13 @@ def reorder_components(
         When used as a decorator with the reverse argument, this will
         return a function as required by the decorator interface.
     """
-    # Must happen outside of inner scope
-    start_with = _normalize_reorder_list(cls, start_with)
-    end_with = _normalize_reorder_list(cls, end_with)
-
     # Special decorator handling
     def inner(cls: type[Device]) -> type[Device]:
-        for cpt_name in reversed(start_with):
+        start_norm = _normalize_reorder_list(cls, start_with)
+        end_norm = _normalize_reorder_list(cls, end_with)
+        for cpt_name in reversed(start_norm):
             cls._sig_attrs.move_to_end(cpt_name, last=False)
-        for cpt_name in end_with:
+        for cpt_name in end_norm:
             cls._sig_attrs.move_to_end(cpt_name, last=True)
         return cls
 
--- a/pcdsdevices/utils.py
+++ b/pcdsdevices/utils.py
@@ -693,7 +693,7 @@ def post_ophyds_to_elog(objs, allow_child=False, hutch_elog=None):
 
 
 def reorder_components(
-    cls: type,
+    cls: type[Device],
     start_with: Optional[List[Union[str, Cpt]]] = None,
     end_with: Optional[List[Union[str, Cpt]]] = None,
 ) -> None:
@@ -705,7 +705,7 @@ def reorder_components(
 
     Parameters
     ----------
-    cls : type
+    cls : Device subclass
         The Device subclass that we'd like to rearrange the order of.
     start_with : list of str, optional
         The component names to bring to the top of the screen.
@@ -721,7 +721,7 @@ def reorder_components(
 
 
 def _normalize_reorder_list(
-    cls: type,
+    cls: type[Device],
     cpts_or_names: Optional[List[Union[str, Cpt]]],
 ) -> List[str]:
     """
@@ -739,7 +739,10 @@ def _normalize_reorder_list(
     return output
 
 
-def move_subdevices_to_start(cls: type, subdevice_cls: type = Device):
+def move_subdevices_to_start(
+    cls: type[Device],
+    subdevice_cls: type[Device] = Device,
+):
     """
     Arrange the component order of a device class to put subdevices first.
 
@@ -750,7 +753,7 @@ def move_subdevices_to_start(cls: type, subdevice_cls: type = Device):
 
     Parameters
     ----------
-    cls : type
+    cls : Device subclass
         The Device subclass that we'd like to rearrange the order of.
     subdevice_cls: type, optional
         A specific class type to move to the front. If omitted, all device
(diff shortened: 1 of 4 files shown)


tangkong (commented):
Looks useful, particularly for screens that end up being serial in how they display components.  

In the short-term I'm all for this as-is.

Longer term I wonder if there's a way to bake this information into the happi such that devices will always be built with the right component order.  Is there a hook that would let us run this function with certain components at the top/bottom?

nrwslac:
I see how we can access these `_sig_attrs_` after instantiating a class in a hutch-python session. Where do we access these attributes before it hits typhos in a call like `typhos <device>`

klauer (commented):
Seems a sensible way around the issue!

I tried to think of alternatives that could be done in the class body or with a wrapper or perhaps a subclass init hook, but in the end I think your method is probably cleanest and most explicit.

I do like the happi idea @tangkong mentions above, but the granularity may be slightly wrong. We probably want per-class and not per-instance, but maybe not?

Couple additional random thoughts:
* Could we allow component instances (e.g., `XPIM.cpt` instead of `"cpt"`) in addition to strings for static checking?
* We could maybe also have a `move_subdevices_to_start`, `sort_components_by_name`, `sort_components_by_kind` helpers that do some of the work for us without being so explicit.

ZLLentz:
> I see how we can access these `_sig_attrs_` after instantiating a class in a hutch-python session. Where do we access these attributes before it hits typhos in a call like `typhos <device>`

@nrwslac the idea is that you can do this to the class object right after defining it, it's defined prior to making an object

ZLLentz:
> Longer term I wonder if there's a way to bake this information into the happi such that devices will always be built with the right component order. Is there a hook that would let us run this function with certain components at the top/bottom? I thought about an optional init kwarg but I don't know if that level of integration is necessary

I wonder what the line is between wanting to do per-device customization via init args and just making a custom screen for it. There could certainly be room for this sort of per-instance adjustment.

ZLLentz:
> Could we allow component instances (e.g., XPIM.cpt instead of "cpt") in addition to strings for static checking?

Do component instances have a way of of checking what name they were bound to the object as? I guess I can iterate through the class's `__dict__` and figure it out. It is certainly much better to do this via component instances for purposes of static checking.

> We could maybe also have a move_subdevices_to_start, sort_components_by_name, sort_components_by_kind helpers that do some of the work for us without being so explicit.

I like this a lot. These are definitely possible.

ZLLentz:
Ok I'm going to implement the suggestions now that I have the all-clear not to completely change course
@nrwslac if you want to use this early, you can copy the current version into your PR into your PR and we'll clean it up later, or you can edit `_sig_attrs` yourself. We can also merge your PR without this and add this reordering in later.

ZLLentz:
Added a few things and made entries in the test suite for this

tangkong (approved):
Looks great to me.  I can see these functions finding a good amount of use.

pcdsdevices/utils.py
```diff
@@ -687,3 +690,151 @@ def post_ophyds_to_elog(objs, allow_child=False, hutch_elog=None):
 
     hutch_elog.post(final_post, tags=['ophyd_status'],
                     title='ophyd status report')
+
+
+def reorder_components(
+    cls: type,
```
klauer, line 696: Maybe
```suggestion
    cls: type[Device],
```

pcdsdevices/utils.py
```diff
@@ -687,3 +690,151 @@ def post_ophyds_to_elog(objs, allow_child=False, hutch_elog=None):
...
+            omitted.append(name)
+        else:
+            raise TypeError(f'Invalid component kind {cpt.kind}')
+    reorder_components(cls, end_with=hinted)
+    reorder_components(cls, end_with=normal)
+    reorder_components(cls, end_with=config)
+    reorder_components(cls, end_with=omitted)
+
+
+def set_standard_ordering(cls: type):
```
klauer, line 822: `set_standard_ordering`, `sort_components_by_kind`, and others that don't take arguments outside of `cls` could easily be converted into class decorators. They would, of course, work as tested here as well.

May or may not be worth it, but it is an easy change.

pcdsdevices/utils.py
```diff
@@ -687,3 +690,151 @@ def post_ophyds_to_elog(objs, allow_child=False, hutch_elog=None):
...
+    """
+    Simplify the user's variable arguments for the component reordering.
+    """
+    if cpts_or_names is None:
+        return []
+    reverse_map = {cpt: name for name, cpt in cls._sig_attrs.items()}
+    output = []
+    for obj in cpts_or_names:
+        if isinstance(obj, Cpt):
+            output.append(reverse_map[obj])
```
klauer, line 736: Maybe: `except KeyError`, and raise an explanatory "you provided a component not in the same class, what are you doing?"

pcdsdevices/utils.py
```diff
@@ -687,3 +690,151 @@ def post_ophyds_to_elog(objs, allow_child=False, hutch_elog=None):
...
+    ----------
+    cls : type
+        The Device subclass that we'd like to rearrange the order of.
+    """
+    hinted = []
+    normal = []
+    config = []
+    omitted = []
+    for name, cpt in cls._sig_attrs.items():
+        if cpt.kind is Kind.hinted:
```
klauer, line 806: `ophyd.Kind` is a flag, so it could be several of these (even if it's not typically the case).
That is, it could be `hinted | config` (implied `normal` due to how the bitmask is set up) which would be entirely valid.

pcdsdevices/utils.py
ZLLentz: I realized I was leaving off the `Device` info but wasn't sure the right way to hint it. I'll do this in all the places.

pcdsdevices/utils.py
ZLLentz: For this to work, all I have to do is make them return `cls`, right?

Can't we do this with all of them? I could have sworn there was a way to do something like:
```
@reorder_components(starts_with=['motor'])
class MyClass(Device):
    ...
```

pcdsdevices/utils.py
ZLLentz: Good call

pcdsdevices/utils.py
klauer: Yes, you can do that as well, just the decorator body gets a bit clunkier. No idea about appropriate annotations, but a common way I've seen/used looks like:

```python
def wrapper(cls=None, **kwargs):
    def wrap_the_class(cls):
        ...  # access kwargs here in the closure
        return cls
    
    if cls is not None:
        # wrapper(cls, starts_with=["motor"])
        return wrap_the_class(cls)
    # @wrapper(**kwargs) decorator on class cls
    return wrap_the_class
```

pcdsdevices/utils.py
ZLLentz: This is a holdover from my willful ignoring of the bitflag behavior, which I've never found a use for.

Ok, so the bitflag choices are the following (and sums thereof):
```
    omitted = 0b000
    normal = 0b001
    config = 0b010
    hinted = 0b101
```

What would the full ordering be then? How do I classify a signal that is `hinted & omitted`, or something like `normal & config`? Do I simply take the highest order component, e.g. all signals with `hinted` on them are the first group? Do I have to sort within the `hinted` group too, like is `hinted/normal` before `hinted/config`? Is `normal/config` before or behind `normal`?

I didn't have any good answers to this while implementing the class. If the numbers here were in order we could simply sort on the kind int number as-is, but they are not (numerically, hinted > config > normal > omitted)

I'm tempted to just group all "hinted" signals without regard for in-hinted ordering, then all "normal" signals (with no hint flag), then all "config" signals (with no hinted or normal flags), and then the omitted signals at the end.

klauer: > I'm tempted to just group all "hinted" signals without regard for in-hinted ordering, then all "normal" signals (with no hint flag), then all "config" signals (with no hinted or normal flags), and then the omitted signals at the end.

I think that's the appropriate path forward here - I similarly have no good answer to your questions above. 
The change would then just be to the effect of replacing `kind is Kind.hinted` with `(kind & kind.Hinted) == kind.Hinted`

ZLLentz: Some additional nuance, a lot of the combinations alluded to above don't actually make sense because the `hinted` flag includes the `normal` flag already. If we limit ourselves to `|` operations, the only possible values are:
```
0b101 == hinted
0b001 == normal
0b010 == config
0b000 == omitted
0b111 == hinted | config
0b011 == normal | config
```
The remaining two bit combinations are:
```
0b110 == ~ normal
0b100 == hinted & ~ normal
```
So this does mean that it's feasible to do a full top to bottom ordering, there are only 6 options. I think I probably still won't though.

ZLLentz:
I think I'm happy with all the updates here- followed a lot of Ken's suggestions. I double-checked that the mr5kr_kbo still looked good after switching it to using the decorator utils. There has been a rebase.

pcdsdevices/utils.py
```diff
@@ -687,3 +690,243 @@ def post_ophyds_to_elog(objs, allow_child=False, hutch_elog=None):
...
+        return []
+    reverse_map = {cpt: name for name, cpt in cls._sig_attrs.items()}
+    output = []
+    for obj in cpts_or_names:
+        if isinstance(obj, Cpt):
+            try:
+                output.append(reverse_map[obj])
+            except KeyError as exc:
+                raise ValueError(
+                    f'Recieved component {obj}, which is not from the device '
```
klauer, line 757:
```suggestion
                    f'Received component {obj}, which is not from the device '
```

pcdsdevices/utils.py
```diff
@@ -687,3 +690,243 @@ def post_ophyds_to_elog(objs, allow_child=False, hutch_elog=None):
...
+                    f'class {cls}. We have components with the following '
+                    f'names: {", ".join(cls._sig_attrs)}'
+                ) from exc
+        elif isinstance(obj, str):
+            output.append(obj)
+        else:
+            raise TypeError(
+                f'Recieved object {obj}, which is not a str or Component.'
```
klauer, line 765:
```suggestion
                f'Received object {obj}, which is not a str or Component.'
```

pcdsdevices/utils.py
```diff
@@ -687,3 +690,243 @@ def post_ophyds_to_elog(objs, allow_child=False, hutch_elog=None):
...
+        The same class from the input, mutated. This is returned so that
+        sort_components_by_kind can be used as a class decorator.
+    """
+    hinted = []
+    normal = []
+    config = []
+    omitted = []
+    for name, cpt in cls._sig_attrs.items():
+        if check_kind_flag(cpt.kind, Kind.hinted):
+            print(cpt.kind)
```
klauer, line 887:
```suggestion
```

klauer (approved):
Just a couple minor stray things. Looks great - and ready for prime time - to me.

pcdsdevices/utils.py
ZLLentz: oops 1

pcdsdevices/utils.py
ZLLentz: oops 2

pcdsdevices/utils.py
ZLLentz: oops 3

PR #996, opened 2022-04-14
Added error checking and clearing functionality to AT2L0

mkestra:
## Description

4 changes:
1. The AttenuatorCalculator_AT2L0_overview.ui gui was changed to have an extra bit array that displays whether each blade has an error. It was further changed to include a "reset errors" button to clear all applicable errors and reset the error bit array.
2. Attenuator.py was changed to add 3 Multiderived signals, one to read the errors and create a bit mask for the aforementioned bitarray, one to to summarize the errors verbally, and one to reset the errors. Command line interface was also added to read errors and reset errors.
3. signals.py, fixed err signal to prevent readbacks from interfering with error reset.  
4. Epics_motor.py, fixed err signal to prevent readbacks from interfering with error reset.
## Motivation and Context

A quality of life improvement for the users of AT2L0. 
https://jira.slac.stanford.edu/browse/LCLSECSD-336?filter=-1
## How Has This Been Tested?

False errors were placed at the PLC level of varying types and they were shown to display in the bit array, appear in the verbal error printout, and reset when the "reset errors" button was pressed.

comments in the code

Errors

## Pre-merge checklist
- [X] Test suite passes on travis


pcdsdevices/attenuator.py
```diff
@@ -1071,6 +1072,114 @@ class AT2L0(FltMvInterface, PVPositionerPC, LightpathInOutMixin):
     # Summary for lightpath view
     num_in = Cpt(InternalSignal, kind='hinted')
     num_out = Cpt(InternalSignal, kind='hinted')
+    
+    """check for errors and return a string indicating any errors verbally"""
```
klauer, line 1076: To make these useful as docstrings, move in the quoted section to the body of your function:

```python
def func():
    """Docstring with information about func."""
```

pcdsdevices/attenuator.py
```diff
@@ -1071,6 +1072,114 @@ class AT2L0(FltMvInterface, PVPositionerPC, LightpathInOutMixin):
...
+    def _check_errors(signals: SignalToValue) ->  str:
+        errors = []
+        #sort out .motor from .motor.plc signals
+        for sig, value in signals.items():
+            if ("motor" in sig.name) ^ ("plc" in sig.name):
```
klauer, line 1081: Been a while since I've seen an xor operator `^` in Python!

klauer, line 1081: Been a while since I've seen an xor operator `^` in Python!

pcdsdevices/attenuator.py
```diff
@@ -1071,6 +1072,114 @@ class AT2L0(FltMvInterface, PVPositionerPC, LightpathInOutMixin):
...
+                value=sig.metadata.get('severity').value
```
klauer, line 1082: The formatting indicates you didn't get install pre-commit. I commented in your other PR but will mention it here in case you read it first: https://confluence.slac.stanford.edu/display/PCDS/How+to+add+a+device+to+pcdsdevices#Howtoaddadevicetopcdsdevices-Pre-commit

klauer, line 1082: The formatting indicates you didn't install pre-commit. I commented in your other PR but will mention it here in case you read it first: https://confluence.slac.stanford.edu/display/PCDS/How+to+add+a+device+to+pcdsdevices#Howtoaddadevicetopcdsdevices-Pre-commit

Also, something like this might be preferable:

```suggestion
                value = int(sig.metadata["severity"])
```

For future reference, since I know you're new to Python - these `IntEnum` values can behave like an integer here. We can compare it directly:

```
>>> at2l0.blade_01.state.state.metadata["severity"] in (0, "")
True
```
The source for these classes is [here](https://github.com/bluesky/ophyd/blob/7909d08e28a671eea4f3499568977ebf8c1d4cc3/ophyd/utils/epics_pvs.py#L28-L57), by the way.

klauer, line 1082: The formatting indicates you didn't get install pre-commit. I commented in your other PR but will mention it here in case you read it first: https://confluence.slac.stanford.edu/display/PCDS/How+to+add+a+device+to+pcdsdevices#Howtoaddadevicetopcdsdevices-Pre-commit

Also, this might be preferable:

```suggestion
                value = int(sig.metadata["severity"])
```

For future reference, too, `IntEnum` behaves like an integer here. We can compare it directly:

```
>>> at2l0.blade_01.state.state.metadata["severity"] in (0, "")
True
```

pcdsdevices/attenuator.py
```diff
@@ -1071,6 +1072,114 @@ class AT2L0(FltMvInterface, PVPositionerPC, LightpathInOutMixin):
...
+        if errors:
+            return "\n".join(["Error summary:"] + errors)
+
+        return "No errors"
+
+    """summarize the errors occurring at any time on any blade via a string"""
+    error_summary = Cpt(
+        MultiDerivedSignalRO,
+        calculate=_check_errors,
+        attrs=sum(
```
klauer, line 1096: A couple things here:
1. I think you make this attribute list twice. You might consider defining it once and reusing it.
2. Alternatively, despite me being the originator of this ugly chunk of code in the other PR, let me suggest a more readable way of generating this list - a [generator](https://wiki.python.org/moin/Generators):

```python
def _get_blade_error_attrs():
    """Get the blade attribute names used for checking errors."""
    for index in range(1,20)
        yield f"blade_{index:02d}.state.error"
        yield f"blade_{index:02d}.state.error_id"
        yield f"blade_{index:02d}.state.error_message"
        yield f"blade_{index:02d}.motor.plc.err_code"
        yield f"blade_{index:02d}.motor.user_readback"
```

Then this component definition becomes

```python
    error_summary = Cpt(
        MultiDerivedSignalRO,
        calculate=_check_errors,
        attrs=list(_get_blade_error_attrs()),
    )
```

I think we could probably agree the above is a bit more readable...

klauer, line 1096: A couple things here:
1. I think you make this attribute list twice. You might consider defining it once and reusing it.
2. Alternatively, despite me being the originator of this ugly chunk of code in the other PR, let me suggest a more readable way of generating this list - a [generator](https://wiki.python.org/moin/Generators):

```python
def _get_blade_error_attrs():
    """Get the blade attribute names used for checking errors."""
    for index in range(1,20)
        yield f"blade_{index:02d}.state.error"
        yield f"blade_{index:02d}.state.error_id"
        yield f"blade_{index:02d}.state.error_message"
        yield f"blade_{index:02d}.motor.plc.err_code"
        yield f"blade_{index:02d}.motor.user_readback"
```

Then this component definition becomes

```python
    error_summary = Cpt(
        MultiDerivedSignalRO,
        calculate=_check_errors,
        attrs=list(_get_blade_error_attrs()),
    )
```

I think we could probably agree the above is a bit more readable...

pcdsdevices/attenuator.py
```diff
@@ -1071,6 +1072,114 @@ class AT2L0(FltMvInterface, PVPositionerPC, LightpathInOutMixin):
...
+                errors.append(0)
+
+            del blade_errors[0:5]
+
+        decimal_value = 0
+        for next_bit in errors:
+            decimal_value = decimal_value * 2 + next_bit
+
+        return decimal_value
+    """summarize errors occurring at any time on any blade via a bitmask"""
```
klauer, line 1140: This information goes below in `error_summary_bitmask` - specifically, `Cpt()` takes a keyword argument `doc=`.

klauer, line 1140: This information goes below in `error_summary_bitmask` - specifically, `Cpt()` takes a keyword argument `doc=`.

pcdsdevices/attenuator.py
```diff
@@ -1071,6 +1072,114 @@ class AT2L0(FltMvInterface, PVPositionerPC, LightpathInOutMixin):
...
+            if ("motor" in sig.name) ^ ("plc" in sig.name):
+                value=sig.metadata.get('severity').value
+            
+            if value not in (0,""):
+                blade_errors.append(1)
+            else:
+                 blade_errors.append(0)
+
+        #remove error messages for blade one
+        del blade_errors[0:5]
```
klauer, line 1125: It's a bit weird (but not incorrect by any stretch!) to delete items from a list like this instead of just iterating over certain indices like:

```python
step = 5
start_index = 0
end_index = step

for _ in range(1, 19):
    error_count = sum(blade_errors[start_index:end_index])
    errors.append(1 if error_count >= 1 else 0)   # or bool(error_count) since False is 0 and True is 1
    start_index += step
    end_index += step
```

klauer, line 1125: It's a bit weird (but not incorrect by any stretch!) to delete items from a list like this instead of just iterating over certain indices like:

```python
step = 5
start_index = 0
end_index = step

for _ in range(1, 19):
    error_count = sum(blade_errors[start_index:end_index])
    errors.append(1 if error_count >= 1 else 0)   # or bool(error_count) since False is 0 and True is 1
    start_index += step
    end_index += step
```

klauer (commented):
This is shaping up and looking much better! I think you're making a lot of progress here.

I added a bunch of suggestions and comments, but please do take them as constructive criticism.

pcdsdevices/attenuator.py
```diff
@@ -1071,6 +1072,114 @@ class AT2L0(FltMvInterface, PVPositionerPC, LightpathInOutMixin):
...
+                    f"blade_{_blade:02d}.motor.plc.err_code",
+                    f"blade_{_blade:02d}.motor.user_readback",
+                ]
+                for _blade in range(1,20)
+            ),
+            [],
+        ),
+    )
+
+    def reset_err(self):
```
ZLLentz, line 1159: Consider maybe a less similar to name to the signal so that the user won't get confused as easily, for example
```suggestion
    def clear_error(self):
```

pcdsdevices/attenuator.py
```diff
@@ -1093,6 +1202,10 @@ class AT2L0(FltMvInterface, PVPositionerPC, LightpathInOutMixin):
     blade_18 = Cpt(FEESolidAttenuatorBlade, ':MMS:18')
     blade_19 = Cpt(FEESolidAttenuatorBlade, ':MMS:19')
 
+    def summarize_err(self):
+        """returns the error summary """
```
ZLLentz, line 1206: This docstring is no longer true, the error summary is printed rather than returned.

pcdsdevices/attenuator.py
```diff
@@ -1155,14 +1268,18 @@ def format_status_info(self, status_info):
...
         ]
 
-        table = '\n'.join(render_ascii_att(cpt_states, start_index=1))
+        table = '\n'.join(render_ascii_att(cpt_states,  start_index=1))
+
         return f"""
 {table}
 Transmission (E={energy} keV): {transmission}
 Transmission for 3rd harmonic (E={energy_3} keV): {transmission_3}
+{self.error_summary.get()}
```
ZLLentz, line 1277: I think `.get()` here shouldn't be used inside `format_status_info`, rather you should be getting the value out of the `status_info` dict like all the other elements here. I think adjusting this will likely fix the unit tests.

The formatting of the info and the getting of the info are separated in these status info formatters so that they can be tested independently and so that a failure to get the data can't possibly completely break the status printout.

pcdsdevices/attenuator.py
```diff
@@ -1155,14 +1268,18 @@ def format_status_info(self, status_info):
...
 """
 
 
+
+
```
ZLLentz, line 1282: Accidental newlines? (1)

pcdsdevices/attenuator.py
```diff
@@ -1256,6 +1373,7 @@ def render_ascii_att(blade_states, *, start_index=0):
     ascii_lines: list of str
         The lines that should be printed to the screen.
     """
+
```
ZLLentz, line 1376: Accidental newlines? (2)

pcdsdevices/attenuator.py
```diff
@@ -1266,8 +1384,9 @@ def render_ascii_att(blade_states, *, start_index=0):
         state_enum = get_blade_enum(state)
         out_line.append(state_enum.as_out_row.center(len(index_str)))
         in_line.append(state_enum.as_in_row.center(len(index_str)))
-
+
```
ZLLentz, line 1387: Accidental newlines? (3)

pcdsdevices/attenuator.py
```diff
@@ -1266,8 +1384,9 @@ def render_ascii_att(blade_states, *, start_index=0):
         state_enum = get_blade_enum(state)
         out_line.append(state_enum.as_out_row.center(len(index_str)))
         in_line.append(state_enum.as_in_row.center(len(index_str)))
-
+    
     separator = '|'
     return [separator.join(filter_line + ['']),
             separator.join(out_line + ['']),
             separator.join(in_line + [''])]
+
```
ZLLentz, line 1392: Accidental newlines? (4)

ZLLentz (commented):
I left a few suggestions of things to tweak, I think the function name is a matter of taste but the rest are important to address.

pcdsdevices/attenuator.py
```diff
@@ -1071,6 +1072,114 @@ class AT2L0(FltMvInterface, PVPositionerPC, LightpathInOutMixin):
...
+            if sum(blade_errors[0:5]) >= 1:
+                errors.append(1)
+            else:
+                errors.append(0)
+
+            del blade_errors[0:5]
+
+        decimal_value = 0
+        for next_bit in errors:
+            decimal_value = decimal_value * 2 + next_bit
```
tangkong, line 1137: An alternate way to do this using python's built-in `int` to interpret a binary string (eg. '1001' into 9).  not really better, I just fell down a small hole on binary representations

```python
decimal_value = int(''.join([str(x) for x in errors]), 2)
```

tangkong:
I don't have much to add on top of Ken and Zach's reviews.  Mostly formatting complaints that should be cleared after running the pre-commit hooks.  Cool to see the MultiDerivedSignal stuff in action!

pcdsdevices/attenuator.py
mkestra: oh interesting, is this generator pattern pythons take on recursion?

mkestra: looking a little closer, they aren't the same, but seemingly similar concepts of repeatedly calling a function to enact iteration.

tangkong: It's certainly a python-ism.  It's not really recursion, but it's a way of dealing with iteration efficiently.   In short, a generator gives you an Iterable object that "yields" its items one at a time.   (rather than having to gather them all at once like a list would).

It's also something that shows up in Bluesky quite often, so I'm sure you'll be come more familiar with it soon. 😄

pcdsdevices/attenuator.py
mkestra: right, I was torn on how to do that, it felt kind of wonky to be deleting from the front and letting the list re-index all the elements. I guess my reasoning at the time is that it would be slightly easier to check if the list was empty by the end to verify that the number of signals in the list returned zero when a mod 5 operator was applied. I also just wanted to go from blade 1 -> blade 2 -> etc. while compiling the error list just to make the logic a little easier in my head. With all that said, you're proposed solution makes sense and less wonky, I'll put it in!

pcdsdevices/attenuator.py
mkestra: I'll add as a string to doc= argument, are these worth keeping in the doc strings as well so someone reviewing the source code only can see it or does that end up making it too cluttered?

pcdsdevices/attenuator.py
mkestra: this change is probably for the best, it's also annoying switching between "err" and "error" just to avoid the signal going crazy.

pcdsdevices/attenuator.py
mkestra: yes, I just went through and loaded the error summary using status_info and the tests don't fail anymore. still need to toy with the formatting when we have multiple lines of error text.

mkestra:
General note, just pushed a new update incorporating the comments here-in. Code runs interactively without error, Planning to test the new error attrb creation logic and a few other things tomorrow. It may be best to save any further code review/comments until after any updates that come from that.

docs/source/upcoming_release_notes/996-ADD_AT2L0_err_support.rst
```diff
@@ -0,0 +1,33 @@
+996 ADD AT2L0 err support
+#################
+
+API Changes
+-----------
+- clear_errors() method to clear errors; e.g. at2l0.clear_errors()
+- print_errors() to print error summary; e.g. at2l0.print_errors()
```
ZLLentz, line 7: Quick clarification: the API change category is for API breaks, e.g. if you removed a method or changed the signature in a backwards-incompatible way. These belong in Device Updates.

Also, the way you've written this here implies that many devices have gotten these methods added, since at2l0 is presented as an example rather than as the only device that has been updated.

pcdsdevices/attenuator.py
ZLLentz: This is redundant in the source code if it's in the doc= argument. The person reading the code will see the doc= arg too.

pcdsdevices/attenuator.py
```diff
@@ -1101,6 +1102,101 @@ class AT2L0(FltMvInterface, PVPositionerPC, LightpathInOutMixin):
...
+    error_summary_bitmask = Cpt(
+        MultiDerivedSignalRO,
+        calculate=_check_errors_bitmask,
+        attrs=list(_get_blade_error_attrs()),
+        doc='summarize errors at any time on any blade via a bitmask',
+    )
+    set_metadata(error_summary_bitmask, dict(variety='bitmask', bits=18))
+
+    def clear_errors(self):
+        """wrap reset errors a function with no arguments """
```
ZLLentz, line 1177: If this is a user-facing function, it should have a clear docstring for the user, who will be able to see it in the hutch python session. I'll make a suggestion:
```suggestion
        """Reset all attenuator errors, making the device ready to move."""
```

pcdsdevices/attenuator.py
```diff
@@ -1101,6 +1102,101 @@ class AT2L0(FltMvInterface, PVPositionerPC, LightpathInOutMixin):
     num_in = Cpt(InternalSignal, kind='hinted')
     num_out = Cpt(InternalSignal, kind='hinted')
```
ZLLentz, line 1104: This isn't something you could have known ahead of time, but if you want a new function to pass through the tab completion filter in `hutch-python` you should add something like this near the top of the class:
```
tab_whitelist = ['clear_errors']
```

mkestra:
Okay, code has been tested and up-to-date with my local branch, have at it!

pcdsdevices/attenuator.py
```diff
@@ -1101,6 +1103,101 @@ class AT2L0(FltMvInterface, PVPositionerPC, LightpathInOutMixin):
...
+        doc='summarize errors at any time on any blade via a bitmask',
+    )
+    set_metadata(error_summary_bitmask, dict(variety='bitmask', bits=18))
+
+    def clear_errors(self):
+        """Reset all attenuator errors, making the device ready to move."""
+        self.reset_errors.put(1)
+
+    def _reset_errors(self, value: OphydDataType) -> SignalToValue:
+        return{sig: 1 for sig in self.parent.reset_errors.signals}
```
klauer, line 1182: This is a-OK according to flake8. New to me that it's acceptable! (the lack of space between `return` and `{` is what I'm pointing at)

klauer (approved):
This looks really good! Great work, @mkestra. 

For one of your first few pcdsdevices contributions, you got one of the most difficult. 😁

pcdsdevices/attenuator.py
```diff
@@ -1101,6 +1103,101 @@ class AT2L0(FltMvInterface, PVPositionerPC, LightpathInOutMixin):
...
+            start_index += step
+            end_index += step
+
+        decimal_value = 0
+        for next_bit in errors:
+            decimal_value = decimal_value * 2 + next_bit
+
+        return decimal_value
+
+    """summarize errors occurring at any time on any blade via a bitmask"""
```
klauer, line 1168:
```suggestion
```

Oops, a couple more minor things to clean that I thought were already gone.

pcdsdevices/attenuator.py
mkestra: I'll add it if it still seems like good practice.

pcdsdevices/attenuator.py
mkestra: I agree, I'll remove them!

mkestra:
added the minor tweaks ken suggested.

PR #998, opened 2022-04-14
Device classes for Apertures in AT1K0

mkestra:
## Description

Added two classes, GattApertureY, GatttApertureX to create new typhos screens that will replace the old screens.
## Motivation and Context

A potentially preventable controls hotline call was taken because the old screens did not show all relevant info.
## How Has This Been Tested?

The typhos screens that these are used to build have been tested to move between the aperture and motor states that they interact with while the SXR didn't have beam on 4/14/2022

comments in class declarations

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/998-add_AT1K0_Apertures.rst
+++ b/docs/source/upcoming_release_notes/998-add_AT1K0_Apertures.rst
@@ -15,8 +15,8 @@ Device Updates
 
 New Devices
 -----------
-The X apertures for AT1K0 now have their own device with 1 state, "centered"
-The Y apertures for AT1K0 now have their own device with 4 states, ["5.5mm","8mm","10mm","13mm"]
+- The X apertures for AT1K0 now have their own device with 1 state, "centered"
+- The Y apertures for AT1K0 now have their own device with 4 states, ["5.5mm","8mm","10mm","13mm"]
 
 Bugfixes
 --------
--- a/docs/source/upcoming_release_notes/998-add_AT1K0_Apertures.rst
+++ b/docs/source/upcoming_release_notes/998-add_AT1K0_Apertures.rst
@@ -1,0 +1,31 @@
+998 add AT1K0 Apertures
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+The X apertures for AT1K0 now have their own device with 1 state, "centered"
+The Y apertures for AT1K0 now have their own device with 4 states, ["5.5mm","8mm","10mm","13mm"]
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- N/A
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -31,7 +31,7 @@ MAX_FILTERS = 12
 class Gatt_Y_aperture(TwinCATStatePMPS):
     """
     Controls the Y axis component of the apertures inside AT1K0
-    Defines the state count as 4 (5.5mm, 8mm, 10mm, 13mm) to limit the number of
+    Defines the state count as 4, (5.5mm, 8mm, 10mm, 13mm) to limit the number of
     config PVs we connect to.
     """
 
@@ -44,7 +44,7 @@ class Gatt_Y_aperture(TwinCATStatePMPS):
 class Gatt_X_aperture(TwinCATStatePMPS):
     """
     Controls the X axis component of the apertures inside AT1K0
-    Defines the state count as 1 (centered) to limit the number of
+    Defines the state count as 1, (centered) to limit the number of
     config PVs we connect to.
     """
 
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -19,6 +19,7 @@ from .device import UnrelatedComponent as UCpt
 from .device import UpdateComponent as UpCpt
 from .epics_motor import BeckhoffAxisNoOffset
 from .inout import InOutPositioner, TwinCATInOutPositioner
+from .pmps import TwinCATStatePMPS
 from .interface import BaseInterface, FltMvInterface, LightpathInOutMixin
 from .signal import InternalSignal
 from .utils import get_status_float, get_status_value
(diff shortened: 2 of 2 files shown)


pcdsdevices/attenuator.py
```diff
@@ -27,6 +28,31 @@
 logger = logging.getLogger(__name__)
 MAX_FILTERS = 12
 
+class Gatt_Y_aperture(TwinCATStatePMPS):
```
klauer, line 31: Trying to avoid the `_` in the name: maybe `GattApertureX` and `GattApertureY`?

pcdsdevices/attenuator.py
```diff
@@ -27,6 +28,31 @@
...
+    config PVs we connect to.
+    """
+
+    states_list = []
+    in_states = []
+    out_states = []
+    _in_if_not_out = True
+    config = UpCpt(state_count=4)
+
+class Gatt_X_aperture(TwinCATStatePMPS):
```
klauer, line 44: You'll need two blank lines in between class definitions.
Did you set up pre-commit?

https://confluence.slac.stanford.edu/display/PCDS/How+to+add+a+device+to+pcdsdevices#Howtoaddadevicetopcdsdevices-Pre-commit

klauer (changes requested):

pcdsdevices/attenuator.py
```diff
@@ -27,6 +28,31 @@
...
 MAX_FILTERS = 12
 
+class Gatt_Y_aperture(TwinCATStatePMPS):
+    """
+    Controls the Y axis component of the apertures inside AT1K0
+    Defines the state count as 4 (5.5mm, 8mm, 10mm, 13mm) to limit the number of
+    config PVs we connect to.
+    """
+
+    states_list = []
```
tangkong, line 38: You've overridden the default values of `in_states` and `out_states`, but aren't `state_list` and `_in_if_not_out` the same as defined in `TwinCATStatePMPS`?  

I'm maybe not well informed about the device we're building, but how does this work?  If there's no in or out state, is `TwinCATStatePMPS` (a subclass of an `InOutPositioner`) the right base class for this?

klauer: I think you may be right that the states lists are unnecessary, @tangkong. Strictly speaking the only thing that should be required here is the `UpCpt`, I believe, but @ZLLentz can correct me if I'm wrong (the machinery is complicated here!).
 
The goal is to mainly change the number of states in the `config` component from the statically-defined default of 10 (if I recall correctly) because it can vary on a device-by-device basis. That is, even if there are PVs that exist for all 10 states, we only want the first 4 here for Y and the first one for X (below).

pcdsdevices/attenuator.py
ZLLentz: https://peps.python.org/pep-0008/#class-names

pcdsdevices/attenuator.py
ZLLentz: You do need to clear most of the defaults to avoid errors like "you don't have an OUT state, what gives?" if your device doesn't have an "OUT" state, so I told Mike to clear them all, but it's likely that you're correct and some of these are not needed.

ZLLentz:
I have nothing to add review-wise, Robert and Ken covered it

tangkong:
As a more general comment, I think it'd be helpful to describe briefly what this device is on the beamline.  My understanding is that this is a gas attenuator, and at first read I thought we were defining a solid slit device.

We can add this information (as well as what it's targeting, AT1K0) into the docstring, if you'd like to keep the class name as `GattApertureY` etc.

pcdsdevices/attenuator.py
mkestra: yes looking back at a couple levels of inheritance, it looks like `states_list=[]` and and `_in_if_not_out=True` are already defined in [TwincatInOutPositioner](https://github.com/mkestra/pcdsdevices/blob/5ae8c2cb7846402001d2dd3593b26d4e1bee84f7/pcdsdevices/inout.py#L221l). I will remove these.

pcdsdevices/attenuator.py
mkestra: I have now!

mkestra:
Hi Robert, I tried to add more description of AT1K0 and the how the apertures are related to it, please tell me if this doesn't seem to address your comment.

pcdsdevices/attenuator.py
```diff
@@ -28,6 +29,34 @@
 MAX_FILTERS = 12
 
 
+class GattApertureY(TwinCATStatePMPS):
+    """
+    AT1K0 is a gas attenuator containing 4 discrete aperture arrays,
```
tangkong, line 34: Much more helpful to the uneducated (like me 😄 ). Is this restricted to AT1K0?  Or are there similar devices that might use this class?

I only ask because I've seen device classes for explicit variants (like [AT2L0](https://github.com/pcdshub/pcdsdevices/blob/f7d03f4aaa3f84512c2ef6e38944a0ae0e0c0de4/pcdsdevices/attenuator.py#L1047)).  If this is a general use class, I'd just suggest writing the docstring to suggest that.  (mentioning AT1K0 as an example implementation rather than **_the_** device, maybe)

klauer (commented):
I think you addressed all of our suggestions and this is looking much better. Would you mind adding release notes as well?

pcdsdevices/attenuator.py
mkestra: ah okay, to my understanding, this is the only place on the beamline that this particular style of aperture of is used.  I believe a different class would be used or written for any other device.

tangkong: Gotcha, sounds good then.  Thanks for humoring my naïveté

ZLLentz (approved):

docs/source/upcoming_release_notes/998-add_AT1K0_Apertures.rst
```diff
@@ -0,0 +1,31 @@
...
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+The X apertures for AT1K0 now have their own device with 1 state, "centered"
+The Y apertures for AT1K0 now have their own device with 4 states, ["5.5mm","8mm","10mm","13mm"]
```
ZLLentz, line 19:
```suggestion
- The X apertures for AT1K0 now have their own device with 1 state, "centered"
- The Y apertures for AT1K0 now have their own device with 4 states, ["5.5mm","8mm","10mm","13mm"]
```

PR #999, opened 2022-04-19
Add device for optics notepad

nrwslac:
## Description
Add device for optics notepad

## Motivation and Context

Create a device to easily generate a screen and interface with python.

## How Has This Been Tested?

happi device optics_homs_notepad and typhos

optics-pitch-notepad

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/999-optics-notepad.rst
+++ b/docs/source/upcoming_release_notes/999-optics-notepad.rst
@@ -7,7 +7,7 @@ API Changes
 
 Features
 --------
-- class for storing pitch positions based on state
+- N/A 
 
 Device Updates
 --------------
@@ -15,7 +15,8 @@ Device Updates
 
 New Devices
 -----------
-- OpticsPitchNotepad
+- OpticsPitchNotepad - a class for storing pitch positions based on state in a notepad IOC
+  for mr1l0, mr2l0, mr1l4, mr1l3, and mr2l3.
 
 Bugfixes
 --------
--- a/docs/source/upcoming_release_notes/999-optics-notepad.rst
+++ b/docs/source/upcoming_release_notes/999-optics-notepad.rst
@@ -7,7 +7,7 @@ API Changes
 
 Features
 --------
-- N/A
+- class for storing pitch positions based on state
 
 Device Updates
 --------------
--- a/docs/source/upcoming_release_notes/999-optics-notepad.rst
+++ b/docs/source/upcoming_release_notes/999-optics-notepad.rst
@@ -1,0 +1,30 @@
+999 optics-notepad
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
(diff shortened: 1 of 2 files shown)


pcdsdevices/mirror.py
```diff
@@ -915,3 +915,26 @@ class XOffsetMirrorState(XOffsetMirror, CoatingState):
     """
     # UI representation
     _icon = 'fa.minus-square'
+
+
+class OpticsPitchNotepad(BaseInterface, Device):
+    """
+        Class for storing pitch positions based on state 
+    """
```
ZLLentz, line 923: Docstring formatting
```suggestion
    """
    Class for storing pitch positions based on state.
    """
```
You should also add what this is.
What is the optics notepad?
What is it used for?
Why does it exist?
How is this different from the PV setpoints on each IOC?

ZLLentz:
I think this just needs a better explanation as to what it is + the pre-release notes
The signals don't use standard naming but I understand why, to an extent

nrwslac:
Should this be something more general? Like a notepad device? Note sure what passing args in through typhos or using multiple prefixes in one Device looks like.

ZLLentz:
> Should this be something more general? Like a notepad device? Note sure what passing args in through typhos or using multiple prefixes in one Device looks like.

That sounds like a really annoying device to make and use, and it doesn't really make a lot of sense for me. The API would basically be "write a device class in the args".

If you have a specific set of PVs that you want access to in screens and in hutch python and potentially elsewhere, making an explicit class for it is best.

The only odd thing about this device as-is is that it doesn't use the naming convention for signal names (lowercase letters with underscores).

PR is just missing the release notes doc stub and two style issues reported by the CI:
```
pcdsdevices/mirror.py:922:53: W291 trailing whitespace
pcdsdevices/mirror.py:924:63: W291 trailing whitespace
```

docs/source/upcoming_release_notes/999-optics-notepad.rst
```diff
@@ -0,0 +1,30 @@
...
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- OpticsPitchNotepad
```
ZLLentz, line 18: The last item on the checklist is:
```
Pre-release docs include context, functional descriptions, and contributors as appropriate
```
You should be slightly more helpful here so that someone who reads the release notes has an understanding of what this means.

nrwslac: added more info under features

klauer: I think something like this would be more appropriate, because it's not obvious what your 'feature' is applying to as you are proposing now:

```
OpticsPitchNotepad - a class for storing pitch positions {in a notepad IOC for mirrors x, y, and z}
```

ZLLentz (approved):
I think we're good

ZLLentz:
merged into dev too

Issue #1002, opened 2022-04-25
Submission guideline tweaks


klauer:
## Current behavior

New PR submitters tend to get confused about specific aspects of the process:
* Using pre-commit, or running isort/flake8 separately
* What the sections in the pre-release notes mean and how to fill them in correctly

## Potential fix

For the pre-release notes section, 
* We already have an excellent explanation of the sections:
* https://github.com/pcdshub/pcdsdevices/blob/master/docs/source/upcoming_release_notes/template-full.rst#maintenance
* However, we do not link this from anywhere that I can see or include it in the documentation
* So: include it in the documentation and link it in the PR template seem to be reasonable options

For pre-commit/isort/flake8, we should, perhaps:
* Link existing documentation
* Make an additional checkbox stating that you read those docs?

Other/better ideas?

ZLLentz:
We can also extend the CI to run the pre-commit to help it not fall through the cracks

ZLLentz:
Thoughts on adding a new job for it vs replacing the normal linter with a pre-commit check?

tangkong:
I remember hearing about some repos applying things like isort and flake8 fixes on commit, rather than relying on the user to apply those changes.  

I don't necessarily like that approach, but it does take the problem partially out of the hands of the submitter.

Optimistically, at some point we'll run out of people to teach about pre-commit hooks.  Realistically I think making a page that explains each step and linking it in the issue / PR template would be helpful.

klauer:
> Thoughts on adding a new job for it vs replacing the normal linter with a pre-commit check?

Given that a flake8 check is part of pre-commit, I don't see the need for both. I'd vote on replacing it. It might also make it more obvious that we specifically want submitters to follow the pre-commit steps.

> I remember hearing about some repos applying things like isort and flake8 fixes on commit, rather than relying on the user to apply those changes.

I think that - despite a bit of frustration on the submitter end and our end for enforcing such things - there is potential for people to learn more about good Python development practices. Personally, I'm OK with continuing to nudge people in that direction, even if it is manually.

> Realistically I think making a page that explains each step and linking it in the issue / PR template would be helpful.

Is [this](https://confluence.slac.stanford.edu/display/PCDS/How+to+add+a+device+to+pcdsdevices#Howtoaddadevicetopcdsdevices-DoIreallyneedtomakeanewdevice?) too complicated/too wordy? I'm open to all input there

ZLLentz:
[...]

PR #1003, opened 2022-04-25
STY/CI: satisfy pre-commit checks

ZLLentz:
## Description

I ran `pre-commit run --all-files`

Skip the one test folder that causes issues if you pre-commit it (pre-commit wants to edit the file that the test suite generates)

## Motivation and Context

With the updated travis CI config, let's clean up this repo so that everyone only sees deviations from their PR, not pre-existing conditions.

https://github.com/pcdshub/pcds-ci-helpers/pull/72

Related to #1002 

## How Has This Been Tested?

This is tests

Not yet, I guess. We'll probably add a link and an item to the pre-merge checklist.

## Pre-merge checklist


ZLLentz:
I might repurpose this PR as a big pre-commit all files PR

ZLLentz:
If I do, quick self-reminder that the lens tests folder needs to be excluded from pre-commit checks

klauer:
> I might repurpose this PR as a big pre-commit all files PR

Less stuff than I expected, to be honest:
https://app.travis-ci.com/github/pcdshub/pcdsdevices/jobs/568162814
I'm :+1: on the overall fixup.



> If I do, quick self-reminder that the lens tests folder needs to be excluded from pre-commit checks

If that's the case, we should update the pre-commit config to exclude that directory. 
You can use syntax similar to this for `files` or `exclude`, by the way. I think it's rather nice: https://github.com/pcdshub/shared-dotfiles/blob/acdadefc1fdf8d841c584759468d1fa1bcf51f0c/.pre-commit-config.yaml#L23-L30

pcdsdevices/ui/AT2L0.detailed.ui
```diff
@@ -1 +1 @@
-detailed_tree.ui
```
tangkong, line 1: What happened here?

tangkong (approved):
Looks good to me.  Very satisfying.

pcdsdevices/ui/AttenuatorSXR_Ladder.detailed.ui
```diff
@@ -1 +1 @@
-detailed_tree.ui
\ No newline at end of file
+detailed_tree.ui
```
tangkong, line 1: What happened here?  Seems like a line was deleted then replaced?

klauer: I think it's the newline at EOF:
```
diff --git a/pcdsdevices/ui/AT2L0.detailed.ui b/pcdsdevices/ui/AT2L0.detailed.ui
index 0711737e06..19e4afe071 120000
--- a/pcdsdevices/ui/AT2L0.detailed.ui
+++ b/pcdsdevices/ui/AT2L0.detailed.ui
@@ -1 +1 @@
-detailed_tree.ui
\ No newline at end of file
+detailed_tree.ui
```

Glad you pointed this out. It appears that pre-commit messed up symlinks here!
image

Excluding `*.ui` from the EOL fixer seems like it could be an okay approach. Or there is a "types" option in pre-commit `symlinks` that could be used to exclude it somehow, it appears.

klauer (changes requested):
flake8 + isort-related changes are nice, and most of the whitespace-at-EOL fixes.

* Symlink issue needs fixing somehow
* It looks like the test suite file imports are at odds with flake8? Or what happened here? https://app.travis-ci.com/github/pcdshub/pcdsdevices/jobs/568172922

pcdsdevices/ui/AttenuatorSXR_Ladder.detailed.ui
ZLLentz: I think excluding .ui and also excluding symlinks will both be necessary. I think these should be excluded from all checks for this repo, since .ui is always generated from designer and therefore there isn't any particular style consideration to figure out. The symlink thing is weird.

ZLLentz:
> the test suite file imports are at odds with flake8

I have no idea what's going on in this CI output. Clearly, the pre-commit I ran locally and the pre-commit running on travis are in disagreement? Or maybe something failed to be committed properly.

Overall:
1. roll-back the STY commit
2. ignore .ui / symbolic links
3. double check what's up with those test imports/isort and find a solution

pcdsdevices/ui/AttenuatorSXR_Ladder.detailed.ui
ZLLentz: RE: types, we would need to exclude symbolic links on a per-hook basis. For now I'll just slap this onto the EOL fixer.

ZLLentz:
```
check for broken symlinks............................(no files to check)Skipped
```
this is a consequence of ignoring .ui files and having all the symbolic links end in .ui... maybe one more revision is needed

ZLLentz:
After the last change, we get this on travis:
```
check for broken symlinks................................................Passed
```
Still skips on windows, which is ok as long as the travis catches it. I don't expect for full feature sets to work on windows but it can be convenient as a dev env for me sometimes.

klauer (approved):
LGTM. Nice work!

PR #1004, opened 2022-04-27
ENH: Fix / add screen methods

tangkong:
## Description
- Adds the `.screen` method to `BaseInterface`, which relies on typhos to auto-generate a screen
- Unifies existing `.screen` methods behind `subprocess`, for more flexibility and consistency
- Fixes bugs in existing IPM screen calls
- Adds camviewer-based `.screen` method to AreaDetector class, with option to open "main" edm screen or the python screen by default.

## Motivation and Context
[Jira](https://jira.slac.stanford.edu/browse/LCLSECSD-282)

There can be times where finding screens is difficult, so opening them from hutch-python can be helpful.  Leaning on existing engineering_tools functions seems to be an efficient path forward.  

IPM devices have proved difficult to universally write a screen method for.  The devices present in ipmConfigEpics do not always match devices in the hutch-python sessions, and  when they do the PV's do not match.  More details in the Jira ticket.  I have chosen to leave the IPM's out of this PR for the time being.

## How Has This Been Tested?
Interactively.  (not really sure how to test this)

This PR, some docstrings

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1004-fix_screen_methods.rst
+++ b/docs/source/upcoming_release_notes/1004-fix_screen_methods.rst
@@ -28,4 +28,4 @@ Maintenance
 
 Contributors
 ------------
-- roberttk
+- tangkong
--- a/docs/source/upcoming_release_notes/1004-fix_screen_methods.rst
+++ b/docs/source/upcoming_release_notes/1004-fix_screen_methods.rst
@@ -1,0 +1,31 @@
+1004 fix_screen_methods
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- adds `.screen()` method to BaseInterface, which opens a typhos screen
+- adds AreaDetector specific `.screen()` method, which calls camViewer
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- fixes calls to ipm_screen
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- roberttk
--- a/pcdsdevices/areadetector/detectors.py
+++ b/pcdsdevices/areadetector/detectors.py
@@ -65,7 +65,15 @@ class PCDSAreaDetectorBase(DetectorBase):
         return port_edges
 
     def screen(self, main=False):
-        """ Try and call camviewer here... """
+        """
+        Open camViewer screen for camera.
+
+        Parameters
+        ----------
+        main : bool, optional
+            Set to True to bring up 'main' edm config screen.
+            Defaults to False, which opens python viewer.
+        """
         if shutil.which('camViewer'):
             arglist = ['camViewer', '-H', str(get_hutch_name()).lower(),
                        '-c', self.name]
--- a/pcdsdevices/areadetector/detectors.py
+++ b/pcdsdevices/areadetector/detectors.py
@@ -64,7 +64,7 @@ class PCDSAreaDetectorBase(DetectorBase):
                           for src, dest in port_edges]
         return port_edges
 
-    def screen(self, main=False):
+    def screen(self, main: bool=False) -> None:
         """
         Open camViewer screen for camera.
 
@@ -74,17 +74,17 @@ class PCDSAreaDetectorBase(DetectorBase):
             Set to True to bring up 'main' edm config screen.
             Defaults to False, which opens python viewer.
         """
-        if shutil.which('camViewer'):
-            arglist = ['camViewer', '-H', str(get_hutch_name()).lower(),
-                       '-c', self.name]
-            if main:
-                arglist.append('-m')
-
-            logger.info('starting camviewer')
-            subprocess.run(arglist, check=False)
-        else:
-            logger.info('no camViewer available')
+        if not shutil.which('camViewer'):
+            logger.error('no camViewer available')
             return
+
+        arglist = ['camViewer', '-H', str(get_hutch_name()).lower(),
+                    '-c', self.name]
+        if main:
+            arglist.append('-m')
+
+        logger.info('starting camviewer')
+        subprocess.run(arglist, check=False)
 
 
 class PCDSHDF5BlueskyTriggerable(SingleTrigger, PCDSAreaDetectorBase):
@@ -391,8 +391,9 @@ class PCDSAreaDetectorTyphos(Device):
                    '--oneline',
                    'GE:16,{0}:IMAGE1;{0},,{0}'.format(self.prefix[0:-1])]
 
-        subprocess.run(arglist, stdout=subprocess.PIPE,
-                       stderr=subprocess.PIPE, check=True)
+        self.log.info('Opening python viewer for camera...')
+        subprocess.run(arglist, stdout=subprocess.DEVNULL,
+                       stderr=subprocess.DEVNULL, check=True)
 
     # Make viewer available in Typhos screen
     cam_viewer = Cpt(AttributeSignal, attr='_open_screen', kind='normal')
--- a/pcdsdevices/areadetector/detectors.py
+++ b/pcdsdevices/areadetector/detectors.py
@@ -5,6 +5,7 @@ All components at the detector level such as plugins or image processing
 functions needed by all instances of a detector are added here.
 """
 import logging
+import shutil
 import subprocess
 import time
 import warnings
@@ -62,6 +63,20 @@ class PCDSAreaDetectorBase(DetectorBase):
             port_edges = [(port_map[src].name, port_map[dest].name)
                           for src, dest in port_edges]
         return port_edges
+
+    def screen(self, main=False):
+        """ Try and call camviewer here... """
+        if shutil.which('camViewer'):
+            arglist = ['camViewer', '-H', str(get_hutch_name()).lower(),
+                       '-c', self.name]
+            if main:
+                arglist.append('-m')
+
+            logger.info('starting camviewer')
+            subprocess.run(arglist, check=False)
+        else:
+            logger.info('no camViewer available')
+            return
 
 
 class PCDSHDF5BlueskyTriggerable(SingleTrigger, PCDSAreaDetectorBase):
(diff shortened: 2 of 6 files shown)


tangkong:
I think this is where the scope of this PR will end.  Other screens are difficult to open given the information in the ophyd device, at least currently.  

I'm sure more screens methods will be added later, on request most likely.

pcdsdevices/areadetector/detectors.py
```diff
@@ -63,6 +64,28 @@ def get_plugin_graph_edges(self, *, use_names=True, include_cam=False):
...
+            Set to True to bring up 'main' edm config screen.
+            Defaults to False, which opens python viewer.
+        """
+        if shutil.which('camViewer'):
+            arglist = ['camViewer', '-H', str(get_hutch_name()).lower(),
+                       '-c', self.name]
+            if main:
+                arglist.append('-m')
+
+            logger.info('starting camviewer')
```
klauer, line 83:
```suggestion
            self.log.info('starting camviewer')
```
The per-Device logger has metadata including the object name, so it's preferable to use it

pcdsdevices/areadetector/detectors.py
```diff
@@ -63,6 +64,28 @@ def get_plugin_graph_edges(self, *, use_names=True, include_cam=False):
                           for src, dest in port_edges]
         return port_edges
 
+    def screen(self, main=False):
```
klauer, line 67: While we're at it:

```suggestion
    def screen(self, main: bool = False) -> None:
```

pcdsdevices/areadetector/detectors.py
```diff
@@ -63,6 +64,28 @@ def get_plugin_graph_edges(self, *, use_names=True, include_cam=False):
...
+        """
+        Open camViewer screen for camera.
+
+        Parameters
+        ----------
+        main : bool, optional
+            Set to True to bring up 'main' edm config screen.
+            Defaults to False, which opens python viewer.
+        """
+        if shutil.which('camViewer'):
```
klauer, line 77: It's a good thing to check for `camViewer` in the path 👍 

I like how you did the "return early if not found" when attempting to run typhos in `pcdsdevices.interface` - might be worth replicating that here, too.

pcdsdevices/interface.py
```diff
@@ -382,6 +384,27 @@ def post_elog_status(self):
...
+                         'screen cannot be opened')
+            return
+
+        arglist = ['typhos', f'{self.name}']
+        logger.info(f'Opening typhos screen for {self.name}...')
+
+        # capture stdout and stderr
+        subprocess.Popen(arglist,
+                         stdout=subprocess.PIPE,
+                         stderr=subprocess.PIPE)
```
klauer, line 406: Piping stdout/stderr to where exactly? Did you by chance just want it to be dumped to `/dev/null`? (https://docs.python.org/3/library/subprocess.html#subprocess.DEVNULL)

Could `open_typhos_for_device` be a utility function?

klauer:
We lack coverage in a _lot_ of pcdsdevices, so it would be weird to single this out as "we must have coverage on this!".

However, if we _really_ wanted to see the lines covered here, you could monkeypatch `subprocess.Popen` in a test suite and ensure that it gets run with the arguments you expect for a given device. Coupling that with your own interactive testing would still be necessary, of course.

I believe the above isn't really worth the extra effort, but it's up to you if you want to go the extra mile.

pcdsdevices/areadetector/detectors.py
tangkong: I suspect I was investigating a different way of opening cameras if `camViewer` didn't exist.  (maybe searching for the `{P}:LAUNCH_EDM` PV?). A good catch nonetheless, `camViewer` does just that.

pcdsdevices/interface.py
tangkong: What were you thinking about for the `open_typhos_for_device` function?  Were you thinking of passing other typhos kwargs through?  I figured that if people wanted to open a typhos screen they would typically just open it from another shell session?

pcdsdevices/areadetector/detectors.py
tangkong: I wanted to make sure that the logging info printed to the terminal, since I'm making these methods as quiet as possible otherwise.  Do we expect the per-Device logger to also print to the hutch-python session?

pcdsdevices/interface.py
klauer: I think what you have is just fine, really. I meant that "open a typhos screen for a device named X in a subprocess" is a generically reusable utility, such that creating something with an API like the following could be useful:

```python
def open_typhos_for_device(device_name: str, **some_options_here):
    ... # insert your .screen() functionality here
```

pcdsdevices/areadetector/detectors.py
```diff
@@ -63,6 +64,28 @@ def get_plugin_graph_edges(self, *, use_names=True, include_cam=False):
                           for src, dest in port_edges]
         return port_edges
 
+    def screen(self, main=False):
+        """
+        Open camViewer screen for camera.
```
ZLLentz, line 69: This is actually the second camviewer open script in this file, I think your implementation may be better than the previous.
https://github.com/pcdshub/pcdsdevices/blob/22c5b9cbfcb3330d0a698d1230903fea26e1dd24/pcdsdevices/areadetector/detectors.py#L361-L372

pcdsdevices/areadetector/detectors.py
```diff
@@ -383,6 +406,10 @@ def _open_screen(self):
     def _open_screen(self, value):
         self.open_viewer()
 
+    def screen(self):
+        """ Lean on open_viewer method """
+        self.open_viewer()
```
ZLLentz, line 411: Ah, didn't see this before making previous comment

pcdsdevices/interface.py
ZLLentz: I'm surprised there isn't already a typhos built-in for this

pcdsdevices/utils.py
```diff
@@ -160,8 +160,11 @@ def ipm_screen(dettype, prefix, prefix_ioc):
     if shutil.which(executable) is None:
         raise EnvironmentError('%s is not on path, we cannot start the screen'
                                % executable)
-    os.system('%s --base %s --ioc %s --evr %s &' %
-              (executable, prefix, prefix_ioc, prefix+':TRIG'))
+
+    logger.info(f'Opening {dettype} screen for {prefix}...')
+    arglist = [executable, '--base', prefix, '--ioc', prefix_ioc,
+               '--evr', prefix+':TRIG']
+    _ = subprocess.Popen(arglist)
```
ZLLentz, line 167: What's the functional difference between binding the Popen output to an unused variable vs no assignment at all?

ZLLentz (approved):

pcdsdevices/areadetector/detectors.py
ZLLentz: I would have expected the per-device logger to show in the hutch-python session. Otherwise we have some sort of bug.

pcdsdevices/utils.py
tangkong: I don't entirely understand why, but in this case `Popen` was returning a value that printed a repr to the terminal.  Each external script behaves a little differently, but I didn't spend a lot of time digging into why.

pcdsdevices/areadetector/detectors.py
tangkong: Yeah, I didn't want to remove this in case I was missing an edge-case where they behaved differently.  To my knowledge, they should be identical.

ZLLentz:
I think I want to tag immediately after this PR, no rush though because we aren't on a strict tagging schedule

tangkong:
Remaining to-do: 
- run through happi and see what screen does for all the devices
- commit some cleanup changes

klauer:
I had been making some issues in https://github.com/pcdshub/device_config/issues - but it's not a repo anyone usually pays attention to, outside of us and Tyler P.

You could further filter your list by only checking devices that are active. (On that note, I think invalid entries should be marked 'inactive' at some point.)

ZLLentz:
At some point we were planning a `happi audit` sort of feature where happi would check its own database for invalid entries, etc., but the person working on it left SLAC partway through.

I don't think we've ever gone through and made a concentrated effort to clean up our db but having a semi-automatic tool would help us quite a bit in doing that.

tangkong:
> You could further filter your list by only checking devices that are active. (On that note, I think invalid entries should be marked 'inactive' at some point.)

```python
In [5]: df[df['screen_method']=='device not valid']['active'].value_counts()
Out[5]: 
True     62
False     3
Name: active, dtype: int64
```
😬 

I agree making some audit tools will be pretty helpful, though certainly outside the scope here.  I'll drop an issue somewhere

ZLLentz:
I think this is looking really good- I'll remove the WIP from the title at least

ZLLentz:
The checklist is a lie though- I don't see a committed pre-release docs page

tangkong:
> The checklist is a lie though- I don't see a committed pre-release docs page

Sorry, it was held up behind my logging confusion.  For posterity: device.log.info is closer to debug level, so here we continue to use the module level logger with the intention to switch the device logger once we push some changes back up to ophyd.

docs/source/upcoming_release_notes/1004-fix_screen_methods.rst
```diff
@@ -0,0 +1,31 @@
...
+--------
+- fixes calls to ipm_screen
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- roberttk
```
ZLLentz, line 31: gh username
```suggestion
- tangkong
```

tangkong: ffff

ZLLentz (approved):

tangkong:
I'll merge mid-day today if there are no further requests.  (I'm still timid around the big green button)

PR #1020, opened 2022-06-10
Add rfof device

wwright-slac:
## Description
Added PCDSDevice for Instrumentation Technologies RF over Fiber device. Users have the option to instantiate base class (ItechRfofBase) which gives access to all PVs (from transmitter and receiver units, as well as status and error PVs), or individually instantiate subclasses for each desired set of PVs.

## Motivation and Context
See above.

## How Has This Been Tested?
Successfully tested in ipython:
rfof base

sub classes

Pre-commit pass:
precommit

Work in progress documentation page:
https://confluence.slac.stanford.edu/pages/viewpage.action?pageId=326514481

## Pre-merge checklist


pcdsdevices/lasers/rfof.py
```diff
@@ -13,17 +13,242 @@
...
+                        doc='Rx: RF control identification failed')
+    rx_rf_sens_err = Cpt(EpicsSignalRO, ':RX_RF_SENS_ERR', kind='normal',
+                         doc='Rx: RF control sensor failure')
+    rx_arm_idn_err = Cpt(EpicsSignalRO, ':RX_ARM_IDN_ERR', kind='normal',
+                         doc='Rx: Arm control identification failed')
+    rx_arm_sens_err = Cpt(EpicsSignalRO, ':RX_ARM_SENS_ERR', kind='normal',
+                          doc='Rx: Arm control sensor failure')
+
+
+class ItechRfofBase(ItechRfofTx, ItechRfofRx,
```
klauer, line 249: This one still needs addressing. Consider this example as to what inheritance does for you:

```python
class MyBaseA:
    a = "this is A"


class MyBaseB:
    b = "this is B"


class MyClass(MyBaseA, MyBaseB):
    """
    MyClass inherits from (or "subclasses") both MyBaseA and MyBaseB. 

    This means all of the elements from MyBaseA and MyBaseB will be included in
    MyClass.

    This is in addition to whatever I put in the class body below.
    """
    c = "this is from MyClass"


print(MyClass().a)   # -> "this is A"
print(MyClass().b)   # -> "this is B"
print(MyClass().c)   # -> "this is from MyClass"
```

With the above in mind, now recall what ophyd does with its "sub-devices" as components:

```python
from ophyd import EpicsSignal, Device, Component as Cpt


class SubDeviceA(Device):
    a_value = Cpt(EpicsSignal, ":suffix")

class SubDeviceB(Device):
    b_value = Cpt(EpicsSignal, ":suffix")

class MyDevice(Device):
    cpt_a = Cpt(SubDeviceA, ":prefix:for:a")
    cpt_b = Cpt(SubDeviceB, ":prefix:for:b")


device = MyDevice("base:prefix", name="device")
print(device.cpt_a.a_value.pvname)   # ->  base:prefix:prefix:for:a:suffix
print(device.cpt_b.b_value.pvname)   # -> base:prefix:prefix:for:b:suffix
print(device.component_names)   # -> ('cpt_a', 'cpt_b')
# device.a_value does not exist (and it shouldn't!) - it's part of the subdevice.
# device.b_value does not exist (and it shouldn't!) - it's part of the subdevice.
```

OK, now what happens if we combine the subclassing _with_ using SubDevice?

```python
class MyDevice(SubDeviceA, SubDeviceB):  # <-- we subclassed
    cpt_a = Cpt(SubDeviceA, "prefix:for:a:")  # <-- and included components
    cpt_b = Cpt(SubDeviceB, "prefix:for:b:")


device = MyDevice("base:prefix:", name="device")
print(device.cpt_a.a_value.pvname)   # ->  base:prefix:prefix:for:a:suffix
print(device.a_value.pvname)  # -> base:prefix::suffix  (!!)
```

Wait, ``a_value`` is on ``device`` AND ``device.cpt_a``, too? This is not good! Look, the prefixes differ confusingly, too...

```python
print(device.component_names)  # -> ('b_value', 'a_value', 'cpt_a', 'cpt_b')
```

These should only be in one spot - and the correct spot would be to keep them as sub-device components (in my opinion) as you could then access components by way of `device.transmitter.tx_*`.

I hope the above makes it a bit clearer why you shouldn't be doing both. Feel free to ask follow-up questions if not!

pcdsdevices/lasers/rfof.py
```diff
@@ -13,17 +13,242 @@
...
+    rx_intp_t = Cpt(EpicsSignalRO, ':RX_INTP_T', kind='normal',
+                    doc='Internal temperature (pressure sensor) (C)')
+
+
+class ItechRfofTx(BaseInterface, Device):
+
+    """Itech RFoF transmitter PVs."""
+
+    # Transmitter adds a single control PV
+    tx_pc1 = Cpt(EpicsSignalRO, ':TX_PC1', kind='normal',
```
klauer, line 89: All of these have an identical prefix `tx_`. You should consider removing this prefix entirely.

It'd also be a good idea to give them names that are sensible for the end-user. You did a great job including the documentation string there. It could be a good hint as to what the attribute names should be.

For example: 
"pc1" means absolutely nothing to me. However, "low-drift optical link loop lock in uV" is very clear. This could mean that instead of `tx_pc1 = Cpt(...)` this could rather be `link_loop_lock = Cpt(...)`

pcdsdevices/lasers/rfof.py
```diff
@@ -13,17 +13,242 @@
...
-    Class for reading back the Cycle RFoF transmitter PVs.
-    """
+
+    """Class for reading back the Cycle RFoF transmitter PVs."""
+
     # Transmitter adds a single control PV
     enable_laser = Cpt(EpicsSignal, ':ENABLELASER', kind='normal')
+
+
+"""Instrumentation Technologies Low-Jitter Continuous Wave"""
```
klauer, line 32: These lines don't do anything - where do they belong?

klauer (changes requested):
Hi @wwright-slac - making some good progress here! 

There's a few things that we need to tackle prior to going forward with merging this.

pcdsdevices/lasers/rfof.py
wwright-slac: Is there a limit on the length of the name of the components? Some of these would be quite long...

klauer: There is no specific limit to component name length. We do, however, keep our code lines to be < 80 characters in general.

Sensible component attribute names are an important thing (in my opinion), and coming up with appropriate ones does take a bit of practice.

I'd suggest formatting these differently if you start increasing the length or amount of information passed to the `Cpt`:
```python
class MyDevice(Device):
    long_component_name = Cpt(
        ComponentClass,
        "Prefix",
        kind="normal",
        doc="This is an explanation about what the component does",
    )
```

I think most would agree that's much more readable than combining it into two lines.

wwright-slac: But for something like the" Variable RF attenuator controller setpoint voltage", there isn't really a way to have that be short, but also descriptive. I could do something like var_rf_att_ctrl_sp_v, I guess?

klauer: `var_rf_att_ctrl_sp_v` has almost lost all meaning with introducing a bunch of new acronyms unnecessarily. A good choice for a name really depends on the device - which I don't know enough about here to make the right suggestion for you.

It does not need to have _everything_ from the `doc` help string, just the important bits to make it uniquely identifiable and useful to the end-user.

Something like `attenuator_setpoint_voltage` or `rf_setpoint_voltage` could be sufficient, if it uniquely identifies its purpose among all of the other components.

pcdsdevices/lasers/rfof.py
wwright-slac: The blank lines? Or the Cycle RFOF lines?

wwright-slac: Ah, yes. I thought the triple quotes also acts as a comment

klauer:
> I ended up removing the base class, as that would make things confusing with the tx and rx monikers missing from the component names.

But they _would_ be present in the component names by virtue of being a component with a sub-device:
```
itech = Itech()
itech.receiver.internal_lock
itech.transmitter.low_drift_opt_link_lock
```

That's part of the reason I mentioned that it was entirely OK to remove `tx_` :)
Or am I still misunderstanding?

wwright-slac:
Ahhhh, I see. I suppose that would be better.

pcdsdevices/lasers/rfof.py
ZLLentz: Sometimes people do put strings in the source code instead of comments, particularly with long multiline comments. I think some Python tutorials mention this. There is a small difference in loading and it's less clear the intent but there isn't much of a functional difference otherwise.

pcdsdevices/lasers/rfof.py
```diff
@@ -1,6 +1,5 @@
-"""
-Module for the RFoF systems used in the LCLS-II timing system.
-"""
```
ZLLentz, line 3: This is going to sound super confusing and contradictory, but strings at the top of modules and at the top of functions or classes have special meanings beyond being comments- they are docstrings and can be picked up as the `__doc__` attribute by inspecting code. So technically this specific documentation was more correct as a string rather than as a comment.

pcdsdevices/lasers/rfof.py
```diff
@@ -11,19 +10,569 @@
...
+        ":TX_PS_V",
+        kind="normal",
+        doc="Electrical phase shifter voltage (V)",
+    )
+    attenuator_setpoint_v = Cpt(
+        EpicsSignalRO,
+        ":TX_ATTS_V",
+        kind="normal",
+        doc="Variable RF attenuator \
+                                controller setpoint voltage (V)",
```
ZLLentz, line 244: I think the result of this is a string that's something like:
```
"Variable RF attenuator                                 controller setpoint voltage (V)"
```
Due to the way the line continuation is handled.

The simplest way to get a string to format correctly while using multiple lines is something like:
```suggestion
        doc=(
            "Variable RF attenuator "
            "controller setpoint voltage (V)"
        ),
```

ZLLentz: Several other doc= kwargs in the module here are continued similarly to this one

ZLLentz (commented):
I'm late to party here but I only see those two minor things I pointed out. Great work!

pcdsdevices/lasers/rfof.py
wwright-slac: Good to know!

pcdsdevices/lasers/rfof.py
wwright-slac: Good to know!

klauer:
I think if you address the couple of formatting issues Zach found and add pre-release notes (*), we should be nearly good to merge here. 👍 

(*) from [here](https://confluence.slac.stanford.edu/display/PCDS/How+to+add+a+device+to+pcdsdevices#Howtoaddadevicetopcdsdevices-Puttingitalltogether)
```bash
# Add a pre-release notes page:
$ cd docs
$ ./pre-release-notes.sh issue_number "add MyDevice"
$ git commit -a -m "DOC: add release notes for MyDevice"
```

docs/source/upcoming_release_notes/issue_number-add_ItechRfof.rst
```diff
@@ -0,0 +1,32 @@
...
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- Will Wright
+- Mike Brown
+- Tyler Johnson
```
ZLLentz, line 32: Github usernames so we can properly credit people for the code they added here
```suggestion
- wwright-slac
- mcb64
- slactjohnson
```

pcdsdevices/lasers/rfof.py
```diff
@@ -11,19 +12,628 @@
...
+    )
+    rx_arm_sens_err = Cpt(
+        EpicsSignalRO,
+        ":RX_ARM_SENS_ERR",
+        kind="normal",
+        doc="Rx: Arm control sensor failure"
+    )
+
+
+class ItechRfofBase(BaseInterface, Device):
```
ZLLentz, line 635: Quick check: why is this called `Base`? I'm usually used to seeing `Base` in class names when they are parent classes for other classes to inherit from, but that's not the case here.  Is there some device-specific thing called a "base" that this represents?

You may consider a short docstring here explaining what this device is, so the next person who browses pcdsdevices understands what this device is for.

docs/source/upcoming_release_notes/issue_number-add_ItechRfof.rst
```diff
@@ -0,0 +1,32 @@
...
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- Instrumentation Technologies RF over Fiber unit
```
ZLLentz, line 18: I suggest including a class name to make it clear what was added. This becomes part of the release notes so it needs to make sense in context of being next to other bullet points.
```suggestion
- Added ItechRfofBase: Instrumentation Technologies RF over Fiber unit
```

ZLLentz (commented):
Just a few questions/comments/suggestions RE: naming and documentation

pcdsdevices/lasers/rfof.py
wwright-slac: You're right. Perhaps, just naming it "ItechRfof" by itself?

ZLLentz: `ItechRfof` is totally :+1:

ZLLentz: The name in your previous diff is totally good too- just want to make sure we've considered the name since it was confusing me.

If we change the name, we also need to change the import in `pcdsdevices.device_types`

ZLLentz:
Excerpt from the automatic test suite run:
```
pcdsdevices/device_types.py:20: in <module>
    from .lasers.rfof import (CycleRfofRx, CycleRfofTx, ItechRfof, ItechRfofErrors,
E   ImportError: cannot import name 'ItechRfof' from 'pcdsdevices.lasers.rfof'
```
Small desync between the import name and the class name

wwright-slac:
Sorry about that Zach! This one should do the trick 🤞

ZLLentz (approved):
In my opinion, this is good to merge!
Ken, have your suggestions been addressed?

klauer (approved):
Yes, this looks really good now. Nice job, @wwright-slac!

wwright-slac:
@ZLLentz @klauer Thanks guys! Next time should be a smoother process, I promise :)

PR #1021, opened 2022-06-14
ENH: experiments with lazy loading for startup speed

ZLLentz:
## Description

- [x] Make components on `BaseInterface` subclasses lazy.
- [x] Set `BaseInterface`'s `lazy_wait_for_connection` to `False`, overriding the ophyd device default.
- [ ] Find other metrics to figure out where to apply lazy loading
- [ ] Rework some of the bigger classes to do fewer big subscriptions on load

## Motivation and Context

It takes ~40s for me to load all of our devices without this patch and ~20s with this patch for a 2x speedup.

`ophyd` has always had a lazy loading mechanism that we've been avoiding because it's less confusing to just have everything loaded at startup, especially because there are a bunch of init-time subscriptions and the whole `lazy_wait_for_connection` thing that don't play nice with each other.

I was doing profiling today and found that we create 32000 components to load our entire collection of beamline devices. This isn't sustainable for something like `lightpath`.

## How Has This Been Tested?

WIP

Some testing will need to be done to see how this behaves live.
There are some annoyances wrt "late" errors that I'm worried about seeing.

WIP

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -193,6 +193,8 @@ class BaseInterface:
         List of string regex to show in autocomplete for non-engineering mode.
     """
 
+    lazy_wait_for_connection = False
+
     tab_whitelist = (OphydObject_whitelist + BlueskyInterface_whitelist +
                      Device_whitelist + Signal_whitelist +
                      Positioner_whitelist)
@@ -215,6 +217,13 @@ class BaseInterface:
             )
 
         cls._class_tab = TabCompletionHelperClass(cls)
+        try:
+            cpts = [getattr(cls, name) for name in cls.component_names]
+        except AttributeError:
+            pass
+        else:
+            for cpt in cpts:
+                cpt.lazy = True
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)


ZLLentz:
I'm going to brainstorm a bit about what we need to do to fully optimize our startup times RE: lazy loading.

It's tempting to skip this step and go straight to returning proxy objects from happi. This might be a good solution for something like `hutch-python`, where the user wants to have lots of objects at once but only uses a handful during a sesion, but this ignores the needs of modules like `lightpath` which want to load lots of devices but only access a subset of their resources.

So, the simplest form of this schema is simply:
- make all the components lazy

This won't hit all the components though, since even a lazy component will be initialized at init time if it has a subscription configured. So, extending a bit:
- make all the components lazy
- make ophydobject subscriptions start only when they are needed

But how do we know when a subscription is needed? We need to add some structure here so that we can set these up with appropriate timings:
- make all the components lazy
- make ophydobject subscriptions start only when they are needed
- add some structure for signalling which signals are needed and when

Another thing to consider: how much of a device do we set up when a resource is accessed? We can completely separate every signal and load them one-by-one on an as-needed basis, but then we run into a problem where we try to run a scan and we set up the signals we need in series and have an awkwardly slow first step.
- make all the components lazy
- make ophydobject subscriptions start only when they are needed
- add some structure for signalling which signals are needed and when
- add some way for us to know which groups to initialize and when so we aren't connecting in series

So here's my outline/pitch of how to accomplish this, glossing over specifics:
- Every component must be lazy.
- Every internal resource in a device that needs a subscription from a signal must be expressed as a signal itself. The lowest-level device subscription would not be set up until the top-level resource signal has been subscribed to or had .get called. This is the structure we need to make it very clear when to set up the subscriptions.
- Every device should define some set of signals to initialize on key methods like "stage" or "move". This makes sure the whole device is set up the first time it is used in a scan or moved interactively. The signals can be initialized in series and then waited on for connection in parallel.

Ok, now let's get more specific in our API/implementation:
- in BaseInterface, enforce global laziness
- make everything in pcdsdevices inherit from BaseInterface
- proliferate MultiDerivedSignal in all cases for resources that need subscriptions
- make a function decorator we can use to initialize signals prior to running the function
- properly apply the function decorator onto move/stage and in other places
- add a function or shortcut for initializing all signals

There might be more to it, but that's where I'm at now.

I'm not going to dive all the way in right away, I want to make sure enough thinking has been done to do all of this correctly.

tangkong:
> proliferate MultiDerivedSignal in all cases for resources that need subscriptions

I arrived at something similar for the lightpath signal subscriptions, but decided I only really wanted a "summary" signal, rather than something that supported some calculate_on_x behavior.  I ended up subclassing `AggregateSignal` and just implemented a simple `_calc_readback` that hashed the values of the sub-signals.  That way the top-level signal varies with the sub-signals.  

I figured that was the least I could do to keep track of a group of signals.  I'm not entirely convinced it's the best approach, but I'm happy to see our thoughts at least align there

ZLLentz:
> I'm happy to see our thoughts at least align there

I agree, it seems to me like this is the most natural/structured way to decide when to propagate subscriptions- when the code asks for a subscription to the calculated/derived resource, and making that resource a signal itself seems to fit.

ZLLentz:
I'm closing this for now- I have a second branch with a cleaner implementation going to the same goals, loosely following the ideas outlined above. I won't delete this branch in case I change my mind.

PR #1024, opened 2022-06-20
REL: v6.2.0

klauer:
## Description

Document and tag the accumulated changes such that the happi test suite will be able to pass.

Reviews are requested from all contributors as a notification that that you commits are being included and as an opportunity to make amendments, but are not mandatory.

Also: I added some quick quality-of-life improvements to `release_notes.py` for this process, ensuring those who don't do it often (like myself) remember the steps.

v6.2.0 (2022-06-20)
===================

Device Updates
--------------
- Add IMS.setup_pmgr as a public API for applications that want to initialize
  pmgr support before the first device uses it. This was previously private
  API at IMS._setup_pmgr.
- Added LED control PVs to CVMI motion class.

New Devices
-----------
- Added ItechRfof class: Instrumentation Technologies RF over Fiber unit

Bugfixes
--------
- Create the pmgr resources when they are first used rather than on IMS
  init, saving 3 seconds of startup time for users that don't need
  pmgr resources.

Maintenance
-----------
- Vendor happi.device.Device as LegacyItem instead of importing it, pending
  deprecation of the happi.device module.

Contributors
------------
- Mbosum
- mcb64
- slactjohnson
- wwright-slac
- zllentz


klauer:
Including #1026 now

ZLLentz (approved):
lgtm

ZLLentz:
We can merge after passing CI
Maarten is on vacation right now

tangkong (approved):
LGTM

ZLLentz (changes requested):
https://github.com/pcdshub/pcdsdevices/blob/master/docs/source/upcoming_release_notes/issue_number-add_ItechRfof.rst

This one didn't get included here

ZLLentz:
We should have caught the "issue number" in the title on review but did not

klauer:
Fixed the malformed filename, re-ran, and force-pushed. _Once more with feeling_, so they say?

klauer:
Moving forward based on previous approval + local testing of pre-release-notes.sh issue number check.

PR #1027, opened 2022-06-20
CLN: remove remaining pre-release notes

klauer:
## Description
The target for #1024 moved a few times and I neglected to delete the pre-release notes files for v6.2.0.

### Why not push to master and avoid judgment?

Instead of pushing to master to quietly correct my mistake, I'll accept the public shaming that comes along with this one.
However, I will not make release notes for this nor check the boxes.


ZLLentz (approved):

tangkong (approved):
Would the direct-push-to-master not get caught and reported by the slack bot? I didn't think there was a sneaky way to do this

ZLLentz:
pushing to master would require some settings edit tomfoolery as well

klauer:
'Twas but a joke, lads.

Travis has yet to see this PR. I don't see a point in cycling it for superfluous testing, and would request that we merge without CI here if you agree.

ZLLentz:
I'm ok with merging this without Travis if Travis has disappointed us today

PR #1028, opened 2022-06-21
ENH: Refactor LightpathMixin and all relevant devices

tangkong:
## Description
Reworks LightpathMixin to have a simpler interface and subscription system.
- Relevant signals are now grouped under a `SummarySignal` (a subclass of AggregateSignal)
- The only required method is now `get_lightpath_status`, which returns a dataclass contianing information needed by lightpath.
- Reworks some existing devices to implement this interface

Will have to be coordinated with a corresponding [lightpath PR](https://github.com/pcdshub/lightpath/pull/133)

This will require a significant overhaul of the existing devices that are to be used in lightpath.  Currently: 
- 177 devices are marked as active and participating in lightpath (active=True, lightpath=True)
  - Of these, only 108 implement the existing LightpathMixin
  - Several devices should be in the lightpath but are not (xpp_lodcm, xcs_lodcm in particular)
- This interface will likely require more specific device classes for branching devices, as their input and output branches will need specificity.

## Motivation and Context
Lightpath is getting an overhaul, and this new interface will be part of it

## How Has This Been Tested?
Eventually in the test suite

This PR

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1028-enh_lp_mixin.rst
+++ b/docs/source/upcoming_release_notes/1028-enh_lp_mixin.rst
@@ -3,7 +3,7 @@
 
 API Changes
 -----------
-- Converts LightpathMixin to the new Lightpath API, consilodating
+- Converts LightpathMixin to the new Lightpath API, consolodating
   reporting into a single LightpathState Dataclass.  The lightpath
   subscription system has also been simplified by using an AggregateSignal
   to monitor all relevant components.
--- a/docs/source/upcoming_release_notes/1028-enh_lp_mixin.rst
+++ b/docs/source/upcoming_release_notes/1028-enh_lp_mixin.rst
@@ -1,0 +1,42 @@
+1028 enh_lp_mixin
+#################
+
+API Changes
+-----------
+- Converts LightpathMixin to the new Lightpath API, consilodating
+  reporting into a single LightpathState Dataclass.  The lightpath
+  subscription system has also been simplified by using an AggregateSignal
+  to monitor all relevant components.
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- Updates LightpathMixin implementation to the new API for all
+  existing lightpath-active devices.  This includes but is not limited to:
+
+  - Mirrors
+  - LODCMs
+  - Attenuators
+
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- Adds happi containers that work with the new Lightpath interface.
+  Notably, these containers allow input_branches and output_branches
+  to be optional kwargs.  This lets these containers work with devices
+  that both do and do not implement the Lightpath interface.
+
+Contributors
+------------
+- tangkong
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -1764,6 +1764,8 @@ class LightpathInOutMixin(LightpathMixin):
         for sig_name, sig_value in lightpath_kwargs.items():
             obj = getattr(self, sig_name)
             if isinstance(obj, LightpathInOutMixin):
+                # TO-DO: deprecate this condition, and rework classes
+                # it applies to.  (attenuators, RTDS)
                 # The inserted/removed are always just a getattr
                 # Therefore, they are safe to call in a callback
                 in_check.append(obj.inserted)
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -11,7 +11,6 @@ import subprocess
 import time
 import typing
 from contextlib import contextmanager
-from dataclasses import dataclass
 from pathlib import Path
 from threading import Event
 from types import MethodType, SimpleNamespace
@@ -21,6 +20,7 @@ from weakref import WeakSet
 import ophyd
 import yaml
 from bluesky.utils import ProgressBar
+from lightpath.path import LightpathState
 from ophyd import Component as Cpt
 from ophyd.device import Device
 from ophyd.ophydobj import Kind, OphydObject
@@ -1563,14 +1563,6 @@ class NullFile:
         pass
 
 
-@dataclass
-class LightpathState:
-    inserted: bool
-    removed: bool
-    transmission: float
-    output_branch: str
-
-
 class LegacyLightpathMixin(OphydObject):
     """
     Mix-in class that makes it easier to establish a lightpath interface.
@@ -1667,37 +1659,42 @@ class LightpathMixin(Device):
     LightpathStatus object
     Create an object similar to the following:
 
-    class MyDevice(LightpathMixin):
-        lp_signals = ['sig1', 'sig2']
+    .. code-block:: python
 
-        def get_lightpath_status(self):
-            # Logic, calculations for transmission, etc
-            status = LightpathStatus(
-                inserted=True, removed=False, transmission=0.0,
-                output_branch='L3'
-            )
-            return status
+        class MyDevice(LightpathMixin):
+            lightpath_signals = ['sig1', 'sig2']
+            sig1 = Cpt(Signal, ':SIG1')
+            sig2 = Cpt(Signal, ':SIG2')
+
+            def calc_lightpath_status(self, sig1=None, sig2=None):
+                # Logic, calculations using sig1, sig2
+                status = LightpathStatus(
+                    inserted=True, removed=False,
+                    transmission=0.0, output_branch='L3'
+                )
+                return status
+
+        dev = MyDevice('PREFIX', name='dev', input_branches=['L0'],
+                       output_branches=['L0'])
     """
     # Component names whose values are relevant for inserted/removed
-    lightpath_cpts = []
+    lightpath_signals = []
 
     # Flag to signify that subclass is another mixin, rather than a device
     _lightpath_mixin = False
 
     # Mixin holds one summary signal that changes with lightpath_cpts
-    lp_summary = Cpt(SummarySignal, name='lp_summary')
+    lightpath_summary = Cpt(SummarySignal, name='lp_summary')
 
     def __init__(self, *args,
                  input_branches=[], output_branches=[], **kwargs):
-        self._lightpath_values = {}
         self._lightpath_ready = False
         self._retry_lightpath = False
         self.input_branches = input_branches
         self.output_branches = output_branches
-        # TODO: grab destination / source lists, if they exist?
         super().__init__(*args, **kwargs)
-        for sig in self.lightpath_cpts:
-            self.lp_summary.add_signal_by_attr_name(sig)
+        for sig in self.lightpath_signals:
+            self.lightpath_summary.add_signal_by_attr_name(sig)
 
         if not self.input_branches or not self.output_branches:
             raise NotImplementedError(
@@ -1711,20 +1708,45 @@ class LightpathMixin(Device):
             # Child of cls will inherit this as False
             cls._lightpath_mixin = False
         else:
-            if not cls.lightpath_cpts:
+            if not cls.lightpath_signals:
                 raise NotImplementedError(
                     'Did not implement LightpathMixin properly.  '
                     'Must supply a list of components (lightpath_cpts)'
                 )
 
-    def get_lightpath_status(self) -> LightpathState:
+    def calc_lightpath_state(self, **kwargs) -> LightpathState:
         """
         Create and return a LightpathState object containing information needed
-        for lightpath
+        for lightpath, given a set of signal values
 
-        Device logic goes here
+        kwargs should be the same as the signal names provided in
+        ``lightpath_signals``
+
+        Device logic goes here.
+
+        Returns
+        -------
+        LightpathState
+            a dataclass containing the Lightpath state
         """
-        raise NotImplementedError('Did not implement LightpathMixin')
+        raise NotImplementedError(
+            'Did not implement LightpathMixin properly.  Must define '
+            'a ``calc_lightpath_state`` method.'
+        )
+
+    def get_lightpath_state(self) -> LightpathState:
+        """
+        Return the current LightpathState
+
+        Returns
+        -------
+        LightpathState
+            a dataclass containing the Lightpath state
+        """
+        kwargs = {sig.name.removeprefix(self.name + '_'): sig.get()
+                  for sig in self.lightpath_summary._signals}
+        status = self.calc_lightpath_state(**kwargs)
+        return status
 
 
 class LightpathInOutMixin(LightpathMixin):
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -1695,6 +1695,9 @@ class LightpathMixin(Device):
         self.input_branches = input_branches
         self.output_branches = output_branches
         super().__init__(*args, **kwargs)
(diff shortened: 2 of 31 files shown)


pcdsdevices/interface.py
```diff
@@ -1646,6 +1656,77 @@ def transmission(self):
                 return 1
 
 
+class LightpathMixin(Device):
```
klauer, line 1659: I am somewhat out of the loop on this, so grain of salt:

I had assumed the mixin itself would go away and we'd end up with a signal like:

`lightpath = Cpt(LightpathSignal, signals=["sig1", "sig2"], status_func=get_lightpath_status)`

Or possibly just retrieving those from the parent class/instance on `__init__`.

I'm not claiming this is in any way _better_. Before updating all of the device classes it may be useful to discuss the benefits of the current approach over other options with an eye toward future maintainability.

pcdsdevices/signal.py
```diff
@@ -422,6 +422,18 @@ def destroy(self):
...
+class SummarySignal(AggregateSignal):
+    """
+    Signal that holds a hash of the values of the constituent signals
+
+    Meant to allow tracking of constituent signals via callbacks.
+    The calculated readback value is useless.
+    """
+    def _calc_readback(self):
+        values = tuple(sig.get() for sig in self._signals)
+        return hash(values)
```
klauer, line 434: Could we just return the tuple itself? Or is there good reason not to?

klauer (commented):
Looking good!

pcdsdevices/interface.py
tangkong: I thought the mixin was a simple way to wrap up the functionality, which boils down to adding the `SummarySignal` and requiring the `get_lightpath_status` method.  I thought it was clearer this way, though I will admit that I didn't think much about a signal that sets a method.  (Component internals scare me slightly).  

I felt the main drawback of the old mixin was that the interface had many too many bits to implement (insert, remove, etc).  Is inheriting from the mixin itself a step that users find onerous? 

I guess the other question is: is the proposed API adequate?  (Outside of the implementation details)

This is the kind of feedback I was looking for, happy to consider other implementation approaches 👍

pcdsdevices/signal.py
tangkong: I wanted to return an ophyd-compatible data type, due to some issue I had that I can't find any notes on.... 

I suppose there's no real reason to not just return the tuple.  I didn't want people accessing this signal and hoping to get any information from it, since I wasn't grouping any metadata with it (what the values correspond to)

ZLLentz (commented):
I like where this is and where it's going- left verbose thoughts. I'll save my other PR review for tomorrow.

pcdsdevices/interface.py
```diff
@@ -1646,6 +1656,77 @@ def transmission(self):
...
+                 input_branches=[], output_branches=[], **kwargs):
+        self._lightpath_values = {}
+        self._lightpath_ready = False
+        self._retry_lightpath = False
+        self.input_branches = input_branches
+        self.output_branches = output_branches
+        # TODO: grab destination / source lists, if they exist?
+        super().__init__(*args, **kwargs)
+        for sig in self.lightpath_cpts:
+            self.lp_summary.add_signal_by_attr_name(sig)
```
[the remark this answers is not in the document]
tangkong: I believe it's just by name ([src](https://github.com/pcdshub/pcdsdevices/blob/06d1fc9d4a71af7b85f6688441995715597a3203/pcdsdevices/signal.py#L369))

pcdsdevices/interface.py
```diff
@@ -1679,3 +1760,11 @@ def _set_lightpath_states(self, lightpath_values):
...
         return dict(in_check=in_check, out_check=out_check,
                     trans_check=trans_check)
+
+    def get_lightpath_status(self):
+        return LightpathState(
+            inserted=self._inserted,
+            removed=self._removed,
+            transmission=self._transmission,
+            output_branch=self._input_branches[0]
+        )
```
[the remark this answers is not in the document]
tangkong: This was more meant to be a reasonable default, which worked for the general case.  I envision these being customized per class, if we do go with this implementation

pcdsdevices/interface.py
```diff
@@ -1646,6 +1656,77 @@ def transmission(self):
...
+
+    def get_lightpath_status(self) -> LightpathState:
+        """
+        Create and return a LightpathState object containing information needed
+        for lightpath
+
+        Device logic goes here
+        """
+        raise NotImplementedError('Did not implement LightpathMixin')
+
```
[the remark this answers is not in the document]
tangkong: > Is that a correct understanding?

Yep, that's exactly what I had in mind.

> splitting get_lightpath_status -> (get_lightpath_status, lightpath_calc)

I like this suggestion, it would make testing easier while also limiting the number of methods users would have to define.  

> making `lp_summary` more useful

I'm wary of bundling too much of the status calculation in the signal itself, since its purpose is more to notify consumers that the signals have changed.  While in the lightpath case, we might want to run the status calculations with every update, I could imagine a case where signals update more frequently than we want to update the gui.  (though I can't imagine the calculations should be that intensive I guess)

I might also be concerned about a single signal containing semi-structured data that needs device knowledge to interpret.  What order do the signal values appear in, do we try to keep that consistent across devices, do users start to use that in other processes, etc.

pcdsdevices/interface.py
```diff
@@ -1561,7 +1563,15 @@ def write(*args, **kwargs):
         pass
 
 
-class LightpathMixin(OphydObject):
+@dataclass
+class LightpathState:
```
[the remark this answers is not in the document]
tangkong: I think you're right.  I can move this to `lightpath`

pcdsdevices/interface.py
klauer: > I thought the mixin was a simple way to wrap up the functionality, which boils down to adding the SummarySignal and requiring the get_lightpath_status method. I thought it was clearer this way, though I will admit that I didn't think much about a signal that sets a method. (Component internals scare me slightly).

I think the internals of component - however scary - aren't so necessary to understand. In this case, looking at it as "Component instantiates a `SummarySignal` that has a `.parent` that leads you back to the relevant Device instance (with the important methods)" is probably good enough?

> Is inheriting from the mixin itself a step that users find onerous?

As much as I tend to (over)use it for its code duplication-reducing aspects, I think multiple inheritance should be avoided wherever possible.

> I guess the other question is: is the proposed API adequate? (Outside of the implementation details)

Given the goals, I think it's sufficient. Including the split of `get` vs `calculate` detailed in other conversations below is important, too, I think.

tangkong: (summarizing external discussions)
After some consideration, I think a mixin is appropriate here.  The key point is that we will be storing branch information (`input_branches` and `output_branches`) externally in the happi database, and requiring them as init kwargs.  In doing so, we need to store that information and use it in the calc_lightpath_state methods; this is a pattern that lends itself to mixins.

tangkong:
I think we've reached a consensus here about the implementation approach, and I've implemented the suggestions noted during this first review round.  

I'm going to start attempting to port over existing lightpath devices to the new interface.  Perhaps more fine tuning will come as a result.

tangkong:
I think I'm going to push for this PR to get merged sooner than later, so we can start resolving merge conflicts and prevent this PR from becoming too lengthy.  Once the corresponding PR from lightpath is merged/tagged, maybe I can get these tests passing and this PR merged.

tangkong:
A rebase happened, apologies for the spam

tangkong:
I've finished applying the lightpath interface to every device marked as `lightpath=True` in the current happi database.  

LightpathMixin has relaxed its requirements slightly, allowing init without branch metadata and deferring implementation checks until `get_lightpath_state()`.  This is because there are some device classes that are both lightpath active and inactive.  Instead of creating an additional mixed-in class each time, I chose to relax the mixin.  

Container-wise, I made the `LCLSLightpathItem` branches optional, since we have several containers that service both lightpath and non-lightpath devices.  This allows the database to pass validation (w.r.t. the lightpath items), with only the lightpath-active devices getting additional metadata

Basically everything is optional now, I guess. 🤷

tangkong:
9e7c57b restores the behavior that delayed lightpath calculations if the state objects were not initialized.  This cleanses a huge amount of errors from the lightpath gui output, so if you're testing with this branch you should pull this update

tangkong:
Some substantial changes this month as we start to wrap up this effort.
- OffsetMirror (cc7713) logic was added, along with needed modifications to the happi container (6a9c290)
  - While the logic of each mirror is similar, but each mirror varies in which components have state positioner components.  To help with this, `calc_lightpath_state` was made to rely on helper methods that can be overloaded in future classes.  
  - Range thresholds are stored in happi as simple arrays.  There is obviously room for confusion here, but I'd love to find a better way without adding huge amounts of init kwargs and happi entries
- LODCM (b798397) logic was added.  LODCM's as currently deployed only have two variants, so separate subclasses were made. 
- LightpathState has been refactored to accept multiple outputs, merging `transmission` and `output_branch` into one field: `output` (changes made in c611124, see https://github.com/pcdshub/lightpath/pull/160)

tangkong:
I'm sorry, I rebased.  This one gets a squash for sure

tangkong:
Tests are passing now, and I'd like to humbly request a re-review.  The main work that was holding this up was updating the relevant devices, but there were also some changes to the `LightpathMixin` itself, mainly in an effort to make sure it was as optional as possible.  

It's a lot to sift through, thanks in advance for your patience with me here

pcdsdevices/mirror.py
```diff
@@ -385,10 +400,216 @@ class XOffsetMirror(BaseInterface, GroupDevice):
...
+            # default case for always-in mirrors
+            return False, True
+
+        if np.shape(self.x_ranges) != (2, 2):
+            # improper ranges for insertion, fail
+            raise MirrorLogicError(
+                'Provided x-ranges are the malformed. '
+                f'got: {np.shape(self.x_ranges)}, expected (2,2)')
+
+        ins_bools = [(limit[0] < x) and (x < limit[1])
```
klauer, line 535: Python syntactic sugar instead, perhaps?
```suggestion
        ins_bools = [limit[0] < x < limit[1]
```

pcdsdevices/mirror.py
```diff
@@ -385,10 +400,216 @@ class XOffsetMirror(BaseInterface, GroupDevice):
...
+
+    def _get_output_branch(self, coating_idx: int, pitch: float) -> str:
+        """
+        get output branch given coating state and pitch
+
+        Parameters
+        ----------
+        coating_idx : int
+            index used to access relevant pitch ranges. First coating
+            is 0-indexed
```
klauer, line 582: Missing `pitch` in docstring

pcdsdevices/mirror.py
```diff
@@ -519,6 +742,60 @@ class XOffsetMirrorBend(XOffsetMirror):
...
+            current lightpath state of the device
+        """
+        try:
+            if np.shape(self.y_ranges) != (2, 2):
+                # improper ranges for insertion, fail
+                raise MirrorLogicError(
+                    'Provided x-ranges are the malformed. '
+                    f'got: {np.shape(self.x_ranges)}, expected (2,2)')
+
+            x_out, x_in = ((limit[0] < y_up) and (y_up < limit[1])
```
klauer, line 776: This same loop is here a few times; you might consider a helper that does the logic and validation:

```python
def find_matching_range_indices(ranges, value) -> List[int]:
  if np.shape...
```

pcdsdevices/spectrometer.py
```diff
@@ -326,8 +341,19 @@ class TMOSpectrometer(BaseInterface, GroupDevice):
     transmission = 1
     SUB_STATE = 'state'
 
+    # dummy signal, state is always the same
+    lightpath_cpts = ['yag_x.user_readback']
 
-class HXRSpectrometer(BaseInterface, GroupDevice):
+    def calc_lightpath_state(self, **kwargs) -> LightpathState:
+        # TODO: get real logic here, instead of legacy hard-coding
```
klauer, line 348: TODO reminder callout

pcdsdevices/mirror.py
```diff
@@ -594,11 +886,18 @@ class KBOMirror(BaseInterface, GroupDevice):
...
-    removed = False
-    transmission = 1
-    SUB_STATE = 'state'
+
+    def calc_lightpath_state(
+        self,
+        x: float,
+        y: float
+    ) -> LightpathState:
+        # TODO: get real logic
```
klauer, line 897: TODO reminder

pcdsdevices/mirror.py
```diff
@@ -763,10 +1062,18 @@ class FFMirror(BaseInterface, GroupDevice):
...
-    SUB_STATE = 'state'
+    lightpath_cpts = ['x.user_readback', 'y.user_readback']
+
+    # Lightpath config: implement inserted, removed, transmission, subscribe
+    def calc_lightpath_state(
+        self,
+        x: float,
+        y: float
+    ) -> LightpathState:
+        # TODO: get real logic
```
klauer, line 1073: TODO reminder

klauer (commented):
I added some code quality/refactor suggestions which are admittedly subjective, so feel free to take them with a grain of salt as usual.

I think this is looking really good as a sweeping first pass. It seems time to give this a real try, and we can revisit as needed.

docs/source/upcoming_release_notes/1028-enh_lp_mixin.rst
```diff
@@ -0,0 +1,42 @@
+1028 enh_lp_mixin
+#################
+
+API Changes
+-----------
+- Converts LightpathMixin to the new Lightpath API, consilodating
```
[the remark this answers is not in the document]
tangkong: 🤦

pcdsdevices/lodcm.py
```diff
@@ -1815,6 +1829,122 @@ def form(left_str, center_str, right_str):
...
+            )
+        # avoid extra get calls in this method
+        state_label = self.tower1.h1n_state.get_state(h1n_state).name
+
+        if state_label in ['C', 'Si']:
+            inserted = True
+            removed = False
+            if state_label == 'C':
+                # Diamond crystal, beam is split between outputs
+                output = {self.output_branches[0]: 0.5,
```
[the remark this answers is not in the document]
tangkong: I think that's reasonable.  I can't think of a good reason not to, at least.

pcdsdevices/spectrometer.py
tangkong: I skipped these devices because they weren't the big branching ones.  I'll make a followup PR when I get that logic sorted out.

tangkong:
@ZLLentz  I don't know if you have time to review this, but I'm hoping to get this in before any more merge conflicts get created.  I think I've addressed most of Ken's points, leaving TO-DO's for later.

ZLLentz (approved):
I think it looks good, I didn't fully evaluate every line because there are so many, but since you've only touched lightpath interfaces the hutch python should continue to work as-is and lightpath should be able to move forward. I suspect we'll have tweaks to make but they won't be obvious until we run with this for a bit.

tangkong:
Ok then, thanks everyone.  Here goes nothing

PR #1030, opened 2022-06-27
MAINT: delay the pint import, it is slow

ZLLentz:
## Description

import `pint` when it is used and no earlier.

## Motivation and Context

importing `pint` can be slow and should be skipped if it is not used.

`pint` is a module that is used in this repository to handle unit conversions.

During the first import of `pint`, it makes several additional imports whether the resources from those imports are used or not. I assume this is done to ensure that those optional dependencies are fully functional when used in conjunction with `pint`, and there is probably some base assumption that these imports are light operations.

The additional imports that pint makes in `pint.compat` on import of `pint` are:
- `pandas`
- `pyarrow` (via pandas)
- `numpy`
- `xarray`
- `dask`
- `scipy` (via dask)
- `distributed` (via dask)
- `fsspec` (via dask)

`pint` source code: https://github.com/hgrecco/pint/blob/30a6dad23f504316f6ede5d50dda97e141519995/pint/compat.py

In total, I observe a speedup from 2.60s to import the mirror classes (which don't use `pint`) to 0.69s, for a savings of about 1.9 seconds.

## How Has This Been Tested?

Interactively only

Not yet

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1030-delay-pint.rst
+++ b/docs/source/upcoming_release_notes/1030-delay-pint.rst
@@ -1,0 +1,31 @@
+1030 delay-pint
+###############
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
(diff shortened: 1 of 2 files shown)


klauer:
Nice

PR #1031, opened 2022-07-08
WIP: performance at init

ZLLentz:
## Description

This is paused briefly while I give some attention to other projects- creating a PR due to the pause.

Overall goals are:
- [x] Do not load any components on init (at least, minimize it)
- [x] Add some tools to avoid common patterns that add init-time steps
- [x] Add some tools to ensure the signals needed for scanning, etc. are can be annotated such that things are set up exactly when they are needed.
- [ ] Investigate how much background loading is feasible given threading constraints and race condition concerns

This was written with some accompanying ophyd changes to skip the init-time getattr on all components rather than using the lazy mechanisms but I might switch back to a setup that doesn't need an ophyd edit yet.

## Motivation and Context

- Startup of python envs is slow.
- Using the lazy mechanisms can help but adds some complications with choosing when to do the loading.
- Some devices may want to be lazy only sometimes, in certain applications, but not in others... making the per-component kwarg potentially insufficient.

## How Has This Been Tested?

- Made the tests pass again
- Interactively
- Need to add tests for the new component types for sure

WIP

## Pre-merge checklist


ZLLentz:
The most interesting parts of this are probably the `AliasComponent` which stops us from needing to getattr components during init to alias them and the `ParentComponent` which allows the lodcm subdevices to reference each other without duplicating all the devices and PVs (while still allowing these devices to work standalone if needed)

ZLLentz:
Next steps here:
- Strip out the lazy stuff for a follow-up PR- this is big enough without that.
- Add tests for `AliasComponent` and `ParentComponent`
- Find other targets for `AliasComponent` e.g. in the slits classes

ZLLentz:
I'll fix the minor merge conflict later, but the tests passed locally on Friday and this could use some opinions I think

pcdsdevices/interface.py
```diff
@@ -217,9 +218,162 @@ def __init_subclass__(cls, **kwargs):
...
+                raise exceptions[0]
+            msg = ", ".join(
+                "{} ({})".format(ex.attr, ex.__class__.__name__)
+                for ex in exceptions
+            )
+            raise ExceptionBundle(
+                "Failed to disconnect all signals ({})".format(msg),
+                exceptions=exceptions,
+            )
+        self.unsubscribe_all()
```
tangkong, line 375: I know this is how it is in ophyd.Device (`super().destroy()` -> `OphydObject.destroy()` -> `self.unsubscribe_all()`), but I found it interesting that if an exception is raised the unsubscribe_all call would get missed

pcdsdevices/interface.py
```diff
@@ -217,9 +218,162 @@ def __init_subclass__(cls, **kwargs):
...
+        """
+        Disconnect and destroy all loaded signals and subdevices.
+
+        This is modified from Device.destroy() to consider that some
+        non-lazy signals may not have been created yet, else we risk
+        creating them during the destroy process.
+        """
+        self._destroyed = True
+        exceptions = []
+        for attr, obj in self._signals.items():
```
tangkong, line 356: I assume we don't use `walk_signals` because something implementing our `BaseInterface` might not be a `Device`, but if it isn't will it have `._signals`?

pcdsdevices/interface.py
```diff
@@ -405,6 +561,59 @@ def screen(self):
...
+    def describe_configuration(self):
+        self.activate()
+        return super().describe_configuration()
+
+    def trigger(self):
+        self.activate()
+        return super().trigger()
+
+
+def needs_activate(func):
```
tangkong, line 594: I do like me some decorators.  👍

tangkong (commented):
I think this looks pretty good.  I left a few questions/comments, but on the whole I like it. 

I spent some time with it, comparing hutch-python and lightpath load times with and without this patch.  

Hutch-python loaded its happi database in ~20s pre-patch, and ~5s post-patch.
Lightpath showed no real performance gain, which I blamed on the fact that we do access components pretty soon after loading a device.  Though we should only be loading a few components.  I'll have to look into this more and see if we're not accidentally `activate`-ing the whole device.

ZLLentz:
> Lightpath showed no real performance gain,

Maybe this is related to the typhos integration? I wonder if typhos is hitting all the components

pcdsdevices/interface.py
```diff
@@ -217,9 +218,162 @@ def __init_subclass__(cls, **kwargs):
...
+            if isinstance(cpt, str):
+                cpt: Component = self._sig_attrs[cpt]
+            if any((
+                cpt._subscriptions and ensure_subs,
+                include_lazy,
+                not cpt.lazy,
+            )):
+                objs.append(getattr(self, cpt.attr))
+        if activation_call:
+            self._active_flag = True
```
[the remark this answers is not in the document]
ZLLentz: It's a bit messy exactly as you've described. The design goal here was to never repeat the "full device activate" sort of thing, but I didn't want to add overhead where I try to figure out which signals still need to be checked... maybe it could be added with none/minimal slowdown.

pcdsdevices/interface.py
```diff
@@ -217,9 +218,162 @@ def __init_subclass__(cls, **kwargs):
...
+        if cpts is None:
+            activation_call = True
+            try:
+                cpts: Iterable[Component] = self._sig_attrs.values()
+            except AttributeError:
+                # No _sig_attrs means not a Device, meaning no cpts
+                self._active_flag = True
+                return
+            if wait_connected and timeout is None:
+                timeout = max(1.0, 0.005 * len(list(self.walk_components())))
```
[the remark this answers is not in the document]
ZLLentz: You're 100% correct, it needs to be configurable

pcdsdevices/interface.py
ZLLentz: You're 100% right. Let's optimize a bit further because we can and clean up as much as possible.

pcdsdevices/interface.py
ZLLentz: The specific reason is that `walk_signals` will instantiate any components that the session has skipped, since it doesn't know what shenanigans I'm up to here, and I didn't want to create components on `destroy`. I also didn't want to change the `walk_signals` logic because I thought that the signal creation would theoretically be a desired behavior in this case.

ZLLentz:
Another consideration is `wait_for_connection`, if that's being called at all in `lightpath`

ZLLentz:
> Things like typhos that inspect every component won't get any benefit, as Robert found (so this is largely hutch-python specific

Yeah, if you need all the signals... then delaying the signal startup doesn't have much merit. I'm banking on most applications not needing all the signals.

> BaseInterface needs to come first in the base class listing, right? Perhaps we should add that check to the test suite if it's not already there.

This check already exists on `BaseInterface` itself

> I can't say I fully grasp the implications of circumventing/patching/replacing...

The original attempt at this was to simply make everything lazy which would be a lot safer, but there are special exceptions in the component instantiation in ophyd for components that use the subscriptions decorator... So basically, anything that makes a subscription on startup would still happen, and my brain said "Unacceptable! Unused background subscriptions are a big offender for performance of big sessions!". I'm not sure if I made the right decisions here and it would not be difficult to persuade me that a simple "for cpt in cpts cpt.lazy=True" would be better.

> For things which access components that don't have ensure_cpts/needs_activate wrapping them (interactive usage falls into this category) we could incur a significant time penalty due to serial instantiation/connection

I still sort of want to try to finish the concept in https://github.com/ZLLentz/pcdsdevices/tree/enh_inspect_experiments as a potential auto-decorator but this also adds some complexity.

klauer:
> I'm not sure if I made the right decisions here and it would not be difficult to persuade me that a simple "for cpt in cpts cpt.lazy=True" would be better.

"Better" in terms of complexity - 100%! Better in terms of everything else? A mixed bag...

> Also, a blanket `needs_activate` on all "public" methods would be another way to help avoid issues with post-startup... maybe mass lazy + mass needs_activate is the sanest attempt here? With some way to opt out of activation for specific methods? (or maybe, we just skip the activate decorator for things that have a subs decorator or a more specific needs_cpts decorator)

That seems worth investigating in parallel with this effort. It's just too difficult to foresee how it would work out otherwise.

ZLLentz:
It's certainly worth a try- putting all the code into `__init_subclass__` basically and adding some liberal usage of the `do_not_wait_for_lazy_connection` decorator, instead of hitting core ophyd methods.

I was also worried about the complexity... I would be so sad if we saved a bunch of startup time for like a month and then lost a bunch of dev time for years due to confusing debugging sessions.

ZLLentz:
I'm going to close this "paused briefly" out of date PR.
I will keep the branch alive so the ideas could be revisited in the future.

PR #1034, opened 2022-07-11
add PID kill at the end of CCM motion

vespos:
## Description
Kill CCM's PID loop at the end of a move command

## Motivation and Context
The PID loop keeps looking for the final position forever. The motor thus runs at too high duty cycles, heats up and causes vacuum spikes in the chamber. This has led to MPS trips. In addition, there is serious potential to fry the motor itself, as it is run too intensively.

## How Has This Been Tested?
Currently running at XPP for CCM experiments. Still waiting on confirmation that this all works fine before finalizing this pull request

Docstring of the new move function.

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1034-Kill_ccm_PID_on_move.rst
+++ b/docs/source/upcoming_release_notes/1034-Kill_ccm_PID_on_move.rst
@@ -1,0 +1,31 @@
+1034 Kill ccm PID on move
+#################
+
+API Changes
+-----------
+Overwrite default move method for the CCMEnergy class to kill the PID loop at the end of each move (default)
+This should prevent the piezo motor from heating up and breaking vacuum or frying itself.
+
+Features
+--------
+- N/A
+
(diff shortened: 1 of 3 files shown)


tangkong (commented):
This looks good to me, and is verbatim what you had implemented / discussed with zach and I.  I assume this works locally, but want to verify that this is a fix that should be applied to all CCM's (which seem to be only `xpp_ccm` and `xcs_ccm`)

tangkong:
Tests are complaining now that we're trying to actually interact with the fake pid in the test suite as well

vespos:
@tangkong The past couple beamtimes did not have any further issues with the CCM. @tangkong  can you please review / approve this again?
I am adding the pre-release doc in an upcoming commit.

tangkong (approved):
Looks good to me, and if this works in production I don't have much to complain about.  We should be ok to merge after pre-release notes

PR #1038, opened 2022-07-14
ENH: USB encoder

vespos:
## Description
Add USB encoder class to pcdsdevices. 

## Motivation and Context
This will allow to remove it from the beamline.py files and ensure better uniformity in handling USB encoder devices. A set_zero method has also been added to the DelayBase class. This will allow reset of the delay to the time zero to use the same command as the encoder.
The USB encoder class can take a linked stage and both can be set to 0 using the `set_zero_with_stage` method.

## How Has This Been Tested?
Tested in xpp3. Behaves as expected.

Docstring

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/pseudopos.py
+++ b/pcdsdevices/pseudopos.py
@@ -798,13 +798,6 @@ class DelayBase(FltMvInterface, PseudoPositioner):
         new_offset = position - self.position[0]
         self.user_offset.put(new_offset)
 
-    def set_zero(self):
-        """
-        Set current position to 0
-        """
-        self.set_current_position(0)
-        return
-
     def format_status_info(self, status_info):
         """
         Use the renderer from the subdevice
--- a/pcdsdevices/pseudopos.py
+++ b/pcdsdevices/pseudopos.py
@@ -785,7 +785,7 @@ class DelayBase(FltMvInterface, PseudoPositioner):
         return self.PseudoPosition(delay=delay_value + self.user_offset.get())
 
     def set_current_position(self, position):
-        '''
+        """
         Calculate and configure the user_offset value, indicating the provided
         ``position`` as the new current position.
 
@@ -793,10 +793,17 @@ class DelayBase(FltMvInterface, PseudoPositioner):
         ----------
         position
             The new current position.
-        '''
+        """
         self.user_offset.put(0.0)
         new_offset = position - self.position[0]
         self.user_offset.put(new_offset)
+
+    def set_zero(self):
+        """
+        Set current position to 0
+        """
+        self.set_current_position(0)
+        return
 
     def format_status_info(self, status_info):
         """
--- a/pcdsdevices/usb_encoder.py
+++ b/pcdsdevices/usb_encoder.py
@@ -8,7 +8,7 @@ from ophyd import Device, EpicsSignal
 from .interface import BaseInterface
 
 
-class USBEncoder(BaseInterface, Device):
+class UsDigitalUsbEncoder(BaseInterface, Device):
     """
     Class for the USB encoder by US Digital.
     https://www.usdigital.com/products/accessories/interfaces/usb/usb4
--- a/pcdsdevices/usb_encoder.py
+++ b/pcdsdevices/usb_encoder.py
@@ -25,9 +25,11 @@ class USBEncoder(BaseInterface, Device):
     scale = Cpt(EpicsSignal, ':SCALE', kind='config')
     offset = Cpt(EpicsSignal, ':OFFSET', kind='config')
 
-    tab_whitelist = ['pos', 'set_zero', 'scale', 'offset']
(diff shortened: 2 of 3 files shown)


pcdsdevices/usb_encoder.py
```diff
@@ -27,6 +27,26 @@ class USBEncoder(BaseInterface, Device):
 
     tab_whitelist = ['pos', 'set_zero', 'scale', 'offset']
 
+    def __init__(self, stage=None):
+        self.stage = stage
```
klauer, line 31: A few things here:
* `stage` is a reserved term in bluesky - you're overriding an existing method by doing this
* `__init__` arguments aren't correct for a `Device`:
    * You'll need to have `prefix` and the usual `__init__` arguments
    * You'll also need to call `super().__init__(args_and_kwargs_here)` in the body

pcdsdevices/usb_encoder.py
```diff
@@ -0,0 +1,52 @@
...
+
+from ophyd import Component as Cpt
+from ophyd import Device, EpicsSignal
+
+from .interface import BaseInterface
+
+
+class USBEncoder(BaseInterface, Device):
+    """
+    USB encoder class
```
klauer, line 13: Can we be more specific: which USB encoder? Which model or which IOC support provides these PVs?

pcdsdevices/usb_encoder.py
vespos: Hi Ken, the code was not quite yet ready for review, sorry about that.
I am going to replace the `stage` by something else, thanks for letting me know

klauer: No problem - I saw you requesting reviews so assumed otherwise :)

pcdsdevices/usb_encoder.py
vespos: I ve added some info and a url

vespos:
Now that we have a method to set 0 on both the encoder and delay stage at the same time, we probably dont need the set_zero method in the DelayBase. Opinions?

pcdsdevices/usb_encoder.py
```diff
@@ -0,0 +1,64 @@
...
+        self.zero.put(1)
+
+    def set_zero_with_axis(self):
+        """
+        Resets the encoder counts to 0 and sets the stage offset
+        so that its position is also 0.
+        """
+        self.zero.put(1)
+        if (
+            self.stage is not None or
```
ZLLentz, line 50: This will always evaluate to `True` because `stage` is a method inherited from `Device`

pcdsdevices/usb_encoder.py
klauer: Great - I think then I'd recommend that this be called something like: `UsDigitalUsbEncoder`
We might not ever get another USB encoder reader, but if we do, it'd be better not have to rename this one.

pcdsdevices/pseudopos.py
```diff
@@ -785,19 +785,26 @@ def inverse(self, real_pos):
         return self.PseudoPosition(delay=delay_value + self.user_offset.get())
 
     def set_current_position(self, position):
```
klauer, line 787: What if instead of the new method, we changed this to:
```suggestion
    def set_current_position(self, position: float = 0.0):
```

Just a thought - I also feel like the `set_zero` method is unnecessary.

vespos: I'd rather have the set_current_position not have a default value. This could lead to bad surprises with the motor position is someones runs add accidentally without a positional argument.

I'll remove the set_zero position for now, it's easy enough to add in the future if we feel we need it

pcdsdevices/usb_encoder.py
vespos: Good catch, fixing it in the next commit

pcdsdevices/usb_encoder.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def set_zero_with_axis(self):
+        """
+        Resets the encoder counts to 0 and sets the linked_axis offset
+        so that its position is also 0.
+        """
+        self.zero.put(1)
+        if (
+            self.linked_axis is not None or
+            hasattr(self.linked_axis, 'set_current_position')
```
ZLLentz, line 51: I think the logic here is still subtly off. Should this be `and`, or else any non-`None` object will pass as `True`? Or should the `is not None` check go away entirely since it's redundant with the `hasattr` check? (`None` won't pass the `hasattr`). Or should the `hasattr` check go away so the user is quickly made aware that they've passed the wrong (some non-motor) device into the class?

vespos: Indeed, the `hasattr` should be enough here. I really just wanna make sure that people get warned if the axis they pass does not have a `set_current_position` method (which I am not even sure every motor has?)
Will fix that in yet another commit!

ZLLentz (approved):
Awesome let's merge now

Issue #1039, opened 2022-07-14
LED / Light control class is in cvmi_motion.py, rather than its own module


wwright-slac:
CvmiLed should be its own module, as there are use cases for controlling LEDs or Fiber-Lites outside of CVMI.
## Expected Behavior
A separate module for light control.

## Current Behavior
CvmiLedinside cvmi_motion.py allows control of LED lights. If someone wants to control an LED through this functionality, they would have to import cvmi_motion and use CvmiLed regardless of location or endstation.

## Context
I want to be able to use a typhos screen to control fiber-lites in RIX.

wwright-slac:
This was completed in https://github.com/pcdshub/pcdsdevices/pull/1048 and https://github.com/pcdshub/pcdsdevices/pull/1040

PR #1017, opened 2022-05-31
Addition of LED PVs to CMVI Motion

Mbosum:
## Description

I added class variables to CVMI for LED control PVs. These include  control and readback PVs for LED rings attached to the viewports on the vacuum chamber. These include illumination percentage, power ON/OFF, and a naming string field.

## Motivation and Context

It enables remote control of LED illumination on CVMI either through hutch python or the CVMI motion typhos screen.

## How Has This Been Tested?

1.  ipython session, loading the class

Screenshot (798)

2. flake8

Screenshot (799)

3. test typhos screen 

Screenshot (800)

## Where Has This Been Documented?

1. Class docstring
2. Confluence page (work in progress): https://confluence.slac.stanford.edu/display/L2SI/IP1+LED+controls

## Pre-merge checklist
- [ ] Code works interactively
- [ ] Code contains descriptive docstrings, including context and API
- [ ] New/changed functions and methods are covered in the test suite where possible
- [ ] Test suite passes locally
- [ ] Test suite passes on travis
- [ ] Ran docs/pre-release-notes.sh and created a pre-release documentation page
- [ ] Pre-release docs include context, functional descriptions, and contributors as appropriate

PR #1040, opened 2022-07-14
Adding separate module for LED (and other light devices) control

wwright-slac:
## Description
Added a separate module, light_control.py, for devices for controlling light brightness. Removed internal module CvmiLed from cvmi_motion.py

## Motivation and Context
https://github.com/pcdshub/pcdsdevices/issues/1039

## How Has This Been Tested?
image

image

A lot of the testing was previously done by @Mbosum here: https://github.com/pcdshub/pcdsdevices/pull/1017

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1039-Create_separate_module_for_light_control.rst
+++ b/docs/source/upcoming_release_notes/1039-Create_separate_module_for_light_control.rst
@@ -29,4 +29,3 @@ Maintenance
 Contributors
 ------------
 - wwright-slac
-- klauer
--- a/docs/source/upcoming_release_notes/1039-Create_separate_module_for_light_control.rst
+++ b/docs/source/upcoming_release_notes/1039-Create_separate_module_for_light_control.rst
@@ -1,0 +1,32 @@
+1039 Create separate module for light control
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- New module for controlling intensity of LEDs or Fiber-Lites, light_control.py
+- Removing CvmiLed from cvmi_motion.py
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- wwright-slac
+- klauer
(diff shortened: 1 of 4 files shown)


pcdsdevices/cvmi_motion.py
```diff
@@ -6,23 +6,10 @@
...
-    Basic control and readback PVs for LED rings attached to the viewports
-    on the vacuum chamber. These include illumination percentage,
-    power ON/OFF, and a naming string field.
-    """
-
-    # LED variables
-    desc = Cpt(PytmcSignal, ':NAME', io='io', kind='normal', string=True)
-    pct = Cpt(PytmcSignal, ':ILL:PCT', kind='normal', io='io')
-    pwr = Cpt(PytmcSignal, ':PWR', kind='normal', io='i')
+from .light_control import LightControl
```
ZLLentz, line 13: Typically this would be considered an API break- but in practice I suspect nobody has imported this name yet. If we were being super pedantic we'd probably want to add a backwards-compatibility line:
```
CvmiLED = LightControl
```

ZLLentz (approved):
This is well-reasoned and looks good. I have no major suggestions and would be happy to merge. The name of the module (light control) is reasonable.

There are a few other references to LEDs in pcdsdevices but those can be refactored later if it makes sense.

docs/source/upcoming_release_notes/1039-Create_separate_module_for_light_control.rst
```diff
@@ -0,0 +1,32 @@
...
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- wwright-slac
+- klauer
```
klauer, line 32:
```suggestion
```
This one was all you, @wwright-slac!

klauer (approved):

pcdsdevices/cvmi_motion.py
wwright-slac: Good idea. Can this go anywhere or should it go in the CVMI class?

wwright-slac: Good idea. Can this go anywhere or should it go in the CVMI class?

ZLLentz: You'd typically want to put it in the same place that the original class was.

The goal would be to make sure that a user who had code structured like:
`from pcdsdevices.cvmi_motion import CvmiLED` would still have their code work after the rename.

I think we can skip it here because it's not really something anyone has imported on their own, but it could also be added.

PR #1045, opened 2022-08-17
WIP: limits handling robustness for poorly behaving IOC/gateway/unknown

ZLLentz:
## Description

- [x] Relocate some of the limit handling overrides from `Newport` to `EpicsMotorInterface` to make them more widely available
- [x] Allow `UpdateComponent` to use the subscription decorators
- [ ] Add some unit tests for limit handling that don't assume well-formed EPICS metadata on `user_setpoint`

## Motivation and Context

We've observed that the majority of motor records in use at LCLS don't properly update the control limit metadata on their `user_setpoint` signals when `HLM` and `LLM` update. At some point I thought this was just the Newports but I'm annoyed and don't want to mess around with differentiating which are behaving normally and which are not. It's not immediately clear to me if these are driver issues or  gateway issues, but rather than muck about in that space I'm opting to use the `HLM` and `LLM` fields directly instead, which I know is more reliable at LCLS. The affected IOCs internally seem to ignore their stated ctrl limits in favor of the values stored in `HLM` and `LLM`.

The `EpicsMotor` class as implemented in `ophyd` should work perfectly fine if the motor record it's connected to updates its limit metadata appropriately.

The dangling repeat of the `_pos_updated` subscription can be removed when we use `UpCpt` since the subscription reference is copied from the original component.

Ref https://jira.slac.stanford.edu/browse/LCLSPC-457

## How Has This Been Tested?

- [x] Interactively with `simioc`
- [ ] Unit tests
- [ ] Usage tests on live hardware

WIP

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/device.py
+++ b/pcdsdevices/device.py
@@ -358,6 +358,19 @@ class UpdateComponent(Component):
         # Use our hostage component from __set_name__
         return self.copy_cpt.create_component(instance)
 
+    # Defer to copy_cpt for subscription handling
+    def subscriptions(self, event_type):
+        return self.copy_cpt.subscriptions(event_type)
+
+    def sub_default(self, func):
+        return self.copy_cpt.sub_default(func)
+
+    def sub_meta(self, func):
+        return self.copy_cpt.sub_meta(func)
+
+    def sub_value(self, func):
+        return self.copy_cpt.sub_value(func)
+
 
 class GroupDevice(Device):
     """
--- a/pcdsdevices/device.py
+++ b/pcdsdevices/device.py
@@ -1,3 +1,4 @@
+import collections
 import copy
 from collections.abc import Iterator
 from typing import Any, Optional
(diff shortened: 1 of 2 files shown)


klauer:
So digging back into ophyd as a refresher: https://github.com/bluesky/ophyd/blob/384d1a96c2a8956b5c3b3df62c7d72379485ba5a/ophyd/epics_motor.py#L87-L104
It says: "Monitor `LLM`/`HLM`: when either updates, query and use the motor record's `.VAL` limits as the new low/high limit settings". That does seem a bit of an odd choice, seeing your solution here.

I wonder if these two solutions may conflict a bit though? Perhaps it's worth fixing upstream.

ZLLentz:
I think `ophyd` is implemented that way because ultimately the `move` actions need to go through the `limits` on the `EpicsSignal` itself anyway, which are sourced from the low/high control limit metadata, hence the need for this PR to dip into the EditMD signal genre so that we can get proper feedback on the signal itself.

A refactor at `ophyd` would probably disable limit checking on `user_setpoint` and handle it in the `EpicsMotor` class directly, which might be a preferable solution, I'm not sure.

It's a hard sell to say "let's redo this device because some motor records are implemented wrong"

klauer:
> I think `ophyd` is implemented that way because ultimately the `move` actions need to go through the `limits` on the `EpicsSignal` itself anyway, which are sourced from the low/high control limit metadata, hence the need for this PR to dip into the EditMD signal genre so that we can get proper feedback on the signal itself.

You're right - I should have realized that prior to commenting.

I still think `ophyd.EpicsMotor` subscribing and hooking on `LLM`/`HLM` updates along with our code doing the same-but-different is at best a duplication of efforts, and at worst something that could race to setting the wrong value. It's something that makes me a bit uncomfortable leaving it up to "however ophyd decides to run subscription callbacks" (which - who knows? - may change in v2)

ZLLentz:
Oh you're absolutely right, it feels iffy.

The md override signal basically ignores all "normal" metadata updates and masks them with the override, which should make it work OK.

Perhaps the `ophyd` change needed is to set up the subscription using the decorator handlers instead of doing it in `__init__` so that we can override it in a subclass.

ZLLentz:
This seems to work, let's merge it and see if complaints about the desync disappear or not. We can always improve the implementation later.

klauer (approved):

tangkong (approved):
Sounds good.  🥇 🧑‍🚀

PR #1046, opened 2022-08-18
ENH: make PVStateSignal report enum_strs and write_access metadata

ZLLentz:
## Description

`PVStateSignal` has been given `enum_strs` and `write_access` metadata that behaves similarly to how the `EpicsSignal` metadata does.

On the first value calculation, which occurs only after the device is fully ready to go, we check the parent states device for its static state enum to use to collect the enum strings and for a setpoint reference that we can subscribe to for write access metadata updates, forwarding these in the `PVStateSignal` as appropriate.

## Motivation and Context

We noticed in the lightpath that any devices with `PVStateSignal` behaved inappropriately at the pcdsdevices level, creating what naively looks like ui bugs in `lightpath`, while actually being sourced to here.

This signal type is used in the legacy stoppers, gate valves, pulse pickers, and movable stands.

## How Has This Been Tested?

Interactively by loading happi devices and via checking the lightpath.
Unit test added.

I need to make the pre-release docs.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/signal.py
+++ b/pcdsdevices/signal.py
@@ -17,7 +17,7 @@ import logging
 import numbers
 import typing
 from threading import RLock
-from typing import Dict, Generator, List, Mapping, Optional, Union
+from typing import Any, Dict, Generator, List, Mapping, Optional, Tuple, Union
 
 import numpy as np
 import ophyd
@@ -439,10 +439,16 @@ class PVStateSignal(AggregateSignal):
         self._has_setpoint_md = False
 
     @property
-    def enum_strs(self):
+    def enum_strs(self) -> Tuple[str, ...]:
+        """
+        Mimic the epics signal property for enum strings.
+        """
         return tuple(state.name for state in self.parent.states_enum)
 
-    def describe(self):
+    def describe(self) -> Dict[str: Dict[str: Any]]:
+        """
+        Make sure a reasonable description exists for bluesky scans.
+        """
         # Base description information
         sub_sigs = [sig.name for sig in self._signals]
         desc = {
@@ -453,7 +459,18 @@ class PVStateSignal(AggregateSignal):
         }
         return {self.name: desc}
 
-    def _calc_readback(self):
+    def _calc_readback(self) -> str:
+        """
+        Implements the PVStatePositioner logic.
+
+        This relies on this signal being a component of a PVStatePositioner
+        class with properly-defined class attributes.
+
+        On the first call, we'll do some setup of metadata values, since
+        this first call happens after the signals connect and we're guaranteed
+        that everything is ready. Doing it earlier can run into some
+        race conditions.
+        """
         # Do some one-time setup here
         # Convenient because we only hit this block when signals are ready
         if (
@@ -505,12 +522,27 @@ class PVStateSignal(AggregateSignal):
         # If all states deferred, report as unknown
         return state_value or self.parent._unknown
 
-    def _setpoint_md_update(self, *args, write_access=None, **kwargs):
+    def _setpoint_md_update(
+        self,
+        *args,
+        write_access: Optional[bool] = None,
+        **kwargs
+    ) -> None:
+        """
+        Metadata callback for us to keep track of write access permissions.
+
+        This metadata is used by ophyd and typhos to help give us better
+        feedback about write access without needing to try to put a value
+        first.
+        """
         if write_access is not None:
             self._metadata['write_access'] = write_access
             self._run_metadata_callbacks()
 
-    def put(self, value, **kwargs):
+    def put(self, value: Union[int, str], **kwargs) -> None:
+        """
+        Redirect puts to moving the parent PVStatePositioner device.
+        """
         self.parent.move(value, **kwargs)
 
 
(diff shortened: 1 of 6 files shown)


pcdsdevices/signal.py
```diff
@@ -428,23 +428,72 @@ class PVStateSignal(AggregateSignal):
...
+
+        This relies on this signal being a component of a PVStatePositioner
+        class with properly-defined class attributes.
+
+        On the first call, we'll do some setup of metadata values, since
+        this first call happens after the signals connect and we're guaranteed
+        that everything is ready. Doing it earlier can run into some
+        race conditions.
+        """
+        # Do some one-time setup here
```
ZLLentz, line 474: I'm not 100% sure if this was the right place to do this

pcdsdevices/signal.py
```diff
@@ -428,23 +428,72 @@ class PVStateSignal(AggregateSignal):
...
+            # Not necessarily available during init though, so do it now
+            self._metadata['enum_strs'] = self.enum_strs
+            self._run_metadata_callbacks()
+        if not self._has_setpoint_md:
+            # One time setup of the setpoint signal's metadata
+            # We need this to apply e.g. write permissions
+            ref_str = self.parent._state_logic_set_ref
+            if isinstance(ref_str, str):
+                setpoint_sig = getattr(self.parent, ref_str)
+                setpoint_sig.subscribe(
```
klauer, line 491: Should stash the subscription identifier and unsubscribe on `destroy()`

klauer (approved):
This all looks good to me 👍

pcdsdevices/signal.py
klauer: The one-time setup in the first callback seems appropriate and efficient to me.

The only thing worth noting is that `_calc_readback` happens inside of an acquired [RLock](https://github.com/klauer/pcdsdevices/blob/408b62f2baf6b8d735609a018629cb7d143cc07f/pcdsdevices/signal.py#L197-L216) . I see nothing in the block that points to a potential deadlock condition though: running subscriptions happens in a separate thread and shouldn't block.

pcdsdevices/signal.py
ZLLentz: I'll make sure to do this

ZLLentz:
Marking as draft again so I don't accidentally hit the green button before resolving the cleanup

pcdsdevices/signal.py
ZLLentz: Since it is in an `RLock` though I might take a moment to see if I can set it up earlier in the chain, both to avoid potential mishaps and to not set a bad example for future people

pcdsdevices/state.py
```diff
@@ -362,6 +362,10 @@ class PVStatePositioner(StatePositioner):
         state. You can set this to 'FIRST' instead to use the first state
         found while traversing the `_state_logic` tree. This means an earlier
         state definition can mask a later state definition.
+
+    _state_logic_set_ref : str or None
+        An optional reference to the component that will be used to set some
+        metadata on the state signal if provided.
```
tangkong, line 368: So this is something that will have to be set on any future `PVStatePositioner` devices if we want the metadata to carry forward correctly.  Is there somewhere we can broadcast that information?  I don't see natural place in the docs for it yet.  

I also can't imagine there's a good way to dynamically guess at the component name without the creator's help.  Throwing an error in its absence might be a bit heavy handed, since it's only a gui-related issue.

tangkong (approved):
This also looks good to me.  Works well in practice (solving my lightpath issues).

pcdsdevices/state.py
ZLLentz: I've actually intentionally omitted it in the `MovableStand` class since the state there isn't settable https://github.com/pcdshub/pcdsdevices/blob/613893f5f709d4ba9066a5054004aa10361d1a4b/pcdsdevices/movablestand.py#L32

One thing I was considering was switching this to a required attribute, defaults to `None`, needs to become a `list`. So then we know that you didn't include it if it's still at None, we know that you meant "this is never settable" if it's an empty list, and if there are multiple set PVs we can summarize their write state together. Maybe it's worth doing.

ZLLentz: And of course, then we could raise an exception during init if it's left at `None`

ZLLentz:
This actually isn't strictly needed for lightpath any more after the full sum of changes leading up to v1.0.0, but I'll still try to get this done for completeness's sake as it could be useful for something else in the future.

pcdsdevices/signal.py
ZLLentz: Apparently `unsubscribe_all()` is called on all signals during `destroy()` whether or not this is handled here.
https://github.com/bluesky/ophyd/blob/0e8acb3df3d17e0ab7aa2cc924831f4a0c580449/ophyd/device.py#L1090-L1093
https://github.com/bluesky/ophyd/blob/0e8acb3df3d17e0ab7aa2cc924831f4a0c580449/ophyd/ophydobj.py#L345-L347

pcdsdevices/signal.py
ZLLentz: Can't quickly find a better way, I'm going to move on

ZLLentz:
old pr, new tests, still passing, gonna merge

Issue #1050, opened 2022-09-07
Intermittent test failure: test_presets (`pcdsdevices/tests/test_interface.py::test_presets `)


klauer:
## Current Behavior
See: https://app.travis-ci.com/github/pcdshub/pcdsdevices/jobs/582139962#L2803
This failed for me during an unrelated change for the BTMS PR, passed when I re-ran it, and finally failed after merging it into master.

```python
    @pytest.mark.skipif(
        sys.platform in ("win32", "darwin"),
        reason="Fails on Windows, no fcntl and different signal handling",
    )
    def test_presets(presets, fast_motor):
        logger.debug('test_presets')
    
        fast_motor.mv(4, wait=True)
        fast_motor.presets.add_hutch('four', comment='four!')
    
        fast_motor.mv(3, wait=True)
        fast_motor.presets.add_hutch('zero', 0, comment='center')
        fast_motor.presets.add_here_user('sample')
        print(fast_motor.presets.positions)
        assert fast_motor.wm_zero() == -3
        assert fast_motor.wm_sample() == 0
        assert fast_motor.wm_four() == 1
        assert fast_motor.presets.state() == 'sample'
    
        # Clear paths, refresh, should still exist
        old_paths = fast_motor.presets._paths
        setup_preset_paths()
        assert not hasattr(fast_motor, 'wm_zero')
        setup_preset_paths(**old_paths)
        assert fast_motor.wm_zero() == -3
        assert fast_motor.wm_sample() == 0
        assert fast_motor.presets.state() == 'sample'
    
        fast_motor.mv_zero(wait=True)
        assert fast_motor.presets.state() == 'zero'
        fast_motor.mvr(1, wait=True)
        assert fast_motor.wm_zero() == -1
        assert fast_motor.wm() == 1
        assert fast_motor.presets.state() == 'Unknown'
    
        # Sleep for one so we don't override old history
        time.sleep(1)
        fast_motor.presets.positions.zero.update_pos(comment='hats')
        assert fast_motor.wm_zero() == 0
        assert fast_motor.presets.positions.zero.pos == 1
    
        assert len(fast_motor.presets.positions.zero.history) == 2
        assert len(fast_motor.presets.positions.sample.history) == 1
    
        repr(fast_motor.presets.positions.zero)
        fast_motor.presets.positions.zero.deactivate()
    
        with pytest.raises(AttributeError):
            fast_motor.wm_zero()
    
        with pytest.raises(AttributeError):
[...]

PR #1053, opened 2022-09-15
Liquid jet

vespos:
## Description
Add liquid_jet class for MFX and XCS liquid jet setups.

## Motivation and Context
Devices used regularly at MFX, XCS, and XPP.

## How Has This Been Tested?
Used at MFX to generate the screens based on the happi entry pointing to this class.

Docstrings

## Pre-merge checklist


pcdsdevices/liquid_jet.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class LiquidJet(BaseInterface, GroupDevice):
+    """
+    Liquid jet motion class.
+
+    This class controls the axis for the liquid jet setups
+    """
+    tab_component_names = True
+    #_icon = 'fa.minus-square'
```
tangkong, line 15:
```suggestion
    # _icon = 'fa.minus-square'
```
Style dictates we put a space after the comment '#'.  Did you want this to have an icon though?

pcdsdevices/liquid_jet.py
```diff
@@ -0,0 +1,30 @@
...
+    #_icon = 'fa.clock-o'
```
tangkong, line 16:
```suggestion
    # _icon = 'fa.clock-o'
```
same as above.  Though if we have one we should delete the other

tangkong (changes requested):
The test suite failed on a test that's seemingly unrelated to this.  looking into it now, but in the meantime there's a couple quick style changes that'd be nice to get in.

ZLLentz:
reference to the flaky test issue, must be unrelated to this PR: https://github.com/pcdshub/pcdsdevices/issues/1050

ZLLentz:
I think we can merge if you uncomment, remove, or fix the style thing on the lines that Robert pointed out above.

Currently this icon string is used in lightpath to pick an icon.

vespos:
Hi thanks both for your feedback. I actually realized that we already have a jet class (jet.py). I'd have to check that it works fine with the newer version of the jet that we have at MFX. Or we might consider updating the old XCS liquid jet to a newer version compatible with the current Beckhoff axis class.

tangkong:
Ah!  You're absolutely right.  It looks like the `pcdsdevices.jet.BeckhoffJet` class should reproduce what you've written in this PR, unless the `BeckhoffAxis_Pre140` is out of date.  A quick look through happi seems to suggest that there aren't any devices using `BeckhoffJet` right now, so updating the `BeckhoffAxis_Pre140` components could be an un-disruptive option.

PR #1056, opened 2022-09-27
ENH: Added OldVonHamos4Crystal and OldVonHamosCrystal classes to pcds…

jortiz-slac:
…devices/spectrometer.py

## Description
Added OldVonHamos4Crystal and OldVonHamosCrystal classes to pcdsdevices/spectrometer.py

## Motivation and Context
This will allow scientists/users to manipulate the old four crystal Von Hamos interactively in a hutch python session.

## How Has This Been Tested?
Classes were tested using simulated motors then tested on actual Von Hamos hardware using a ipython session.
All axes were tested.

Doctrings used in classes.
modification and tests are documented in JIRA https://jira.slac.stanford.edu/browse/LCLSECSD-1513

## Pre-merge checklist
- [NA] Test suite passes locally
- [NA ] Test suite passes on travis

### DIFF
--- a/pcdsdevices/device_types.py
+++ b/pcdsdevices/device_types.py
@@ -42,7 +42,7 @@ from .sample_delivery import (HPLC, PCM, CoolerShaker, FlowIntegrator,
 from .sensors import RTD, TwinCATThermocouple
 from .sequencer import EventSequencer
 from .slits import Slits
-from .spectrometer import Kmono, VonHamos4Crystal
+from .spectrometer import Kmono, VonHamos4Crystal, OldVonHamos4Crystal
 from .timetool import Timetool, TimetoolWithNav
 from .valve import GateValve, Stopper
 from .wfs import WaveFrontSensorTarget
--- a/pcdsdevices/device_types.py
+++ b/pcdsdevices/device_types.py
@@ -42,7 +42,7 @@ from .sample_delivery import (HPLC, PCM, CoolerShaker, FlowIntegrator,
 from .sensors import RTD, TwinCATThermocouple
 from .sequencer import EventSequencer
 from .slits import Slits
-from .spectrometer import Kmono, OldVonHamos4Crystal, VonHamos4Crystal
+from .spectrometer import Gen1VonHamos4Crystal, Kmono, VonHamos4Crystal
 from .timetool import Timetool, TimetoolWithNav
 from .valve import GateValve, Stopper
 from .wfs import WaveFrontSensorTarget
--- a/pcdsdevices/device_types.py
+++ b/pcdsdevices/device_types.py
@@ -42,7 +42,7 @@ from .sample_delivery import (HPLC, PCM, CoolerShaker, FlowIntegrator,
 from .sensors import RTD, TwinCATThermocouple
 from .sequencer import EventSequencer
 from .slits import Slits
-from .spectrometer import Kmono, VonHamos4Crystal, OldVonHamos4Crystal
+from .spectrometer import Kmono, OldVonHamos4Crystal, VonHamos4Crystal
 from .timetool import Timetool, TimetoolWithNav
 from .valve import GateValve, Stopper
 from .wfs import WaveFrontSensorTarget
--- a/pcdsdevices/spectrometer.py
+++ b/pcdsdevices/spectrometer.py
@@ -364,7 +364,7 @@ class HXRSpectrometer(BaseInterface, GroupDevice):
     SUB_STATE = 'state'
 
 
-class OldVonHamosCrystal(BaseInterface, GroupDevice):
+class Gen1VonHamosCrystal(BaseInterface, GroupDevice):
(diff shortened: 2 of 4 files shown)


tangkong (commented):
Looks good on the whole to me, thanks for documenting everything so nicely. 

From what I can see here, it looks like the main change is in how the PV's are structured.  The python-side interface of this `OldVonHamosCrystal` is similar to the existing `VonHamos4Crystal`

Just a few clarifying comments from me here.  If it's been tested and works against the real device I don't have too many other concerns

pcdsdevices/device_types.py
```diff
@@ -42,7 +42,7 @@
 from .sensors import RTD, TwinCATThermocouple
 from .sequencer import EventSequencer
 from .slits import Slits
-from .spectrometer import Kmono, VonHamos4Crystal
+from .spectrometer import Kmono, VonHamos4Crystal, OldVonHamos4Crystal
```
tangkong, line 45:
```suggestion
from .spectrometer import Kmono, OldVonHamos4Crystal, VonHamos4Crystal
```

The pre-commit checks failed here, asking for these to be listed in alphabetical order

silkenelson:
Regarding the name: I veto'ed the original name ( as I did not think it was meaningful ), but did not have a much better idea, so I could use help:
This is a 4-crystal von Hamos that uses Kohzu stages. It uses the very first Beckhoff motion PLC w/ IOC that Zach built (pre ADS) etc, so it is for an 'old' Beckhoff motion setup.

pcdsdevices/spectrometer.py
```diff
@@ -361,3 +362,45 @@ class HXRSpectrometer(BaseInterface, GroupDevice):
     removed = False
     transmission = 1
     SUB_STATE = 'state'
+
+
+class OldVonHamosCrystal(BaseInterface, GroupDevice):
```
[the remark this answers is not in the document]
jortiz-slac: I spoke with Vincent and Silke about the name and agreed 'old' was okay - definitely better than using 'Mfx' 
This is an old PLC system that will not be re-used, but this legacy Von Hamos will be used for some time.

tangkong:
Is the existing vonHamos class the one being used for MFX?  Or will there be a third class created for that system?

Putting a "generation 1 PLC" note in the docstring would probably satisfy me.   Maybe we name the class `Gen1VonHamosCrystal` or something similar?

jortiz-slac:
I do like Gen1, If everyone agrees, we can change it to that

silkenelson:
I second Gen1, looks clearly more professional than 'old'.

vespos:
Did we try the generation of the auto-generated typhos screen from this class? That may just be my ignorance here, but I am not familiar with how the nested devices would show up on the screen. Also, shouldn't the kind be set to "hinted" for the component to appear in the screen?

Else this all looks good to me! The screen is definitely a nice to have, but not critical as this is mostly used at XCS, where the CLI is generally used.

pcdsdevices/spectrometer.py
```diff
@@ -361,3 +362,45 @@ class HXRSpectrometer(BaseInterface, GroupDevice):
...
+class Gen1VonHamosCrystal(BaseInterface, GroupDevice):
+    """Pitch, yaw, and translation motors for control of a single crystal."""
+
+    tab_component_names = True
+
+    pitch = FCpt(EpicsMotorInterface, '{prefix}{_pitchAxis}', kind='normal')
+    yaw = FCpt(EpicsMotorInterface, '{prefix}{_yawAxis}'  , kind='normal')
+    trans = FCpt(EpicsMotorInterface, '{prefix}{_transAxis}', kind='normal')
+
+    def __init__(self, prefix, pitchAxis, yawAxis, transAxis, **kwargs):
```
ZLLentz, line 376: Small quibble here, these variable names violate the pep8 conventions
https://peps.python.org/pep-0008/#function-and-variable-names

ZLLentz:
> Did we try the generation of the auto-generated typhos screen from this class? That may just be my ignorance here, but I am not familiar with how the nested devices would show up on the screen. Also, shouldn't the kind be set to "hinted" for the component to appear in the screen?

Here's a few screenshots of what typhos shows for `typhos --fake-device pcdsdevices.spectrometer.Gen1VonHamos4Crystal`
It essentially puts the common motor first, then makes a subheading for each crystal that each contain the motors for that crystal. I think we want the auto-generation to do these sorts of devices a bit differently but that's how it runs for now.
I guess you don't need to put subdevices as hinted.

image
image

pcdsdevices/spectrometer.py
jortiz-slac: Hi Zach, no problem. Changed them to be PEP 8 – Style Guide for Python Code compliant.
Thanks!

ZLLentz (approved):
Looks good to me, I see no reason not to merge

PR #1062, opened 2022-10-20
MNT: pull branches from attached happi md, rather than as kwargs

tangkong:
## Description
In order to be backwards compatible with past happi db's, we should not require `input_branches` and `output_branches` as kwargs for a lightpath-compatible devices.  Instead of grabbing those parameters from happi and setting them as attributes on the ophyd object, instead we read them from the attached metadata namespace.  

If that namespace does not exist, we look for any branches provided at init, finally raising an exception if that fails.

The `lightpath_summary` signal is now initialized when either:
- `input_branches` and `output_branches` are provided at init
- the `md` attribute is set by happi, after initialization

## Motivation and Context
We realized that changing the kwargs for all lightpath devices would break the hutch python implementations, since they might be locked at earlier tags.  This is basically restoring the old initialization interface, or the option to use it at least.

## How Has This Been Tested?
`happi audit` checking device instantiation, get_lightpath_state, and the types of the dataclass fields
- this only fails on some devices due to unreachable PV's, which  might be fixed with 🌟 **the great device_config merge** 🌟 
lightpath opens for all paths, as well as in sim mode
test suite passes as well

This PR

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -1693,20 +1693,22 @@ class LightpathMixin(Device):
         self._lightpath_ready = False
         self._retry_lightpath = False
         self._cached_state = None
-        self.input_branches = input_branches
-        self.output_branches = output_branches
+        self._md = None
 
         super().__init__(*args, **kwargs)
 
-        if self.input_branches and self.output_branches:
+        if input_branches and output_branches:
+            self._input_branches = input_branches
+            self._output_branches = output_branches
             self._init_summary_signal()
 
     def _init_summary_signal(self) -> None:
-        for sig in self.lightpath_cpts:
-            self.lightpath_summary.add_signal_by_attr_name(sig)
+        if self.lightpath_cpts:
+            for sig in self.lightpath_cpts:
+                self.lightpath_summary.add_signal_by_attr_name(sig)
 
-        self.lightpath_summary.subscribe(self._calc_cache_lightpath_state,
-                                         run=False)
+            self.lightpath_summary.subscribe(self._calc_cache_lightpath_state,
+                                             run=False)
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
@@ -1763,6 +1765,41 @@ class LightpathMixin(Device):
         Intended for use as a callback subscribed to lightpath_summary
         """
         self.get_lightpath_state(use_cache=False)
+
+    @property
+    def input_branches(self):
+        """
+        return input_branches from happi metadata, unless provided at init
+        """
+        md = getattr(self, 'md', None)
+        if md:
+            return md.get('input_branches')
+        else:
+            return getattr(self, '_input_branches')
+
+    @property
+    def output_branches(self):
+        """
+        return output_branches from happi metadata, unless provided at init
+        """
+        md = getattr(self, 'md', None)
+        if md:
+            return md.get('output_branches')
+        else:
+            return getattr(self, '_output_branches')
+
+    @property
+    def md(self):
+        if self._md is None:
+            raise AttributeError('Device does not have an attached md, '
+                                 'and was likely not initialized from happi')
+        return self._md
+
+    @md.setter
+    def md(self, new_md):
+        """ initialize lightpath when md is set """
+        self._md = new_md
+        self._init_summary_signal()
 
 
 class LightpathInOutMixin(LightpathMixin):
--- a/pcdsdevices/ccm.py
+++ b/pcdsdevices/ccm.py
@@ -1022,8 +1022,8 @@ class CCM(BaseInterface, GroupDevice, LightpathMixin, CCMConstantsMixin):
 
         Compares the x position with the saved in and out values.
         """
-        self._inserted = np.isclose(x_up, self._in_pos)
-        self._removed = np.isclose(x_up, self._out_pos)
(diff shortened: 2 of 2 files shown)


docs/source/upcoming_release_notes/1062-mnt_lp_branch_property.rst
```diff
@@ -0,0 +1,32 @@
+1062 mnt_lp_branch_property
+#################
+
+API Changes
+-----------
+- ``input_branches`` and ``output_branches`` are no longer required
+  for ``LightpathMixin``.  ``LightpathMixin`` will attempt to grab
+  this information from the ``md`` namespace attached by happi.
```
ZLLentz, line 8: I don't think this is a breaking API change

pcdsdevices/interface.py
```diff
@@ -1764,6 +1766,41 @@ def _calc_cache_lightpath_state(self, *args, **kwargs) -> None:
...
+        if md:
+            return md.get('output_branches')
+        else:
+            return getattr(self, '_output_branches')
+
+    @property
+    def md(self):
+        if self._md is None:
+            raise AttributeError('Device does not have an attached md, '
+                                 'and was likely not initialized from happi')
```
ZLLentz, line 1795: I like the helpful extended message here

pcdsdevices/interface.py
```diff
@@ -1764,6 +1766,41 @@ def _calc_cache_lightpath_state(self, *args, **kwargs) -> None:
...
+    @property
+    def input_branches(self):
+        """
+        return input_branches from happi metadata, unless provided at init
+        """
+        md = getattr(self, 'md', None)
+        if md:
+            return md.get('input_branches')
+        else:
+            return getattr(self, '_input_branches')
```
ZLLentz, line 1778: Another way to do this would have been to set `input_branches` and `output_branches` as part of the md setter and drop the leading underscore variants

tangkong: That's true.  I can get rid of some @property-ing here

docs/source/upcoming_release_notes/1062-mnt_lp_branch_property.rst
```diff
@@ -0,0 +1,32 @@
...
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- ``input_branches`` and ``output_branches`` are no longer required
+  for ``LightpathMixin``.  ``LightpathMixin`` will attempt to grab
+  this information from the ``md`` namespace attached by happi.A
```
klauer, line 28:
```suggestion
  this information from the ``md`` namespace attached by happi.
```

docs/source/upcoming_release_notes/1062-mnt_lp_branch_property.rst
```diff
@@ -0,0 +1,32 @@
...
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- ``input_branches`` and ``output_branches`` are no longer required
```
klauer, line 26:
```suggestion
- ``input_branches`` and ``output_branches`` keyword arguments are no longer required
```

klauer (approved):
This looks good to me - and I'm glad it was easier than I was imagining. 👍

docs/source/upcoming_release_notes/1062-mnt_lp_branch_property.rst
tangkong: vim has betrayed me 😢

pcdsdevices/interface.py
```diff
@@ -1693,20 +1693,22 @@ def __init__(self, *args,
...
-        for sig in self.lightpath_cpts:
-            self.lightpath_summary.add_signal_by_attr_name(sig)
+        if self.lightpath_cpts:
+            for sig in self.lightpath_cpts:
+                self.lightpath_summary.add_signal_by_attr_name(sig)
 
-        self.lightpath_summary.subscribe(self._calc_cache_lightpath_state,
-                                         run=False)
+            self.lightpath_summary.subscribe(self._calc_cache_lightpath_state,
+                                             run=False)
```
ZLLentz, line 1711: Do we need to guard against this running twice? Or does it not matter?

ZLLentz (approved):
Looks clean. One more side-thought but otherwise :+1:

pcdsdevices/interface.py
tangkong: That was meant to prevent the subscription if `lightpath_cpts` was empty, but that might be excessive.

ZLLentz: specifically: `_init_summary_signal` will be called twice if the branches are passed in init via happi, is that a problem?

klauer: The scenario where happi is misconfigured to send these settings as kwargs to `__init__` could lead to double subscriptions and confusion. I think Zach is onto something - it's a good idea to add a guard here.

That's the only scenario that seems plausible to me. We won't have someone creating an instance on their own (with kwargs) and patching in `.md`.

tangkong: Oh I see what you mean.  I can add a check for that real quick.  The signals are stored as a dictionary and shouldn't be a problem, but duplicate subscriptions might be

ZLLentz:
I think we're good to go here!

PR #1069, opened 2022-10-25
FIX: do not init summary signal unless we have valid branches

ZLLentz:
## Description

When initializing the lightpath branches from happi, guard against bad input_branches or output_branches.

## Motivation and Context

If one of these lightpath-compatible devices is instantiated from an old happi db without input_branches and output_branches, it initializes the summary signal but everything fails in a spammy way.

## How Has This Been Tested?

Worked in dev

Need to add release notes

## Pre-merge checklist


tangkong (approved):
This looks good to me.  bugfixes up until the end

PR #1071, opened 2022-10-26
DOC: v7.0.1 release notes

ZLLentz:
## Description

Small fixes we noticed after the v7.0.0 tag

## Motivation and Context

Let's include the fixes in our conda release


tangkong (approved):
I like not having bugs in our releases

PR #1073, opened 2022-11-01
Changing quadratic equation constants and limits

wwright-slac:
VLS focusing mirror testing resulted in finding a new linear to angle conversion equation. This PR input these new coefficients into the crix_motion python module, in addition to changing the resulting soft angular soft limit values. This has been tested in production, this PR just ensures that the changes stick when the next tag is released.

### DIFF
--- a/docs/source/upcoming_release_notes/1073-New_quadratic_equation_and_soft_limtis_for_VLS_focus_mirror.rst
+++ b/docs/source/upcoming_release_notes/1073-New_quadratic_equation_and_soft_limtis_for_VLS_focus_mirror.rst
@@ -1,0 +1,30 @@
+1073 New quadratic equation and soft limtis for VLS focus mirror
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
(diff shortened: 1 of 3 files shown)


ZLLentz (approved):

ZLLentz:
The tests fail here because the calc checks are no longer valid for the grating (I think)
https://github.com/pcdshub/pcdsdevices/blob/ddca02ce329ca5ce28f36365910ef555f4d33479/pcdsdevices/tests/test_crix_motion.py#L8-L13

wwright-slac:
Ah, good catch. Didn't realize these were actually calculated values

wwright-slac:
@ZLLentz Improper number of sigfigs?

wwright-slac:
Oh, I think I forgot that there is an offset

ZLLentz:
Oh no it looks like this was a exercise in frustration

wwright-slac:
@ZLLentz I made **exactly** the same mistake I made last week

PR #1076, opened 2022-11-02
ENH: add pre-move check for beckhoff axis nonzero velo

ZLLentz:
## Description

When the user requests a `BeckhoffAxis` move, check that the velocity is nonzero before attempting the move.

## Motivation and Context

If you execute a move with velocity=0, nothing happens and it can be confusing.
This makes it so at least an exception is raised.

This exception shows up in the typhos screen too.

The BeckhoffAxis EPICS motor records default to velocity=0 for safety.

## How Has This Been Tested?

Added unit tests and tested interactively with test PLC

I'll make a pre-release doc

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -1038,6 +1038,26 @@ class BeckhoffAxis(EpicsMotorInterface):
 
         return status
 
+    def check_value(self, pos: float) -> None:
+        """
+        Check that the motor can reach the position.
+
+        Inherits limits and related checks from parent classes.
+        Adds one additional check: the velocity must be nonzero
+        for a move to succeed. Otherwise, the move fails silently.
+
+        Beckhoff EPICS motors at LCLS that have never been initialized
+        before default to a zero velocity for safety. This check
+        lets the user know why the motor does not move.
+        """
+        super().check_value(pos)
+        velo = self.velocity.get()
+        if velo <= 0:
+            raise RuntimeError(
+                f'Motor velocity is {velo}, which is not valid. '
+                'Please configure a nonzero, positive velocity.'
+            )
+
 
 class BeckhoffAxisPLC_Pre140(BeckhoffAxisPLC):
     """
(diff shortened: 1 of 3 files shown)


tangkong (approved):
LGTM.  hooray for showing up in the typhos screen too

ZLLentz:
> hooray for showing up in the typhos screen too

Small caveat: it shows up at a pop-up using the normal pydm "if there is an exception raised, put the exception text into a pop-up". I'm thinking about doing a typhos PR that puts this into the move status field instead so that it's harder to miss (if the pop-up is somewhere weird on your control room many monitors setup, for example)

tangkong:
Are those popups not tied to the screen that generates them?  I feel like they should be centered on their parents by default...  Maybe a discussion for that typhos PR

ZLLentz:
I'm not entirely sure. At my SLAC desk's 2 monitor setup the popup actually appears _between_ the two monitors regardless of where the screen is...

pcdsdevices/epics_motor.py
```diff
@@ -1038,6 +1038,26 @@ def home(self, direction=None, wait=True, **kwargs):
...
+
+        Beckhoff EPICS motors at LCLS that have never been initialized
+        before default to a zero velocity for safety. This check
+        lets the user know why the motor does not move.
+        """
+        super().check_value(pos)
+        velo = self.velocity.get()
+        if velo <= 0:
+            raise RuntimeError(
+                f'Motor velocity is {velo}, which is not valid. '
```
klauer, line 1057: It may be wise to add the axis name in the exception message as which motor it refers to may be less clear when used outside of a GUI specifically for the axis (e.g., used in an N-D scan, in a nested function, etc).

klauer (approved):
This one has tripped up many of us. Nice fix.

pcdsdevices/epics_motor.py
ZLLentz: Good call

ZLLentz:
Merge conflict because I added a test to same file in both of my PRs, should be easy to resolve

PR #1077, opened 2022-11-02
TST: add option to skip device classes in find_all_device_classes

tangkong:
## Description
Adds an optional argument to `find_all_classes` (and by extension `find_all_device_classes`) that skips certain class names.  

## Motivation and Context
Filtering based on subclass has worked in the past, but now searching for all subclasses of ophyd.Device can return the lightpathmixins, which need a little more work to subclass properly.  This only appeared as a problem in https://github.com/pcdshub/atef/pull/132

## How Has This Been Tested?
Test suite runs, atef test suite can be modified to pass now too

This PR

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/tests/conftest.py
+++ b/pcdsdevices/tests/conftest.py
@@ -8,7 +8,7 @@ import sys
 import warnings
 from pathlib import Path
 from types import ModuleType, SimpleNamespace
-from typing import Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional
 
 import ophyd
 import pytest
@@ -113,7 +113,7 @@ def find_pcdsdevices_submodules() -> Dict[str, ModuleType]:
     return modules
 
 
-def find_all_classes(classes, skip: Optional[List[str]] = None) -> list:
+def find_all_classes(classes, skip: Optional[List[str]] = None) -> List[Any]:
     """Find all device classes in pcdsdevices and return them as a list."""
     skip = skip or []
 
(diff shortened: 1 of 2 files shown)


pcdsdevices/tests/conftest.py
```diff
@@ -113,14 +113,15 @@ def find_pcdsdevices_submodules() -> Dict[str, ModuleType]:
     return modules
 
 
-def find_all_classes(classes) -> list:
+def find_all_classes(classes, skip: List[str] = []) -> list:
```
klauer, line 116: This will seem pedantic, however it's been drilled into me in the past and I think it's been for the better:
-> Never include mutable default arguments. It doesn't matter if `skip` is actually mutated here or not.  It doesn't matter if the function is simple or not.

Future refactors may forget about this and introduce bugs - so let's reinforce good habits and side-step that issue now by just not doing it at all.

(`skip: Optional[List[str]] = None` and `skip = skip or []` is a quick path forward, of course).

klauer (approved):
Looks good to me (forgive the nitpick...)

pcdsdevices/tests/conftest.py
tangkong: It's something I should remember by now.  Good catch.

pcdsdevices/tests/conftest.py
```diff
@@ -135,9 +138,13 @@ def sort_key(cls):
     return list(sorted(set(devices), key=sort_key))
 
 
-def find_all_device_classes() -> list:
-    """Find all device classes in pcdsdevices and return them as a list."""
-    return find_all_classes(ophyd.Device)
+def find_all_device_classes(skip: Optional[List[str]] = None) -> list:
```
ZLLentz, line 141: Another tiny tiny nitpick: pick either `List` or `list` for your annotation and stick to it imo

ZLLentz (approved):

PR #1078, opened 2022-11-04
DOC: v7.1.0

ZLLentz:
## Description

Release notes
Reviewers includes all contributors as a notice. Reviewing is not mandatory.

## Motivation and Context

Tag is needed to include some features for `atef` and for general deployment to make apps like `lightpath` work properly for the xrt mirrors and to bring out some error handling optimization for `BeckhoffAxis` motors.

PR #1081, opened 2022-11-05
MNT: rework for the 4-blade attenuator screens

### LINKED ISSUE #1079: SXR attenuator screens: "custom" energy does not toggle widget visibility
## Current Behavior
For at1k4_calc and at2k2_calc, these screens exist:

image

These are embedded as [`AttenuatorCalculatorSXR_FourBlade_calc.ui`](https://github.com/pcdshub/pcdsdevices/blob/master/pcdsdevices/ui/AttenuatorCalculatorSXR_FourBlade_calc.ui).

## Expected Behavior

When the user clicks "Actual", the current LCLS photon energy should be used.
When the user clicks "Custom", a line edit widget allowing you to customize the photon energy for the calculation should be shown.

This is the rule:
image
Where the enum is `ca://${prefix}:SYS:EnergySource`

## Possible Solution
* Unclear why the rule isn't firing to me.
* Could remove the rule and just always show the widget?

### LINKED ISSUE #1080: SXR attenuator screens: 4-blade calculator results not intuitive or clear
## Expected Behavior

It should be obvious which filters the SXR solid attenuator calculator will insert from the provided user interface in pcdsdevices.
 
Currently, the calculator only indicates if _any_ filter will be inserted for the blade or not.

## Current Behavior

Here are two obvious examples, where the result of the calculator should be to either insert all filters or remove all filters:

A transmission of 0.0:

image

This is _okay_ at best, we still don't know which filter of the 4 per blade is going to be inserted.

A transmission of 1.0:

image

This is the only scenario where we can say the display is unambiguous. All filters are removed. Full beam.

And a less obvious example - a transmission of 0.01:

image

1 filter inserted (but which one?)

## Possible Solution

Currently, the IOC only returns integer values for the filter number to be inserted. This does not include material or thickness information.

```
$ caget AT1K4:CALC:SYS:BestConfiguration_RBV
AT1K4:CALC:SYS:BestConfiguration_RBV 4             1 1 9 8
```
This means for the 4 blades: removed, removed, and filter 9, filter 8.

So:
* The IOC itself could be modified at some point
* For now, we could show the number of the filter to be inserted instead of the green/gray box. This isn't great from the user's perspective as they have to look at the filter options for that blade but is at least better than the current solution.

klauer:
## Description
Updates to the attenuator screens for AT1K4 and AT2K2.

## Motivation and Context
Closes #1079 by removing the rule entirely
Closes #1080 by showing all the filters and the number that will be inserted

## How Has This Been Tested?
Interactively

PR text

image

Inserted filters and the configuration matrix of per-blade filters:
image

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1080-sattscreens.rst
+++ b/docs/source/upcoming_release_notes/1080-sattscreens.rst
@@ -1,0 +1,34 @@
+1080 sattscreens
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- Four blade SXR solid attenuator (AT1K4 and AT2K2) screens have been updated
+  to include all of the filters installed on each blade. It will also show the
+  per-blade filters that the calculator will insert when "Apply Configuration"
+  is clicked. The custom energy line edit will now remain visible regardless of
+  the "Actual/Custom" Photon Energy selection.
+
+Contributors
+------------
+- klauer
--- a/pcdsdevices/ui/AttenuatorCalculatorSXR_Blade_config.embedded.ui
+++ b/pcdsdevices/ui/AttenuatorCalculatorSXR_Blade_config.embedded.ui
@@ -1,0 +1,245 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>Form</class>
+ <widget class="QWidget" name="Form">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>816</width>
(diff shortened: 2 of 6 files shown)


tangkong (approved):
This worked well for me, though the problems reported in issues previously did make it harder to actually see the changes.  I like this a lot though

- The ui file that showed all the filters was one of three identically named files, strangely enough none of them had 'all_blades' in the name)
image

klauer:
> The ui file that showed all the filters was one of three identically named files, strangely enough none of them had 'all_blades' in the name)

The added `all_blades` ones are just embedded in the top-level `*FourBlade.detailed.ui` one, so you don't get the option to view them in isolation in typhos. It seemed like a good choice when I was putting it together, at least. (Side note: I had trouble coming up with a filename for those `all_blade` ones, so they are confusing even to me.)

ZLLentz (approved):
Looks good to me

I'll pull the doc here into the notes for #1078 after the merge

PR #1082, opened 2022-11-08
ENH: create RTDSX0ThreeStage for RTDSK0 solid drilling experiments

mkestra:
## Brief Description

Created a new, lightpath enabled class for the 3-stage actuator on RTDSK0 that will be used for solid drilling experiments.

##In-Depth Description

Added class, RTDSX0ThreeStage to rtds_ebd.py. This will track the lighpath status of the solid drilling experiment samples that will be held by the actuator. There are three stages but the only time there is potential to interfere with the beam is when the Y stage is not at it's outer limit. This is additionally tracked as a fast fault in the PMPS system. 

## Motivation and Context

This  setup will be used to conduct solid drilling experiments for SC readiness, we need a way to easily move multiple samples into and out of beam as needed, warranting the addition of this actuator. We further need some indication that it is not intering with the beam, hence inclusion of this device in lightpath and PMPS systems.

## How Has This Been Tested?
The class has been debugged, instantiated and used to create a dummy class that was tested and confirmed to show it was not interfering with beam in a local clone of light path. It will need to be further debugged when there is no beam. This is set to trigger as 'inserted' when the outer limit switch is not hit, the same logic as PMPS. If we tried to confirm that it logically switches while the beam was running on KFE, we would trigger a fast fault and potentially interfere with beam time.

update 11/9: tested and verified with PMPS and lightpath that this class functions to effectively determine if the y-stage is inserted.

A brief docstring has been included in the class definition. Will add pre-lease-notes in the next day or two, but want to get feedback on the code in the mean time!

-->

## Pre-merge checklist
- [X] Test suite passes on travis

### DIFF
--- a/docs/source/upcoming_release_notes/1082-Add_3-stage_motion_on_RTDSK0.rst
+++ b/docs/source/upcoming_release_notes/1082-Add_3-stage_motion_on_RTDSK0.rst
@@ -1,0 +1,30 @@
+1082 Add 3-stage motion on RTDSK0
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+------------
+- RTDSX0ThreeStage class; 3DoF motion stage for SOlid Drilling experiments 
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- Mike Estrada (mkestra)
--- a/docs/source/upcoming_release_notes/1082-Add_3-stage_motion_on_RTDSK0.rst
+++ b/docs/source/upcoming_release_notes/1082-Add_3-stage_motion_on_RTDSK0.rst
@@ -15,7 +15,7 @@ Device Updates
 
 New Devices
 ------------
-- RTDSX0ThreeStage class; 3DoF motion stage for SOlid Drilling experiments 
+- RTDSX0ThreeStage class; 3DoF motion stage for Solid Drilling experiments
 
 Bugfixes
 --------
--- a/docs/source/upcoming_release_notes/1082-Add_3-stage_motion_on_RTDSK0.rst
+++ b/docs/source/upcoming_release_notes/1082-Add_3-stage_motion_on_RTDSK0.rst
@@ -27,4 +27,4 @@ Maintenance
 
 Contributors
 ------------
-- Mike Estrada (mkestra)
+- mkestra
--- a/pcdsdevices/rtds_ebd.py
+++ b/pcdsdevices/rtds_ebd.py
@@ -54,7 +54,7 @@ class RTDSX0ThreeStage(BaseInterface, LightpathMixin):
     mmsz = Cpt(BeckhoffAxis, ':MMS:Z', kind='normal')
 
     # Determine if the y stage is at it's outer limit switch
-    open_limit = Cpt(EpicsSignalRO, ':MMS:Y:PLC:bLimitBackwardEnable_RBV', kind='normal',
+    open_limit = Cpt(EpicsSignalRO, ':MMS:Y.HLS', kind='normal',
                      doc='Reads if the y-stage is at its outer limit.')
     lightpath_cpts = ['open_limit']
 
--- a/pcdsdevices/rtds_ebd.py
+++ b/pcdsdevices/rtds_ebd.py
@@ -1,10 +1,12 @@
+from lightpath import LightpathState
 from ophyd import Component as Cpt
 from ophyd import Signal
 
 from .device import GroupDevice
+from .epics_motor import BeckhoffAxis
 from .inout import InOutPositioner
-from .interface import BaseInterface, LightpathInOutCptMixin
-from .signal import PytmcSignal
+from .interface import BaseInterface, LightpathInOutCptMixin, LightpathMixin
+from .signal import EpicsSignalRO, PytmcSignal
 
 
 class PneumaticActuator(InOutPositioner):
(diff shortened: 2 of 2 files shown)


pcdsdevices/rtds_ebd.py
```diff
@@ -44,6 +46,27 @@ def _do_move(self, state):
...
 
+class RTDSX0ThreeStage(BaseInterface, LightpathMixin):
+    """Three stages X,Y,Z, for solid drilling experiments"""
+
+    mmsx = Cpt(BeckhoffAxis, ':MMS:X', kind='normal')
+    mmsy = Cpt(BeckhoffAxis, ':MMS:Y', kind='normal')
+    mmsz = Cpt(BeckhoffAxis, ':MMS:Z', kind='normal')
+
+    # Determine if the y stage is at it's outer limit switch
+    open_limit = Cpt(EpicsSignalRO, ':MMS:Y.HLS', kind='normal',
```
ZLLentz, line 57: Not much wrong with a single duplicated PV connection here, but I'm curious: is there any way with the `LightpathMixin` to access subcomponents? The Y high limit is a PV connection that already existed under `mmsy.high_limit_switch`

pcdsdevices/rtds_ebd.py
```diff
@@ -44,6 +46,27 @@ def _do_move(self, state):
...
+                     doc='Reads if the y-stage is at its outer limit.')
+    lightpath_cpts = ['open_limit']
+
+    def calc_lightpath_state(self, open_limit=None):
+        # Logic, calculations using open_limit for y-stage
+        status = LightpathState(
+            inserted=not open_limit, removed=open_limit,
```
ZLLentz, line 64: This matches the fast fault logic, so this is probably appropriate. The only way this can fail is with a fake "inserted" result with the FFO overridden, right? As in, if the stage is "almost out" and the lightpath says "I'm blocking beam!" It's probably ok.

pcdsdevices/rtds_ebd.py
```diff
@@ -44,6 +46,27 @@ def _do_move(self, state):
...
+            output={'K0' : 0.0}
```
ZLLentz, line 65: Not fully used to the newest lightpath interface yet: does this mean it _always_ shows as blocking the `K0` line in `lightpath`, regardless of inserted or removed?

Does the hard-coding of `K0` here mean that we need to change this when this gets moved to the `L0` line later? Is there a way to not hard-code this?

pcdsdevices/rtds_ebd.py
```diff
@@ -44,6 +46,27 @@ def _do_move(self, state):
             self.out_cmd.put(1)
 
 
+class RTDSX0ThreeStage(BaseInterface, LightpathMixin):
+    """Three stages X,Y,Z, for solid drilling experiments"""
+
+    mmsx = Cpt(BeckhoffAxis, ':MMS:X', kind='normal')
+    mmsy = Cpt(BeckhoffAxis, ':MMS:Y', kind='normal')
+    mmsz = Cpt(BeckhoffAxis, ':MMS:Z', kind='normal')
```
ZLLentz, line 54: This is a unique component naming compared to other devices in the repo. That doesn't make it wrong, but it's worth thinking on it for a moment. The component names here are acronyms that nominally stand for "motion motor stepper x", for example.

ZLLentz (commented):
I'm not objecting to any of this, it works and looks clean. I've asked a bunch of design questions as food for thought.

I can't click approve until it has the release notes stub.

pcdsdevices/rtds_ebd.py
mkestra: Interesting question, I guess it didn't occur to me that the high_limit_switch PV may be accessible without separately reading it into Cpt as I did here. @tangkong, does doing it the way Zach mentions here seem feasible/preferable?

pcdsdevices/rtds_ebd.py
mkestra: right, this is encoded relatively such that the position is estimated only as a product of the lead screw and [expected] rotation. We've had some issues with this stage where it can stall due to some combination of unexpected friction in the drive and limited current draw observed in the testing phase. There's probably one or two other things contributing to this too. All that to say, the only reasonable "100% safe" state I can think to employ is the outer limit switch being hit.

pcdsdevices/rtds_ebd.py
mkestra: right, I based this nominally off of the naming convention in RTDSBase, where the pneumatic cylinders are labeled by their CCC; mpa1, mpa2, etc. That said, reviewing the more standard naming convention across other device types something like x_motor, y_motor, z_motor would probably be more in alignment with standard practice. I'll leave this for now as I wouldn't want to change the name without testing again, which would require waiting for another opportunity. I'll try to pay closer attention to package-wide naming standards moving forward.

pcdsdevices/rtds_ebd.py
mkestra: tbh I'm not entirely sure, this is more-or-less following a standard example and some additional assistance from @tangkong. I would appreciate to hear the answer to this as well if you get a moment Robert!

pcdsdevices/rtds_ebd.py
tangkong: Yes, you can list `['mmsy.high_limit_switch']` as one of the `lightpath_cpts`.  Some pcdsdevices do this already I think
https://github.com/pcdshub/pcdsdevices/blob/3bd3b5c121c2cabac4238a4da05bfd5f9cb2b2b0/pcdsdevices/ccm.py#L941

I do think that would be preferable, if only to clean up the class definition a bit

pcdsdevices/rtds_ebd.py
ZLLentz: matching the other rtds devices is a good enough reason in my book

pcdsdevices/rtds_ebd.py
tangkong: (sorry for the tardiness)

Zach is correct, this would show as 0 transmission (blocking beam) no matter the insertion state.  this would also need to be changed if we moved the device as Zach notes.  I'd suggest something like
```python
trans = 0.0 if not open_limit else 1.0
status = LightpathState(
    inserted = not open_limit,
    removed = open_limit,
    output = {self.output_branches[0]: trans}
)
```

if there's additional nuance to the transmission here, we you could code it in prior to creating the `LightpathState`

docs/source/upcoming_release_notes/1082-Add_3-stage_motion_on_RTDSK0.rst
```diff
@@ -0,0 +1,30 @@
...
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- Mike Estrada (mkestra)
```
tangkong, line 30:
```suggestion
- mkestra
```
We usually keep these to just the github username

pcdsdevices/rtds_ebd.py
tangkong: In the end I guess we only check the transmission if the device is inserted, but having the LightpathState return sensible transmissions is easy to do and simple enough I think.

I assume that because of this, lightpath behaved as expected while you were testing?

pcdsdevices/rtds_ebd.py
mkestra: ah okay, for the same reason as before, making this change now would require waiting for another chance to test, so I think If it seems okay I will leave it as-is for now, and keep this in mind moving forward!

tangkong (approved):
I think this looks good to me at this point.  Since we won't get any more testing time soon, we can probably defer further changes until then.

The remaining suggested changes do not change the functionality of the device

pcdsdevices/rtds_ebd.py
ZLLentz: I largely wanted to make sure this was at least possible :+1:
One extra PV in a device with dozens isn't a big deal

ZLLentz:
I'm going to change the title because it's not descriptive at all.

The "K0" hard-coded will guarantee that this must be revisited in the future. If there are reasons to not fix it now it's ok.

I think this is good to merge once it has a better title.

ZLLentz (approved):

PR #1085, opened 2022-11-16
add single blades control by making a new LusiSlitsWithBlades class

vespos:
## Description
Add controls of the single blades in a slit as per scientists request.

## Motivation and Context
When aligning the slits, it is sometimes desirables to move a single blade a a time. For example when doing precise clipping of all four slits.

## How Has This Been Tested?
Tested with s4 at XPP.

TBD

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/slits.py
+++ b/pcdsdevices/slits.py
@@ -475,6 +475,8 @@ class LusiSlitsWithBlades(LusiSlits):
     blade_north = FCpt(IMS, '{self._mms_prefix}:{self._blade_pvs[2]}')
     blade_south = FCpt(IMS, '{self._mms_prefix}:{self._blade_pvs[3]}')
 
+    tab_whitelist = ['blade_up', 'blade_down', 'blade_north', 'blade_south']
+
     def __init__(self, prefix, blade_pvs_idx, **kwargs):
         self._blade_pvs = blade_pvs_idx
         self._mms_prefix = prefix.removesuffix(':JAWS')+':MMS'
--- a/pcdsdevices/slits.py
+++ b/pcdsdevices/slits.py
@@ -29,9 +29,9 @@ from ophyd.status import wait as status_wait
 from .areadetector.detectors import PCDSAreaDetectorTyphosTrigger
 from .device import GroupDevice
 from .device import UpdateComponent as UpCpt
+from .epics_motor import BeckhoffAxis, BeckhoffAxisNoOffset, PCDSMotorBase
 from .interface import (BaseInterface, FltMvInterface, LightpathInOutCptMixin,
                         LightpathMixin, MvInterface)
-from .epics_motor import BeckhoffAxis, BeckhoffAxisNoOffset, IMS
 from .pmps import TwinCATStatePMPS
 from .sensors import RTD, TwinCATTempSensor
 from .signal import NotImplementedSignal, PytmcSignal
@@ -424,12 +424,19 @@ class LusiSlits(SlitsBase):
     xcenter = Cpt(LusiSlitPositioner, '', slit_type='XCENTER', kind='normal')
     ycenter = Cpt(LusiSlitPositioner, '', slit_type='YCENTER', kind='normal')
 
+    # Individual blade aliases
+    blade_top = Cpt(PCDSMotorBase, ':TOP', kind='normal')
+    blade_bottom = Cpt(PCDSMotorBase, ':BOTTOM', kind='normal')
+    blade_north = Cpt(PCDSMotorBase, ':NORTH', kind='normal')
+    blade_south = Cpt(PCDSMotorBase, ':SOUTH', kind='normal')
+
     # Local PVs
     blocked = Cpt(EpicsSignalRO, ':BLOCKED', kind='omitted')
     open_cmd = Cpt(EpicsSignal, ':OPEN', kind='omitted')
     close_cmd = Cpt(EpicsSignal, ':CLOSE', kind='omitted')
     block_cmd = Cpt(EpicsSignal, ':BLOCK', kind='omitted')
 
+    tab_whitelist = ['blade_top', 'blade_bottom', 'blade_north', 'blade_south']
     lightpath_cpts = ['xwidth.readback', 'ywidth.readback']
 
     def open(self):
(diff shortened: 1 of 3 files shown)


vespos:
@tangkong I have tried so many different ways to pass a list of pvs to the SlitBLades subdevice, and that is the only way I found that works... This is clearly ugly, is there a better way to do something like that?

Leaving it as a draft for now and will clean-up, flake8 etc once we think we have a good way of doing it.

pcdsdevices/slits.py
```diff
@@ -453,6 +455,38 @@ def calc_lightpath_state(
...
                                             ywidth=ywidth_readback)
 
+class SlitsBlades(BaseInterface, GroupDevice):
+
+    up = FCpt(IMS,'{self._pv_up}')
+    down = FCpt(IMS,'{self._pv_down}')
+    north = FCpt(IMS,'{self._pv_north}')
+    south = FCpt(IMS,'{self._pv_south}')
+
+    def __init__(self, blade_pvs, **kwargs):
```
ZLLentz, line 465: Another approach is to pass them separately rather than in a single list argument. We have a lot of devices that do this using the `UnrelatedComponent` e.g. https://github.com/pcdshub/pcdsdevices/blob/23a747b0cb4b6d7c3719870e16a57e0cbbdd106a/pcdsdevices/ccm.py#L792-L810 but you can also do this without by making extra arguments.

Make sure you also pass *args into the super().__init__ too or the rest of the device won't work properly.

vespos: Thanks, that's good to know! The lack of dynamic in Ophyd is really frustrating at times...
After discussing with @tangkong , I however think the syntax overhead is not worth it. I'm going to keep the blades at the top level.

vespos:
Why is the buildd failing? There seem to be an issue with the pre-commit checks but I cant quickly figure out what's wrong. @tangkong @ZLLentz

klauer:
Our continuous integration checks to see that you actually used pre-commit on your code so we don't accidentally commit non-compliant code into the repository.

Travis says:
```
isort....................................................................Failed
- hook id: isort
- files were modified by this hook
Fixing /home/travis/build/pcdshub/pcdsdevices/pcdsdevices/slits.py
Fixing /home/travis/build/pcdshub/pcdsdevices/pcdsdevices/device_types.py
```

This means that your imports are not sorted properly. The tool we use (`isort`) sorts them for you and keeps things consistent and clean. If you use `pre-commit` when developing, things get a lot easier. It would automatically fix the issue prior to pushing to GitHub. That is, `pre-commit` would run `isort` every time you commit, avoiding the failure on GitHub.

See here: https://confluence.slac.stanford.edu/pages/viewpage.action?spaceKey=PCDS&title=How+to+add+a+device+to+pcdsdevices and scroll to `pre-commit`

It's primarily just a `pre-commit install` and then `git commit` assuming you have changes to be committed. If you've already accidentally committed changes without pre-commit, you can use `pre-commit run` (as in the docs above)

vespos:
Thanks @klauer , I've run it

New commit which changes sustantially how we do things: there is a PV that aliases the blade axis. A trick is nonetheless needed, as the IMS class does not work on the PV. Using PCDSMotorBase works and should be fine for this application.

klauer:
Looks mostly good to me. Noting a few things:
- Confirming for myself that this change will work for all `LusiSlits` in our database:
```
(pcds-5.5.1) [klauer@psbuild-rhel7-02  ~]$ happi search --names 'device_class=*LusiSlits'
xpp_sb3_slits
```
Only one! OK. And XPP aliases this as follows: `from xpp.db import xpp_sb3_slits as s4`

- I assume the tab behavior is as you desire, so 👍 
- The pre-release changelog will require some modification as the approach is no longer the same

tangkong (commented):
Does every LusiBlades IOC have these aliases?  If so I'm ok with adding them to the existing LusiBlades class, but otherwise we should make a separate class as to not break the existing classes.

Looking very briefly into this, it seems not all devices that use the `LusiSlits` (or the equivalent `LusiSlits` classes have these aliases. I think we should make this a new subclass and change the happi entry for the relevant XPP devices

silkenelson:
Under the assumption that Lusi Slits are the JJ-slits with the 4 separate motors that also have a JAWS record, please do NOT make subclasses. If anyone is running a LusiSLit IOC without the aliases, they are doing it wrong and need to upgrade. This should have happened in the lead to Run 18, but we were understaffed.
Please let me know if there are Lusi slits instances that are different for a real reason and we can talk subclassing.

vespos:
A lot of the LusiSlits slits use the alias class "Slits" for their class in happi, so there are many more than xpp_sb3_slits in fact. I've checked a number of them (not all though), they all have the relevant PVs. Let me clean up the doc and release notes.

I would agree with Silke here, it's best to include it in the LusiSlits class, so we don't have to change every happi entry, as this is something every LusiSlits should have.

I will spread words to other POCs to check their slits IOCs or at least be mindful next time they update their pcdsdevice repo.

tangkong:
There are quite a few slits classes that collide with this, as `pcdsdevices.device_types.Slits` = `pcdsdevices.slits.Slits` = `pcdsdevices.slits.LusiSlits` (as Vincent points out).  

A quick audit of happi shows 6 slits devices currently failing to connect: 
* xrt_dg3m_slits
* um6_slits
* sl2k4
* qrix_det_slits
* sl2k2
* sl1k4
* qrix_opt_slits 
* hxd_dg2_slits

[Full (painful to read) output](https://gist.github.com/tangkong/a0ac7904f76a95434dbeeb88b0c99a6e)

ZLLentz:
The main consequence is that you can't call read on the device and neither can bluesky (without an error message) so you wouldn't be able to scan through the slit's gap size with bluesky.

I created some github issues about sl2k2, qrix_det_slits, and qrix_opt_slits which might be set up wrong in the Beckhoff PLCs. The TMO slits are also BeckhoffSlit instances but the IOC was simply off earlier.

xrt_dg3m_slits, um6_slits, and hxd_dg2_slits are the IMS slits that would want updates.

ZLLentz (approved):
Should be good with the understanding that the three slit IOCs should be updated when they can be

tangkong (approved):
Having come to an agreement, I'm good to go too

vespos:
Thanks I'll be taking care of the IMS IOCs.
Merging now.

PR #1086, opened 2022-11-18
ENH: Adds support for Two Bladed SXR Attenuators and AT1K2 class

wwright-slac:
## Description
Adding a class for AT1K2 Solid Attenuator, which has two standard 8 filter blades, and one Linear Bellows Drive which requires simple in/out functionality. 

## How Has This Been Tested?
This has not been tested yet with existing PVs, however the screens generate fine.

image

image

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1086-Adding_classes_for_AT1K2.rst
+++ b/docs/source/upcoming_release_notes/1086-Adding_classes_for_AT1K2.rst
@@ -11,7 +11,8 @@ Features
 
 Device Updates
 --------------
-- Attenuator module now has classes for two blade attenuators
+- Adds a preliminary attenuator class AT1K2 and base classes for similar
+   two-blade ladder attenuators designed by JJ X-ray.
 
 New Devices
 -----------
--- a/docs/source/upcoming_release_notes/1086-Adding_classes_for_AT1K2.rst
+++ b/docs/source/upcoming_release_notes/1086-Adding_classes_for_AT1K2.rst
@@ -1,0 +1,30 @@
+1086 Adding classes for AT1K2
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- Attenuator module now has classes for two blade attenuators
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- wwright-slac
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -1166,7 +1166,7 @@ class AttenuatorSXR_LadderTwoBladeLBD(FltMvInterface, PVPositionerPC,
     tab_component_names = True
 
     # Register that all blades are needed for lightpath calc
-    lightpath_cpts = [f'blade_{idx:02}.state.state' for idx in range(1, 3)]
+    lightpath_cpts = [f'blade_{idx:02}.state.state' for idx in range(1, 4)]
 
     # Summary for lightpath view
     num_in = Cpt(InternalSignal, kind='hinted')
@@ -1175,7 +1175,8 @@ class AttenuatorSXR_LadderTwoBladeLBD(FltMvInterface, PVPositionerPC,
     calculator = UCpt(AttenuatorCalculatorSXR_TwoBlade)
     blade_01 = Cpt(SXRLadderAttenuatorBlade, ':MMS:01')
     blade_02 = Cpt(SXRLadderAttenuatorBlade, ':MMS:02')
-    lbd_blade = Cpt(FEESolidAttenuatorBlade, ':MMS:03')
+    # LBD Stage
+    blade_03 = Cpt(FEESolidAttenuatorBlade, ':MMS:03')
 
     def __init__(self, *args, limits=None, **kwargs):
         UCpt.collect_prefixes(self, kwargs)
--- a/pcdsdevices/attenuator.py
+++ b/pcdsdevices/attenuator.py
@@ -902,6 +902,41 @@ class AttenuatorCalculatorSXR_Blade(AttenuatorCalculatorFilter):
 '''
 
 
+class AttenuatorCalculatorSXR_TwoBlade(AttenuatorCalculatorBase):
+    """
+    2 blade x 8 filter solid attenuator variant from the L2SI project.
+
+    Parameters
+    ----------
+    prefix : str
+        Full Solid Attenuator base PV.
+
+    name : str
+        Alias for the Solid Attenuator.
+    """
+
+    tab_component_names = True
+    first_filter = 1
+    num_filters = 4
+    # Not using "DDC" here, so the parent is `self`:
+    _filter_parent = None
+    _filter_index_to_attr = {
+        1: 'blade_01',
+        2: 'blade_02',
+    }
+
+    blade_01 = Cpt(AttenuatorCalculatorSXR_Blade, ':AXIS:01:', index=1)
+    blade_02 = Cpt(AttenuatorCalculatorSXR_Blade, ':AXIS:02:', index=2)
+
+    def format_status_info(self, status_info):
+        """
+        Override status info handler to render the attenuator.
+        """
+        return utils.combine_status_info(
+            self, status_info, self._filter_index_to_attr.values(),
+        )
+
+
 class AttenuatorCalculatorSXR_FourBlade(AttenuatorCalculatorBase):
     """
     4 blade x 8 filter solid attenuator variant from the L2SI project.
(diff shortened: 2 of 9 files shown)


wwright-slac:
@klauer Not sure if things like this need to be tested with a device if it's mostly just using pre-existing code

pcdsdevices/attenuator.py
```diff
@@ -902,6 +902,41 @@ def format_status_info(self, status_info):
...
+    prefix : str
+        Full Solid Attenuator base PV.
+
+    name : str
+        Alias for the Solid Attenuator.
+    """
+
+    tab_component_names = True
+    first_filter = 1
+    num_filters = 4
```
klauer, line 920:
```suggestion
    num_filters = 2
```

This may be vestigial, but probably best to keep it correct regardless

klauer (commented):
I think it's probably okay overall to have a copy, though more code reuse could be preferable.

What we'll need is to:
* Make a page like this: https://confluence.slac.stanford.edu/display/L2SI/AT1K4-SOLID+SXR+Solid+Attenuator for AT1K4 to detail the filters and report results from the checkout procedure (has it been checked out yet? where is it? this PR is the first I've heard of it!)
* Deploy a new calculator IOC for this which will definitely require testing and some (hopefully minor) Python code for the startup script (based on [this](https://github.com/pcdshub/solid-attenuator/blob/master/solid_attenuator/ioc_kfe_at1k4_calc/__main__.py) for at1k4)
* Add new screens to this repository for the 2-blade variant of the calculator
* Test those new screens with these device classes
* Verify that the calculator is doing what it should with those screens + hutch-python

I think we should _probably_ do all of the above before considering to merge this one in.

pcdsdevices/attenuator.py
ZLLentz: Note for myself: this is only used in the 18 blade att to help construct a mapping, and then is set to the new correct value elsewhere to maintain interfaces despite not being used again.

pcdsdevices/attenuator.py
```diff
@@ -1104,6 +1139,170 @@ def format_status_info(self, status_info):
...
+
+    def _setup_move(self, position):
+        """(PVPositioner compat) - calculate, then move."""
+        # Do not call `calculator.calculate()` here to respect the current
+        # calculator settings:
+        self.calculator.desired_transmission.put(position)
+        self.calculator.run_calculation.put(1, wait=True)
+        return super()._setup_move(position)
+
+    def get_lightpath_state(self, use_cache=True) -> LightpathState:
```
ZLLentz, line 1209: It's ok to duplicate here as needed- it can be a future task to refactor this module to remove code duplication. Duplicating once or twice isn't a big deal.

pcdsdevices/attenuator.py
wwright-slac: Yes, I thought this was weird, but I kept it there since the AT2K2 also has it

wwright-slac:
@klauer Do you think there is anything special to do regarding the fact that AT1K2 has a non-attenuating blade? Perhaps blocking the functionality of the calculator when the mirror blade is in?

klauer:
@wwright-slac Yes, I think we probably need to modify the lightpath calculation code here to be aware of the inspection mirror.

The calculator _IOC_ may need to be aware of the state of the inspection mirror, too. The calculator currently does the following:
* Taking in the current beam energy (`x` eV)
* Reading in the state of the per-blade filters and matching them with the installed filter material/thickness
* Calculate transmission at `x` eV (`ActualTransmission_RBV`) and `3*x` eV (the third harmonic; `Actual3OmegaTransmission_RBV`) as the product of the transmission of all N blades

With the inspection mirror in, that should drop to 0 (or show an error state, perhaps).
Would we also want the calculator to remove the inspection mirror when 'apply configuration' is hit? That is, the user says "I want a transmission of 0.1 at this energy; make it happen -> (calculator removes inspection mirror and inserts appropriate filters)".

However, if we request a transmission of 0, we wouldn't want the calculator inserting the inspection mirror as an attenuation method. If these changes make it to the IOC, we'll have to be careful about these scenarios...

I think the answer there is at least a "maybe". I'm curious about others' opinions here.

ZLLentz:
For the IOC, I don't think we should code the inspection mirror as a "full block target" for the edge cases mentioned by Ken. Currently for AT2L0 we ignore the inspection mirror entirely in the calculator (it's on blade 1, the screen even skips this blade). That inspection mirror is essentially a separate device as far as the calculation is concerned.

For every attenuator with an inspection mirror I think mirror in makes sense to include it as an error/0 transmission state, (but not as a normal part of the calculation)

For the Python class, I tried to understand what we did for AT2L0 as a precedent but I'm coming to the conclusion that it's implemented incorrectly there and I'll be taking a closer look to understand it better.

I suspect the correct behavior is mirror in -> transmission 0 (beam is blocked) for the purposes of `lightpath`.

(I also suspect we should put a fast fault in the PLC for mirror in...)

tangkong:
The at2l0 lightpath logic is functionally the same as the other `LightpathInOutCptMixin` classes, and was duplicated to give the class access to the number of blades in/out.  This was done to keep the attributes that existed before the lightpath rework working.  Otherwise the logic is the same as in the mixin.

To that end, there's nothing in at2l0's lightpath behavior that deals with a special inspection mirror, so that would need to be added.

ZLLentz:
AT2L0's class actually does show blocking with the mirror in, but it's incidental. It uses the mixin as you say Robert but the mixin is a bit busted for the attenuator blades. They don't check in with the calculator in any way so every blade will report the default full blocking when inserted. So the AT2L0 class might do everything correctly but the behavior it relies on is broken.

wwright-slac:
image
image

IOC working with new screens

docs/source/upcoming_release_notes/1086-Adding_classes_for_AT1K2.rst
```diff
@@ -0,0 +1,30 @@
...
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- Attenuator module now has classes for two blade attenuators
```
klauer, line 14:
```suggestion
- Adds a preliminary attenuator class AT1K2 and base classes for similar
   two-blade ladder attenuators designed by JJ X-ray.
```

I say "preliminary" because the motion axes don't exist yet as far as EPICS is concerned (i.e., there's no IOC)

pcdsdevices/attenuator.py
```diff
@@ -1104,6 +1139,170 @@ def format_status_info(self, status_info):
...
+                cpt_name = sig.name.removeprefix(self.name + '_')
+                cpt_name = cpt_name.removesuffix('_state_state')
+                lightpath_kwargs[cpt_name] = sig.get()
+
+            self._cached_state = self.calc_lightpath_state(**lightpath_kwargs)
+
+        return self._cached_state
+
+    def calc_lightpath_state(self, **lightpath_kwargs) -> LightpathState:
+        # Repeat lightpath logic to extract num_in, num_out
```
klauer, line 1229: The logic here needs a bit of work as we've been discussing. Something along the lines of "use calculator transmission value if blade 3 is removed or 0.0 if blade 3 is inserted" - meaning that the number of lightpath components drops to 2 (calculator transmission and blade 3 state)

(Hmm, it could also be OK to merge this and then make a follow-up PR since this one is rather large...)

ZLLentz: Yeah it could be OK to handle this later alongside #1092 which needs much the same adjustment. It's possible that others also need a small tweak here.

pcdsdevices/attenuator.py
```diff
@@ -1125,6 +1324,29 @@ class AT1K4(AttenuatorSXR_Ladder):
...
+    ----------
+    prefix : str
+        Solid Attenuator base PV.
+
+    name : str
+        Alias for the Solid Attenuator.
+
+    calculator_prefix : str
+        The prefix for the calculator PVs.
+    """
```
ZLLentz, line 1347: Would it be possible to add this to the test suite fixture? I'm not actually sure how this works but it seems like we're somehow (successfully?) testing that the move commands work for the other attenuators.

https://github.com/pcdshub/pcdsdevices/blob/b66e6808a58679b1ed3cbb906ec6821b131acaf5/pcdsdevices/tests/test_attenuator.py#L171-L181

klauer (approved):
So `solid-attenuator` has a built-in simulator IOC for 4-blade SXR solid attenuators. AT1K2 uses a subset of the PVs required for that, meaning we can actually use the built-in simulator to test out AT1K2:

```bash
$ python -m solid_attenuator.ioc_sim_sxr --list-pvs --system=AT1K2
```

Careful if you run this on psbuild as these PVs "look like" production ones. If you accidentally leave it running when you get your PLC project in place, it could cause confusion down the line (e.g., you think you're moving your motor in lucid, but you're moving the simulator axis).

We may want to make a symlink to `detailed_tree.ui` in the `pcdsdevices/ui` directory, because this was necessary:
image

I'm able to move to filter configurations with this setup:

image

With that, I think this is good to go for the first pass.

pcdsdevices/attenuator.py
ZLLentz: 😢

wwright-slac: @ZLLentz Ah! Missed that. Submitting a new PR for this

Issue #1083, opened 2022-11-09
Need a class for DUT_MotionPneumaticActuator, FB_MotionPneumaticActuator


ZLLentz:
## Expected Behavior

We need a class for controlling beckhoff pneumatic devices that are implemented using https://github.com/pcdshub/lcls-twincat-motion/blob/2644528a99a1857cd0cdaefd62f2612f10e35b9e/lcls-twincat-motion/Library/DUTs/DUT_MotionPneumaticActuator.TcDUT.

This is the crucial datatype embedded in https://github.com/pcdshub/lcls-twincat-motion/blob/2644528a99a1857cd0cdaefd62f2612f10e35b9e/lcls-twincat-motion/Library/POUs/Motion/FB_MotionPneumaticActuator.TcPOU

This is used for ST3K2, ST4K4, and ST5K4, so it only would control stoppers for now, but the underlying code is intended to be generic for pneumatic motion.

The class should contain helpful methods for inserting and removing these devices and lightpath support (inserted = blocking)

## Current Behavior

No such class exists. Existing stopper classes are not suitable for these devices.

## Possible Solution

Create a new device.

## Context

James Cryan in TMO tried to instantiate an ST4K4 device and found that no suitable class existed

## Your Environment

latest

ZLLentz:
I suspect we want a generic beckhoff pneumatic-type device with references to the beckhoff motion lib and insert/remove helpers, then a specific stopper class that adds the lightpath info

nrwslac:
This looks like a good issue to get back into some python. Mind if I take a stab at it? @ZLLentz

ZLLentz:
@nrwslac go for it if it interests you, it's 100% fair game!

PR #1088, opened 2022-11-28
1083 pneumatic actuator

nrwslac:
## Description
Adds a new module pneumatic.py to support all pneumatics in python. 
Added class `BeckhoffPneumatic`
This class includes `lightpath` support and `insert` and `remove` functions.

The `BeckhoffPneumatic` class is initially available to support beckhoff pneumatic devices that are implemented using https://github.com/pcdshub/lcls-twincat-motion/blob/2644528a99a1857cd0cdaefd62f2612f10e35b9e/lcls-twincat-motion/Library/DUTs/DUT_MotionPneumaticActuator.TcDUT.
## Motivation and Context

User tried to instantiate a pcdsdevice for ST4K4 using class `Stopper`, This class was insufficient for ST4K4. https://slac.slack.com/archives/C9RP2PQHM/p1667963741318729

github issue: https://github.com/pcdshub/pcdsdevices/issues/1083

## Screenshots:
st4k4-insert
st4k4-remove

## Pre-merge checklist


pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,77 @@
...
+			#throw?
+		
+	def retract(self):
+		if self.retract_ok:
+			self.retract_signal.put(1)
+		else:
+			pass
+			#throw?
+
+	trans = 0.0 if limit_switch_in and not limit_switch_out else 1.0
```
tangkong, line 61: This should certainly be inside of `calc_lightpath_state`, if you want to calculate it based on the current values of `limit_switch_in` and `limit_switch_out`.

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,77 @@
...
+	position_state = Cpt(EpicsSignalRO, ':nPositionState')
+
+	def insert(self):
+		"""
+		Method for inserting Beckhoff Pneumatic Actuator
+		"""
+		if self.insert_ok:
+			self.insert_signal.put(1)
+		else:
+			pass
```
tangkong, line 51: You could raise an exception here, depending on how you want to use this method.  Lightpath no longer uses these methods, so it's up to you.  I might raise a `RuntimeError`, personally.

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,77 @@
...
+			self.retract_signal.put(1)
+		else:
+			pass
+			#throw?
+
+	trans = 0.0 if limit_switch_in and not limit_switch_out else 1.0
+
+	def calc_lightpath_state(self, limit_switch_in=None, limit_switch_out=None):
+		status = LightpathState(
+			inserted=limit_switch_in and not limit_switch_out,
```
tangkong, line 65: While this is technically fine, I think it's possible to just let `inserted = bool(limit_switch_in)` and `removed=bool(limit_switch_out)`.  Lightpath will report the state as inconsistent or unknown if both inserted and removed are true or false respectively.  

In fact, it's important to capture that behavior.  As it is now, the inconsistent state (True-True) would resolve as unknown (False-False).  If you don't want this device to ever report "inconsistent", I guess this could be reasonable

tangkong (commented):
Left a couple thoughts, but most notably you should be using 4 spaces instead of tabs.

ZLLentz (commented):
I think this is the right direction. We can use this to handle all the pneumatics that use these FBs, such as some of the stoppers. James Cryan in tmo could use this class as-is to `.insert` and `.retract` his in-hutch stopper, for example.

`.remove` is the common name for the "get the thing out" function but it doesn't matter that much to me.

I'm not such sure how much of the e.g. interlock, error handling, etc. needs to be implemented in the first pass.

silkenelson:
I care a little about standard names: if we 'remove' is standard, could we change 'retract' to 'remove'?

nrwslac:
> I care a little about standard names: if we 'remove' is standard, could we change 'retract' to 'remove'?

Will do.

pcdsdevices/pneumatic.py
nrwslac: this is done.

pcdsdevices/pneumatic.py
nrwslac: I threw some exceptions.

pcdsdevices/pneumatic.py
nrwslac: Light path is doing stuff here, no need to check it manually. Updated.

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,78 @@
...
+	error = Cpt(EpicsSignalRO, ':PLC:bError')
+	error_id = Cpt(EpicsSignalRO, ':PLC:nErrorId')
+	error_message = Cpt(EpicsSignalRO, ':PLC:sErrorMessage')
+	position_state = Cpt(EpicsSignalRO, ':nPositionState')
+
+	def insert(self):
+		"""
+		Method for inserting Beckhoff Pneumatic Actuator
+		"""
+		if self.insert_ok:
```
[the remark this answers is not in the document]
nrwslac: fixed this bug.

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,78 @@
...
+			self.insert_signal.put(1)
+		else:
+			pass
+			#throw?
+		
+	def retract(self):
+		if self.retract_ok:
```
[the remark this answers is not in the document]
nrwslac: fixed this bug.

ZLLentz:
This is really good as-is. I'll annotate two more small things but this would already be useful in the tmo/rix.

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,69 @@
...
+"""
+
+from lightpath import LightpathState
+from ophyd import Component as Cpt
+from ophyd import EpicsSignal, EpicsSignalRO
+
+from pcdsdevices.interface import LightpathMixin
+
+
+class BeckhoffPneumatic(LightpathMixin):
```
ZLLentz, line 15: Adds some standard utilities like the terminal pretty-print
```suggestion
class BeckhoffPneumatic(BaseInterface, LightpathMixin):
```

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,69 @@
...
+Pneumatic Classes.
+
+This Module contains all the classes relating to Pneumatic Actuators
+"""
+
+from lightpath import LightpathState
+from ophyd import Component as Cpt
+from ophyd import EpicsSignal, EpicsSignalRO
+
+from pcdsdevices.interface import LightpathMixin
```
ZLLentz, line 12: import for the interface utilities
```suggestion
from pcdsdevices.interface import BaseInterface, LightpathMixin
```

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,69 @@
...
+    position_state = Cpt(EpicsSignalRO, ':nPositionState')
+
+    def insert(self):
+        """
+        Method for inserting Beckhoff Pneumatic Actuator
+        """
+        if self.insert_ok.get():
+            self.insert_signal.put(1)
+        else:
+            raise RuntimeError("Insertion not allowed")
```
ZLLentz, line 53: More precise message here, maybe?

https://github.com/pcdshub/lcls-twincat-motion/blob/447abc92b08f778b56b2ef828a904e352f9b0410/lcls-twincat-motion/Library/DUTs/DUT_MotionPneumaticActuator.TcDUT#L62
```suggestion
            raise RuntimeError("Insertion not permitted by PLC")
```

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def remove(self):
+        if self.retract_ok.get():
+            self.retract_signal.put(1)
+        else:
+            raise RuntimeError("Removal not allowed")
```
ZLLentz, line 59: Same precision suggestion as above
```suggestion
            raise RuntimeError("Removal not permitted by PLC")
```

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,69 @@
...
+    insert_signal = Cpt(EpicsSignal, 'CMD:IN')
+    retract_signal = Cpt(EpicsSignal, 'CMD:OUT')
+
+    # returns
+    busy = Cpt(EpicsSignalRO, ':bBusy')
+    done = Cpt(EpicsSignalRO, ':bDone')
+    reset = Cpt(EpicsSignal, ':bReset')
+    error = Cpt(EpicsSignalRO, ':PLC:bError')
+    error_id = Cpt(EpicsSignalRO, ':PLC:nErrorId')
+    error_message = Cpt(EpicsSignalRO, ':PLC:sErrorMessage')
```
ZLLentz, line 43: It's not really needed for today, but you could imagine some future version of this that waits for the busy/done signals to finish when you insert or retract and reports the error_message from the PLC

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,69 @@
...
+    position_state = Cpt(EpicsSignalRO, ':nPositionState')
```
ZLLentz, line 44: Maybe raise this one to hinted since it's the most important thing to know (says inserted/retracted/moving/invalid)
```suggestion
    position_state = Cpt(EpicsSignalRO, ':nPositionState', kind='hinted')
```

ZLLentz (commented):
Small suggestion barrage

pcdsdevices/pneumatic.py
nrwslac: done.

pcdsdevices/pneumatic.py
nrwslac: done.

pcdsdevices/pneumatic.py
nrwslac: done.

pcdsdevices/pneumatic.py
nrwslac: done.

pcdsdevices/pneumatic.py
nrwslac: done.

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,69 @@
+
```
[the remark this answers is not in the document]
nrwslac: done.

pcdsdevices/pneumatic.py
nrwslac: Would this be blocking? or some sort of thread code that code goes to sleep?

nrwslac: it should be pretty quick to write a loop with a time limit to do this.

nrwslac: implemented a loop with a timeout. wdy think?

ZLLentz: A full implementation would have something like
```
def insert(self, wait: bool = False, timeout: float = 10.0) -> Status:
    ...
```
And then you'd use the ophyd status objects to communicate the error states.

The loop is solid but:
- you'd want it to be optional
- re-use ophyd move kwargs like `wait` and `timeout` for familiarity
- should be a subroutine that's used in both insert and remove

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,84 @@
...
+            time_elapsed = 0
+            while (time_elapsed < time_out):
+                time_elapsed = time.time() - start_time
+                if not self.done.get() and self.busy.get():
+                    continue
+                else:
+                    break
+            if self.error.get():
+                raise RuntimeError(self.error_message.get())
+            if time_elapsed == time_out:
```
ZLLentz, line 64:
```suggestion
            if time_elapsed >= time_out:
```

nrwslac: updated this.

pcdsdevices/pneumatic.py
nrwslac: Okay, let me take a look at these suggestions. Thank you.

pcdsdevices/pneumatic.py
```diff
@@ -45,27 +43,32 @@ class BeckhoffPneumatic(BaseInterface, LightpathMixin):
...
-            if self.error.get():
-                raise RuntimeError(self.error_message.get())
-            if time_elapsed >= time_out:
-                raise RuntimeError("Insertion Failed Timed Out")
-
+            # three ways to leave this loop
+            # 1 true done and not busy
+            # 2 error
+            # 3 timeout
+            while (not self.done.get() and self.busy.get()):
```
nrwslac, line 61: If I understand status correctly, one instantiated, it will keep its own count and raise an exception if it expires.

pcdsdevices/pneumatic.py
nrwslac: I re-implemented this using status. Will port this so a callable routine so remove and also use it if this looks good.

pcdsdevices/pneumatic.py
ZLLentz: Yeah, that's correct. `Status` does a lot of the timeout stuff for you and has helpful methods like `status.wait`. It can also be used to communicate error status without blocking, since you can make a background thread or a subscription callback check for errors and apply them to the `Status` object outside of the main thread.

A common pattern is:
- subscribe to self.done with a function that marks `Status` as finished or sets the exception, then unsubscribes itself
- optionally wait on the status object
- return the status object

Another common pattern is:
- create a thread the waits for done in a loop with some sleeps thrown in, then marks `Status` as finished or sets the exception
- optionally wait on the status object
- return the status object

All of this isn't needed to just make the pneumatic work but it's the most complete way in ophyd to report error states. Otherwise, our main asynchronous error reporting tool (for the `wait=False` case) would be logging and print statements which aren't super useful for responding to errors programatically.

nrwslac: Added a sketch of what I think the first pattern should look like. Not sure how I access the PV value in the call back subscription.

pcdsdevices/pneumatic.py
```diff
@@ -46,28 +46,28 @@ class BeckhoffPneumatic(BaseInterface, LightpathMixin):
...
-            # 1 true done and not busy
-            # 2 error
-            # 3 timeout
-            while (not self.done.get() and self.busy.get()):
+        def callback(value):
+            if value:
                 if self.error.get():
                     error = Exception(self.error_message.get())
                     status.set_exception(error)
-                    return status
```
nrwslac: exception raised, no need to return.

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,87 @@
...
+    error_message = Cpt(EpicsSignalRO, ':PLC:sErrorMessage')
+    position_state = Cpt(EpicsSignalRO, ':nPositionState', kind='hinted')
+
+    def insert(self, wait: bool = False, timeout: float = 10.0) -> Status:
+        """
+        Method for inserting Beckhoff Pneumatic Actuator
+        """
+        status = Status(self, timeout)
+
+        def callback(value):
```
nrwslac, line 52: def check_value(*, old_value, value, **kwargs): (should add default call back args)

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,87 @@
...
+            if value:
+                if self.error.get():
+                    error = Exception(self.error_message.get())
+                    status.set_exception(error)
+                else:
+                    status.set_finished()
+                    self.done.clear_sub(callback)
```
ZLLentz, line 59: This should clear on both success and on failure

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,87 @@
...
+
+        self.done.subscribe(callback)
```
ZLLentz, line 61: This should be done after the insert_ok check and right before we put to the insert_signal
You want to avoid starting the subscription here if we don't even do a move

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,87 @@
...
+            self.insert_signal.put(1)
+        else:
+            error = Exception("Insertion not permitted by PLC")
+            status.set_exception(error)
+
+        if wait:
+            status.wait()
+        return status
+
+    def remove(self):
```
ZLLentz, line 73: last thing to consider is doing the checks on both insert or remove, maybe with some shared methods to deduplicate

pcdsdevices/pneumatic.py
nrwslac: good point.

pcdsdevices/pneumatic.py
nrwslac: make status an attribute so it can be accessed by the callback.

ZLLentz (commented):
The logic looks fully correct to me!

I guess all that's left is to find a pneumatic to try it on
Bonus points would be for figuring out how to set up a useful unit test (one that's worth the hassle)

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,97 @@
...
+    # commands
+    insert_signal = Cpt(EpicsSignal, 'CMD:IN')
+    retract_signal = Cpt(EpicsSignal, 'CMD:OUT')
+
+    # returns
+    busy = Cpt(EpicsSignalRO, ':bBusy')
+    done = Cpt(EpicsSignalRO, ':bDone')
+    reset = Cpt(EpicsSignal, ':bReset')
+    error = Cpt(EpicsSignalRO, ':PLC:bError')
+    error_id = Cpt(EpicsSignalRO, ':PLC:nErrorId')
```
klauer, line 42: Are these not pytmc-generated? If they are, all these read-only signals should have a `_RBV` at the end of them or you'll get the setpoint at best (or a pv-not-found at worst)

ZLLentz: One way to handle this is to use the PytmcSignal https://github.com/pcdshub/pcdsdevices/blob/525a89eec8a54fd71d784052542daec8813a9d29/pcdsdevices/signal.py#L38

nrwslac: I went with PytmcSingal. Thanks for catching that Ken and Zach.

nrwslac:
I think this is ready to go. Would like to know how to better simulate the tests, but that can happen later.

pcdsdevices/pneumatic.py
```diff
@@ -0,0 +1,98 @@
...
+    insert_signal = Cpt(PytmcSignal, ':CMD:IN', io="io")
+    retract_signal = Cpt(PytmcSignal, ':CMD:OUT', io="io")
+
+    # returns
+    busy = Cpt(PytmcSignal, ':bBusy', io="i")
+    done = Cpt(PytmcSignal, ':bDone', io="i")
+    reset = Cpt(PytmcSignal, ':bReset', io="io")
+    error = Cpt(PytmcSignal, ':PLC:bError', io="i")
+    error_id = Cpt(PytmcSignal, ':PLC:nErrorId', io="i")
+    error_message = Cpt(PytmcSignal, ':PLC:sErrorMessage', io="i")
```
ZLLentz, line 44: It looks to me like this is returning as an array of ints instead of as a string, probably because the PV is a char waveform
```suggestion
    error_message = Cpt(PytmcSignal, ':PLC:sErrorMessage', io="i", string=True)
```

docs/source/upcoming_release_notes/1083-DUT_MotionPneumaticActuator.rst
```diff
@@ -0,0 +1,31 @@
+1083 DUT_MotionPneumaticActuator
+#################
+
+API Changes
+-----------
+- Add module pneumatic.py
+- Add test test_pneumatic.py
```
ZLLentz, line 7: Nitpick: this section is really for API breaks, like removing modules. Maybe I should update the template.
```suggestion
- N/A
```

ZLLentz (approved):
I think I'm all aboard on merging this one

nrwslac:
Is Travis down? @ZLLentz. I see a PR going on: https://github.com/pcdshub/pcdsdevices/pull/1107

ZLLentz:
I'm going to merge this as-is despite the inbetween-CI-state sort of thing going on.
- the pip tests pass, which indicates that there's no real problem with the code
- the CI builds are transitioning over to GHA
- other tests that are failing are unrelated to the updates here

I'll take responsibility and make a follow-up PR if need be for CI-related issues.

PR #1090, opened 2022-12-05
Added .Leviton as a container, created device classes for Leviton PVs…

ljansen7:
… in fms.py

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/fms.py
+++ b/pcdsdevices/fms.py
@@ -31,8 +31,10 @@ CXI:SETRA:01:CH5_MASS 60 monitor
 """
 
 
-from ophyd import Component as Cpt, Device, EpicsSignalRO
-from ophyd import FormattedComponent as FCpt, EpicsSignalRO, Device
+from ophyd import Component as Cpt
+from ophyd import Device, EpicsSignalRO
+from ophyd import FormattedComponent as FCpt
+
 
 class Setra5000(Device):
     Timestamp = Cpt(EpicsSignalRO, ':TIME', kind='hinted')
@@ -42,13 +44,13 @@ class Setra5000(Device):
     SampleDuration = Cpt(EpicsSignalRO, ':SAMPLE_TIME_RBV', kind='hinted')
     SampleMode = Cpt(EpicsSignalRO, ':SAMPLE_MODE_RBV', kind='hinted')
     MassMode = Cpt(EpicsSignalRO, ':MASS_MODE_RBV', kind='hinted')
-    
+
     Temperature = Cpt(EpicsSignalRO, ':TEMP', kind='hinted')
     RH = Cpt(EpicsSignalRO, ':RH', kind='hinted')
     BarometricPressure = Cpt(EpicsSignalRO, ':BP', kind='hinted')
     FlowRate = Cpt(EpicsSignalRO, ':FLOW_RATE', kind='hinted')
     CO2 = Cpt(EpicsSignalRO, ':CO2', kind='hinted')
-    TVOC = Cpt(EpicsSignalRO, ':TVOC', kind='hinted')    
+    TVOC = Cpt(EpicsSignalRO, ':TVOC', kind='hinted')
 
     '''
     ch0_size = Cpt(EpicsSignalRO, ':CH0_SIZE', kind='normal')
@@ -70,31 +72,30 @@ class Setra5000(Device):
     # kind=hinted, normal, config, omitted
 
 
-
- #Power Distribution Units: Temperature, Humidity, and Load Feed
-
-#PDU_Temp1: Will be for racks with 2 PDUs, with total of 3 sensors for monitoring purposes
+# Power Distribution Units: Temperature, Humidity, and Load Feed
+# PDU_Temp1: Will be for racks with 2 PDUs, with total of 3 sensors for monitoring purposes
 class PDU_Temp1(Device):
     Sensor1 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[0]}:Sensor:1:GetTempValue', kind='hinted')
     Sensor2 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[1]}:Sensor:2:GetTempValue', kind='hinted')
     Sensor3 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[2]}:Sensor:1:GetTempValue', kind='hinted')
     Sensor4 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[3]}:Sensor:2:GetTempValue', kind='hinted')
 
-    def __init__(self, *args, elevations = [], **kwargs):
+    def __init__(self, *args, elevations=[], **kwargs):
         self.elevation = elevations
         super().__init__(*args, **kwargs)
 
-#PDU_Temp2: Will be for portable racks or Racks that have a single PDU, with a total of 2 sensors for monitoring
+
+# PDU_Temp2: Will be for portable racks or Racks that have a single PDU, with a total of 2 sensors for monitoring
 class PDU_Temp2(Device):
     Sensor1 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[0]}Sensor:1:GetTempValue', kind='hinted')
     Sensor2 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[1]}Sensor:2:GetTempValue', kind='hinted')
 
-    def __init__(self, *args, elevations = [], **kwargs):
+    def __init__(self, *args, elevations=[], **kwargs):
         self.elevation = elevations
         super().__init__(*args, **kwargs)
 
-#PDU_Temp3: Will be for PDUs that are linked, 1=Master, 2=Child
 
+# PDU_Temp3: Will be for PDUs that are linked, 1=Master, 2=Child
 class PDU_Temp3(Device):
     Sensor1 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[0]}:Sensor:1:GetTempValue', kind='hinted')
     Sensor2 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[1]}:Sensor:2:GetTempValue', kind='hinted')
@@ -102,10 +103,11 @@ class PDU_Temp3(Device):
     Sensor4 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[3]}:Sensor:2:GetTempValue', kind='hinted')
     Sensor5 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[4]}:Sensor:1:GetTempValue', kind='hinted')
     Sensor6 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[5]}:Sensor:2:GetTempValue', kind='hinted')
-       
-    def __init__(self, *args, elevations = [], **kwargs):
+
+    def __init__(self, *args, elevations=[], **kwargs):
         self.elevation = elevations
         super().__init__(*args, **kwargs)
+
 
 class PDU_Temp4(Device):
     Sensor1 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[0]}:Sensor:1:GetTempValue', kind='hinted')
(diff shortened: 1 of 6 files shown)


pcdsdevices/fms.py
```diff
@@ -0,0 +1,259 @@
...
+CXI:SETRA:01:CH1_MASS 60 monitor
+CXI:SETRA:01:CH2_SIZE 60 monitor
+CXI:SETRA:01:CH2_MASS 60 monitor
+CXI:SETRA:01:CH3_SIZE 60 monitor
+CXI:SETRA:01:CH3_MASS 60 monitor
+CXI:SETRA:01:CH4_SIZE 60 monitor
+CXI:SETRA:01:CH4_MASS 60 monitor
+CXI:SETRA:01:CH5_SIZE 60 monitor
+CXI:SETRA:01:CH5_MASS 60 monitor
+"""
```
tangkong, line 31: We should replace these notes with a more informative docstring for the module.  Maybe something like:
```
"""
Module to define classes for the Facility Monitoring System (FMS).
"""
```

Though more detail is always appreciated

pcdsdevices/fms.py
```diff
@@ -0,0 +1,259 @@
...
+    ch2_size = Cpt(EpicsSignalRO, ':CH2_SIZE', kind='normal')
+    ch2_mass = Cpt(EpicsSignalRO, ':CH2_MASS', kind='normal')
+    ch3_size = Cpt(EpicsSignalRO, ':CH3_SIZE', kind='normal')
+    ch3_mass = Cpt(EpicsSignalRO, ':CH3_MASS', kind='normal')
+    ch4_size = Cpt(EpicsSignalRO, ':CH4_SIZE', kind='normal')
+    ch4_mass = Cpt(EpicsSignalRO, ':CH4_MASS', kind='normal')
+    ch5_size = Cpt(EpicsSignalRO, ':CH5_SIZE', kind='normal')
+    ch5_mass = Cpt(EpicsSignalRO, ':CH5_MASS', kind='normal')
+    Total_Mass = Cpt(EpicsSignalRO, ':TOT_MASS', kind='normal')
+    '''
```
tangkong, line 67: If these components are no longer being used, we can just delete them.

pcdsdevices/fms.py
```diff
@@ -0,0 +1,259 @@
...
+
+    # kind, doc are important keywords for later
+    # kind=hinted, normal, config, omitted
+
+
+
+ #Power Distribution Units: Temperature, Humidity, and Load Feed
+
+#PDU_Temp1: Will be for racks with 2 PDUs, with total of 3 sensors for monitoring purposes
+class PDU_Temp1(Device):
```
tangkong, line 77:
```suggestion
class PDU_Temp1(Device):
    """ A rack with 2 PDUs, with total of 3 sensors for monitoring purposes """
```

Moving these comments into the class as a docstring will let us connect this information to the device when it's created later on.  This should be done throughout the file.

pcdsdevices/fms.py
```diff
@@ -0,0 +1,259 @@
...
+    Sensor1 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[0]}:Sensor:1:GetTempValue', kind='hinted')
+    Sensor2 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[1]}:Sensor:2:GetTempValue', kind='hinted')
+    Sensor3 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[2]}:Sensor:1:GetTempValue', kind='hinted')
+    Sensor4 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[3]}:Sensor:2:GetTempValue', kind='hinted')
+
+    def __init__(self, *args, elevations = [], **kwargs):
+        self.elevation = elevations
+        super().__init__(*args, **kwargs)
+"""
+#PDU_Temp1(elevations = [15, 15, 32])
```
tangkong, line 135: Old comments / notes can be deleted

pcdsdevices/happi/containers.py
```diff
@@ -665,3 +665,15 @@ class EnvironmentalMonitor(LCLSItem):
                        optional=True, enforce=str)
     ioc_base = EntryInfo('Base PV of the EK9000 IOC', optional=True,
                          enforce=str)
+
+
+from pcdsdevices.happi.containers import LCLSItem
```
tangkong, line 670:
```suggestion
```

`LCLSItem` is defined in this module, you don't need to import it.

tangkong (changes requested):
Thanks for jumping through the github hoops!  I'm relatively certain that the lucid screens are working to your liking, which is always the main target.  

There's a bit of formatting / style issues to address to get this PR ready to merge, but on the whole I think the approach we took here is the one that makes sense for us.  

First, you should run the pre-commit on your files.  these should be configured for pcdsdevices already, so in your local pcdsdevices repo run ([confluence docs](https://confluence.slac.stanford.edu/display/PCDS/How+to+add+a+device+to+pcdsdevices)):

```console
(dev) $ pre-commit install  # run this only once
(dev) $ git add pcdsdevices/my_device.py   # add new things to the staging area
(dev) $ pre-commit run  # manually run pre-commit
(dev) $ git commit -m "ENH: add my device"
# This may fail here because pre-commit says you have issues to solve.# You may need to go and fix those issues and then re-add my_device.py
# If you still can't figure out the problem and want to commit regardless of the failures, you can use the following:
(dev) $ git commit -n -m "ENH: add my device
```

pcdsdevices/fms.py
```diff
@@ -0,0 +1,259 @@
...
+
+from ophyd import Component as Cpt, Device, EpicsSignalRO
+from ophyd import FormattedComponent as FCpt, EpicsSignalRO, Device
+
+class Setra5000(Device):
+    Timestamp = Cpt(EpicsSignalRO, ':TIME', kind='hinted')
+    Date = Cpt(EpicsSignalRO, ':DATE', kind='hinted')
+
+    SampleState = Cpt(EpicsSignalRO, ':SAMP_STATE', kind='hinted')
+    SampleDuration = Cpt(EpicsSignalRO, ':SAMPLE_TIME_RBV', kind='hinted')
```
ZLLentz, line 42: Are these NamedLikeThis for the screens or something like that?
Usually we do e.g. `sample_duration` for these python attributes. But maybe there's a special reason for it? I'm interested because maybe we want to change how typhos displays these names...

pcdsdevices/fms.py
ZLLentz: It's also possible to keep them around but set `kind='omitted'` if we think "someone" would find them useful (but not us)

pcdsdevices/fms.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+ #Power Distribution Units: Temperature, Humidity, and Load Feed
+
+#PDU_Temp1: Will be for racks with 2 PDUs, with total of 3 sensors for monitoring purposes
+class PDU_Temp1(Device):
+    Sensor1 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[0]}:Sensor:1:GetTempValue', kind='hinted')
+    Sensor2 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[1]}:Sensor:2:GetTempValue', kind='hinted')
+    Sensor3 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[2]}:Sensor:1:GetTempValue', kind='hinted')
+    Sensor4 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[3]}:Sensor:2:GetTempValue', kind='hinted')
```
ZLLentz, line 81: Optional simplification: `FormattedComponent` is smart enough to assume self. So self doesn't need to be included in any of these. Optional of course, it works as-is so you don't have to change anything.

https://github.com/bluesky/ophyd/blob/f9be38c9b1de6338b4586b55eea55e74f3e18b2b/ophyd/device.py#L382-L383

pcdsdevices/fms.py
```diff
@@ -0,0 +1,259 @@
...
+#PDU_Temp2: Will be for portable racks or Racks that have a single PDU, with a total of 2 sensors for monitoring
+class PDU_Humidity2(Device):
+    Sensor1 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[0]}Sensor:1:GetHumidValue', kind='hinted')
+    Sensor2 = FCpt(EpicsSignalRO, '{self.prefix}:{self.elevation[1]}Sensor:2:GetHumidValue', kind='hinted')
+
+    def __init__(self, *args, elevations = [], **kwargs):
+        self.elevation = elevations
+        super().__init__(*args, **kwargs)
+
+class PDU_Humidity3(Device):
```
ZLLentz, line 156: The class numbering scheme here is a bit confusing

pcdsdevices/fms.py
```diff
@@ -0,0 +1,259 @@
...
+   WaterFlowRate_gpm = Cpt(EpicsSignalRO, ':WaterFlowRate_CALC', kind='hinted')
+   AirTempInTop_F = Cpt(EpicsSignalRO, ':AirTempInTop_CALC', kind='hinted')
+   AirTempInMid_F = Cpt(EpicsSignalRO, ':AirTempInMid_CALC', kind='hinted')
+   AirTempInBot_F = Cpt(EpicsSignalRO, ':AirTempInBot_CALC', kind='hinted')
+   AirTempOutTop_F = Cpt(EpicsSignalRO, ':AirTempOutTop_CALC', kind='hinted')
+   AirTempOutMid_F = Cpt(EpicsSignalRO, ':AirTempOutMid_CALC', kind='hinted')
+   AirTempOutBot_F = Cpt(EpicsSignalRO, ':AirTempOutBot_CALC', kind='hinted')
+   FrontDoorAccess = Cpt(EpicsSignalRO, ':FrontDoorAccess_CALC', kind='hinted')
+   RearDoorAccess = Cpt(EpicsSignalRO, ':RearDoorAccess_CALC', kind='hinted')
+
```
ZLLentz, line 259: We can probably cut this file's repetition down a lot with some inheritance. A lot of these devices are just "that existing device, but more", which can be expressed as:
```
class Device1ButMore(Device1):
    more = Cpt(...)
```
Sometimes this makes the file harder to read and debug, so it's a judgement call as to when it's appropriate.

pcdsdevices/happi/containers.py
```diff
@@ -665,3 +665,15 @@ class EnvironmentalMonitor(LCLSItem):
...
+  
+class Leviton(LCLSItem):
+    kwargs = deepcopy(LCLSItem.kwargs)
+    kwargs.default['elevations'] = "{{elevations}}"
+    elevations = EntryInfo(doc = 'List of elevation numbers for rack',
```
ZLLentz, line 675: It took me a while to understand how you were able to cut down the class count so much. This elevations arg is a good solution to that :+1:.

ZLLentz (commented):
This is a huge improvement organizationally over the previous unmerged Leviton1, Leviton2 Leviton3, ... Leviton83 that leaked into confluence and grabbed Alex's attention.

I put in some suggestions/questions/food for thought and I agree with Robert's comments.

Setups like this make it pretty clear that ophyd's rigidity can be a bummer sometimes and make me wonder if we should make a more flexible version (that loses the advantages of rigidity in favor of letting you decide how many Sensors there should be on the fly).

We can decide whether the best path forward is:
- Merge now and update in a follow-up
- Update first and merge once comments are addressed

pcdsdevices/fms.py
ljansen7: What does it mean for self to be assumed by FormattedComponent?

pcdsdevices/fms.py
ljansen7: What aspect of it is confusing/counterintuitive?

pcdsdevices/fms.py
ZLLentz: It means you can leave it out and it knows to check `self.`. So, for example, Sensor4 here could have equivalently been:
```
Sensor4 = FCpt(EpicsSignalRO, '{prefix}:{elevation[3]}:Sensor:2:GetTempValue', kind='hinted')
```

pcdsdevices/fms.py
ZLLentz: These are ordered sequentially by where they are placed in the file with no class function meanings.

tangkong: I took the number in the class name to mean the number of components in the class.  Is that right?

ZLLentz: As in: class 1 has 4 elements, class 2 has 2, class 3 has 6, etc.

ljansen7: Yes, you're right that its counterintuitive in that regard for both the Humidity and Temp classes. A solution for that could be switching the class 1 to have 2 elements, class 2 to have 4 elements, and class 3 to have 6? However, I am worried that switching the numbers would mean re-editing the device_class for the majority of the happi entries, which is quite a bit of work. Would this be an important edit?

tangkong: It wouldn't be too much work if you edited the db file directly, it could be as simple as a series of find-replace commands.  I think it's important to try to be as logical as possible with naming, as any small confusions we have now will be exaggerated in the future.  

I can help you poke around the actual database file

ljansen7: Ok good to know, thank you! So then what naming scheme would be best for temp and humidity? Humidity1 for 2 entries, Humidity2 for 4, and Humidity3 for 6? Or should we make it so that the number directly corresponds to the number of sensors we are describing in that device class (Humidity2, Humidity4, Humidity6)?

ZLLentz: This is always a judgement call. I think making them the number of sensors is the most intuitive because then your class name is self-describing instead of relying on the context of the other classes. Either way is OK though because they seem to come in pairs anyway.

pcdsdevices/fms.py
ljansen7: I don't think there is a particular reason for it. I changed the names in the fms.py file to match the standard you are referring to (sample_duration). What would need to be changed in typhos/what is typhos for?

ZLLentz:
Hi Lana, I see you've marked all the conversations as resolved, but I don't see any updates. Did you commit and push everything?

pcdsdevices/fms.py
ZLLentz: Typhos automatically generates a screen based on the text here and doesn't need any specific changes.

tangkong (approved):
This looks good to me.  A few nitpicks that we can fix up later, but for now this works in conjunction with modifications to the deployed happi database to load the fms LUCID screen.  👍 
I added pre-release notes to help us with documentation in the future.

ZLLentz:
Yeah there's some cleanup to do e.g. pep8 https://peps.python.org/pep-0008/#function-and-variable-names and some wayward comments and ordering but otherwise it's sound and shouldn't need any structural changes

Still much sadness that it wasn't simpler to do variable-sized classes in ophyd

tangkong:
Alright, looks good to merge then.  Thanks @ljansen7 !

PR #1097, opened 2022-12-20
FIX: pseudopos readback desync issues

ZLLentz:
## Description

Make `PseudoPositioner` devices defined in this module only write to the `ophyd_readback` signal if they were the most recent device to write to the `ophyd_setpoint` signal.

## Motivation and Context

If multiple python sessions are run with slightly different versions of the same pseudo positioner, they can fight over control of the ophyd_readback signal.

see https://jira.slac.stanford.edu/browse/LCLSECSD-1761

## How Has This Been Tested?

I added a unit test here. This should also be tested in a hutch. It's possible that I'll need to make some tweaks.

Adding some release notes

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/pseudopos.py
+++ b/pcdsdevices/pseudopos.py
@@ -125,6 +125,9 @@ class PseudoPositioner(ophyd.pseudopos.PseudoPositioner):
     """ + ophyd.pseudopos.PseudoPositioner.__doc__
 
     def __init__(self, *args, **kwargs):
+        self.my_move = False
+        self._move_time = 0
+        self._my_move_timeout = 10
         super().__init__(*args, **kwargs)
 
         if len(self.RealPosition._fields) == 1:
@@ -132,10 +135,6 @@ class PseudoPositioner(ophyd.pseudopos.PseudoPositioner):
 
         if len(self.PseudoPosition._fields) == 1:
             self.PseudoPosition.__float__ = _as_float
-
-        self.my_move = False
-        self._move_time = 0
-        self._my_move_timeout = 10
 
     def _update_notepad_ioc(self, position, attr):
         """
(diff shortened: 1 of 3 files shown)


ZLLentz:
I'll ask for reviews in the new year as to not distract vacation time!

klauer (approved):
This is pretty simple and looks good to me 👍 

I had a longer comment thinking through how this would work and possible edge cases, but the short of it is that I don't think there's any real concern and this is about as simple/effective as it could get.

ZLLentz:
I think I fixed the test suite issue that I caused, it might take some time to be allowed test this on real hardware for edge cases but I'm pretty confident that the testing is sufficient.

tangkong (approved):
This looks good to me too.  I was halfway to reviewing this earlier but got distracted, sorry

pcdsdevices/pseudopos.py
```diff
@@ -221,6 +230,35 @@ def set_current_position(self, position):
...
+        for positioner in self._pseudo:
+            try:
+                signal = getattr(positioner, 'notepad_setpoint', None)
+                if signal is None:
+                    continue
+                signal.subscribe(self._our_move_check)
+            except Exception as ex:
+                self.log.debug(
+                    'Failed to set up _our_move_check subscriptions.',
+                    exc_info=ex,
```
ZLLentz, line 259: Haha one last small thing on self-review
This is a bool kwarg and I'm passing an exception instance
This "works" but feels wrong

ZLLentz: TIL this actually also optionally takes an exception instance which in this case does the same thing but in other cases may not

klauer: Right, for reference - https://docs.python.org/3/library/logging.html
```
Changed in version 3.5: The exc_info parameter can now accept exception instances.
```

PR #1099, opened 2022-12-21
ENH: VAT590 needle valve

ZLLentz:
## Description

Add a class for controlling the VAT590 needle valve.
For now, it's just the EPICS control points.

## Motivation and Context

There are data types for this in TwinCAT and one valve installed in TMO: https://jira.slac.stanford.edu/browse/LCLSECSD-1026

## How Has This Been Tested?

Not at all, the IOC for this is offline

## Where Has This Been Documented?

Pre-release notes only

## Pre-merge checklist
- [ ] Code works interactively
- [x] Code contains descriptive docstrings, including context and API
- [x] New/changed functions and methods are covered in the test suite where possible
- [x] Test suite passes locally
- [ ] Test suite passes on travis
- [x] Ran docs/pre-release-notes.sh and created a pre-release documentation page
- [x] Pre-release docs include context, functional descriptions, and contributors as appropriate

PR #1100, opened 2023-01-04
CI: try numpy 1.23 per recommendation

klauer:
## Description
See #1099 (https://github.com/pcdshub/pcdsdevices/pull/1099#issuecomment-1370374728)

## Motivation and Context
Closes #1100

## How Has This Been Tested?
On Travis CI. The first build passed, indicating this should be a suitable (temporary) bandage.

Pre-release notes

## Pre-merge checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ env:
     - CONDA_RECIPE_FOLDER="conda-recipe"
     # Extra dependencies needed to run the tests which are not included
     # at the recipe and dev-requirements.txt. E.g. PyQt
-    - CONDA_EXTRAS="pip"
+    - CONDA_EXTRAS="pip numpy=1.23"
     # Requirements file with contents for tests dependencies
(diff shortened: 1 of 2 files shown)


tangkong (approved):
My hope was that we wouldn't have to wait too long for this to be fixed, since we would end up reverting the change in a couple days.  But perhaps that's too long to let these PR's fail CI.

ZLLentz (approved):
Should be good to merge, I'll add an additional comment for context based on our slack discussions

ZLLentz:
Some dependency tree context:

- `numba` is required by `sparse` which is required by `tiled` as of late November last year
- `tiled` is required by `databroker` which is required by `bluesky` which is required by `pcdsdevices`
- `pcdsdevices` uses `bluesky` just to grab the progress bar
- `bluesky` depends on `databroker` only in the conda forge recipe

PR #1102, opened 2023-01-11
REF: GitHub Actions and migration to setuptools-scm, pyproject.toml

klauer:
## Description
Migration to GitHub Actions and updating to the latest standards.

This was a test case for https://github.com/pcdshub/pcds-python-migration-tools/pull/3 that turned into a reviewable/merge-able migration of the repository. The branch name and commit noise are the main unfortunate parts at this point...

```
- pcdsdevices no longer uses Travis CI and has migrated to GitHub Actions for
  continuous integration, testing, and documentation deployment.
- pcdsdevices has been migrated to use setuptools-scm, replacing versioneer, as
  its version-string management tool of choice.
- pcdsdevices has been migrated to use the modern ``pyproject.toml``, replacing
  ``setup.py`` and related files.
- Older language features and syntax found in the repository have been updated
  to Python 3.9+ standards by way of ``pyupgrade``.
- Sphinx 6.0 is now supported for documentation building.
- ``docs-versions-menu`` replaces ``doctr-versions-menu`` and ``doctr`` usage
  for documentation deployment on GitHub Actions.  The deployment key is now
  no longer required.
```

## How Has This Been Tested?
Continuous integration, test suite, and a bit locally as well.

## Pre-merge checklist


.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
-exclude = .git,__pycache__,build,dist,versioneer.py,pcdsdevices/_version.py,pcdsdevices/areadetector/*,docs/*
+exclude = .git,__pycache__,build,dist,pcdsdevices/_version.py,docs/source/conf.py
 max-line-length = 88
```
tangkong, line 3: There are a bunch of flake8 that suggest this config wasn't used by GHA.  I need to take a closer look but I imagine we should pull this file in each repo like the pre-commit-config is

klauer: Hmm, good catch. Odd...

Old: https://github.com/pcdshub/pcds-ci-helpers/blob/ffd48516b17c7ef3ac8fe075a5fc29b1947c1873/travis/pre_commit.sh#L14
New: https://github.com/pcdshub/pcds-ci-helpers/blob/ffd48516b17c7ef3ac8fe075a5fc29b1947c1873/.github/workflows/pre-commit.yml#L90-L93

On the plus side, GHA is _for some reason_ reporting status in this PR unlike when I opened the pcdsutils migration PR.

.pre-commit-config.yaml
```diff
@@ -23,11 +24,11 @@ repos:
     -   id: debug-statements
 
 -   repo: https://github.com/pycqa/flake8.git
-    rev: 3.9.2
+    rev: 6.0.0
```
klauer, line 27: Let's go back to 3.9.2 and try again:
```suggestion
    rev: 3.9.2
```

klauer:
Auto-converted `pyproject.toml` when installed in my local environment appears to still be OK with respect to happi containers:

```
$ happi add
...
pcdsdevices.happi.containers.Acromag
pcdsdevices.happi.containers.AreaDetector
pcdsdevices.happi.containers.Attenuator
pcdsdevices.happi.containers.BaslerCamera
...

$ wherep pcdsdevices
pcdsdevices= /Users/klauer/mc/envs/other/lib/python3.9/site-packages/pcdsdevices/__init__.py
version= 7.1.1.dev72+g92e09b8d2d
```

klauer:
OK, fixed the conda recipe + documentation building with pyproject.toml/setuptools-scm.

Where do we want to go from here? Should I clean this up in a new PR with or without some of the changes here (e.g., revert to `setup.py`)? (@ZLLentz @tangkong )

ZLLentz:
The 3.11 conda build fails here but I can make a pcdsdevices/py311 build locally... and the error is nonsensical here so that seems like a pain unravel (though it is far and away low low priority)

In general I'm :+1: on moving to "the newest thing", I'm happy to review this one as-is

pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+classifiers = [ "Programming Language :: Python :: 3",]
+description = "Ophyd Device definitions for LCLS Beamline components"
+dynamic = [ "version", "readme", "dependencies", "optional-dependencies",]
+keywords = []
+name = "pcdsdevices"
+requires-python = ">=3.9"
+
+[options]
+zip_safe = false
+include_package_data = true
```
ZLLentz, line 15: One oddity in the build output that might be worth noting:

<details>

```
packages/setuptools/command/build_py.py:[202](https://github.com/pcdshub/pcdsdevices/actions/runs/3905733272/jobs/6673094192#step:11:203): SetuptoolsDeprecationWarning:     Installing 'pcdsdevices.tests' as data is deprecated, please list it in `packages`.
      !!


      ############################
      # Package would be ignored #
      ############################
      Python recognizes 'pcdsdevices.tests' as an importable package,
      but it is not listed in the `packages` configuration of setuptools.

      'pcdsdevices.tests' has been automatically added to the distribution only
      because it may contain data files, but this behavior is likely to change
      in future versions of setuptools (and therefore is considered deprecated).

      Please make sure that 'pcdsdevices.tests' is included as a package by using
      the `packages` configuration field or the proper discovery methods
      (for example by using `find_namespace_packages(...)`/`find_namespace:`
      instead of `find_packages(...)`/`find:`).

      You can read more about "package discovery" and "data files" on setuptools
      documentation page.


  !!

    check.warn(importable)
  /home/runner/micromamba/conda-bld/pcdsdevices_1673554832717/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/lib/python3.11/site-packages/setuptools/command/build_py.py:202: SetuptoolsDeprecationWarning:     Installing 'pcdsdevices.tests.test_lens_sets' as data is deprecated, please list it in `packages`.
      !!


      ############################
      # Package would be ignored #
      ############################
      Python recognizes 'pcdsdevices.tests.test_lens_sets' as an importable package,
      but it is not listed in the `packages` configuration of setuptools.

      'pcdsdevices.tests.test_lens_sets' has been automatically added to the distribution only
      because it may contain data files, but this behavior is likely to change
      in future versions of setuptools (and therefore is considered deprecated).

      Please make sure that 'pcdsdevices.tests.test_lens_sets' is included as a package by using
      the `packages` configuration field or the proper discovery methods
      (for example by using `find_namespace_packages(...)`/`find_namespace:`
      instead of `find_packages(...)`/`find:`).

      You can read more about "package discovery" and "data files" on setuptools
      documentation page.


  !!

    check.warn(importable)
  /home/runner/micromamba/conda-bld/pcdsdevices_1673554832717/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/lib/python3.11/site-packages/setuptools/command/build_py.py:202: SetuptoolsDeprecationWarning:     Installing 'pcdsdevices.ui' as data is deprecated, please list it in `packages`.
      !!


      ############################
      # Package would be ignored #
      ############################
      Python recognizes 'pcdsdevices.ui' as an importable package,
      but it is not listed in the `packages` configuration of setuptools.

      'pcdsdevices.ui' has been automatically added to the distribution only
      because it may contain data files, but this behavior is likely to change
      in future versions of setuptools (and therefore is considered deprecated).

      Please make sure that 'pcdsdevices.ui' is included as a package by using
      the `packages` configuration field or the proper discovery methods
      (for example by using `find_namespace_packages(...)`/`find_namespace:`
      instead of `find_packages(...)`/`find:`).

      You can read more about "package discovery" and "data files" on setuptools
      documentation page.

  ```

</details>

**Comment by tangkong** [comment]
I was trying to think of reasons not to transition now, and keep coming back to the confusion PoC's/scientists will have when their branches fall out of date.  But if there's a time to tackle this it might be now, during a "lull".

File: pyproject.toml (hunk 3 above)
Comment (klauer) [reply pyproject.toml:15]: I think I've resolved it. Good catch! See 48ed02cbb1

File: .pre-commit-config.yaml [hunk 4]
```diff
@@ -1,17 +1,18 @@
...
+    rev: v4.4.0
     hooks:
     -   id: no-commit-to-branch
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
-        exclude: |
-                (?x)^(
-                    pcdsdevices/tests/test_lens_sets/.*|
-                    .*\.ui$
-                )$
```
ZLLentz, line 14: This needs to be reverted or else you get a changed file every time you run the test suite that gets changed back again when you run the pre-commit

pcdsdevices/tests/test_lens_sets/test
```diff
@@ -1 +1 @@
-[2, 0.0002, 4, 0.0005]
```
ZLLentz, line 1: This is the file that will get repeatedly changed back and forth by pre-commit vs the test suite in an endless battle

.pre-commit-config.yaml
klauer: I'm glad you're cautiously checking these - I got lost in rabbit holes and neglected the real intent of the PR at some point.

pcdsdevices/state.py
```diff
@@ -639,7 +641,7 @@ def state_config_dotted_names(state_count: int) -> list[Optional[str]]:
     ]
 
 
-def state_config_dotted_velos(state_count: int) -> list[Optional[str]]:
+def state_config_dotted_velos(state_count: int) -> list[str | None]:
```
ZLLentz, line 644: is `typing.Optional` out of style now?

pcdsdevices/tests/test_entrypoints.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import annotations
```
ZLLentz, line 1: Good call adding this test file in the same pr as the build changes

pcdsdevices/tests/test_lens_sets/test
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+[2, 0.0002, 4, 0.0005]
```
ZLLentz, line 1: I'm also confused as to why the build is happier with an `__init__.py` added into this directory- this directory didn't have any python files in it prior

pcdsdevices/tests/test_utils.py
```diff
@@ -26,7 +26,7 @@
 @pytest.fixture(scope='function')
 def sim_input(monkeypatch):
     master, slave = pty.openpty()
-    with open(slave, 'r') as fake_stdin:
+    with open(slave) as fake_stdin:
```
ZLLentz, line 29: should I be omitting the mode when it's `'r'` from now on?

pcdsdevices/tests/test_lens_sets/test
klauer: A directory without an `__init__.py` is an implicit namespace package. The presence of `__init__.py` in there disables that (see https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages).

I don't think the implicit namespace packages have a place for us, and the addition of `__init__.py` makes it obvious what should be importable (for setuptools and for us, I think). Maybe my perspective is dated, though.

pcdsdevices/state.py
klauer: I'm a bit more partial to `list[Optional[str]]` than `list[str | None]`, personally. 

`pyupgrade` uses https://peps.python.org/pep-0604/ wherever possible, though

pcdsdevices/tests/test_utils.py
klauer: It's the default (`open(..., mode='r')` so it's at least technically unnecessary

pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+
+[[project.authors]]
+name = "SLAC National Accelerator Laboratory"
+
+[project.license]
+file = "LICENSE.md"
+
+[project.scripts]
+
+[tool.setuptools_scm]
```
ZLLentz, line 25: This line from the output raised a flag and confused me but might be nothing:
https://github.com/pcdshub/pcdsdevices/actions/runs/3907251443/jobs/6676280720#step:11:800
```
Created wheel for pcdsdevices: filename=pcdsdevices-0.0.0-py3-none-any.whl
```

ZLLentz: This is only a mid step in the conda builds- in the pip builds it never throws out a 0.0.0

klauer: Another good catch - this may be an actual issue... 

I thought I had resolved these in the cookiecutter, but perhaps not... I'll double-check the build output for the latest update.

conda-recipe/meta.yaml
```diff
@@ -1,20 +1,23 @@
-{% set data = load_setup_py_data() %}
+{% set package_name = "pcdsdevices" %}
+{% set import_name = "pcdsdevices" %}
+{% set version = load_file_regex(load_file=os.path.join(import_name, "_version.py"), regex_pattern=".*version = '(\S+)'").group(1) %}
```
ZLLentz, line 4: Another small weird thing that isn't causing an issue in the gha conda build:

```
ClobberWarning: Conda was asked to clobber an existing path.
  source path: /home/runner/conda_pkgs_dir/libjpeg-turbo-2.1.4-h166bdaf_0/share/man/man1/rdjpgcom.1
  target path: /home/runner/micromamba/conda-bld/pcdsdevices_1673570096661/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/share/man/man1/rdjpgcom.1
```
https://github.com/pcdshub/pcdsdevices/actions/runs/3907251443/jobs/6676280720#step:11:1421

pyproject.toml
klauer: No, something's missing in this repository, hmm: https://github.com/pcdshub/cookiecutter-pcds-python/actions/runs/3907026295/jobs/6675800569#step:13:210

.pre-commit-config.yaml
klauer: Reverted below

pcdsdevices/tests/test_lens_sets/test
ZLLentz: The question is: there isn't anything to import here, it's just files that include data, so why should we want/need to add an importable file?

conda-recipe/meta.yaml
```diff
@@ -1,20 +1,23 @@
...
 
 package:
-  name: pcdsdevices
-  version : {{ data.get('version') }}
+  name: {{ package_name }}
+  version : {{ version }}
```
ZLLentz, line 8: And here's just a little bit more detail about py311 build failure, which should be out of scope for this PR:
https://github.com/pcdshub/pcdsdevices/actions/runs/3907251443/jobs/6676280916#step:11:846
```
Mamba failed to solve:
 - pcdsdevices 7.1.1.dev84+g5b42573cbd py_0
 - bluesky >=1.6.4
 - happi
 - jsonschema
 - lightpath >=1.0.0
 - matplotlib-base
 - numpy
 - ophyd >=1.5.1
 - pcdscalc >=0.2.0
 - pcdsutils >=0.4.0
 - pint
 - prettytable
 - pyepics >=3.4.2
 - python >=3.9
 - pytmc >=2.7.0
 - pyyaml
 - schema
 - scipy
 - python
 - python 3.11.*

with channels:
 - file:///home/runner/conda-bld

The reported errors are:
- Encountered problems while solving:
-   - package matplotlib-base-3.5.2-py37he341ac4_0 requires python_abi 3.7 *_pypy37_pp73, but none of the providers can be installed
```

The last error message is nonsensical, the only odd thing I see is `python` being specified twice

ZLLentz:
I did my best to look really really closely at the full diff and the gha output for any oddities since a lot of this was templated/automated. I'm happy to help find solutions if anything looks like a real issue.

pcdsdevices/state.py
ZLLentz: In my head, `Optional` is more intuitive as an argument, but not as a return type, where they are about the same to me.
I'll need to sit on it to form a stronger opinion myself.

pcdsdevices/tests/test_lens_sets/test
klauer: For myself, I've always just included `__init__.py` so I can be sure that regardless of the `setup.py` / `MANIFEST` / `setup.cfg` / etc settings, the directory I intended to include will be available in the distributed package. 

It's possible that setting `namespaces` to `True` and removing extraneous `__init__.py` files could be fine - I'm just not confident about it. The best way of doing things is likely buried somewhere in here or the link above: https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

pcdsdevices/tests/test_utils.py
klauer: Resolving this to clear the PR comments; we discussed this at length

conda-recipe/meta.yaml
klauer: The weird thing there is that matplotlib-base is on py37 (see the build string) for some reason instead of py311...

Matplotlib-base does have 3.11 builds, it seems:
https://anaconda.org/conda-forge/matplotlib-base/files

klauer:
Updated the branch protection settings of this repository with https://github.com/pcdshub/pcds-python-migration-tools/pull/5

Note that `Required` is now showed for everything but Python 3.11

klauer:
So the ❌  in the actions status is the subject of a continuously-requested feature over multiple years that GitHub apparently does not care to really investigate - so it seems to me. There's a lot of frustrated commenters in the following links:

https://github.com/actions/runner/issues/2347
https://github.com/orgs/community/discussions/15452

Any "solution" requires working around the system. We could:
1. Live with seeing ❌ for optional checks (and having to click to see what passed on a per-commit basis)
2. Remove our experimental tests from the workflow entirely (not a great choice)
3. Investigate a solution like this that replaces status information for "continue-on-error" jobs with a bot comment: https://github.com/mainmatter/continue-on-error-comment 
4. Keep Travis CI until GitHub Actions comes around on the issue (... no)

Though all undesirable, I'd vote for (1) until we come up with a better solution.

With that, this is ready for a comment/review/merge

tangkong (approved):
Unfortunate ❌  aside, I think this looks good.  A quick glance through the test logs doesn't  raise any red flags to me, and the changes all look good to me.  I'm sure more refinements will be added as we roll this out to other projects, but I think it might be good to tackle those in another PR.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -24,6 +24,6 @@
 - [ ] Code contains descriptive docstrings, including context and API
 - [ ] New/changed functions and methods are covered in the test suite where possible
 - [ ] Test suite passes locally
-- [ ] Test suite passes on travis
+- [ ] Test suite passes on GitHub Actions
```
ZLLentz, line 27: :tada:

.pre-commit-config.yaml
```diff
@@ -23,11 +23,11 @@ repos:
...
 
 -   repo: https://github.com/pycqa/flake8.git
-    rev: 3.9.2
+    rev: 6.0.0
     hooks:
     -   id: flake8
 
 -   repo: https://github.com/timothycrosley/isort
-    rev: 5.10.1
+    rev: 5.11.4
```
ZLLentz, line 31: Maybe we should bump this and then close #1107 ?

pcdsdevices/version.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class VersionProxy(UserString):
+    def __init__(self):
+        self._version = None
+
+    def _get_version(self) -> Optional[str]:
+        # Checking for directory is faster than failing out of get_version
+        repo_root = Path(__file__).resolve().parent.parent
+        if (repo_root / ".git").exists() or (repo_root / ".git_archival.txt").exists():
```
ZLLentz, line 19: I was at some point worried that this would slow down the lookup, but after thinking harder about it I think there is no case where the `.git_archival.txt` exists, is not subbed, and the git directory is missing. The tarball won't have the `.git` directory and and the install won't have the `.git_archival.txt` file or the `.git` directory.

ZLLentz (approved):
I'm happy to press the button on this any time

PR #1109, opened 2023-01-31
BLD: use compatible xraydb and sqlalchemy versions

klauer:
## Description
Specified compatible xraydb and sqlalchemy versions in requirements files.

See related issue: https://github.com/pcdshub/pcdscalc/issues/21

## Pre-merge checklist

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -41,6 +41,10 @@ requirements:
     - pmgr >=2.0.2
     - pyqt <5.15
     - typhos >=2.4.0
+    # Note: these should be removed once a newer xraylib is available on conda-forge
+    # or moved to pcdscalc after testing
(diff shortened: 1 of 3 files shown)


ZLLentz (approved):
The linked context issue makes it clear that we should merge this, as annoying as it is to lug around dependencies.

The only real alternative I see are:
- Waiting and hoping
- Put the pins on pcdscalc and then pin pcdscalc here

tangkong (approved):
I feel like the proper place to pin it is in pcdscalc, but the pragmatic approach is probably to make pcdsdevices the pin-cushion since it's the package we deal with directly and frequently.  

All aboard merging this so we can do more pcdsdevices work

ZLLentz (approved):

tangkong (approved):

PR #1111, opened 2023-02-04
MNT: Update xpp_lodcm lightpath logic

tangkong:
## Description
Fixes a typo and makes the xpp lodcm never attenuate any beam unless blocking

## Motivation and Context
In a quest for a more accurate lightpath displays

## How Has This Been Tested?
Interactively, though more testing should be done

Here

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/lodcm.py
+++ b/pcdsdevices/lodcm.py
@@ -1947,11 +1947,14 @@ class XPPLODCM(LODCM):
         removed = self.tower1.h1n_state.check_removed(h1n_state)
 
         if inserted and not removed:
-            output = {'L0': 0.5,
-                      'L1': 0.5}
+            # ignore attenuation from the LODCM splitting beam
+            output = {'L0': 1,
+                      'L2': 1}
         elif not inserted and removed:
+            # if removed, full transmission through L
             output = {'L0': 1}
         else:
+            # for unknown states, mark as blocking
             output = {'L0': 0}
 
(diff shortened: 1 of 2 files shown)


vespos:
Looks good! If we want accurate transmission estimates for this device, we need to have a discussion, probably including the scientists too. I feel that this is, however, good enough, not sure this would be worth the effort. The LODCM is always receives special attention when being aligned and is unlikely to be overlooked.

vespos (approved):

pcdsdevices/lodcm.py
```diff
@@ -1947,11 +1947,14 @@ def calc_lightpath_state(
         removed = self.tower1.h1n_state.check_removed(h1n_state)
 
         if inserted and not removed:
-            output = {'L0': 0.5,
-                      'L1': 0.5}
+            # ignore attenuation from the LODCM splitting beam
+            output = {'L0': 1,
+                      'L2': 1}
```
klauer, line 1952: Looks like the typo was L1 -> L2

klauer (approved):
If @vespos is happy, we're happy.

The extra comments are nice - LGTM 👍

ZLLentz (approved):
Should work much better than before :+1:

I think the original has a missed opportunity to differentiate between the diamond crystal that splits the beam and the si crystal that only sends to the mono line, but that shouldn't be in scope here

PR #1112, opened 2023-02-07
RSpowersupply 

tongju12:
## Old RS powersupply is working well but we want to add several more PVs as description of this device, remote connect

Add connect, device_id, device_desc and device_mode into happi database
## Motivation and Context

Make scientists easier to use this device by python code

## How Has This Been Tested?

We tested all other PVs already, but not the new ones yet. I believe those should work. 
## Pre-merge checklist
- [x ] Ran docs/pre-release-notes.sh and created a pre-release documentation page


tangkong (commented):
I might rename some of these fields to be a bit more self-explanatory, but the syntax here seems fine.  My suggestions would be as follows, but if the scientists want these I will acquiesce:
* `connect` -> `reconnect`
* `device_id` -> `id`  (it's not really the device but a single channel and there aren't any other ids on this component to worry about?)
* `device_desc` -> `desc` (same reason)
* `device_mode` -> `remote_mode`

I tried loading up this device but found many of these PV's missing (`connect`, `device_id`, `device_desc`, `device_mode`) for each channel.  Are these the right PVs or are they just down currently?

tangkong (approved):
Looks good.  Tong has tested this against the real device.  We should just need pre-release notes to close this one out

PR #1116, opened 2023-02-13
add smarpod as a device

patoppermann:
## Description
Add a new device class for smarpod and detailed  UI.

## Motivation and Context
We have a new device from SmarAct, this time a SmarPod.

## How Has This Been Tested?
IOC connection was tested.

Doc in the Code

## Pre-merge checklist


tangkong:
I think this looks mostly good aside from styling issues.  Try running pre-commit checks locally:  [Confluence page](https://confluence.slac.stanford.edu/display/PCDS/How+to+add+a+device+to+pcdsdevices#Howtoaddadevicetopcdsdevices-Pre-commit)

``` console
(dev) $ pre-commit install  # run this only once
(dev) $ git add pcdsdevices/my_device.py   # add new things to the staging area
(dev) $ pre-commit run  # manually run pre-commit
```

pcdsdevices/smarpod.py
```diff
@@ -0,0 +1,266 @@
...
+    """
+    cmd_accel = Cpt(EpicsSignal, ':CMD:ACCEL', kind='normal')
+    accel = Cpt(EpicsSignal, ':ACCEL', kind='normal')
+
+
+
+    """
+    Read and Set Poses, Pose 1-5:
+    """
+    pose_1_name = Cpt(EpicsSignal, ':POSE_1:NAME', kind='normal', doc='Pose 1 name input')
```
klauer, line 212: Any reason not to make a single pose a sub-device?

pcdsdevices/smarpod.py
```diff
@@ -0,0 +1,266 @@
...
+    ver_product  = Cpt(EpicsSignal, ':VER:PRODUCT', kind='normal')
+    ver_firmware = Cpt(EpicsSignal, ':VER:FIRMWARE', kind='normal')
+
+
+
+    """
+    Status
+    """
+    cmd_read_status = Cpt(EpicsSignal, ':CMD:READ_STATUS', kind='normal')
+    status_temp = Cpt(EpicsSignal, ':STATUS:TEMP', kind='normal')
```
klauer, line 159: I don't think these should be "normal" but rather omitted in most cases.

See https://confluence.slac.stanford.edu/display/PCDS/How+to+Create+an+Ophyd+Device#HowtoCreateanOphydDevice-Choosingtheright%22kind%22

pcdsdevices/smarpod.py
```diff
@@ -0,0 +1,266 @@
...
+    Stop
+    """
+    cmd_stop = Cpt(EpicsSignal, ':CMD:STOP', kind='normal')
+
+
+
+    """
+    Pose reachable?
+    """
+    cmd_reachable = Cpt(EpicsSignal, ':CMD:REACHABLE', kind='normal')
```
klauer, line 68: These need some documentation strings attached to the component

pcdsdevices/smarpod.py
```diff
@@ -0,0 +1,266 @@
...
+Module for the SmarPod and related devices.
+"""
+from ophyd.device import Component as Cpt
+from ophyd.device import Device
+from ophyd.signal import EpicsSignal
+from .interface import BaseInterface
+
+
+
+class SmarPod(BaseInterface, Device):
```
klauer, line 11: This class will not play nicely with bluesky for the purposes of scanning as you've implemented it as a generic `ophyd.Device`.

There needs to be a `Positioner` somewhere for the smarpod.

More details here: https://confluence.slac.stanford.edu/display/PCDS/How+to+Create+an+Ophyd+Device#HowtoCreateanOphydDevice-PVPositioner

Consider the upstream source of this SmarPod IOC implementation and its corresponding ophyd device: https://github.com/NSLS-II-HXN/profile_collection/blob/d724c469b8c4ac1784ca755b40b1992b3e5d0640/startup/15-zp.py#L15-L38 (and further down where it's being used as a sub-device)

klauer (changes requested):

pcdsdevices/smarpod.py
ZLLentz: Interesting, it looks like NSLS-II splits this kind of device into many different 1D positioners

pcdsdevices/smarpod.py
patoppermann: It was incorporated according to Ken's suggestion.

pcdsdevices/smarpod.py
patoppermann: the importance of the channels has been revised.

pcdsdevices/smarpod.py
patoppermann: doc was added everywhere.

pcdsdevices/smarpod.py
patoppermann: PVPositioner was chosen as the class.
from ophyd import PVPositioner

pcdsdevices/smarpod.py
```diff
@@ -0,0 +1,133 @@
+"""
+Module for the SmarPod and related devices.
+"""
+from ophyd import PVPositioner
+from ophyd.device import Component as Cpt
+from ophyd.device import Device
+from ophyd.signal import EpicsSignal, EpicsSignalRO
+
+
+class SmarPod(PVPositioner):
```
klauer, line 10: Have you given this a try, @patoppermann? Like running it in Python and instantiating a `SmarPod` instance?

I ask because `PVPositioner` requires some components to be defined, which I don't see here: https://confluence.slac.stanford.edu/display/PCDS/How+to+Create+an+Ophyd+Device#HowtoCreateanOphydDevice-PVPositioner

`PVPositioner` should raise an exception if those are not set: see [here](https://github.com/bluesky/ophyd/blob/master/ophyd/pv_positioner.py#L118)

I'm not sure if a single `PVPositioner` will work for the SmarPod, since it has 6 DOF (x/y/z/rx/ry/rz) to check. Normally `PVPositioner` works with scalars, but _maybe_...

pcdsdevices/smarpod.py
```diff
@@ -0,0 +1,133 @@
...
+    cmd_calibrate = Cpt(EpicsSignal, ':CMD:CALIBRATE', kind='normal')
+    cmd_read_ver = Cpt(EpicsSignal, ':CMD:READ_VER', kind='normal')
+    ver_sys = Cpt(EpicsSignalRO, ':VER:SYS', kind='omitted')
+    # Device Info
+    ver_sn = Cpt(EpicsSignalRO, ':VER:SN', kind='omitted')
+    ver_product = Cpt(EpicsSignalRO, ':VER:PRODUCT', kind='omitted')
+    ver_firmware = Cpt(EpicsSignalRO, ':VER:FIRMWARE', kind='omitted')
+    # Status
+    cmd_read_status = Cpt(EpicsSignal, ':CMD:READ_STATUS', kind='normal')
+    status_temp = Cpt(EpicsSignalRO, ':STATUS:TEMP', kind='omitted')
```
klauer, line 86: All of this is still really jumbled together. Consider that it might make more sense to shove status-related things into a SmarPodStatus device, which could then be instantiated with a `Cpt(SmarPodStatus, ":STATUS")` much like how I was talking about the poses.

pcdsdevices/smarpod.py
```diff
@@ -0,0 +1,133 @@
...
+    y = Cpt(EpicsSignal, ':Y', kind='normal', doc='Pose Y Position Data')
+    z = Cpt(EpicsSignal, ':Z', kind='normal', doc='Pose Z Position Data')
+    rx = Cpt(EpicsSignal, ':RX', kind='normal', doc='Pose RX Position Data')
+    ry = Cpt(EpicsSignal, ':RY', kind='normal', doc='Pose RY Position Data')
+    rz = Cpt(EpicsSignal, ':RZ', kind='normal', doc='Pose RZ Position Data')
+    record = Cpt(EpicsSignal, ':RECORD', kind='normal', doc='Fanout for pose')
+    execute = Cpt(EpicsSignal, ':EXECUTE', kind='normal', doc='Enacts Pose Positions')
+
+
+class Smarpod(Device):
```
klauer, line 128: There's `SmarPod` above and `Smarpod` here - two different classes.
The above one has everything but the poses. I think these should be merged somehow?

pcdsdevices/ui/smarpod.detailed.ui
ZLLentz, line 1: I think you should remove this file unless you're planning to make a screen

pcdsdevices/smarpod.py
```diff
@@ -0,0 +1,133 @@
+"""
+Module for the SmarPod and related devices.
+"""
+from ophyd.device import Component as Cpt
+from ophyd.device import Device
+from ophyd.signal import EpicsSignal, EpicsSignalRO
+
+
+class SmarPodPose(Device):
```
ZLLentz, line 9: Out of the loop question: What exactly does `Pose` mean in this context? I think some class level docstrings would make this file more friendly to people like me who don't know what a smarpod is.

e.g. something like this but with correct info
```
"""
SmarPodPose is a class that controls a smarpod positioner.

It is included in the more general SmarPod object, which represents a full controller.
A smarpod is a hexapod-like 6-axis positioner.
"""
```

Onto each of the classes to just state what they are.

ZLLentz (commented):
Hi, I have two minor suggestions.

It won't be possible to use this in a normal scan, but it's good that you've accumulated all the PVs. Perhaps in a follow-up we can add scanning behavior.

ZLLentz:
This should be good to go
There may be more adjustments later for various hutch needs as they arise but there is no reason to let it sit for another year

ZLLentz (approved):

PR #1117, opened 2023-02-14
Fixing AT1K2 Ui

wwright-slac:
AT1K2 screens were not generating with the cascading screen effect, similar to with the standard SXR Attenuator screens. I added symbolic links to the detailed_tree UI files to alleviate this issue.
 

## How Has This Been Tested?
Trying out these screens in my work area showed that this change was successful. 

NA

<!--
image

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1117-Adding_symbolic_links_for_AT1K2.rst
+++ b/docs/source/upcoming_release_notes/1117-Adding_symbolic_links_for_AT1K2.rst
@@ -1,0 +1,30 @@
+1117 Adding symbolic links for AT1K2
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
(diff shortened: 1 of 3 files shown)


tangkong:
This works nicely for me, replaces the existing behavior that brings up the positioner widget.  
image

wwright-slac:
Matplotlib build error?

tangkong:
matplotlib seems to not be playing nice with py3.11 on our builds.  For now we've been punting the issue; don't worry about it for this PR 👍

ZLLentz (approved):
This looks good to me

ZLLentz:
Funny story, I sourced the build error down to our pyqt version being unavailable in 3.11 and conda is just particularly bad at explaining which package is at fault in these cases. I think the solver tries older and older versions of python and then hits the last one and gives up, then finds a random library that isn't compatible with the old version and presents that as the error instead. Hence, "matplotlib is not compatible with py37 something something" (working theory)

PR #1123, opened 2023-03-17
Ophyd class for square-one tri-sphere motion system

baljamal:
Ophyd class for square-one tri-sphere motion system

## Motivation and Context
TMO  needs Ophyd class for square-one to be able to do scans rather than doing that manually.

## How Has This Been Tested?
An actual test has been done on TMO LAMP square-one table.
X-axis translation moved using ophyd class and cross-validated with LabVIEW interface.

some inline comments

## Pre-merge checklist


pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,43 @@
...
+from ophyd.device import Device
+from ophyd.device import FormattedComponent as FCpt
+from ophyd.pv_positioner import PVPositionerIsClose
+from ophyd.signal import EpicsSignal, EpicsSignalRO
+
+logger = logging.getLogger(__name__)
+
+
+class SQR1Axis(PVPositionerIsClose):
+    setpoint = FCpt(EpicsSignal, '{self.prefix}:TARGET:{self._axis}', kind='normal')
```
ZLLentz, line 16: (Optional) There's a shorthand for these: "self." can be omitted.
```suggestion
    setpoint = FCpt(EpicsSignal, '{prefix}:TARGET:{_axis}', kind='normal')
```

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,43 @@
...
+    readback = FCpt(EpicsSignal, '{self.prefix}:TARGET:{self._axis}:RBV', kind='normal')
```
ZLLentz, line 17: I suspect the readback should be "hinted" here, it's the most important component for these scans. This will make it show up in the automatic table/plotting callbacks during bluesky scans.
```suggestion
    readback = FCpt(EpicsSignal, '{self.prefix}:TARGET:{self._axis}:RBV', kind='hinted')
```

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,43 @@
...
+    stop_value = 1
+
+    llm = FCpt(EpicsSignalRO, '{self.prefix}:TARGET:{self._axis}:LLM', kind='normal')
+    hlm = FCpt(EpicsSignalRO, '{self.prefix}:TARGET:{self._axis}:HLM', kind='normal')
+
+    def __init__(self, axis, **kwargs):
+        self._axis = axis  # axis values {X,Y,Z,rX, rY, rZ}
+        super().__init__(**kwargs)
+
+        self.setpoint.put(self.readback.get())  # sync position setpoint with readback
```
ZLLentz, line 30: We should be really careful doing this, for three reasons:
1. We're blocking on two PV connections during `__init__`, and we're doing it 6 times for a maximum total of 12 possible timeouts.
2. We're causing a change to the control system on `__init__`. If a second user opens a session that contains a square 1 device while the first user is doing a scan, depending on the timing we can get very strange behavior.
3. This device will error out on `__init__` if run from anywhere outside of tmo, which may be disruptive.

What exactly is gained from doing this sync? Is there any way we can accomplish this during moves only instead of at initialization?

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,43 @@
...
+        self._limits = ((self.llm.get(), self.hlm.get()))
```
ZLLentz, line 31: Same as above: does this need to happen now, during `__init__`? This is a 2x timeout + 6 axis = 12 timeouts, potentially. It might be better to do this when the limits are checked or by using a subscription callback.

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,43 @@
...
+class SQR1(Device):
+    def __init__(self, prefix, name) -> None:
+        self.prefix = prefix
+        self.name = name
+        self.x = SQR1Axis(axis="X", prefix=prefix, name=f"{name}_x", rtol=0.01)
+        self.y = SQR1Axis(axis="Y", prefix=prefix, name=f"{name}_y", rtol=0.01)
+        self.z = SQR1Axis(axis="Z", prefix=prefix, name=f"{name}_z", rtol=0.01)
+        self.rx = SQR1Axis(axis="rX", prefix=prefix, name=f"{name}_rx", rtol=0.01)
+        self.ry = SQR1Axis(axis="rY", prefix=prefix, name=f"{name}_ry", rtol=0.01)
+        self.rz = SQR1Axis(axis="rZ", prefix=prefix, name=f"{name}_rz", rtol=0.01)
```
ZLLentz, line 43: You've skipped the `Device.__init__` (`super().__init__(prefix, name=name)`) entirely and implemented everything as an `__init__`-time attribute instead of as a component.

These individual subdevices will work, but SQR1 will not function properly as a top-level device.

You can instead include components in the class definition and skip overriding `__init__` entirely.

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,43 @@
...
+    actuate_value = 1
+    stop_signal = FCpt(EpicsSignal, '{self.prefix}:KILL', kind='normal')
+    stop_value = 1
+
+    llm = FCpt(EpicsSignalRO, '{self.prefix}:TARGET:{self._axis}:LLM', kind='normal')
+    hlm = FCpt(EpicsSignalRO, '{self.prefix}:TARGET:{self._axis}:HLM', kind='normal')
+
+    def __init__(self, axis, **kwargs):
+        self._axis = axis  # axis values {X,Y,Z,rX, rY, rZ}
+        super().__init__(**kwargs)
```
ZLLentz, line 28: This behaves a bit differently to all other devices because the positional `prefix` argument has been made keyword-only and `axis` is now the first argument instead of `prefix`. I recommend undoing this so that people don't get confused when trying to instantiate this device, and to make it easier to include this device as a component.

```suggestion
    def __init__(self, prefix, axis, **kwargs):
        self._axis = axis  # axis values {X,Y,Z,rX, rY, rZ}
        super().__init__(prefix, **kwargs)
```

ZLLentz (changes requested):
I have some questions and suggestions here, to try to avoid subtle issues/confusions in the future. 

It looks like it is good and functional for this use case right now but I want to make sure we'll be OK later too.

pcdsdevices/sqr1.py
baljamal: This system has one move command PV for the six motors, therefore I initialize each motor setpoint value with its readback value to avoid moving one motor while other motor(s) setpoints are zeros for example, then they will move to unwanted positions.

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,141 @@
...
+    actuate_value = 1
+    stop_signal = FCpt(EpicsSignal, '{prefix}:KILL', kind='normal')
+    stop_value = 1
+
+    def __init__(self,
+                 prefix,
+                 axis: str,
+                 sync_setpoints: typing.function = None,
+                 **kwargs
+                 ):
```
ZLLentz, line 32: Optional suggestion: run this through `black` to see what kind of formatting the python community is favoring for arg inputs these days. This sort of layout is somewhat hard to read.

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,141 @@
...
+        self.prefix = prefix
```
ZLLentz, line 33: Optional suggestion: This is ok but is actually redundant because it is set during `super().__init__`, you could remove it

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,141 @@
...
+                         )
+        self.x._sync_setpoints = self.sync_setpoints
+        self.y._sync_setpoints = self.sync_setpoints
+        self.z._sync_setpoints = self.sync_setpoints
+        self.rx._sync_setpoints = self.sync_setpoints
+        self.ry._sync_setpoints = self.sync_setpoints
+        self.rz._sync_setpoints = self.sync_setpoints
+
+    def sync_setpoints(self):
+        print("sync_setpoint")
```
ZLLentz, line 83: Question: are you finding this print statement useful? If you are, great, if not, you could remove it.

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,141 @@
...
+                                 timeout=timeout,
+                                 sync_enable=False)
+        ry_status = self.ry.move(ry_sp,
+                                 wait=False,
+                                 timeout=timeout,
+                                 sync_enable=False)
+        rz_status = self.rz.move(rz_sp,
+                                 wait=False,
+                                 timeout=timeout,
+                                 sync_enable=False)
```
ZLLentz, line 132: Does this actually work smoothly with no hiccups? We're writing to the shared actuate signal 6 times in the code here.

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,141 @@
...
+
+        status = AndStatus(
+            x_status,
+            AndStatus(y_status,
+                      AndStatus(z_status,
+                                AndStatus(rx_status,
+                                          AndStatus(ry_status,
+                                                    rz_status)))))
```
ZLLentz, line 140: There's funny syntax here that's equivalent and fairly clear/compact
```
status = x_status & y_status & z_status & rx_status & ry_status & rz_status
```

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,141 @@
...
+        return status_wait(status) if wait else status
```
ZLLentz, line 141: This makes it so we sometimes return `None` and sometimes return a status object. Usually it's best to always return the same type when possible.

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,141 @@
...
+        return super().move(position, wait, timeout, moved_cb)
+
+
+class SQR1(Device):
+    x = FCpt(SQR1Axis, '{prefix}', axis="X", rtol=0.001)
+    y = FCpt(SQR1Axis, '{prefix}', axis="Y", rtol=0.001)
+    z = FCpt(SQR1Axis, '{prefix}', axis="Z", rtol=0.001)
+    rx = FCpt(SQR1Axis, '{prefix}', axis="rX", rtol=0.001)
+    ry = FCpt(SQR1Axis, '{prefix}', axis="rY", rtol=0.001)
+    rz = FCpt(SQR1Axis, '{prefix}', axis="rZ", rtol=0.001)
```
ZLLentz, line 56: Optional suggestion: this isn't wrong, but you don't need to use an FCpt here. Usually FCpt is for when you need to put non-prefix things into the prefix.

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,141 @@
...
+                         configuration_attrs=configuration_attrs,
+                         parent=parent,
+                         **kwargs
+                         )
+        self.x._sync_setpoints = self.sync_setpoints
+        self.y._sync_setpoints = self.sync_setpoints
+        self.z._sync_setpoints = self.sync_setpoints
+        self.rx._sync_setpoints = self.sync_setpoints
+        self.ry._sync_setpoints = self.sync_setpoints
+        self.rz._sync_setpoints = self.sync_setpoints
```
ZLLentz, line 80: I think this and the other associated code is a clean solution to the setpoint synchronization problem :+1:

pcdsdevices/sqr1.py
baljamal: I tried black to auto-format it basically, it put the whole method definition in one line which violates PEP8 80 characters rule

pcdsdevices/sqr1.py
baljamal: you are right, I deleted it

pcdsdevices/sqr1.py
baljamal: these PVs are living in the same CA server, so writing set point values to them is not a heavy process

pcdsdevices/sqr1.py
baljamal: I agree, operator overloading looks more neat. I followed your suggestion

pcdsdevices/sqr1.py
baljamal: I followed your suggestion

pcdsdevices/sqr1.py
baljamal: thanks for compliment

pcdsdevices/sqr1.py
ZLLentz: I guess I was more curious about how the IOC behaves. The sequence here is something like:

- call x move:
  - write to the x setpoint
  - set the actuate signal
- call y move:
  - write to the y setpoint
  - set the actuate signal
... etc.

Does this encounter any problems with e.g. starting a move to x before the rz parameter is available?

ZLLentz: I think this is the last functional thing to consider before merging

pcdsdevices/sqr1.py
ZLLentz, line 1: Aside from the sequencing question, this PR is missing the standard procedure of including release notes.
cd docs -> run `./pre-release-notes.sh 1123 square-one` -> put your username in and say you added the square one device

If the pre-merge checklist from the description template is completed we can merge soon

pcdsdevices/sqr1.py
baljamal: I tried to move three motors at the same time and it is working fine. I got your point but it doesn't seem a problem.

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,142 @@
...
+    readback = FCpt(EpicsSignal, '{prefix}:TARGET:{_axis}:RBV', kind='hinted')
+    actuate = FCpt(EpicsSignal, '{prefix}:MOV', kind='normal')
+    actuate_value = 1
+    stop_signal = FCpt(EpicsSignal, '{prefix}:KILL', kind='normal')
+    stop_value = 1
+
+    def __init__(self,
+                 prefix,
+                 axis: str,
+                 sync_setpoints: typing.Callable = None,
```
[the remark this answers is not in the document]
baljamal: you are right, I should pay more attention to types.
I went with this: sync_setpoint: Callable | None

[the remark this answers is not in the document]
ZLLentz: This is done elsewhere in the file too

docs/source/upcoming_release_notes/1123-square-one_tri-sphere_motion_system.rst
```diff
@@ -0,0 +1,30 @@
...
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- SQR1
```
ZLLentz, line 18: Can you please include context information here? Even if it's just the header of your module docstring?

At this point in the review I'm kind of just saying "read and follow the pre-merge checklist" repeatedly which is probably unpleasant for both of us.

pcdsdevices/sqr1.py
ZLLentz: Alright, if you've tested the code as-is and there hasn't been a problem it's probably OK and shouldn't hold up the PR
I'm still worried about if of course since it's an unnecessary risk

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,142 @@
+"""
+    Ophyd class for square-one tri-sphere motion system
+    (https://www.sqr-1.com/tsp.html)
+"""
+
```
ZLLentz, line 5: Your code doesn't work any more, needs to add
```
from __future__ import annotations
```
at the top here to allow the annotation | syntax in py39

ZLLentz: A small unit test would catch things like this, even one that simply imports the module

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,142 @@
...
+             moved_cb: float = None,
+             sync_enable: bool = True,
+             ):
+        if sync_enable and self._sync_setpoints:
+            self._sync_setpoints()
+        return super().move(position, wait, timeout, moved_cb)
+
+
+class SQR1(Device):
+    x = Cpt(SQR1Axis, '{prefix}', axis="X", rtol=0.001)
```
ZLLentz, line 50: These are all malformed now:
```
In [1]: import pcdsdevices.sqr1

In [2]: dev = pcdsdevices.sqr1.SQR1("MCQ:TMO:01", name="dev")

In [3]: dev.component_names
Out[3]: ('x', 'y', 'z', 'rx', 'ry', 'rz', 'stop_signal')

In [4]: dev.x.prefix
Out[4]: 'MCQ:TMO:01{prefix}'
```
the curly braces setup only works for FCpt, to refactor to Cpt you would have needed to drop this down to (I think):
```suggestion
    x = Cpt(SQR1Axis, '', axis="X", rtol=0.001)
```

docs/source/upcoming_release_notes/1123-square-one_tri-sphere_motion_system.rst
```diff
@@ -0,0 +1,30 @@
...
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- Basil Aljamal (baljamal)
```
ZLLentz, line 30:
```suggestion
- baljamal
```

pcdsdevices/sqr1.py
```diff
@@ -1,8 +1,23 @@
 """
-    Ophyd class for square-one tri-sphere motion system
-    (https://www.sqr-1.com/tsp.html)
+Ophyd class for square-one tri-sphere motion system (https://www.sqr-1.com/tsp.html) # noqa: E501
```
tangkong, line 2: For the record you shouldn't just ignore line length conventions because you feel like it.  It's perfectly possible to break this up into two lines

ZLLentz (commented):
I like this as-is and want to merge it
A simple import/instantiation test would be useful but I'm not going to force you to do it

One more thing I need to check before the approval: I want to see how these look in the generated docs, the docstrings look kind of nonstandard but it shouldn't necessarily be a problem

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,280 @@
...
+                        rz_sp=None,
+                        wait=True,
+                        timeout=10.0):Move multiple axes simultaneously
+        to specified setpoints.
+        stop(): Stop the motion of the entire tri-sphere motion system.
+
+    Example:
+        >>> tri_sphere = SQR1(prefix="SQR1:", name="tri_sphere")
+        >>> tri_sphere.multi_axis_move(x_sp=1.0, y_sp=2.0, z_sp=3.0)
+        >>> tri_sphere.stop()
```
ZLLentz, line 185: These examples render really well in the html docs, we should do this more often :+1:
image

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+    This class represents the entire square-one tri-sphere motion system,
+    which consists of multiple axes (X, Y, Z, rX, rY, rZ) that can be
+    controlled individually. It aggregates instances of the `SQR1Axis` class
+    for each axis and provides methods for multi-axis movement and
+    stopping the motion.
+
+    Attributes:
+        x (SQR1Axis): An instance of the `SQR1Axis` class representing
+        the X-axis.
```
ZLLentz, line 141: Each of these is rendering kind of strangely in the html docs:
image

ZLLentz: I'm not sure the right way to reformat this because I've only used the numpy style docstrings
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/#google-vs-numpy

baljamal: I think the problem with second line indentation

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+    This class represents and controls a single axis of the square-one
+    tri-sphere motion system using EPICS PVs (Process Variables).
+    It is a subclass of PVPositionerIsClose and includes attributes for setting
+    the setpoint, reading back the current position,
+    actuating the motion, and stopping the motion.
+
+    Attributes:
+        setpoint (EpicsSignal): An EPICS signal for setting the desired
+        position of the axis.
```
ZLLentz, line 66: Your indentation commit below fixes the docs build for the SQR1 class
If you do the same thing for the docstring here in the SQRT1Axis then this is ready to merge

pcdsdevices/sqr1.py
```diff
@@ -0,0 +1,281 @@
...
+
+class SQR1Axis(PVPositionerIsClose):
+    """
+    This class represents and controls a single axis of the square-one
+    tri-sphere motion system using EPICS PVs (Process Variables).
+    It is a subclass of PVPositionerIsClose and includes attributes for setting
+    the setpoint, reading back the current position,
+    actuating the motion, and stopping the motion.
+
+    Attributes:
```
klauer, line 64: If you look at our style guide [here](https://pcdshub.github.io/style.html#docstrings), or any of the other code in this repository (or any of our other Python repositories), you'll see we follow the numpy docstring convention.

I think it would be good for you to understand how to format them for your own future reference/submissions, but we can reformat them for you if necessary to get this merged sooner rather than later.

baljamal: thanks for sharing the link, I wasn't aware of it.

klauer: Thanks for fixing it up 👍

ZLLentz (approved):
Let's finally merge this

PR #1126, opened 2023-04-13
ENH: Self-seeding

vespos:
## Description
Make mono classes compatible with self-seeded operation of the accelerator.
The self-seed bandwidth is much narrower and we anticipate that we'll need to optimize the mono energies to maximize the throughput.
A first optimization of a simple monotonic search is currently implemented

## How Has This Been Tested?
To be tested with and without beam.
CCM self-seed energy class can be instantiated with:
`CCMEnergyWithSelfSeed(name='ccm_test', prefix='XPP:MON:MPZ:07A', ipms_prefix='XPP:TT:01:IPMS.VALA', hutch='XPP')`

## Pre-merge checklist

## To do
- [x] CCM monotonic search
- [ ] Test CCM without beam
- [ ] Test CCM with beam
- [ ] LODCM monotonic search
- [ ] Test LODCM without beam
- [ ] Test LODCM with beam

### DIFF
--- a/docs/source/upcoming_release_notes/1273-adding_pa1k4-pf_device.rst
+++ b/docs/source/upcoming_release_notes/1273-adding_pa1k4-pf_device.rst
@@ -1,0 +1,30 @@
+1273 adding pa1k4-pf device
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- adding new device PA1K4-PF
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- tongju12
--- a/pcdsdevices/tmo_ip1.py
+++ b/pcdsdevices/tmo_ip1.py
@@ -1,0 +1,51 @@
+"""TMO sample calibration foil to work with spectrometer"""
+from lightpath import LightpathState
+from ophyd.device import Component as Cpt
+
+from .device import GroupDevice
+from .device import UpdateComponent as UpCpt
+from .epics_motor import BeckhoffAxis
+from .interface import BaseInterface, LightpathMixin
+from .pmps import TwinCATStatePMPS
+
+
+class CalibrationAxis(TwinCATStatePMPS):
+    """
+    Sample calibration 1D State Setup
+    Here, we specify 5 states, and 1 motor, for Y
+    axe.
+    """
+    config = UpCpt(state_count=5, motor_count=1)
+
+
+class SCaFoil(BaseInterface, GroupDevice, LightpathMixin):
+    """
+    TMO PMPS1D sample calibration class.
+    Parameters:
+    -----------
+    prefix : str
+        Base PV for the motion system
+    name : str
+        Alias for the device
+    """
+    # UI Representation
+    _icon = 'fa.minus-square'
+    tab_component_names = True
+
+    # Sample calibration Y axis
+
+    pf_state = Cpt(CalibrationAxis, ':STATE', kind='normal')
+    pf_y = Cpt(BeckhoffAxis, ':MMS:Y', doc="Y-axis of photo filter pa1k4", kind='normal')
+    removed = False
+    transmission = 1
+    SUB_STATE = 'state'
+
+    # dummy signal, state is always the same
+    lightpath_cpts = ['cali_foil.user_readback']
+
+    def calc_lightpath_state(self, **kwargs) -> LightpathState:
+        return LightpathState(
+            inserted=True,
+            removed=False,
+            output={self.output_branches[0]: 1}
+        )


vespos:
@ZLLentz @tangkong , before I move ahead with testing and implementing a similar class for the LODCM, what do you think of the approach?
In particular on 
- The way the arguments for the optimization method are handled? 
- Does the framework seems ok to easily add more optimization methods, should we want to?
- The components' definition?
- Would that work ok with bluesky scans?

pcdsdevices/ccm.py
```diff
@@ -773,6 +773,178 @@ def inverse(self, real_pos: namedtuple) -> namedtuple:
...
+            'energy_offset' : offset,
+            'n_average' : n,
+            'ipm_ratio_threshold' : ratio_th,
+            'energy_step_size' : step_size,
+            'max_step' : max_step
+        }
+        self.method = 'monotonic'
+        return
+
+    def move(self, new_pos , *args, **kwargs):
```
tangkong, line 942: For this positioner to work with bluesky plans, you'll likely need to hijack the `.set()` method as well.  That's the method that plans like `bp.scan` will use, I believe

pcdsdevices/ccm.py
```diff
@@ -773,6 +773,178 @@ def inverse(self, real_pos: namedtuple) -> namedtuple:
...
+        if self.method == 'monotonic':
+            self.monotonic_increase_search(new_pos, **self.optimization_args)
```
tangkong, line 944: If you also had a mapping from `method_name` to `optimization_method`, you could get rid of ifs and do something like:

``` python
def move(self, new_pos, *args, **kwargs):
    self.method_map[self.method](new_pos, self.opt_args[self.method])
```

pcdsdevices/ccm.py
```diff
@@ -773,6 +773,178 @@ def inverse(self, real_pos: namedtuple) -> namedtuple:
...
+        else:
+            self.hutch = 'TST'
+
+        prefixes = {'ipms_prefix': ipms_prefix}
+        UCpt.collect_prefixes(self, prefixes)
+        super().__init__(prefix, **kwargs)
+
+        self.ipm_idx = ipm_idx
+
+        self.method = 'monotonic'
```
tangkong, line 844: Just spitballing here.  Maybe it'd be easier to have something like:
``` python
self.opt_args = {
    'method_name_1': {'method1_arg1': 1, 'method1_arg2': 2},
    'method_name_2': {'method2_arg1': 1, 'method2_arg2': 2},
}
```
Such that you can access the optimization args by method name.  That just gives you a place to put different args as you add more methods

tangkong (commented):
I left a couple of thoughts.  Is the target plan for this something like `bp.scan`?  My main concern is that all the messages sent to the RunEngine resolve properly, but I probably have to dig deeper to be sure

pcdsdevices/ccm.py
```diff
@@ -773,6 +773,178 @@ def inverse(self, real_pos: namedtuple) -> namedtuple:
...
+            'energy_step_size' : step_size,
+            'max_step' : max_step
+        }
+        self.method = 'monotonic'
+        return
+
+    def move(self, new_pos , *args, **kwargs):
+        if self.method == 'monotonic':
+            self.monotonic_increase_search(new_pos, **self.optimization_args)
+        return
```
ZLLentz, line 945: For this to work in bluesky, it specifically needs a `set` method that returns a `Status` object that is marked as done once the optimization routine is complete. It's worth taking a moment to climb up the inheritance tree to see what `set` is defined as in the parent class here.

ZLLentz: Ultimately, due to inheritance, you have a wrapper on top of this set method:
https://github.com/bluesky/ophyd/blob/27e2934de0157e3c0d6eb5ddb3e06ab15ce53038/ophyd/positioner.py#L48-L93

So to make this work properly in bluesky, you need to include and return an appropriate `MoveStatus`

Note that the intent for bluesky is for `set` to be called and returned immediately with the `MoveStatus`, and then all the motion should happen in the background, though this isn't strictly required.

vespos: I guess I could do smth silly like `return super().move(current_position)` after the optimization step. This would return the status of a move, without moving anything

pcdsdevices/ccm.py
```diff
@@ -773,6 +773,178 @@ def inverse(self, real_pos: namedtuple) -> namedtuple:
...
+    _enable_warn_constants: bool = False
+    hutch: str
+
+    tab_component_names = True
+
+    def __init__(
+        self,
+        prefix: str,
+        ipms_prefix: str = None,
+        ipm_idx: dict = {'upstream': 0, 'downstream': 1},
```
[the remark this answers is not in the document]
vespos: Interesting! I don't think it would have dramatic effect here, but it is definitely worth sticking to good practices! Thanks for calling me out on something I do all the time :-D

pcdsdevices/ccm.py
vespos: See Zach's comment for solution

pcdsdevices/ccm.py
vespos: Good idea, that would simplify the addition of new methods

pcdsdevices/ccm.py
ZLLentz: If you don't want to do a fully-featured status, you can create one, mark it as done, and return it immediately.
```
from ophyd.status import Status
<snip>
def move(<args>):
    <snip>
    status = Status()
    status.set_finished()
    return status
```
This should be good enough to have bluesky accept it, with the caveat that it we're just matching the API without considering how the run engine wants our move to be asynchronous.

vespos: I ended using the status from the move in the underlying method function. let me know if you think that would work, and we can close this thread

pcdsdevices/ccm.py
vespos: Implemented it, let me know what you think

pcdsdevices/ccm.py
```diff
@@ -773,6 +774,198 @@ def inverse(self, real_pos: namedtuple) -> namedtuple:
...
+            'n_average' : n,
+            'ipm_ratio_threshold' : ratio_th,
+            'energy_step_size' : step_size,
+            'max_step' : max_step
+        }
+        self.method = 'monotonic'
+        return
+
+    def move(self, new_pos , *args, **kwargs):
+        status = self.method_map[self.method](new_pos, self.optimization_args[self.method])
```
tangkong, line 965:
```suggestion
        status = self.method_map[self.method](new_pos, **self.optimization_args[self.method])
```

You'll have to unpack the optimization kwargs here, but otherwise I think this should work

pcdsdevices/ccm.py
```diff
@@ -773,6 +774,198 @@ def inverse(self, real_pos: namedtuple) -> namedtuple:
...
+            ipm_ratio = self.ipm_ratio(intensities)
+            if ipm_ratio > self.ipm_ratio_threshold:
+                print('Optimization complete!\n')
+                break
+            status = super().mvr(stepsize, wait=True)
+        return status
+
+    def get_ipms_intensities(self, n_shots=10):
+        intensities = []
+        status = Status(timeout=5)
```
tangkong, line 910: Do you want this to be configurable from the method call?  I imagine 5s is more than enough time, but it does seem a bit arbitrary

vespos: Yes it is a bit arbitrary. I just don't want the think to hang forever in case something is wrong. Talking to Takahiro, we probably want to average 10-50 shots, less than 1s worth of data, so I could probably reduce that timeout further. I hope we never run into the timeout here tbh

pcdsdevices/ccm.py
```diff
@@ -773,6 +774,198 @@ def inverse(self, real_pos: namedtuple) -> namedtuple:
...
+            than the self-seed bandwidth for good results.
+        max_steps: int
+            If no good point is reached after max_step points, the optimization is
+            aborted.
+        """
+        self.self_seed.move(setpoint)
+        # move slightly below the setpoint
+        status = super().move(setpoint-offset, wait=True)
+        # wait for self-seed to settle
+        time.sleep(1)
```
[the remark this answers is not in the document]
vespos: It is still unclear how AD would be able to notify us of the completion of the self-seed motion. I hope they can provide a status PV, or a DONE flag, but if that's not possible, I will adjust the sleep time to give enough time for their task to complete and settle. In the case of the Vernier energy scan (the closest production case we have), we don't really rely on their status either, although I think there is an implementation in the beam energy request class.

pcdsdevices/ccm.py
```diff
@@ -773,6 +774,198 @@ def inverse(self, real_pos: namedtuple) -> namedtuple:
...
+
+        for step in range(max_steps):
+            intensities = self.get_ipms_intensities(n_average)
+            ipm_ratio = self.ipm_ratio(intensities)
+            if ipm_ratio > self.ipm_ratio_threshold:
+                print('Optimization complete!\n')
+                break
+            status = super().mvr(stepsize, wait=True)
```
ZLLentz, line 905: `mvr` doesn't return a `status`, unfortunately. I was asked to make it return nothing to avoid cluttering the terminal output.

I think maybe the cleanest way here is to compute the positions and use `super().move` again? Otherwise I guess you can just manually do the relative move using `super().move` as well.

pcdsdevices/ccm.py
```diff
@@ -773,6 +774,198 @@ def inverse(self, real_pos: namedtuple) -> namedtuple:
...
+        return status
```
ZLLentz, line 906: I think returning this status and using it in `bluesky` should work, though note that this will stall out the `asyncio` thread for the duration of the optimization which could be a problem if there were other background tasks. You don't need to prematurely do this though, it might not become a problem at all.

If it becomes a problem, you'd usually return the incomplete status immediately and do all the steps in a background thread and have the thread mark the status as finished at the appropriate time.

pcdsdevices/ccm.py
vespos: Oh that is unfortunate! As you suggest I will just compute the new position in the function and use super().move instead.

vespos: Just fixed it. It now uses move() instead so that a proper status is returned

PR #1130, opened 2023-04-17
REL: v7.3.0

ZLLentz:
## Description

As always, I've added contributors as reviewers.
Reviewing is not mandatory unless you think I messed up your notes in post.

- Small release including release notes
- An extra commit I ended up reverting because I thought there was a docs bug but there was not
- Commit the generated api page since I incidentally updated it during the extra commit

## Motivation and Context

- Synchronize with the next small pcds-envs update today/tomorrow

## How Has This Been Tested?

docs build ok

is docs


klauer (approved):

docs/source/api.rst
```diff
@@ -323,8 +360,9 @@ pcdsdevices.interface
 
     pcdsdevices.interface.BaseInterface
     pcdsdevices.interface.FltMvInterface
-    pcdsdevices.interface.LightpathInOutMixin
+    pcdsdevices.interface.LegacyLightpathMixin
```
tangkong, line 363: I guess we haven't regenerated these in a while

ZLLentz: Yeah, it generates during the docs `make` and we probably don't do it manually very often

tangkong (approved):

docs/source/api.rst
klauer: It's still generated each time in the actions, though, so this is only relevant for the master branch recording of `api.rst`:

(A link from last week)
https://github.com/pcdshub/pcdsdevices/blob/f7d38c582a48086cbdb0813d98504b66d2c7e0f7/master/_sources/api.rst.txt?plain=1#L363

PR #1132, opened 2023-04-25
ENH: Added ability for IMS configs to be read from questionnaire + extra functions

spenc333:
## Description

Added to the IMS class in epics_motor.py the ability to grab questionnaire stage identities and apply them to happi created motor objects that then applies it to the parameter manager for the given PV. Also added were functions that allow the user to see the given configuration parameters or the current parameters of that base PV.

## Motivation and Context

Fixes jira issue: https://jira.slac.stanford.edu/browse/LCLSECSD-552

## How Has This Been Tested?

Tested on a hutch machine (xcs-control, xpp-control) in a basic hutch-python environment in pcds-5.7.1 using active questionnaires and a simple conf.yml file that specifies a specific hutch.

Docstrings have been added to code where necessary.

## Pre-merge checklist


pcdsdevices/epics_motor.py
```diff
@@ -735,6 +737,27 @@ def flag_is_cleared(value=None, **kwargs):
...
+                                 'and was likely not initialized from happi')
+        return self._md
+
+    @md.setter
+    def md(self, new_md):
+        """ initialize attributes when md is set """
+        self._md = new_md
+        self._extra = self._md.extraneous
+        self._id = self._extra['_id']
+        self._stageidentity = self._extra['stageidentity']
```
tangkong, line 753: Correct me if I'm wrong, but `stageidentity` and `pvbase` seem to be fields you're assuming are in your `new_md` argument.  With such a general looking setter property, I'm worried others might try to set metadata information that does not fit this use case and cause errors.  

There's a few things you could do off the top of my head:
* Re-name these methods to make them more specific to the pmgr use-case  (or at least flesh out the docstring to describe this more spefically)
* add some logic to deal with missing keys
``` python
self._stageidentity = self._extra.get('stageidentity')  # returns None if key does not exist 
if self._stageidentity is None:
    # raise some error message?
    return
```

unfortunately `HappiItem` doesn't let you add info to it, so if you really wanted to preserve the container structure you'd have to make a new container.  I'm starting to think you might want to just pull the info you need from the happi container and store your pmgr metadata in a different field.

pcdsdevices/epics_motor.py
```diff
@@ -735,6 +737,27 @@ def flag_is_cleared(value=None, **kwargs):
...
 
+    @property
+    def md(self):
+        if self._md is None:
+            raise AttributeError('Device does not have an attached md, '
+                                 'and was likely not initialized from happi')
+        return self._md
+
+    @md.setter
+    def md(self, new_md):
```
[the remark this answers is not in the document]
ZLLentz: Maybe we should add a `post_happi_md` hook that `happi` looks for to make stuff like this easier to manage and more clear.

I think the expectation here is that a `HappiItem` comes in from the `happi` questionnaire load.

pcdsdevices/epics_motor.py
ZLLentz: I think the most likely `KeyError` here will be from someone adding an IMS motor to the non-questionnaire `happi`. I agree that this case needs to be handled.

Any thoughts on only allowing this to be set exactly once to avoid people trying to set over it?

pcdsdevices/epics_motor.py
```diff
@@ -735,6 +737,27 @@ def flag_is_cleared(value=None, **kwargs):
...
+        """ initialize attributes when md is set """
+        self._md = new_md
+        self._extra = self._md.extraneous
+        self._id = self._extra['_id']
+        self._stageidentity = self._extra['stageidentity']
+        self._pvbase = self._extra['pvbase']
+        if self._stageidentity == "NEW":
+            print(f"This is a new stage, parameter configuration does not exist yet. Please manually configure {self._id}.")
```
ZLLentz, line 756: Note that this print will end up being interleaved into the big loading block, so it might be easy to miss. Is there another method that could be more helpful to show this during? Or would that be too spammy?

pcdsdevices/epics_motor.py
```diff
@@ -735,6 +737,27 @@ def flag_is_cleared(value=None, **kwargs):
...
+        else:
+            IMS._setup_and_check_pmgr()
+            self._pm.set_config(self._pvbase, self._stageidentity)
```
ZLLentz, line 759: Note that this will cause pmgr database changes every time anyone starts a new hutch python session, even from dev machines etc.
I'm not sure if this is a problem or not, but I thought I'd note it.

pcdsdevices/epics_motor.py
```diff
@@ -758,6 +781,47 @@ def get_configuration(self):
...
+        Return the current configuration parameters of the motor
+        in the parameter manager
+
+        Parameters
+        ----------
+        cfgname : str
+               The name of the configuration
+
+        Returns
+        _______
```
ZLLentz, line 795: Docs  build expects hyphens
```suggestion
        -------
```

pcdsdevices/epics_motor.py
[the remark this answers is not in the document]
tangkong: We thought about something (sort of) [similar with lightpath](https://github.com/pcdshub/pcdsdevices/blob/02b295e236a21b4d8fcdc6969041f5b8a3b5fa03/pcdsdevices/interface.py#L1787-L1793), I think. 

I see I got this confused.  Happi uses this setter when it [creates the device ](https://github.com/pcdshub/happi/blob/c46a5e9e81741b01e5f98629cae460662be33d1c/happi/loader.py#L200)
and the questionaire has those expected fields.  

This is basically identical to what I did so I guess I can't complain 😆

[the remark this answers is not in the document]
tangkong: Now that we've hit upon two cases where we want a `post_happi_md` hook, I think it's probably worth adding.  I could see the argument being made that devices should have the ability to do something with the metadata they're given, and as I demonstrated, the intention of the md-setter can be easily misconstrued

pcdsdevices/epics_motor.py
```diff
@@ -810,6 +874,7 @@ def diff_configuration(self, cfgname=None):
         Raises an exception if the comparison fails for any reason.
         """
         IMS._setup_and_check_pmgr()
+        print("Actual Values, Configuration Values")
```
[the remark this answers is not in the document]
spenc333: The function that follows prints the contents of a two "columns" for comparing the values in each row but there's no header so the print statement basically servers as the header.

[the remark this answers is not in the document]
tangkong: If you assign the output of this method to a variable, does the table actually print?
e.g.: `var = motor.find_configuration(pattern)`

The following method calls [`pmgr.diff_config`](https://github.com/pcdshub/pmgr/blob/ff674f427819381b0880ef45706486465449514f/pmgr/pmgrAPI.py#L197)  -> [`pmgrobj.diffConfig`](https://github.com/pcdshub/pmgr/blob/ff674f427819381b0880ef45706486465449514f/pmgr/pmgrobj.py#L1053) don't seem to have prints in them, rather they just pass a dictionary through.

pcdsdevices/epics_motor.py
spenc333: I tested this and it looks like the changes only get applied when you are on the specific hutch machine that the change is meant to happen in.

pcdsdevices/epics_motor.py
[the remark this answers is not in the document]
tangkong: Opened https://github.com/pcdshub/happi/pull/308 to address this, if that gets merged we should loop back on this and the lightpath mixins to use that new hook.  This is outside the scope of this PR though, and the implementation here should work fine in the interim.  I'll resolve this

pcdsdevices/epics_motor.py
spenc333: It was expected to print in the big block but instead of a plain print statement, I will do a "logger.warning(...)" to at least add some differentiation.

ZLLentz: You also have access to `self.log.warning(...)` for ophyd devices, which adds in some metadata about the calling device (I think?)

pcdsdevices/epics_motor.py
```diff
@@ -735,6 +737,27 @@ def flag_is_cleared(value=None, **kwargs):
...
+        """ initialize attributes when md is set """
+        self._md = new_md
+        self._extra = self._md.extraneous
+        self._id = self._extra['_id']
+        self._stageidentity = self._extra['stageidentity']
+        self._pvbase = self._extra['pvbase']
+        if self._stageidentity == "NEW":
+            print(f"This is a new stage, parameter configuration does not exist yet. Please manually configure {self._id}.")
+        else:
+            IMS._setup_and_check_pmgr()
```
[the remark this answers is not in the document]
tangkong: To Zach's point here, we should probably wrap this in a `try: except:` to catch these exceptions.  Otherwise if we can't connect to the pmgr the loading process will get interrupted.  

In this function block, it seems like we should just return if an error is thrown from this point.  We got some metadata from happi but could not connect to pmgr, so we don't continue with setting the config (next line)

pcdsdevices/epics_motor.py
tangkong: To provide a suggestion here as to how you might handle this, you could choose to only continue with this setter method if the `stageidentity` key exists.  My original suggestion from above still applies here.  

I think future modifications to `happi` are positioned to better solve the only-use-method-once problem.

pcdsdevices/epics_motor.py
```diff
@@ -874,8 +875,15 @@ def diff_configuration(self, cfgname=None):
...
         """
         IMS._setup_and_check_pmgr()
-        print("Actual Values, Configuration Values")
-        return self._pm.diff_config(self.prefix, cfgname)
+
+        d = self._pm.diff_config(self.prefix, cfgname)
+        table = PrettyTable()
+        table.field_names = ["Parameter", "Actual", "Configuration"]
+        for key, value in d.items():
+            actual = d[key][0]
```
tangkong, line 883: `dict.items()` gives you the `key` and `value` as you've set out in the for loop, so you can actually just access the value specifically via: `value[0]` or `value[1]`.  Here you're just assigning `value` and tossing it to the side

tangkong (commented):
I think this is looking good on the whole.  I may seem like I'm nitpicking on exception handling, but just take that as a sign that you did a good job with the big things, leaving only the finer points to poke at.  😄 

I do think handling exceptions / missing metadata keys gracefully here is important.  Since this `.md` setting / pmgr querying happens when the device is loaded, if it errors out it can prevent subsequent devices/objects from loading

pcdsdevices/epics_motor.py
spenc333: I made changes to the first point made here but I guess I forgot to push it. This is what the change looks like. 
Screen Shot 2023-05-01 at 12 38 43

tangkong: Looks good 👍   One last suggestion is that you also put `self._id` in the first logger.warning method so we know exactly which object failed.  (also it's technically more readable / "pythonic" to use `is` to check if a value is None)

``` python
if self._stageidentity is not None:
    logger.warning(f"Stage Identity has not been set for object: {self._id}.  Configure manually.")
```

tangkong:
Oh the last-last thing is to add pre-release notes.

pcdsdevices/epics_motor.py
spenc333: Wouldn't it be "is None"?

pcdsdevices/epics_motor.py
```diff
@@ -735,6 +738,31 @@ def flag_is_cleared(value=None, **kwargs):
...
+                                 'and was likely not initialized from happi')
+        return self._md
+
+    @md.setter
+    def md(self, new_md):
+        """ initialize attributes when md is set """
+        self._md = new_md
+        self._extra = self._md.extraneous
+        self._id = self._extra.get('_id')
+        self._pvbase = self._extra['pvbase']
```
ZLLentz, line 754: There's one more dict access here, this one could also raise a `KeyError`.
I suggest either changing this to a .get or using self.prefix instead.
```suggestion
        self._pvbase = self._extra.get('pvbase')
```

ZLLentz (approved):
I think you've handled everything except for the possible error case on IMS._setup_and_check_pmgr on line 763.

If you're done with this and would rather let it go and do other things I'll make an issue for it and we can tackle it later.

ZLLentz:
Last tiny thing, the pre-release file needs to be renamed:
```
Error: Artifact path is not valid: /_sources/upcoming_release_notes/1132-ENH:_Added_ability_for_IMS_configs_to_be_read_from_questionnaire_+_extra_functions.rst.txt. Contains the following character:  Colon :
```
TIL there were rules for this

PR #1133, opened 2023-05-04
Hxrsss filter wheel update

spenc333:
## Description

Changed previous filter wheel PV that first defined the motor PV to the states PV.

## Motivation and Context

Wanted to add the ability to change states using enums/state names from a hutch-python session instead of remembering motor positions.

## How Has This Been Tested?

Tested from a hutch machine (xcs-control) in a hutch python session using a simplified conf.yml file and custom happi device file that contained the spectrometer device . Also tested physically to make sure the device moved between the desired states.

Updated doc name within filter attribute.

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1133-Hxrsss_filter_wheel_update.rst
+++ b/docs/source/upcoming_release_notes/1133-Hxrsss_filter_wheel_update.rst
@@ -1,0 +1,31 @@
+1133 Hxrsss filter wheel update
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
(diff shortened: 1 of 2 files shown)


ZLLentz:
Looks like a strictly positive change to me :+1:

Note that we didn't actually remove any features here, because the original motor is now embedded in the state device e.g. `hxrsss.filter.motor` as defined here: https://github.com/pcdshub/pcdsdevices/blob/02b295e236a21b4d8fcdc6969041f5b8a3b5fa03/pcdsdevices/state.py#L448, so 

We can merge this if you add a pre-release notes page with a device updates section for the hxrsss

ZLLentz (approved):
lgtm

PR #1142, opened 2023-07-10
DOC/BLD: release notes for v7.4.3, remove pins

ZLLentz:
## Description

- Squeeze another bugfix tag in before the pcds conda release
- Remove the xraydb and sqlalchemy pins

## Motivation and Context

- There's no reason to delay this tag, and I'd like to minimize dev versions at start of run resume
- The pins are no longer needed, since the most recent versions of these packages are compatible with each other.


tangkong (approved):

ZLLentz:
Hang on, actually, I'm going to slip in one packaging fix here before merging that deserves some review

requirements.txt
```diff
@@ -13,5 +13,3 @@ pytmc>=2.7.0
 pyyaml
 schema
 scipy
-sqlalchemy<2.0.0
```
tangkong, line 16: For whatever reason this isn't being picked up by our pip tests?...
https://github.com/pcdshub/pcdsdevices/actions/runs/5524933875/jobs/10077909012?pr=1142#step:10:135

ZLLentz: Some other package must still be constraining them

ZLLentz: aha https://github.com/pcdshub/pcdscalc/blob/master/requirements.txt

tangkong: And the conda recipe in pcdshub/pcdscalc is still pinned as well, but the conda feedstock is not.  

So many recipes

ZLLentz:
I think this should be all correct now- good to merge + tag before the end of the day?

tangkong (approved):
This does build, and tests pass, so this should be good to go.  I reran a pip test in the offchance there were more hidden pins.

ZLLentz:
I reran the pypi build and verified that there's no longer this pin chain in the loop

PR #1144, opened 2023-07-26
ENH: add a setter to the epics motor interface limits attribute.

ZLLentz:
## Description

Add a setter to the epics motor interface limits attribute.

## Motivation and Context

XCS was having a problem with this line in `LookupTablePositioner`:
https://github.com/pcdshub/pcdsdevices/blob/af187b6b26235f95056cf9b168bb6decf6627c4c/pcdsdevices/pseudopos.py#L990
Where it sets the limits of the real motor.

Rather than consider ways that this could be done differently to avoid EPICS puts on init, I opted to add the missing limits setter as a quick fix.

## How Has This Been Tested?

Live in XCS

Release notes only

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -177,16 +177,26 @@ Limit Switch: {switch_limits}
 """
 
     @property
-    def limits(self):
+    def limits(self) -> tuple[float, float]:
         """Override the limits attribute"""
         return self._get_epics_limits()
 
-    def _get_epics_limits(self):
+    def _get_epics_limits(self) -> tuple[float, float]:
         limits = self.user_setpoint.limits
         if limits is None or limits == (None, None):
             # Not initialized
             return (0, 0)
         return limits
+
+    @limits.setter
+    def limits(self, lims: tuple[float, float]):
+        """
+        Write to the epics limits on setattr
+
+        Expects a tuple of two elements, the low and high limits to set.
+        """
+        self.low_limit_travel.put(min(lims))
+        self.high_limit_travel.put(max(lims))
 
     def enable(self):
         """
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -195,8 +195,24 @@ Limit Switch: {switch_limits}
 
         Expects a tuple of two elements, the low and high limits to set.
         """
-        self.low_limit_travel.put(min(lims))
-        self.high_limit_travel.put(max(lims))
+        # Wrong len is probably a bigger mistake
+        if len(lims) != 2:
+            raise ValueError("Limits should be a tuple of 2 values.")
+        # Adjust for moment of dsylexia
+        low = min(lims)
+        high = max(lims)
+        if low == high:
+            # User probably meant to disable limits
+            low = 0
+            high = 0
+        orig_high = self.high_limit_travel.get()
+        if low > orig_high:
+            # Set high first so we make room for low
+            self.high_limit_travel.put(high)
+            self.low_limit_travel.put(low)
+        else:
+            self.low_limit_travel.put(low)
+            self.high_limit_travel.put(high)
 
     def enable(self):
         """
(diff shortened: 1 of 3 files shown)


pcdsdevices/epics_motor.py
```diff
@@ -177,17 +177,27 @@ def format_status_info(self, status_info):
...
         return limits
 
+    @limits.setter
+    def limits(self, lims: tuple[float, float]):
+        """
+        Write to the epics limits on setattr
+
+        Expects a tuple of two elements, the low and high limits to set.
+        """
+        self.low_limit_travel.put(min(lims))
```
klauer, line 198: Maybe `if not len(lims) == 2: raise ValueError(...)`

And _maybe_ have special handling for `if min(lims) == max(lims)` ? That implies "disable limits or low, high = (0, 0)" in the motor record world I believe

pcdsdevices/epics_motor.py
```diff
@@ -177,17 +177,27 @@ def format_status_info(self, status_info):
...
 
-    def _get_epics_limits(self):
+    def _get_epics_limits(self) -> tuple[float, float]:
         limits = self.user_setpoint.limits
         if limits is None or limits == (None, None):
             # Not initialized
             return (0, 0)
         return limits
 
+    @limits.setter
```
klauer, line 191: Setters that put to PVs have been divisive in the past...
I think in this case the utility of it is justified and am on board

klauer (approved):

pcdsdevices/epics_motor.py
```diff
@@ -177,17 +177,27 @@ def format_status_info(self, status_info):
...
+    def limits(self, lims: tuple[float, float]):
+        """
+        Write to the epics limits on setattr
+
+        Expects a tuple of two elements, the low and high limits to set.
+        """
+        self.low_limit_travel.put(min(lims))
+        self.high_limit_travel.put(max(lims))
```
tangkong, line 199: I know this is an entirely logical assumption to make, but are we sure we can be certain of this?  We don't have any whacky motors that have the limits switched and would require the order to matter?

(a genuine question, I could see some zany IOC that swaps things around for some unreasonable reason)

Maybe my imagination is getting the better of me

tangkong (approved):
LGTM despite my harebrained comment

pcdsdevices/epics_motor.py
klauer: That line of questioning reminds me - it won't be a problem for `high_limit_travel < low_limit_travel` on the IOC side as you've implied but there _is_ a potential issue there:
You can't set a high-limit below the low limit, and you can't set the low limit above the current high limit. That needs some special handling to my recollection.

pcdsdevices/epics_motor.py
ZLLentz: Both of these are worth doing

pcdsdevices/epics_motor.py
ZLLentz: I agree, it's a lot less annoying than requiring some other method that every motor-like needs for changing the limits value.

pcdsdevices/epics_motor.py
ZLLentz: I think we're guaranteed this by the way the IOC works, but it's a good question

ZLLentz: oh no you're so right Ken, there's some awful stateful behavior here depending on the original limit settings

ZLLentz:
This became a bit more involved so I added some unit tests too

pcdsdevices/epics_motor.py
klauer: Glad I commented then, despite the haziness... Fix looks good, I think

klauer (approved):
Updated fix LGTM - bonus points for the nice test

tangkong (approved):
LGTM 👍

PR #1164, opened 2023-09-21
add energy_c for first tower only

vespos:
## Description
Add energy control for the first crystal only

## Motivation and Context
When sending the beam to the XRT while not multiplexing with XPP, we want to be able to change the tower1 energy only, without changing alignment of the second tower, which is not in the way of the beam.

## How Has This Been Tested?
On xpp3 with XPP LODCM.
Waiting for scientist's confirmation that it works as intended for them.

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1164-Add_E1C_to_LODCM.rst
+++ b/docs/source/upcoming_release_notes/1164-Add_E1C_to_LODCM.rst
@@ -1,0 +1,30 @@
+1164 Add E1C to LODCM
+#################
+
+API Changes
+-----------
+- self.E1C now available to move first tower energy only on LODCMs
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- LODCM: Add the E1C pseudo-interface
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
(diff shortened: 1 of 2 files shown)


vespos:
I wonder if we could make the code cleaner by having a main energy be a combination of the individual energies on tower1 and 2? By doing so we would loose the direct link between the real variables and the pseudo-positioner of the main energy though... Probably not desirable

docs/source/upcoming_release_notes/1164-Add_E1C_to_LODCM.rst
```diff
@@ -0,0 +1,30 @@
+1164 Add E1C to LODCM
+#################
+
+API Changes
+-----------
+- self.E1C now available to move first tower energy only on LODCMs
```
ZLLentz, line 6: Should we rename this section to API breaks? The intent here is to flag things that would need a major release increment. This probably all should have been in the `Device Updates` section.

ZLLentz (approved):
This looks reasonable to me, pending testing
I think this whole module could use some cleanup but that's not the burden of your PR

PR #1167, opened 2023-09-27
MAINT: include a precision on the state_velo signal

ZLLentz:
## Description

Set the precision of state_velo to 3, instead of the default 0.

This signal shows up in typhos with a default precision of 0, which means the ui doesn't show any details past the decimal point.

This is also fixed generically in typhos itself, but it's good to set it here too for the benefit of other applications and to make it work cleanly with older versions of typhos.

## Motivation and Context

We need to see floats as floats and not as ints

## How Has This Been Tested?

Interactively only, see the following which is this PR + typhos's tag without the typhos-side fix:
image

pre-release notes

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1167-maint_state_velo_prec.rst
+++ b/docs/source/upcoming_release_notes/1167-maint_state_velo_prec.rst
@@ -1,0 +1,33 @@
+1167 maint_state_velo_prec
+##########################
+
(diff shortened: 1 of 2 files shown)


klauer (approved):
Seems correct to me

tangkong (approved):

PR #1173, opened 2023-10-04
FIX: tmo spec satt has 6 states, not 7

ZLLentz:
## Description

Switch state count on this device from 7 to 6

## Motivation and Context

This causes an error when we load the device and breaks the device in general
There is some deeper bug to fix in the library but this at least patches the surface-level behavior

## How Has This Been Tested?

Interactively only (the big traceback goes away, the state enum gets set up properly)

Here only

## Pre-merge checklist


tongju12 (approved):
Yes, should be 6 instead of 7. Thanks Zach!

klauer (approved):

PR #1174, opened 2023-10-09
DOC: specify k2700 cpts are for ranges

KaushikMalapati:
## Description

I added 'range' to the doc attributes and names of the components dcv, acv, dci, and aci of the K2700 class.

## Motivation and Context

I originally thought that the pvs associated with these components were to readback the values in their names, i.e. GetDCV would read the direct current voltage, but they actually get the range configured for each measurement. You use :Reading to get the actually measurement.

## How Has This Been Tested?

N/A

New upcoming release notes

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/pim.py
+++ b/pcdsdevices/pim.py
@@ -572,14 +572,14 @@ class K2700(BaseInterface, Device):
               doc='Identity (name) of this device')
     reading = Cpt(EpicsSignalRO, ":Reading", kind="normal",
                   doc='Trigger and return a new measurement')
-    dcv = Cpt(EpicsSignalRO, ":GetDCV", kind="normal",
-              doc='DC voltage')
-    acv = Cpt(EpicsSignalRO, ":GetACV", kind="normal",
-              doc='AC voltage')
-    dci = Cpt(EpicsSignalRO, ":GetDCI", kind="normal",
-              doc='DC current')
-    aci = Cpt(EpicsSignalRO, ":GetACI", kind="normal",
-              doc='AC current')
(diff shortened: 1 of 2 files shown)


pcdsdevices/pim.py
```diff
@@ -572,14 +572,14 @@ class K2700(BaseInterface, Device):
...
                   doc='Trigger and return a new measurement')
-    dcv = Cpt(EpicsSignalRO, ":GetDCV", kind="normal",
-              doc='DC voltage')
-    acv = Cpt(EpicsSignalRO, ":GetACV", kind="normal",
-              doc='AC voltage')
-    dci = Cpt(EpicsSignalRO, ":GetDCI", kind="normal",
-              doc='DC current')
-    aci = Cpt(EpicsSignalRO, ":GetACI", kind="normal",
-              doc='AC current')
+    dcv_range = Cpt(EpicsSignalRO, ":GetDCV", kind="normal",
```
tangkong, line 575: So this is technically a sort of "API Break", since we're changing the name of a component on this device.  If there are older scripts that expect `K2700.dcv`, they won't be able to find that name any longer. 

Practically, if you **know** you're the only one using this device and you won't break others' usage of the device, this is fine.  I just want to make sure this is the case.  

If it were me I'd update the documentation alone, but that may be out of an abundance of caution

KaushikMalapati: Thanks for the clarification, release notes have been updated to note the API breaks. I recently added this device class and there aren't any happi entries using it other than one I made so it should be safe to change the component names.

tangkong (approved):
LTGM, thanks for keeping the docs clear 👍

PR #1176, opened 2023-10-16
DOC/REL: v8.1.0

ZLLentz:
## Description

This is a release notes PR, incorporating the work from the reviewers. Reviewing is optional, but it's a good idea to make sure I didn't misrepresent your work since I always slightly tweak them.

Done here:
- Run the release notes script
- Hand-edit a lot: there were a lot of good, detailed notes that ultimately no longer make sense in the context of v8.0.0 -> v8.1.0 because they reference names or features that never made it into a tag. For example, the API break in #1174 doesn't need to be reported here because the API that got broken was not present at all in v8.0.0.

## Motivation and Context

I'm going to rebuild pcds-5.8.0 to incorporate some unrelated fixes, so we might as well bring in new features from `pcdsdevices` at the same time.


ZLLentz:
I'm going to double-check that the CI-generated docs look good once the CI is done running

tangkong (approved):

nrwslac (approved):
lgtm!

PR #1179, opened 2023-11-02
 TMO GMD device classes

KaushikMalapati:
## Description

Adding devices from /cds/group/pcds/pyps/apps/hutch-python/tmo/tmo/tmo_beamline_control.py

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-4058

## How Has This Been Tested?

Has not been tested, but will do so programmatically

Pre-release notes

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1179-tmo_gmd_classes.rst
+++ b/docs/source/upcoming_release_notes/1179-tmo_gmd_classes.rst
@@ -1,0 +1,31 @@
+1179 tmo_gmd_classes
+#################
+
+API Breaks
+----------
+- Moving K2700 and IM3L0_K2700 to keithley.py
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- K6514, GMD (previously unimplemented), GMDPreAmp, SXRGasAtt
+- Taken from /cds/group/pcds/pyps/apps/hutch-python/tmo/tmo/tmo_beamline_control.py with some modifications
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- KaushikMalapati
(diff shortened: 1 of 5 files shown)


KaushikMalapati:
I haven't added the undulator devices or beamline devices since they seem fairly tmo specific. Does it make sense to add them to pcdsdevices?

ZLLentz:
Sorry for the delay on reviews here, I'll be going through my github backlog today, starting with this PR.

ZLLentz:
> I haven't added the undulator devices or beamline devices since they seem fairly tmo specific.

I think the TMO devices are definitely best-off left in TMO (there's a lot of them here)
We might want to bring in the undulator devices since they are shared, but maybe not in the same PR. There's some weird code in there that we should evaluate first.

For this PR, everything looks normal and good- a solid, faithful port. I'm going to go through it by hand and check if there's anything James put in there that I think is slightly off, such as the kinds and the docs

ZLLentz:
Actually, thinking about it more lucidly- maybe it's more productive to merge this as-is, and then do a follow-up myself with a bunch of banal changes? It seems silly to ask you to change minor things when the original device source isn't you, and the decisions involved probably need more than just me to figure out anyway. The port is clean.

KaushikMalapati:
I added many of the docs and changed some of the kinds (most were originally hinted). Should I change back the kinds to what James had and then merge? I was thinking that having most kinds as hinted would have too much information on the device screen but that might not have been a good idea.

ZLLentz:
I think if you have/had specific ideas about kinds/docs it's good to include the changes- I had the same opinion as you re: too many hinted, etc.
I suspect that if I brought this list of docs/kinds in front of a few other engineers we'd be able to accumulate some additional ideas for some of the empty docs and help us identify read-only variables and figure out which things should be kind=config

KaushikMalapati:
In that case, feel free to merge. I will watch out for another pr related to this to see what sort of changes are made.

ZLLentz (approved):

PR #1185, opened 2023-11-13
ENH: Add OPCPA lxt class

### LINKED ISSUE #1184: LCLS-II Laser Locker "lxt" class
## Expected Behavior
The LCLS-I laser locker has a class, referred to as the lxt motor, for adjusting laser arrival time via python. We need the same capability for the new laser locker system. 

## Context / environment
TMO would like to try scanning timing during their commissioning shift.

slactjohnson:
## Description
PR to add a class supporting the new OPCPA laser locker system. This is essentially a copy-paste of the old class, with new PVs, to keep a consistent user interface for the operators. 

I _would_ like to remove the notepad PV setup at some point, because I plan on having a user offset built in to the IOC eventually, but for now it's not there so I'll leave that in. Future PR. 

## Motivation and Context
closes #1184 

## How Has This Been Tested?
Tested on the OPCPA system in the laser hall this weekend.

Comments in the code, with some testing details here: https://confluence.slac.stanford.edu/pages/viewpage.action?pageId=407260134

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/lxe.py
+++ b/pcdsdevices/lxe.py
@@ -414,6 +414,170 @@ Current position (user, dial): {position} [{units}] {dial_pos} [s]
 """
 
 
+class Lcls2LaserTiming(FltMvInterface, PVPositioner):
+    """
+    A "lxt" motor for the LCLS-II laser locker, AKA the OPCPA laser locker.
+    """
+
+    tab_component_names = True
+
+    verbose_name = 'Laser X-ray Timing'
+
+    _tgt_time = Cpt(EpicsSignal, ':PHASCTL:DELAY_SET', auto_monitor=True,
+                    kind='omitted',
+                    doc='The internal nanosecond-expecting signal.'
+                    )
+    setpoint = Cpt(_ScaledUnitConversionDerivedSignal,
+                   derived_from='_tgt_time',
+                   derived_units='s',
+                   original_units='ns',
+                   kind='hinted',
+                   doc='Setpoint which handles the timing conversion.',
+                   limits=(-100e-6, 100e-6),
+                   )
+    notepad_setpoint = Cpt(NotepadLinkedSignal, ':lxt:OphydSetpoint',
+                           notepad_metadata={'record': 'ao',
+                                             'default_value': 0.0},
+                           kind='omitted'
+                           )
+    notepad_readback = Cpt(NotepadLinkedSignal, ':lxt:OphydReadback',
+                           notepad_metadata={'record': 'ao',
+                                             'default_value': 0.0},
+                           kind='omitted'
+                           )
+    user_offset = Cpt(NotepadLinkedSignal, ':lxt:OphydOffset',
+                      notepad_metadata={'record': 'ao', 'default_value': 0.0},
+                      kind='normal',
+                      doc='A Python-level user offset.'
+                      )
+
+    # The phase shifter will be moved after the above record is touched, so
+    # use its done status:
+    done = Cpt(EpicsSignal, ':PHASCTL:DELAY_MOVING', auto_monitor=True, kind='omitted')
+    done_value = 0
+
+    def __init__(self, prefix='', *, egu=None, **kwargs):
+        if egu not in (None, 's'):
+            raise ValueError(
+                f'{self.__class__.__name__} is pre-configured to work in units'
+                f' of seconds.'
+            )
+        super().__init__(prefix, egu='s', **kwargs)
+
+    @user_offset.sub_value
+    def _offset_changed(self, value, **kwargs):
+        """
+        The user offset was changed.  Update the setpoint attribute.
+        """
+        self.setpoint.user_offset = value
+
+    def _setup_move(self, position):
+        """Update the notepad setpoint, move, and do not wait."""
+        try:
+            signal = self.notepad_setpoint
+            if signal.connected and signal.write_access:
+                if signal.get(use_monitor=True) != position:
+                    signal.put(position, wait=False)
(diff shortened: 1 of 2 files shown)


slactjohnson:
Assigned to Ken and Zach since I think that they're the only remaining people who have touched this file previously. I also had some issues this weekend that I could use some advice on. 

I found that the class worked fine in interactive mode, but seemed to get out of sync while using a sim detector in a scan. I used the available settle_time parameter (e.g. `timing = Lcls2LaserTiming("LAS:LHN:LLG2:01", name="Bay 3 Locker", settle_time=2.0)`), and this seemed to help somewhat, but I think that the phase shifter may be reporting "done" too early, or perhaps inconsistently. This makes scanning via bluesky rather fragile. 

See additional notes referenced in the PR. TMO would like to do time scans during their commissioning shift the evening of 11/13. I think that they could work around this, but if we can get it in a better state before then that would be great.

slactjohnson:
While working on this, there was something I didn't quite understand: why was the scaling inverted? See here: https://github.com/pcdshub/pcdsdevices/pull/1185/files#diff-4e3e1bccacabb3b1a5d073e8d9247a4e650ce3e7cc9d59b1ad10519ef89f895dR219-R231

I'm also not entirely sure why a scale of -1 is "Perhaps the only scale that will make sense". This negation confused me for a bit, but the code still works. I'm curious if anyone knows the context surrounding this decision.

ZLLentz:
I think the reasoning for that comment in the linked scaled unit conversion signal is that the purpose of the parent signal class is to convert unit a to unit b, so any scaling on top of that done in this child class (other than a change in direction) is redundant/clashes with the unit rescaling.

pcdsdevices/lxe.py
```diff
@@ -414,6 +414,170 @@ def format_status_info(self, status_info):
...
+                    signal.put(position, wait=False)
+        except Exception as ex:
+            self.log.debug('Failed to update notepad setpoint to position %s',
+                           position, exc_info=ex)
+        super()._setup_move(position)
+
+        # Something is wrong with done signal, just sleep and pretend
+        time.sleep(1)
+        self._move_changed(value=1-self.done_value)
+        self._move_changed(value=self.done_value)
```
ZLLentz, line 489: I notice this suspicious bit is copied from the old class. Is this behavior still needed with the new setups? Could this be responsible for some of the strange behavior you're seeing?

pcdsdevices/lxe.py
```diff
@@ -414,6 +414,170 @@ def format_status_info(self, status_info):
...
+        dial_pos = self.dial_pos
+        position = get_status_float(
+            status_info, 'position', precision=3, format='e')
+        units = get_status_value(status_info, 'setpoint', 'units')
+        return f"""\
+Virtual Motor {self.verbose_name} {self.prefix}
+Current position (user, dial): {position} [{units}] {dial_pos} [s]
+"""
+
+
```
ZLLentz, line 580: I've verified that the methods here are exactly the same as in `LaserTiming`. Is there any consideration for reducing the code duplication? There are probably clean ways to source the common functionality from the same place. Maybe we can refactor in a follow-up PR.

pcdsdevices/lxe.py
slactjohnson: Ah, perhaps. I will try removing that old behavior when I can.

pcdsdevices/lxe.py
slactjohnson: Sounds good. As mentioned in a previous comment, I'd like to remove some of the additional functionality that I think will be taken up by the IOC in a future PR. I can refactor the remaining common at the same time.

slactjohnson:
TMO will not use the python class tonight; I will try removing the odd sleep from previous class before merging.

pcdsdevices/lxe.py
ZLLentz: Yeah, I think all three lines here should be removed if the moving/done signal value is accurate.

slactjohnson:
I finally got a chance to test this more. I think this is ready for review.

pcdsdevices/lxe.py
```diff
@@ -414,6 +413,165 @@ def format_status_info(self, status_info):
...
+                    kind='omitted',
+                    doc='The internal nanosecond-expecting signal.'
+                    )
+    setpoint = Cpt(_ScaledUnitConversionDerivedSignal,
+                   derived_from='_tgt_time',
+                   derived_units='s',
+                   original_units='ns',
+                   kind='hinted',
+                   doc='Setpoint which handles the timing conversion.',
+                   limits=(-100e-6, 100e-6),
```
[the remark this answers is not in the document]
slactjohnson: This is a good thing to think about. Off-hand, I think there really isn't something we can determine easily _a priori_, in part because the limit would be related to laser rep-rate, which could change even during the course of a single experiment. 
The 100us limits seemed OK to me, since they're wide enough for almost all experiments, and can be modified on the fly if truly necessary.

ZLLentz:
re-running the failed check, it looks like a web resource timeout thing

slactjohnson:
@ZLLentz Thanks. How does one do that? I didn't see an option to do that when clicking through this web interface.

ZLLentz:
You can click "details" to navigate to the github actions tab and there should be a re-run combo box in the top right area with options to re-run all or re-run failed workflows only

I'm not entirely sure what the permission sets are for seeing/using the option
image

slactjohnson:
Yep, it's there, I just missed it. Thanks! I'll merge then since everything is passing.

PR #1187, opened 2023-11-14
ENH: set alarm config to hinted only for beckhoff motors

ZLLentz:
## Description

For the typhos templates that contain BeckhoffAxis instances, set the alarm widget to only check "hinted" signals.

## Motivation and Context

1. All of the sub-fields of this motor record tend to go into an alarm state at the same time. This is very spammy and redundant.
2. Due to the bug described in https://jira.slac.stanford.edu/browse/ECS-847, a bunch of the fields can remain stuck in an alarm state indefinitely, even if the alarm is cleared. This can leave a lot of phantom errors that aren't real.
3. Today, I heard that this perma-alarm was upsetting people.

Due to the nature of the bug, the alarm states clear when the associated value updates.

One setting I can easily change in these widgets is to drop the alarm widget from checking all normal signals to only checking hinted signals. Typically, this is just the `.RBV` field which tends to have the correct alarm state due to how often it updates.

## How Has This Been Tested?

Interactively only
- I compared sp1k1_mono before/after
- I found motors that still had an error state even after this change (see rtdsx01, still major but less spammy)
- I double-checked that the VLSOptics screens still ran and gave reasonable alarms (those look better now too, though still a major alarm)

Before and after for sp1k1_mono:
image

pre-release docs

## Pre-merge checklist
- [ ] ~Code contains descriptive docstrings, including context and API~

### DIFF
--- a/pcdsdevices/ui/BeckhoffAxis.embedded.ui
+++ b/pcdsdevices/ui/BeckhoffAxis.embedded.ui
@@ -79,19 +79,22 @@
      <property name="show_expert_button" stdset="0">
       <bool>true</bool>
      </property>
+     <property name="alarmKindLevel" stdset="0">
+      <enum>TyphosPositionerWidget::HINTED</enum>
+     </property>
     </widget>
    </item>
   </layout>
  </widget>
  <customwidgets>
   <customwidget>
-   <class>TyphosPositionerWidget</class>
-   <extends>QWidget</extends>
+   <class>TyphosPositionerRowWidget</class>
+   <extends>TyphosPositionerWidget</extends>
    <header>typhos.positioner</header>
   </customwidget>
   <customwidget>
-   <class>TyphosPositionerRowWidget</class>
-   <extends>TyphosPositionerWidget</extends>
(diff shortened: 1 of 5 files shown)


pcdsdevices/ui/BeckhoffAxis.embedded.ui
```diff
@@ -79,19 +79,22 @@
...
     </widget>
    </item>
   </layout>
  </widget>
  <customwidgets>
   <customwidget>
-   <class>TyphosPositionerWidget</class>
-   <extends>QWidget</extends>
+   <class>TyphosPositionerRowWidget</class>
+   <extends>TyphosPositionerWidget</extends>
```
ZLLentz, line 92: These random other rearrangements happened automatically by designer, I decided to leave them in instead of fighting them

nrwslac:
LGTM, I didn't see any typos. I can't tell a lot about the background logic of this from the XML, but the explanation in the PR description sounds right to me.

nrwslac (approved):

PR #1189, opened 2023-11-16
Add j120k flowswitch, insert flow switch in appropriate classes.

nrwslac:
## Description

- added new module digital_signals.py
- added new class `J120K`
- added `J120K` to `SxrTestAbsorber`, `XPIM`, `IM2K0`, `PowerSlits`
- created new class `PPMCoolSwitch`, `WaveFrontSensorTargetCool`
## Motivation and Context

Where all devices in the happi database will get the cooling switch added `J120K` to the class. For devices who otherwise aren't all the same I inherited the base class. 

## How Has This Been Tested?

- Opened typhos screen from happi database where class name didn't change with local python environment and checked for flow_ok signal
- Opened typhos screens with manual module declaration where class name was different than happi database.
- All screens opened correctly.
- Opened a hutch python session with local python module and checked flow switch property reports:
-  `In [8]: sw.flow_switch.get_flow_ok`
-  `Out[8]: 1`

[https://github.com/pcdshub/lcls-plc-kfe-motion/pull/120](https://github.com/pcdshub/lcls-plc-kfe-motion/pull/120)
[https://docs.google.com/spreadsheets/d/1UAVYuG5my1ic51wRrpq9xJfvT5X7NX-nF4QzPtAUaPg/edit#gid=0](https://docs.google.com/spreadsheets/d/1UAVYuG5my1ic51wRrpq9xJfvT5X7NX-nF4QzPtAUaPg/edit#gid=0)

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1189-flow-switch.rst
+++ b/docs/source/upcoming_release_notes/1189-flow-switch.rst
@@ -17,7 +17,7 @@ New Devices
 -----------
 - `PPMCoolSwitch` ppms with cooling switch not a meter.
 - `WavefrontSensorTargetCool` WaveFrontSensors with a cooling switch.
-- `J120k` a device class for a cooling switch.
+- `J120K` a device class for a cooling switch.
 
 Bugfixes
 --------
--- a/docs/source/upcoming_release_notes/1189-flow-switch.rst
+++ b/docs/source/upcoming_release_notes/1189-flow-switch.rst
@@ -7,7 +7,7 @@ API Breaks
 
 Features
 --------
-- N/A
+- pcdsdevices now has a `digital_signals.py` module for simple digital io.
 
 Device Updates
 --------------
--- a/docs/source/upcoming_release_notes/1189-flow-switch.rst
+++ b/docs/source/upcoming_release_notes/1189-flow-switch.rst
@@ -1,0 +1,31 @@
+1189 flow-switch
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- added `J120K` to `SxrTestAbsorber`, `XPIM`, `IM2K0`, `PowerSlits`
+
+New Devices
+-----------
+- `PPMCoolSwitch`
+- `WavefrontSensorTargetCool`
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- N/A
--- a/docs/source/upcoming_release_notes/1189-flow-switch.rst
+++ b/docs/source/upcoming_release_notes/1189-flow-switch.rst
@@ -7,7 +7,7 @@ API Breaks
 
 Features
 --------
-- pcdsdevices now has a `digital_signals.py` module for simple digital io.
+- pcdsdevices now has a `digital_signals` module for simple digital io.
 
 Device Updates
 --------------
@@ -16,7 +16,7 @@ Device Updates
 New Devices
 -----------
 - `PPMCoolSwitch` ppms with cooling switch not a meter.
-- `WavefrontSensorTargetCool` WaveFrontSensors with a cooling switch.
+- `WaveFrontSensorTargetCool` WaveFrontSensors with a cooling switch.
 - `J120K` a device class for a cooling switch.
 
 Bugfixes
@@ -29,4 +29,4 @@ Maintenance
 
 Contributors
 ------------
-- N/A
+- nrwslac
--- a/docs/source/upcoming_release_notes/1189-flow-switch.rst
+++ b/docs/source/upcoming_release_notes/1189-flow-switch.rst
@@ -15,8 +15,9 @@ Device Updates
 
 New Devices
 -----------
-- `PPMCoolSwitch`
-- `WavefrontSensorTargetCool`
+- `PPMCoolSwitch` ppms with cooling switch not a meter.
+- `WavefrontSensorTargetCool` WaveFrontSensors with a cooling switch.
+- `J120k` a device class for a cooling switch.
 
 Bugfixes
 --------
--- a/pcdsdevices/digital_signals.py
+++ b/pcdsdevices/digital_signals.py
@@ -1,0 +1,20 @@
+from ophyd import Component as Cpt
+from ophyd import Device
+
+from .interface import BaseInterface
+from .signal import PytmcSignal
+
+
+class J120K(BaseInterface, Device):
+    """
+    A class representing the J120K 24V dry contact cooling flow switch.
+    """
+    flow_ok = Cpt(PytmcSignal, 'FSW:FLOW_OK', io='i',
+                  kind='normal', doc='flow rate nominal')
+
+    @property
+    def get_flow_ok(self):
+        """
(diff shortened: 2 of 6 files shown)


tangkong:
Is there a list of device that will have their `device_class` changed in happi?  (will there be?  I assumed so with the addition of new devices)

I went through and looked at devices whose class had the flow switch added, and the following didn't have active `*:FLOW_OK_RBV` pvs: 
* im1l0
* im2l0
* im4l0
* sl1l0

pcdsdevices/digital_signals.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    flow_ok = Cpt(PytmcSignal, ':FSW:FLOW_OK', io='i',
+                  kind='normal', doc='flow rate nominal')
+
+    @property
+    def get_flow_ok(self):
+        """
+        returns True if flow rate is nominal
+        """
+        return self.flow_ok.get()
```
tangkong, line 20: I might cast this to a boolean, especially since your docstring says it returns True.  If you're going to have a getter here you may as well standardize the output
```suggestion
        return bool(self.flow_ok.get())
```

nrwslac:
> Is there a list of device that will have their `device_class` changed in happi? (will there be? I assumed so with the addition of new devices)
> 
> I went through and looked at devices whose class had the flow switch added, and the following didn't have active `*:FLOW_OK_RBV` pvs:
> 
> * im1l0
> * im2l0
> * im4l0
> * sl1l0

These aren't implement yet, but will be in a month or two. I guess this will break the PV connection temporarily.

pcdsdevices/digital_signals.py
nrwslac: thats `True`

tangkong:
> These aren't implement yet, but will be in a month or two. I guess this will break the PV connection temporarily.

This didn't prevent the devices from loading, so it's not a huge deal.  It'll just fail if we try to `wait_for_connection` on the devices

docs/source/upcoming_release_notes/1189-flow-switch.rst
```diff
@@ -0,0 +1,32 @@
...
+
+Device Updates
+--------------
+- added `J120K` to `SxrTestAbsorber`, `XPIM`, `IM2K0`, `PowerSlits`
+
+New Devices
+-----------
+- `PPMCoolSwitch` ppms with cooling switch not a meter.
+- `WavefrontSensorTargetCool` WaveFrontSensors with a cooling switch.
+- `J120k` a device class for a cooling switch.
```
ZLLentz, line 20:
```suggestion
- `J120K` a device class for a cooling switch.
```

docs/source/upcoming_release_notes/1189-flow-switch.rst
```diff
@@ -0,0 +1,32 @@
...
+- pcdsdevices now has a `digital_signals.py` module for simple digital io.
+
+Device Updates
+--------------
+- added `J120K` to `SxrTestAbsorber`, `XPIM`, `IM2K0`, `PowerSlits`
+
+New Devices
+-----------
+- `PPMCoolSwitch` ppms with cooling switch not a meter.
+- `WavefrontSensorTargetCool` WaveFrontSensors with a cooling switch.
```
ZLLentz, line 19:
```suggestion
- `WaveFrontSensorTargetCool` WaveFrontSensors with a cooling switch.
```
spelling fix to make the sphinx auto link work

docs/source/upcoming_release_notes/1189-flow-switch.rst
```diff
@@ -0,0 +1,32 @@
+1189 flow-switch
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- pcdsdevices now has a `digital_signals.py` module for simple digital io.
```
ZLLentz, line 10: Not 100% sure but I think this also makes a sphinx auto link without the .py
```suggestion
- pcdsdevices now has a `digital_signals` module for simple digital io.
```

docs/source/upcoming_release_notes/1189-flow-switch.rst
```diff
@@ -0,0 +1,32 @@
...
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- N/A
```
ZLLentz, line 32:
```suggestion
- nrwslac
```

docs/source/upcoming_release_notes/1189-flow-switch.rst
nrwslac: sphixed

docs/source/upcoming_release_notes/1189-flow-switch.rst
nrwslac: sphinxed

docs/source/upcoming_release_notes/1189-flow-switch.rst
nrwslac: sphinxed

docs/source/upcoming_release_notes/1189-flow-switch.rst
nrwslac: authored.

pcdsdevices/digital_signals.py
nrwslac: boolified

nrwslac:
> Is there a list of device that will have their `device_class` changed in happi? (will there be? I assumed so with the addition of new devices)
> 
> I went through and looked at devices whose class had the flow switch added, and the following didn't have active `*:FLOW_OK_RBV` pvs:
> 
> * im1l0
> * im2l0
> * im4l0
> * sl1l0

Entries for pf1k0 and im1k3 will currently need to be updated. Corresponding to `WaveFrontSensorTargetCool` and `PPMCoolSwitch`

pcdsdevices/digital_signals.py
```diff
@@ -0,0 +1,20 @@
...
+
+class J120K(BaseInterface, Device):
+    """
+    A class representing the J120K 24V dry contact cooling flow switch.
+    """
+    flow_ok = Cpt(PytmcSignal, ':FSW:FLOW_OK', io='i',
+                  kind='normal', doc='flow rate nominal')
+
+    @property
+    def get_flow_ok(self):
```
ZLLentz, line 16: A few more thoughts, though I'm totally splitting hairs at this point:

- Why does this property exist? Is `flow_ok.get()` insufficient as an API? (I do see that it returns an int and not a bool)
- If there's a reason for it to exist, why is it a property and not a method?
- If there's a good reason for it to be a property, why is it named like a function? I think people in an interactive session will very likely do a `get_flow_ok()` call which will give an annoying "type `bool` is not callable" sort of error.

nrwslac: This is actually totally true, I did this myself. I'll delete it. I thought people would want a designated function, but in this case it should drop the property tag. I also did think about the fact that you can just call get on the signal.

ZLLentz: ok, sounds good- I would have been fine with any outcome here, just trying to make sure we consider these small details

ZLLentz (approved):
looks good to me

PR #1194, opened 2024-01-22
Add ACR status option for CCM energy moves

vespos:
## Description
Add the ACR status PV for cases where we are moving the undulator or the K, which need to be waited on.

## How Has This Been Tested?
In XPP3, with ACR changing the K value.

N/A

## Pre-merge checklist


pcdsdevices/ccm.py
```diff
@@ -773,6 +773,41 @@ def inverse(self, real_pos: namedtuple) -> namedtuple:
...
+    hutch : str, optional
+        The hutch we're in. This informs us as to which vernier
+        PVs to write to. If omitted, we can guess this from the
+        prefix.
+    acr_status_sufix : str
+        Prefix to the SIOC PV that ACR uses to report the move status.
+        For HXR this usually is 'AO805'.
+    """
+    vernier = FCpt(BeamEnergyRequest, '{hutch}',
+                   acr_status_suffix='AO805',
```
vespos, line 796: How do I use the acr_status_suffix here instead of hardcoding the 'AO805'?

tangkong: It looks like `FormattedComponent` only allows you to format the suffix.  `DynamicDeviceComponent` also doesn't quite fit the bill.  I don't see a Component type in ophyd or pcdsdevices that lets you format an arbitrary kwarg, but I feel like we must have done something like this before.
Maybe @ZLLentz has better memory than I do?

ZLLentz: There's a argument from `Component` called `add_prefix` that defaults to `("suffix", "write_pv")`, which is used to determine which string arguments we should prepend with the prefix. This is re-used in `FormattedComponent` to determine which string arguments should be formatted. So, you could do:
```
    vernier = FCpt(BeamEnergyRequest, '{hutch}',
                   acr_status_suffix='{acr_suffix}',
                   add_prefix=('suffix', 'acr_status_suffix')

        self.acr_suffix = 'AO805'
```

tangkong: Ah thanks, that's a neat backdoor.  (or at least something I've never seen in the Component class)

pcdsdevices/ccm.py
```diff
@@ -773,6 +773,41 @@ def inverse(self, real_pos: namedtuple) -> namedtuple:
...
+                   kind='normal',
+                   doc='Requests ACR to move the energy.')
+
+    def __init__(
+        self,
+        prefix: str,
+        hutch: typing.Optional[str] = None,
```
tangkong, line 803: As is, `hutch` is ignored and not passed to `.super().__init__`.  This will work since the parent class looks at the prefix, but nothing will happen if you provide `hutch`

pcdsdevices/ccm.py
```diff
@@ -903,11 +938,20 @@ class CCM(BaseInterface, GroupDevice, LightpathMixin, CCMConstantsMixin):
...
             'also requesting a vernier move.'
         ),
     )
+    energy_with_acr_status = Cpt(
+        CCMEnergyWithACRStatus, '', kind='normal',
+        doc=(
+            'PseudoPositioner that moves the alio in '
+            'terms of the calculated CCM energy while '
+            'also requesting an energy change to ACR. '
+            'This will wait on ACR to complete the move.'
```
tangkong, line 952: So as far as I can see, these two energy components are identical except one sends and waits on a request to ACR.  Is there ever a case where we wouldn't want to send a request to ACR?  If ACR is already at the setpoint, is there an issue with sending the request anyway?

I'm mostly asking since having two components might be confusing down the line.  The docs are nice and clear, I'm just wondering if we can remove opportunities for mistakes.

pcdsdevices/ccm.py
```diff
@@ -773,6 +773,41 @@ def inverse(self, real_pos: namedtuple) -> namedtuple:
         return self.PseudoPosition(energy=energy)
 
 
+class CCMEnergyWithACRStatus(CCMEnergyWithVernier):
+    """
+    CCM energy motor and ACR beam energy request with status.
+    Note that in this cae vernier indicates any ways that ACR will act on the
```
tangkong, line 779:
```suggestion
    Note that in this case vernier indicates any ways that ACR will act on the
```

I think this is what you meant?  unless "cae" is yet another acronym I need to learn

pcdsdevices/ccm.py
vespos: Yes there is. When we do fast energy scans over small ranges, we do a pseudo fly scan, where we don't need / want to wait on ACR. When doing larger ranges, when the K needs to move, things are much slower and we need to wait.

vespos:
@ZLLentz @tangkong  I have pushed some more changes. We are going to test things this week with ACR so there will probably be more to come.
 
I changed some naming, as to better reflect the reality of the use cases. While i feel this is making things less confusing in the long run, I don't know if that is not breaking things elsewhere. Can you advise? The changes are:
- vernier -> acr_energy: the energy may or may not be the Vernier. This is particularly true for the case when we wait for the acr status
- bunch -> pv_index: we are using the different PVs not only for multi-color / multi bunch, but also like here for combined Vernier and K motion

ZLLentz (commented):
This is tested working live, right? I think it looks good. The vernier name change is still an api break but it's definitely more correct and probably ok.

This is just missing pre-release docs I think.

vespos:
Addressed the comments. I am also working on a confluence page to go more in depth about the energy scanning schemes that we are using. 
Next step for after the actual tests with ACR.

vespos:
So we did some self-seeding test recently, and I have added a couple commits related to that.
Pending the test fix and the release notes, I think we are close to being able to merge.

Two things that I would like your input on:
* Was there any reason to have the Vernier tolerance set to 5eV? I assume this is only driven by the SASE beandwidth (~10eV), and it needs to be reduced to 0.5eV for the self-seed (bandwidth ~1eV). But is there any other consideration to take into account?
* Are we ok with having the reargs function live in the ccm.py? Do we want to have that as a most consistent way to re-arg functions should we need that?

ZLLentz:
> Was there any reason to have the Vernier tolerance set to 5eV?

I had remembered this being inherited from an older version of the code, but when I went to check I found that the original source used 3eV. My limited understanding is that if the move is small enough we don't want to wait for the energy to change, which can be a timely process.

> Are we ok with having the reargs function live in the ccm.py?

It probably belongs in a utils module instead, but it could always be moved in a later refactor if needed.

vespos:
Ok, so moving to 0.5eV is fine. The Vernier change takes 0.2-0.3s, so it's comparable if not faster than the CCM.
It's been tested with beam, and in the self-seed case, the 5eV tolerance definitely breaks things. In SASE, the Vernier move is not a bottleneck.

I'll move the reargs function to the utils.py here, it makes indeed more sense to have it there.

PR #1195, opened 2024-01-25
Custom EPS screens 

jozamudi:
## Description

Added BeckhofAxisEPSCustom class  in epics_motor.py and BeckhofAxisEpicsCusom.ui  to allow the use of a custom screen to show EPS information.

## Motivation and Context

We want to be able to protect motion stages from running into each other. 
https://github.com/pcdshub/lcls-plc-crixs-motion/pull/71

we want to be able to display this EPS information on they typhos screen to make accessible for the users.

https://jira.slac.stanford.edu/browse/ECS-3966

## How Has This Been Tested?

Used typhos 'pcdsdevices.epics_motor.BeckhoffAxisEPSCustom' to open  typhos screen

EPS Screens can be placed here.
reg/g/pcds/epicsdev/screens/pydm/eps_screens/${beamline}/${name}

image

## Pre-merge checklist


ZLLentz:
I think this is a good idea, but before we merge I think we need to:

- finish the pre-merge checklist items
- add a screenshot of a screen in-use rather than with a "path not found" error for documentation purposes

More thoughts/questions:
- Is the standard path included here where we're going to start stashing EPS screens?
- Is there some reason why we cannot or should not use https://github.com/pcdshub/pcdswidgets/blob/master/pcdswidgets/eps_byteindicator.py in a more direct way here? Motors should theoretically have consistent PVs for `enable`/`forward enable`/`backward enable`

jozamudi:
@ZLLentz  The standard path was documented  in the BeckhofAxisEPSCustom class declaration. I added the path in the PR. Is there a better path we can use?

I don't think there is a major reason why we could not use the eps_widget more directly. The only reason would be that some information might not be relevant to a motor that does not have those eps conditions implemented in the PLC.

pcdsdevices/epics_motor.py
```diff
@@ -1299,6 +1299,18 @@ class BeckhoffAxisEPS(BeckhoffAxis):
               doc='PLC error handling, aux functions, and EPS.')
 
 
+class BeckhoffAxisEPSCustom(BeckhoffAxis):
+    """
+    A beckhoff axis where custom EPS Logic has been implemented in
+    a PLC and needs a custom screen to display EPS information
+
+    EPS screens are found in
+    reg/g/pcds/epics-dev/screens/pydm/eps_screens/${beamline}/${name}
```
ZLLentz, line 1308: I think this is a reasonable place for these screens, but a few things to note:

- This forces us to always use epics-dev in this version of the code. In the future I intend for people to be able to tag and release their screens, so this will need to be revised in some way in a later revision.
- We should make sure there's a reasonable readme in this directory that explains what it is for.
- We should stop proliferating the deprecated `/reg` paths, preferring the newer `/cds` paths instead (and, we should always have the preceding `/` for correctness.

```suggestion
    /cds/group/pcds/epics-dev/screens/pydm/eps_screens/${beamline}/${name}
```

pcdsdevices/ui/BeckhoffAxisEPSCustom.ui
```diff
@@ -0,0 +1,363 @@
...
+           </rect>
+          </property>
+          <layout class="QVBoxLayout" name="verticalLayout_4">
+           <item>
+            <widget class="PyDMEmbeddedDisplay" name="PyDMEmbeddedDisplay">
+             <property name="toolTip">
+              <string/>
+             </property>
+             <property name="filename" stdset="0">
+              <string>/reg/g/pcds/epics-dev/screens/pydm/eps_screens/${beamline}/${name}.ui</string>
```
ZLLentz, line 325:
```suggestion
              <string>/cds/group/pcds/epics-dev/screens/pydm/eps_screens/${beamline}/${name}.ui</string>
```

docs/source/upcoming_release_notes/1195-Added_BeckhofAxisEPSCustom_class_in_epics_motor.py_and_BeckhofAxisEpicsCusom.ui_to_allow_the_use_of_a_custom_screen_to_show_EPS_information..rst
```diff
@@ -0,0 +1,32 @@
...
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- Josue Zamudio Estrada
```
ZLLentz, line 32: You should use your github name for proper release notes credit
```suggestion
- jozamudi
```

ZLLentz (commented):
Sorry for my late reply- just small things to adjust here

ZLLentz:
This fell off my radar for two reasons: the github notification interface dropped the mention of "review requested", and I got distracted with Monday things yesterday. I'm not sure why github demoted my involvement here from "review requested" to "mentioned", but it made my brain skip over this in the notifications list in favor of the other tickets... my bad.

ZLLentz (approved):

PR #1196, opened 2024-01-30
TST/BUG: tests and bugfixes for at2l0

ZLLentz:
## Description

Fix regressions in at2l0 in the mds signals

- [x] Add a unit test for each signal
- [x] Fix the bug or the test as appropriate

I also ran pre-commit autoupdate and fixed the "new" flake8 things. I did this because the existing version was giving nonsense errors on my machine.

## Motivation and Context

See https://jira.slac.stanford.edu/browse/ECS-4470

## How Has This Been Tested?

This is tests

Here

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/tests/test_attenuator.py
+++ b/pcdsdevices/tests/test_attenuator.py
@@ -216,3 +216,46 @@ def test_new_attenuator_smoke(fake_new_attenuator):
             fake_new_attenuator.status_info()
         )
     )
+
+
+@pytest.fixture(scope="function")
+def at2l0(request):
+    at2l0 = make_fake_device(AT2L0)("AT2L0:", name=f"fake_at2l0_for_{request.function}")
+    for walk in at2l0.walk_signals():
+        try:
+            if "message" in walk.dotted_name:
+                walk.item.sim_put("")
+            else:
+                walk.item.sim_put(0)
+        except AttributeError:
+            ...
+        walk.item._metadata["severity"] = 0
+    return at2l0
+
+
+def test_at2l0_error_summary(at2l0):
+    assert at2l0.error_summary.get() == "No Errors"
+    at2l0.blade_05.state.error_message.sim_put("test error")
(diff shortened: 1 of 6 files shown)


ZLLentz:
The IOC is currently down. I'm confident the fix works but after what happened list time I'll opt to wait until I can run it on the real thing.

pcdsdevices/tests/test_attenuator.py
```diff
@@ -216,3 +216,50 @@ def test_new_attenuator_smoke(fake_new_attenuator):
...
+    at2l0.blade_19.motor.plc.err_code.sim_put(1)
+    # bitmask intentionally reversed (better for ui)
+    assert at2l0.error_summary_bitmask.get() == 0b1
+    at2l0.blade_17.motor.plc.err_code.sim_put(1)
+    assert at2l0.error_summary_bitmask.get() == 0b101
+
+
+def test_at2l0_clear_errors(at2l0):
+    samples = (
+        at2l0.blade_01.motor.plc.cmd_err_reset,
```
tangkong, line 256: There's one of these for every blade right?  Is there a reason we don't just... test them all?

ZLLentz: mostly because I didn't want to write a reflexive test (checking all the pvs in at2l0's list of pvs to check, for example), and I didn't feel like copy/pasting 20 times

maybe I should suck it up and check them all

tangkong: I was just thinking something like 
```python
for i in range(20):
    samples.append(getattr(at2l0, f'blade_{str(i).zfill(2)}.motor.plc.cmd_err_reset'))
```

Or is there something that I'm missing there

ZLLentz: yeah I can do this

tangkong (approved):
LGTM

ZLLentz:
I've confirmed this function runs without errors in the screen and in the ipython terminal

PR #1199, opened 2024-02-16
Add flow sensor to devices

nrwslac:
## Description

- in `atm.py` add `flow_meter` to `ArrivalTimeMonitor`, add new class for `MFXATM`
- in `attenuator.py` add `flow_meter` to `AttenuatorSXR_Ladder`
- in `attenuator.py` add `flow_meter` to `AttenuatorSXR_LadderTwoBladeLBD`
- in `device_types.py` add `WaveFrontSensorTargetCool` , `WaveFrontSensorTargetPF2K2`
- in `mirror.py` add flow sensor variable to `FFMirror`
- create a new sub module `pc.py` and class `PhotonCollimator` to readout `flow_switch`
- in `wfs.py` add new class: `WaveFrontSensorTargetFDQ` with `flow_meter` 

## Motivation and Context

- newly installed flow sensors.

## How Has This Been Tested?

- opened all typhos screens, verified working.

https://jira.slac.stanford.edu/browse/ECS-1248
https://jira.slac.stanford.edu/browse/ECS-1186
https://jira.slac.stanford.edu/browse/ECS-1514
https://jira.slac.stanford.edu/browse/ECS-1482

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1199-cool-updates.rst
+++ b/docs/source/upcoming_release_notes/1199-cool-updates.rst
@@ -1,0 +1,36 @@
+1199 cool-updates
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- create a new sub module `pc.py`
+
+Device Updates
+--------------
+- in `atm.py` add `flow_meter` to `ArrivalTimeMonitor`
+- in `attenuator.py` add `flow_meter` to `AttenuatorSXR_Ladder`
+- in `attenuator.py` add `flow_meter` to `AttenuatorSXR_LadderTwoBladeLBD`
+- in `device_types.py` add `WaveFrontSensorTargetCool` , `WaveFrontSensorTargetPF2K2`
+- in `mirror.py` add flow sensor variable to `FFMirror`
+- in `wfs.py` add `flow_meter` to `WaveFrontSensorTarget`
+
+New Devices
+-----------
+- add class `PhotonCollimator` to readout `flow_switch`
+- add class `WaveFrontSensorTargetPF2K2` to readout `flow_switch`
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- nrwslac
--- a/docs/source/upcoming_release_notes/1199-cool-updates.rst
+++ b/docs/source/upcoming_release_notes/1199-cool-updates.rst
@@ -7,21 +7,21 @@ API Breaks
 
 Features
 --------
-- create a new sub module `pc.py`
+-
 
 Device Updates
 --------------
 - in `atm.py` add `flow_meter` to `ArrivalTimeMonitor`
 - in `attenuator.py` add `flow_meter` to `AttenuatorSXR_Ladder`
 - in `attenuator.py` add `flow_meter` to `AttenuatorSXR_LadderTwoBladeLBD`
-- in `device_types.py` add `WaveFrontSensorTargetCool` , `WaveFrontSensorTargetPF2K2`
+- in `device_types.py` add `WaveFrontSensorTargetCool` , `WaveFrontSensorTargetFDQ`
 - in `mirror.py` add flow sensor variable to `FFMirror`
 - in `wfs.py` add `flow_meter` to `WaveFrontSensorTarget`
 
 New Devices
 -----------
-- add class `PhotonCollimator` to readout `flow_switch`
-- add class `WaveFrontSensorTargetPF2K2` to readout `flow_switch`
+- add class `PhotonCollimator` to readout `flow_switch` in new module `pc.py`
+- add class `WaveFrontSensorTargetFDQ` to read out `flow_meter`
 - add class `MFXATM` in `atm.py`
 
 Bugfixes
--- a/docs/source/upcoming_release_notes/1199-cool-updates.rst
+++ b/docs/source/upcoming_release_notes/1199-cool-updates.rst
@@ -7,7 +7,7 @@ API Breaks
 
 Features
 --------
--
+- N/A
 
 Device Updates
 --------------
@@ -16,7 +16,6 @@ Device Updates
 - in `attenuator.py` add `flow_meter` to `AttenuatorSXR_LadderTwoBladeLBD`
 - in `device_types.py` add `WaveFrontSensorTargetCool` , `WaveFrontSensorTargetFDQ`
 - in `mirror.py` add flow sensor variable to `FFMirror`
-- in `wfs.py` add `flow_meter` to `WaveFrontSensorTarget`
 
 New Devices
 -----------
--- a/docs/source/upcoming_release_notes/1199-cool-updates.rst
+++ b/docs/source/upcoming_release_notes/1199-cool-updates.rst
@@ -22,6 +22,7 @@ New Devices
 -----------
 - add class `PhotonCollimator` to readout `flow_switch`
 - add class `WaveFrontSensorTargetPF2K2` to readout `flow_switch`
+- add class `MFXATM` in `atm.py`
 
 Bugfixes
 --------
--- a/pcdsdevices/mirror.py
+++ b/pcdsdevices/mirror.py
@@ -1132,9 +1132,9 @@ class FFMirror(BaseInterface, GroupDevice, LightpathMixin):
     y_enc_rms = Cpt(PytmcSignal, ':ENC:Y:RMS', io='i', kind='normal')
     pitch_enc_rms = Cpt(PytmcSignal, ':ENC:PITCH:RMS', io='i', kind='normal')
 
-    cool_flow1 = Cpt(EpicsSignalRO, ':FWM:1_RBV', kind='normal')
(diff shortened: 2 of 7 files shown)


pcdsdevices/pc.py
tangkong, line 1: We should be better about this across pcdsdevices, but do you mind adding a little docstring blurb to the top of this file describing what devices go here?  Just something like:
```
"""
Device classes for Photon collimator and related devices.  These devices will ...
"""
```

This is a little more important for short module names like `pc.py`

pcdsdevices/mirror.py
```diff
@@ -1132,6 +1132,10 @@ class FFMirror(BaseInterface, GroupDevice, LightpathMixin):
     y_enc_rms = Cpt(PytmcSignal, ':ENC:Y:RMS', io='i', kind='normal')
     pitch_enc_rms = Cpt(PytmcSignal, ':ENC:PITCH:RMS', io='i', kind='normal')
 
+    cool_flow1 = Cpt(EpicsSignalRO, ':FWM:1_RBV', kind='normal')
```
tangkong, line 1135: Is there any interesting information we could put in the doc for these Components?  Something like where these readouts are?

tangkong:
I'm going to run some checks on the devices affected here, since the changes seem pretty wide-reaching.  Otherwise I'm largely happy with this

tangkong:
A brief audit of the affected devices, searching happi for device_class matching the following classes:
* `ArrivalTimeMonitor`
    * devices: `mfx_atm` (98/103 signals connected, flow_meter succeeds), `tm1k2` (all connected), `tm2k2` (all connected)
* `AttenuatorSXR_Ladder`
    * devices: `at2k2` (all connected)
* `AttenuatorSXR_LadderTwoBladeLBD`
    * devices: N/A 
* `FFMirror`
    * devices: `mr2k2_flat` (all connected), `mr4k4_kbo` (all connected), `mr5k4_kbo` (all connected)
* `WaveFrontSensorTarget`
    * devices: `pf1k0` (103/106 flow_meter FQD not connected),  `pf1k2` (all connected), `pf1k4` (103/106 flow_meter FQD not connected), `pf1l0` (103/106 flow_meter FQD not connected), `pf2k2` (all connected), `pf2k4` (103/106 flow_meter FQD not connected)

pcdsdevices/pc.py
nrwslac: yea, I can add some details here.

pcdsdevices/mirror.py
nrwslac: yea, I can add some details here.

nrwslac:
@tangkong 
- mfx_atm will need a new class that sets `flow_meter` to `None`, ill add this to this PR.
- pf1k0 will be assigned to class ` WaveFrontSensorTargetCool` in happi.
- pf1k4 needs the PV declared in the PLC, then it will be connected.
- pf1l0 needs to be assigned to class ` WaveFrontSensorTargetCool`  in happi and still needs hardware wired.
- pf2k4 needs the PV declared in the PLC, then it will be connected.

A couple of these will show disconnected in the screen temporarily.

tangkong:
Assigning new classes to devices in happi is a bit of a tricky task.  Many hutches still use old versions of pcdsdevices, and we ideally want to keep backwards compatibility with those.  (as is, if we change the happi device class name and old hutch-python doesn't have that class in their pcdsdevices, we get an error)

This is a bit more of a tricky task, since it requires us to determine if devices are actively used in those hutches.  

best case scenario the outdated hutches never touch these devices.  I wonder what the best way to look for these conflicts is

ZLLentz:
I'm planning to tag pcdsdevices and pcds_conda this week. Would you like me to wait for this PR to be completed?

nrwslac:
> I'm planning to tag pcdsdevices and pcds_conda this week. Would you like me to wait for this PR to be completed?

yes please.

ZLLentz:
>  if they don't update then nothing changes and nothing breaks, but they don't get any new features?

If they don't update but happi does, and the happi update requires changing the classes, then the old code breaks because it cannot access the new classes. I also don't have a good short-term solution to this with the currently implemented happi/pcdsdevices scheme. Long-term we could consider swapping more devices to self-named classes. There's some consideration for e.g. adding fallback classes or versioning to happi as well.

nrwslac:
> > if they don't update then nothing changes and nothing breaks, but they don't get any new features?
> 
> If they don't update but happi does, and the happi update requires changing the classes, then the old code breaks because it cannot access the new classes. I also don't have a good short-term solution to this with the currently implemented happi/pcdsdevices scheme. Long-term we could consider swapping more devices to self-named classes. There's some consideration for e.g. adding fallback classes or versioning to happi as well.

I see. In this case I should suspect that any device updated to a new class might break. If it does break, I roll it back to the previous class.

docs/source/upcoming_release_notes/1199-cool-updates.rst
```diff
@@ -0,0 +1,37 @@
+1199 cool-updates
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- create a new sub module `pc.py`
```
ZLLentz, line 10: nitpick: I think this should be mentioned when talking about the PhotonCollimator device, since it isn't a new feature and isn't understandable without that context.

docs/source/upcoming_release_notes/1199-cool-updates.rst
```diff
@@ -0,0 +1,37 @@
...
+- in `attenuator.py` add `flow_meter` to `AttenuatorSXR_Ladder`
+- in `attenuator.py` add `flow_meter` to `AttenuatorSXR_LadderTwoBladeLBD`
+- in `device_types.py` add `WaveFrontSensorTargetCool` , `WaveFrontSensorTargetPF2K2`
+- in `mirror.py` add flow sensor variable to `FFMirror`
+- in `wfs.py` add `flow_meter` to `WaveFrontSensorTarget`
+
+New Devices
+-----------
+- add class `PhotonCollimator` to readout `flow_switch`
+- add class `WaveFrontSensorTargetPF2K2` to readout `flow_switch`
```
ZLLentz, line 24: I don't understand the note here. It seems like this class exists in order to not read out any flow information? Are the notes wrong, the code wrong, or my understanding wrong?

pcdsdevices/pc.py
ZLLentz, line 1: I had to double-check that there weren't any name conflicts here for pc- it seems ok. I was worried there would be a pc module for some computing thing.

ZLLentz (commented):
Looks normal to me

The happi/pcdsdevices thing is sort of just something we have to live with until we have a good normal solution, for handling it, I guess? This isn't the first PR that's had this issue.

I had a few nitpicks/confusion on the release notess

pcdsdevices/pc.py
nrwslac: Thats a good point. Should I make it more verbose? PhotonCol.py?

ZLLentz: It needs to stay pep8 compliant, so e.g. photon_col.py, but I think pc.py is every bit as acceptable as atm.py.
Is there some sort of larger class of components we could name the file after? I imagine that pc.py will not end up with very many devices in it.
Otherwise it should be ok, googling "python pc module" doesn't give any interesting hits.

docs/source/upcoming_release_notes/1199-cool-updates.rst
nrwslac: That's right, I could make a new class from the base class instead of what I've done here. I guess the idea is that this device will eventually use the base class and this class can be deleted. If I do it the former way, we end up with all classes inheriting a base class and no one using the base class. On the other hand, we might benefit one day from having the base class available. Probably what I am doing is more confusing than necessary. I can make a derived class and make this more paradigmatic.

ZLLentz: I don't understand your response here: my question isn't about the implementation, it's about the docs which don't match the implementation

nrwslac: I see the confusion. Let me update the doc.

docs/source/upcoming_release_notes/1199-cool-updates.rst
nrwslac: moved doc to new devices section.

docs/source/upcoming_release_notes/1199-cool-updates.rst
nrwslac: Decided to make a new class with flow meter, so this PF2K2 class is gone and this doc is gone as well.

tangkong (approved):
I think this looks good, thanks for dealing with our documentation nitpicks.  We'll keep the potential pcdsdevices/happi issues in mind moving forward.

pcdsdevices/pc.py
nrwslac: I think we can drop this in lens.py if we expand the scope of that module from be lenses to lenses in general and accept the claim that a photon collimator is a type of lens system.

nrwslac: I'll leave this for future migration consideration.

PR #1216, opened 2024-04-12
BUG: Tpr ns_delay throwing error

KaushikMalapati:
## Description

Changed _get_delay, _get_width, and _put_last in tpr.py to signals elements instead of attrs elements as keys for items, since it maps from signal objects to values, not the names of signals

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-5093

## How Has This Been Tested?

Interactively, I made a tpr object and manually called the calculate_on_get/put functions for delay and width to make sure they gave the right values and didn't throw errors

Pre-release notes made

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/tests/test_tpr.py
+++ b/pcdsdevices/tests/test_tpr.py
@@ -26,7 +26,7 @@ def test_disconnected_trigger():
 
 def put_equals_setpoint(mds, setpoint, ns_time):
     mds.put(ns_time)
-    return setpoint.value == ns_time
+    return setpoint.get() == ns_time
 
 
 def mds_get(mds, tick_signal, tap_signal, num_ticks, num_taps):
--- a/pcdsdevices/tests/test_tpr.py
+++ b/pcdsdevices/tests/test_tpr.py
@@ -1,7 +1,7 @@
 import pytest
 from ophyd.sim import make_fake_device
 
-from ..tpr import TimingMode, TprTrigger
+from ..tpr import TPR_TAP_NS, TPR_TICK_NS, TimingMode, TprTrigger
 
 
 @pytest.fixture(scope='function')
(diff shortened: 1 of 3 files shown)


KaushikMalapati:
@ZLLentz, I'm using this helper function in two unit tests 
`def put_equals_setpoint(mds, setpoint, ns_time):
    mds.put(ns_time)
    return setpoint.get() == ns_time`
to use MultiDerivedSignal.put() and confirm that it updates the setpoint signal, which is what its calculate_on_put is supposed to do. This works interactively, and the two unit tests that use put_equals_setpoint pass locally, but I think they fail in ci/cd. I think that there is a race condition for setpoint's value being updated since a commit where I was printing out the value before getting it passed. I've tried cannibalizing some code from test_multi_derived_rw_sub in test_signal but I never enter value_callback and end up spinning in wait_until_value. Do you have any suggestions on what I could try or do you know if there's a way to force ophyd to update a signal so you can't see stale values?

pcdsdevices/tests/test_tpr.py
```diff
@@ -22,3 +22,29 @@ def test_enable(fake_trigger):
...
+    return setpoint.get() == ns_time
+
+
+def mds_get(mds, tick_signal, tap_signal, num_ticks, num_taps):
+    tick_signal.put(num_ticks)
+    tap_signal.put(num_taps)
+    return mds.get() == (TPR_TICK_NS * num_ticks + TPR_TAP_NS * num_taps)
+
+
+def test_ns_delay(fake_trigger):
```
tangkong, line 38:
```suggestion
@pytest.mark.timeout(5)
def test_ns_delay(fake_trigger):
```
Adding a timeout here makes this run consistently locally.  We do this a lot in the pcdsdevices test suite, and you'll probably need it for the `test_width` test as well

KaushikMalapati: With the timeout added, it still fails locally sometimes and failed the the github job. This is the error message
`E       AssertionError: assert False
E        +  where False = put_equals_setpoint(MultiDerivedSignal(name='trig_a_ns_delay', parent='trig_a', value=None, timestamp=1713216587.5022318), FakeEpicsSignal(name='trig_a_delay_setpoint', parent='trig_a', value=0, timestamp=1713216587.5021243), 100)
E        +    where MultiDerivedSignal(name='trig_a_ns_delay', parent='trig_a', value=None, timestamp=1713216587.5022318) = FakeTprTrigger(TST:TPR, name=trig_a).ns_delay
E        +    and   FakeEpicsSignal(name='trig_a_delay_setpoint', parent='trig_a', value=0, timestamp=1713216587.5021243) = FakeTprTrigger(TST:TPR, name=trig_a).delay_setpoint

test_tpr.py:40: AssertionError
`
I don't know why a timeout would help since I'm not hanging on anything. Sleeping for two seconds after the put in put_equals_setpoint (between lines 28 and 29) makes it pass consistently locally but I imagine there there's a better way to ensure that you get the setpoint pv after it's been written to by the MultiDerivedSignal. I tried using code from test_multi_derived_rw_sub and wait_until_value in test_signal, but I never entered the callback after subscribing for reasons I don;'t understand. That test and mine are both using MultiDerivedSignals.

tangkong: Sorry, I assumed here that we just needed to wait for the callbacks to resolve.  I guess the problem here is that the signals already have values, and we need them to update.  It should be easy enough to just poll until the condition is true, given a timeout:

```python
def wait_until(condition, interval=0.1, timeout=1, *args):
  start = time.time()
  while not condition(*args) and time.time() - start < timeout:
    time.sleep(interval)
```

where `condition` is a function wrapping your assert statement or something similar.

tangkong: This time I can get it very consistently passing locally for me, previously on heavy repeats (100) I was getting failures

KaushikMalapati: I ended up using Status.wait for the put return along with a timeout on the tests. It works and feels cleaner than spinning or a callback in this case

ZLLentz (approved):
I'm happy with the fix and the tests! Hopefully this will catch any future regressions (e.g. if we accidentally break this by mucking with MultiDerivedSignal)
I'm going to click the update with master button, the tests are going to pass again, and then we'll merge and proceed to tagging

PR #1219, opened 2024-04-29
BUG: Prevent lightpath logic from over-scheduling threads

tangkong:
## Description
Summary: In the case where the mirror is not initialized, prevent pcdsdevices from scheduling a task every time lightpath_summary is updated.  

## Motivation and Context
https://jira.slac.stanford.edu/browse/ECS-5186

In some more detail: 
We encountered a case where `XOffsetMirrorXYState` would never initialize, so every call to `._get_insertion_state()` would try to schedule a retry thread.  This helper method gets called whenever `lightpath_summary` updates, so for a quickly changing set of `lightpath_cpts`, this would create threads until we hit the limit.  This isn't a problem if the IOC is properly responding, but we can make failure here less catastrophic

`LightpathMixin._retry_lightpath` seems to be a vestigial attribute I reclaimed for this PR.  It was used in the legacy lightpath behavior, for a similar purpose.

## How Has This Been Tested?
Added a test.  Live testing would require us to modify the device to never initialize its state and log thread creation events.  (assuming that disabling the IOC is out of the question)

This PR, and the jira ticket

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1219-bug_lp_thread_overschedule.rst
+++ b/docs/source/upcoming_release_notes/1219-bug_lp_thread_overschedule.rst
@@ -22,7 +22,7 @@ Bugfixes
 - Prevent some devices from creating threads at high frequency when
   trying to get the lightpath state.  These devices classes include
   `XOffsetMirrorXYState`, `AttenuatorSXR_Ladder`,
-  `AttenuatorSXR_LadderTwoBladeLBD``, `AT2L0`, `XCSLODCM`, and `XPPLODCM`
+  `AttenuatorSXR_LadderTwoBladeLBD`, `AT2L0`, `XCSLODCM`, and `XPPLODCM`
 
 Maintenance
 -----------
--- a/docs/source/upcoming_release_notes/1219-bug_lp_thread_overschedule.rst
+++ b/docs/source/upcoming_release_notes/1219-bug_lp_thread_overschedule.rst
@@ -1,0 +1,31 @@
+1219 bug_lp_thread_overschedule
+###############################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- Prevent ``XOffsetMirrorXYState`` from creating threads at high frequency when
+  trying to get the lightpath state.
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- tangkong
--- a/docs/source/upcoming_release_notes/1219-bug_lp_thread_overschedule.rst
+++ b/docs/source/upcoming_release_notes/1219-bug_lp_thread_overschedule.rst
@@ -19,8 +19,10 @@ New Devices
 
 Bugfixes
 --------
-- Prevent ``XOffsetMirrorXYState`` from creating threads at high frequency when
-  trying to get the lightpath state.
+- Prevent some devices from creating threads at high frequency when
+  trying to get the lightpath state.  These devices classes include
+  `XOffsetMirrorXYState`, `AttenuatorSXR_Ladder`,
+  `AttenuatorSXR_LadderTwoBladeLBD``, `AT2L0`, `XCSLODCM`, and `XPPLODCM`
 
 Maintenance
 -----------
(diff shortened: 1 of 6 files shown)


tangkong:
In this case we do check whenever EPICS updates; this call chain originates from a callback.  The problem was that each call failed and scheduled a retry, which never stopped retrying.  As I type this I realize now that this failure mode (essentially a memory leak) is not fully addressed by this PR, just delayed

ZLLentz:
Ohh, so each time any lightpath-relevant PV updates, as long as this one is still not ready, we add a new chain of queue entries...

ZLLentz:
That's why the queue was overflowing and breaking the session

ZLLentz:
Offline we identified similar issues in the attenuator, lodcm, and basic lightpath mix-in classes

tangkong:
Ok I think this is a better way to go about this.  Now we have one retry allotted for the entire instance, which only resets if the lightpath state is fetched successfully (if we don't hit the `if not self.state_initialized` block).

if this looks good I can apply similar logic to the other affected device classes

ZLLentz:
I think this is the way to go if it's working for a prod mirror. This prevents the proliferation of queue additions from getting out of hand in this case we've identified with a partial IOC.

tangkong:
The hotfix in prod doesn't do any of these checks, rather just bypasses the block entirely
```python
# < snip pcdsdevices.mirror.XOffsetMirrorXYState._get_insertion_state > 
        if not self.insertion._state_initialized:
            # 27/04/2024 roberttk turning off the task scheduling in an attempt
            # to deal with unresponsive PVs
            self.log.debug('insertion state not initialized, '
                           'scheduling lightpath calcs for later')

            # schedule_task(self._calc_cache_lightpath_state, delay=2.0)
            logger.warning(f'mirror {self} could not find lightpath state, '
                           'providing dummy state')
            # raise MirrorLogicError('insertion state not initialized')
            return True, True
```

ZLLentz:
For my review I'm going to test that this doesn't disrupt prod behavior for one example of each (working) device.
I'm not going to go as far as to try to reproduce the odd condition that caused this issue.

ZLLentz:
My notes are:

at2l0 ok
at2k2 ok
no elements of AttenuatorSXR_LadderTwoBladeLBD exist in happi
im3l0 ok
xpp_lodcm ok
xcs_lodcm ok

lightpath gui to XPP, MFX, TMO ok

docs/source/upcoming_release_notes/1219-bug_lp_thread_overschedule.rst
```diff
@@ -0,0 +1,31 @@
...
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- Prevent ``XOffsetMirrorXYState`` from creating threads at high frequency when
+  trying to get the lightpath state.
```
ZLLentz, line 23: Maybe this needs a slight extension now

ZLLentz (approved):
I had a docs quibble but the fix is simple and effective, it brings us out of the looping/memory leak hell without completely abandoning the need for the extra check.

docs/source/upcoming_release_notes/1219-bug_lp_thread_overschedule.rst
```diff
@@ -0,0 +1,33 @@
...
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- Prevent some devices from creating threads at high frequency when
+  trying to get the lightpath state.  These devices classes include
+  `XOffsetMirrorXYState`, `AttenuatorSXR_Ladder`,
+  `AttenuatorSXR_LadderTwoBladeLBD``, `AT2L0`, `XCSLODCM`, and `XPPLODCM`
```
ZLLentz, line 25:
```suggestion
  `AttenuatorSXR_LadderTwoBladeLBD`, `AT2L0`, `XCSLODCM`, and `XPPLODCM`
```

ZLLentz (approved):

docs/source/upcoming_release_notes/1219-bug_lp_thread_overschedule.rst
tangkong: d'oh thank you

PR #1221, opened 2024-05-14
TXI Mirror Cooling, SP1K1 Mono RTDs

nrwslac:
## Description

- `XOffsetMirrorNoBend` in mirror.py gets 3 new cooling readout components.
- `Mono` in spectrometer.py gets 4 new RTD components and re-named RTDs 1-8. Make cooling component names consistent with mirror cooling component names.

## Motivation and Context

- Wired up cooling panel sensors.
- Found new information about the RTD positions on SP1K1 MONO.

## How Has This Been Tested?

- Ran typhos screens locally.

https://jira.slac.stanford.edu/browse/ECS-4588

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/mirror.py
+++ b/pcdsdevices/mirror.py
@@ -752,7 +752,7 @@ class XOffsetMirrorNoBend(XOffsetMirror):
 
     name : str
 
-    Currently (10/11/2023) services: mr1l1, mr1k3, mr2k3
+    Currently (5/15/2024) services: mr1l1, mr1k3, mr2k3
 
     """
     bender = None
--- a/pcdsdevices/mirror.py
+++ b/pcdsdevices/mirror.py
@@ -752,11 +752,15 @@ class XOffsetMirrorNoBend(XOffsetMirror):
 
     name : str
 
-    Currently (10/11/2023) services: mr1l1
+    Currently (10/11/2023) services: mr1l1, mr1k3, mr2k3
 
     """
     bender = None
     bender_enc_rms = None
+
+    cool_flow1 = Cpt(EpicsSignalRO, ':FWM:1_RBV', kind='normal', doc='Mirror cooling panel loop flow sensor')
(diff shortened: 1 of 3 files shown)


tangkong (commented):
My guess is that nobody relies on these, but I'm always wary of changing existing names in case someone's script assumes it will be there.  Aside from that I'm in favor of the more descriptive names (rtd might be an understandable acronym to everyone)

nrwslac:
> My guess is that nobody relies on these, but I'm always wary of changing existing names in case someone's script assumes it will be there. Aside from that I'm in favor of the more descriptive names (rtd might be an understandable acronym to everyone)

I could be convinced to drop rtd for temp.

tangkong:
Now that you mention it, I'd be much happier with something like `grating_temp_1`.  At the point where someone's using an ophyd device I'm not sure the exact temperature measurement method/device is too important?

nrwslac:
> Now that you mention it, I'd be much happier with something like `grating_temp_1`. At the point where someone's using an ophyd device I'm not sure the exact temperature measurement method/device is too important?

yea, I think temp is the way to go. I'll update.

pcdsdevices/mirror.py
```diff
@@ -752,12 +752,16 @@ class XOffsetMirrorNoBend(XOffsetMirror):
 
     name : str
 
-    Currently (10/11/2023) services: mr1l1
+    Currently (10/11/2023) services: mr1l1, mr1k3, mr2k3
```
ZLLentz, line 755: Is the old date still correct in this comment?
```suggestion
    Currently (5/15/2024) services: mr1l1, mr1k3, mr2k3
```

ZLLentz:
I'll drop my approval once you update the names

nrwslac:
Updated mono temperature readout names and mirror doc string date.

ZLLentz (approved):
Looks good to me!

ZLLentz:
I'm going to roll through with an admin merge due to the current docs build issues

PR #1227, opened 2024-05-22
ENH: The set_many utility now works with Ophyd positioners as well as…

canismarko:
… signals.

## Description

Previously, the function ``pcdsdevices.util.set_many()`` worked with Signal objects, but failed with Positioner objects. This is because Positioner.set() does not accept a ``settle_time`` argument. This change inspects each object's ``set()`` method to determine which keyword arguments it should provide.

## Motivation and Context

I like the MultiDerivedSignal that depends on set_many, but it is limited to being derived from signals. With this change it can alse derive from positioners.

## How Has This Been Tested?

I added a test for the ``set_many`` function. The ``set_many()`` call used in this test includes both a positioner and a regular Ophyd Signal.

I still need to test this in my client library, so for now it's a draft PR for feedback and discussion.

The interface has not changed, so some comments are included in the code and the test. Client code should not notice a difference.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/utils.py
+++ b/pcdsdevices/utils.py
@@ -33,24 +33,24 @@ except ImportError:
 
 logger = logging.getLogger(__name__)
 
-arrow_up = '\x1b[A'
-arrow_down = '\x1b[B'
-arrow_right = '\x1b[C'
-arrow_left = '\x1b[D'
-shift_arrow_up = '\x1b[1;2A'
-shift_arrow_down = '\x1b[1;2B'
-shift_arrow_right = '\x1b[1;2C'
-shift_arrow_left = '\x1b[1;2D'
-alt_arrow_up = '\x1b[1;3A'
-alt_arrow_down = '\x1b[1;3B'
-alt_arrow_right = '\x1b[1;3C'
-alt_arrow_left = '\x1b[1;3D'
-ctrl_arrow_up = '\x1b[1;5A'
-ctrl_arrow_down = '\x1b[1;5B'
-ctrl_arrow_right = '\x1b[1;5C'
-ctrl_arrow_left = '\x1b[1;5D'
-plus = '+'
-minus = '-'
+arrow_up = "\x1b[A"
+arrow_down = "\x1b[B"
+arrow_right = "\x1b[C"
+arrow_left = "\x1b[D"
+shift_arrow_up = "\x1b[1;2A"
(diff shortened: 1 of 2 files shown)


pcdsdevices/utils.py
```diff
@@ -527,15 +548,16 @@ def set_many(
     """
     statuses = []
     log = owner.log if owner is not None else logger
+    set_kw = dict(timeout=timeout, settle_time=settle_time)
     for signal, value in to_set.items():
+        # Decide which keywords are suitable for this signal
+        params = inspect.signature(signal.set).parameters
+        _kw = {key: val for key, val in set_kw.items() if key in params.keys()}
```
ZLLentz, line 555: The test suite is failing because this approach breaks the kwargs-passing from this function for any signals who accept `timeout` and `settle_time` via `**kwargs`. Since the parameters aren't explicitly in the signature they get skipped, and the test that checks that they've been called with the given arguments fails.

canismarko: Ah yes. Okay, I've nested the ``set()`` in another ``try…except TypeError`` block that retries the ``set()`` call without the *settle_time* argument. I don't really love the nested ``try…except``s, but I think this describes what I'm actually trying to do.

I didn't include it as another ``except TypeError:`` for the outer ``try…except`` block because if the second attempt also raised a ``TypeError``, then this would bypass the logger in ``except Exception:``.

pcdsdevices/utils.py
```diff
@@ -528,14 +549,18 @@ def set_many(
     statuses = []
     log = owner.log if owner is not None else logger
     for signal, value in to_set.items():
+        # Set the actual signal value
         try:
-            st = signal.set(
-                value, timeout=timeout, settle_time=settle_time
-            )
+            try:
```
tangkong, line 554: This is definitely the simpler solution, but I was left wondering if we could inspect the signature and only pass the relevant kwargs through.  I think we could (`inspect.signature()` or `inspect.fullargspec()`), but there's no reason to over-complicate things here.

tangkong (approved):
This looks good to me.

pcdsdevices/utils.py
ZLLentz: The previous implementation used `inspect` but ran into issues

ZLLentz (approved):
I think this is a solid improvement and a good implementation, thank you for the contribution!

The black reformat is always welcome too

PR #1228, opened 2024-05-22
FIX: IMS class set_current_position limits issue

ZLLentz:
## Description

In the `IMS` class and in similar motor classes, `set_current_position` had a bug where it would refuse to let you set a position that is outside of the limits span.
See https://jira.slac.stanford.edu/browse/ECS-5354
This has been fixed by replacing the deprecated `set_and_wait` call (which threw a warning) with a `put(x, force=True)`, which ignores the limit. This is followed by a manual confirmation because `put` doesn't return a status object and the waiting doesn't work anyway for these specific devices (hence the special function override). I would have preferred to do something like `set(x, force=True).wait(timeout=1)` but `set` doesn't currently support the `force` kwarg. I think this is also a bug, but one I'm less motivated to look into.

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-5354

## How Has This Been Tested?

Interactively in XPP
I'll think at least for a moment if there's a reasonable unit test to add but there might not be

Here, jira, + later release notes

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -5,8 +5,10 @@ import functools
 import logging
 import shutil
 import subprocess
+import time
 from typing import Callable, ClassVar, Optional
 
+import numpy as np
 from ophyd.device import Component as Cpt
 from ophyd.device import Device
 from ophyd.device import FormattedComponent as FCpt
@@ -16,7 +18,7 @@ from ophyd.signal import EpicsSignal, EpicsSignalRO, Signal
 from ophyd.status import DeviceStatus, MoveStatus, SubscriptionStatus
 from ophyd.status import wait as status_wait
 from ophyd.utils import LimitError
-from ophyd.utils.epics_pvs import raise_if_disconnected, set_and_wait
+from ophyd.utils.epics_pvs import raise_if_disconnected
 from pcdsutils.ext_scripts import get_hutch_name
 from prettytable import PrettyTable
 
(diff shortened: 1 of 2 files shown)


ZLLentz:
Writing a unit test for this requires emulating some IOC behavior so it seems somewhat fruitless. I'm going to mark this as ready for review.

pcdsdevices/epics_motor.py
```diff
@@ -601,7 +603,13 @@ def set_current_position(self, pos):
         """
         self.set_use_switch.put(1, wait=True)
         try:
-            set_and_wait(self.user_setpoint, pos, timeout=1)
+            # Force to ignore limits
+            self.user_setpoint.put(pos, force=True)
+            # Instead of waiting on the put (broken), manually check
```
tangkong, line 608: For my own understanding, what we're doing here is putting the motor into set-mode and setting the user_setpoint to something that could be outside of the limits.  

Does this have the possibility of putting us into a weird state where the setpoint is outside the limits?

ZLLentz: When you change the set position using set mode, the limits also get updated appropriately to maintain the motor's apparent distance from them.

tangkong: And python doesn't pick up on this until after the set move I guess.  I see

tangkong (approved):
I agree that the ideal would be waiting on some status instead of just assuming that the process will resolve within 0.3s, but this seems to work and is good enough for now

pcdsdevices/epics_motor.py
ZLLentz: Yeah, the Python is actually unnecessarily limiting the setpoint update here- it sees that there are limits on the setpoint PV, but the IOC has no intention of enforcing them while in the setter mode.

ZLLentz: It's also worth noting also that all other version of this setter code use `force=True`, it must have snuck in as a mistake when re-implementing this here.

ZLLentz:
> I agree that the ideal would be waiting on some status

I'm a bit confused that the built-in waiting mechanisms don't allow me to pass `force` as a kwarg. Might look into this at a later date.

ZLLentz:
Updating branch and then merging if it passes again

PR #1234, opened 2024-05-24
Add VBAS and VMAX signals to PCDSMotorBase

### LINKED ISSUE #1224: Base Velocity and Velocity Max missing from EpicsMotor base class in pcdsdevices.epics_motor
## Current Behavior
`<PV>:m<N>.VMAX` and `<PV>:m<N>.VBAS` are not ported over in `EpicsMotor` parent class. This means that if either of those PVs get set (like in the `edm` screens, for example...) then you run into unclear limitations to changing the velocities in inherited screens. This is particularly troublesome for SmarActs, as you become unable to uncap the velocities, or may even be inexplicably stuck at certain velocities if the `VBAS` was set erroneously.

## Expected Behavior
`VMAX` and `VBAS` PVs from `EpicsMotor` parent class available in a `kind=config` type signal in Typhos screens.

## Context / environment
After changing over the out-of-loop mechanical delay lines in the FTL to the typhos screens, users (myself included) were unable to change the velocities on several stages with no clear indicator as to why.

1. `caput` the `VBAS` or `VMAX` to some nominal value (e.g. 1 mm/s)
2. Try to change the `velocity` signal in a `pcdsdevices.epics_motor.SmarAct` screen that is below either of those values

## Suggested Solution
Bring these two signals over in the `tab_whitelist` in the parent class to prevent this issue in the future.

slactjohnson:
## Description
Moves VBAS and VMAX signals to PCDSMotorBase for access by more sub-classes. 
Change SmarAct to use PCDSMotorBase. 

## Motivation and Context
Closes #1224 

## How Has This Been Tested?
Tested using the same test motor @aberges-SLAC had in his submitted issue. 

This PR.

## Screenshots:
Typhos suite of that motor with "omitted" signals also shown. 

image

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -108,6 +108,9 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
     EpicsMotor.low_limit_travel.kind = Kind.config
     EpicsMotor.direction_of_travel.kind = Kind.normal
 
+    velocity_base = Cpt(EpicsSignal, '.VBAS', kind='omitted')
+    velocity_max = Cpt(EpicsSignal, '.VMAX', kind='config')
+
     _alarm_filter_installed: ClassVar[bool] = False
     _moved_in_session: bool
     _egu: str
@@ -503,8 +506,6 @@ class PCDSMotorBase(EpicsMotorInterface):
     motor_spg = Cpt(EpicsSignal, '.SPG', kind='omitted')
 
     velocity = Cpt(EpicsSignal, '.VELO', kind='config')
-    velocity_base = Cpt(EpicsSignal, '.VBAS', kind='omitted')
-    velocity_max = Cpt(EpicsSignal, '.VMAX', kind='config')
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -1473,7 +1474,7 @@ class SmarActOpenLoopPositioner(PVPositionerComparator):
         return setpoint-self.atol < readback < setpoint+self.atol
 
 
-class SmarAct(PCDSMotorBase):
+class SmarAct(EpicsMotorInterface):
     """
     Class for encoded SmarAct motors controlled via the MCS2 controller.
     """
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -83,7 +83,8 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
 
     tab_whitelist = ["set_current_position", "home", "velocity",
                      "check_limit_switches", "get_low_limit",
-                     "set_low_limit", "get_high_limit", "set_high_limit"]
+                     "set_low_limit", "get_high_limit", "set_high_limit",
+                     "velocity_base", "velocity_max"]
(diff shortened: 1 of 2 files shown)


pcdsdevices/epics_motor.py
```diff
@@ -1470,7 +1473,7 @@ def done_comparator(self, readback, setpoint):
         return setpoint-self.atol < readback < setpoint+self.atol
 
 
-class SmarAct(EpicsMotorInterface):
+class SmarAct(PCDSMotorBase):
```
ZLLentz, line 1476: Does the SmarAct class have the same oddities that the PCDSMotorBase was created to fix?
```
        1. The 'TDIR' field does not exist on the PCDS implementation. This
        indicates what direction the motor has travelled last. To account for
        this difference, we use the :meth:`._pos_changed` callback to keep
        track of which direction we believe the motor to be travelling
        and store this in a simple :class:`~ophyd.signal.Signal`.
        2. The 'SPG' field implements the three states used in the LCLS
        motor record.  This is a reduced version of the standard EPICS
        'SPMG' field.  Setting to 'STOP', 'PAUSE' and 'GO'  will respectively
        stop motor movement, pause a move in progress, or resume a paused move.
        3. In some cases, puts to the setpoint PV cannot be waited on. In fact
        this seems to brick the whole python session, requiring a slightly
        modified set_current_position method.
```
If not, we should instead be moving these all the way up to `EpicsMotorInterface`.

slactjohnson: Ah, you're right. It does not. I'll move them.

aberges-SLAC:
If you want to be my hero, you could also fix line 1516 for me and close [issue #1235](https://github.com/pcdshub/pcdsdevices/issues/1235) :]]

slactjohnson:
@aberges-SLAC I could do that, but we generally don't do things that way. Usually we're just fixing one or more related bugs, in one PR. That particular bug is not related to the original intent of the PR. But I get it, it's just two letters. 

@ZLLentz What do you think about bundling Adam's requested fix in a separate PR vs. this PR? Not just with this PR, but rather what is our general stance on that kind of thing?

aberges-SLAC:
> @aberges-SLAC I could do that, but we generally don't do things that way. Usually we're just fixing one or more related bugs, in one PR. That particular bug is not related to the original intent of the PR. But I get it, it's just two letters.
> 
> @ZLLentz What do you think about bundling Adam's requested fix in a separate PR vs. this PR? Not just with this PR, but rather what is our general stance on that kind of thing?

It's most definitely bad practice. :] I should really just do the right thing and open a separate PR against the issue and have it tracked, just in case.

ZLLentz:
In general, the code is much easier to review and discuss in a focused way if it only addresses a limited number of related things. Bundling should be avoided when possible.

I wouldn't reject a PR for bundling unless it was excessive (tons of lines changed with multiple unrelated goals)

slactjohnson:
@ZLLentz Sounds good, thanks for confirming my suspicions.
@aberges-SLAC I will finish up this PR and make a separate one to fix the tip-tilt class.

tangkong (approved):
This looks good to me 👍

Issue #1238, opened 2024-05-31
Create SmarActPicoscale subclass for OL-MDL positioners


aberges-SLAC:
## Current Behavior
Currently, the OL-MDL in FTL are run in SmarAct typhos screens, which lack important PVs related to the Picoscale (PS) interferometer & sensor heads that actually dictate the 

## Expected Behavior
Be able to read PS related RBV on device state, PS state, working distances, etc. 

## Context / environment
Useful for OL-MDL in the FTL and other potential deployments of SmarAct Picoscale systems.

## Suggested Solution
Subclass the SmarAct device class in `pdsdevices.epics_motor` and include the necessary PVs.

PR #1239, opened 2024-05-31
Add SmarAct Picoscale device

aberges-SLAC:
## Description
Added a SmarActPicoscale device as a subclass of SmarAct in `pcdsdevices.epics_motor`. This pulls some Picoscale specific PVs to the display that are necessary for debugging. 

## Motivation and Context
The Out of Loop Mechanical Delay Lines use the Picoscale interferometers and sensor heads for the position encoding in their systems. Picoscale has it own quirks that will render the stages inoperable if violated (i.e. beam interrupted). This is not transparent to the user in the current SmarAct device class. This new subclass of a device brings the necessary diagnostic information forward.

Closes: https://github.com/pcdshub/pcdsdevices/issues/1238

## How Has This Been Tested?
I tested this screen on a currently deployed stage in the FTL and was able to retain normal functionality of the SmarAct and read the Picoscale specific PVs. 

I did the following in las-console:
```bash
$ cd /cds/group/pcds/epics-dev/aberges/pcdshub/pcdsdevices
$ source pcds_conda; export PYTHONPATH=$PWD
$ typhos "pcdsdevices.epics_motor.SmarActPicoscale[{'prefix': 'LAS:FTL:MCS:02:m2', 'name':'OL-MDL 5', 'ioc_base': 'LAS:FTL:MCS:02'}]" &
```

In this pull request.

image

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1238-Add_SmarActPicoscale.rst
+++ b/docs/source/upcoming_release_notes/1238-Add_SmarActPicoscale.rst
@@ -1,0 +1,33 @@
+1238 Add SmarActPicoscale subclass
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- Includes new PV RBVs for Picoscale at motor level:
+- pico_present, pico_exists, pico_sig_qual, pico_enable
+- Includes new PV RBVs for Picoscale at controller level:
+- pico_stable, pico_name, pico_wmin (working distance min), pico_wmax (working distance max)
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- added SmarActPicoscale subclass of SmarAct
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- aberges-SLAC
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -1526,28 +1526,51 @@ class SmarActEncodedTipTilt(Device):
 class SmarActPicoscale(SmarAct):
     """
     Class for encoded SmarAct motors controller via the MCS2 controller
-    which use the Picoscale sensor heads for encoded positions.
-    Child of SmarAct device.
+    which use the PicoScale sensor heads for encoded positions.
+    Instantiating a device requires inclusion of the ioc_base in order to
+    properly access PicoScale properties.
+
+    Example
+    -------
+    SmarActPicoscale('prefix': 'LAS:MCS2:02:m1', 'name': 'Delay 1',
+                     'ioc_base': 'LAS:MCS2:02').
     """
     # configuration settings and readbacks
-    pico_present = Cpt(EpicsSignalRO, ':PS_PRESENT', kind='config')
-    pico_exists = Cpt(EpicsSignalRO, ':HAVE_PS', kind='config')
-    pico_valid = Cpt(EpicsSignalRO, ':PS_DATA_VALID', kind='config')
-    pico_sig_qual = Cpt(EpicsSignalRO, ':PS_SIG_QUALITY', kind='config')
+    pico_present = Cpt(EpicsSignalRO, ':PS_PRESENT', kind='config',
+                       doc='PicoScale is on and connected')
+    pico_exists = Cpt(EpicsSignalRO, ':HAVE_PS', kind='config',
+                      doc='Does this channel have a PicoScale assigned to it?')
+    pico_valid = Cpt(EpicsSignalRO, ':PS_DATA_VALID', kind='config',
+                     doc='Is the data valid and ready for accurate msmts?')
+    pico_sig_qual = Cpt(EpicsSignalRO, ':PS_SIG_QUALITY', kind='config',
+                        doc='Quality of interferometer signal. Higher = better')
+
+    # auto and manual adjustment related PVs
+    pico_adj_state = FCpt(EpicsSignal, '{self._ioc_base}:PS:CUR_ADJ_STATE',
+                          write_pv='{self._ioc_base}:PS:REQ_ADJ_STATE',
+                          kind='config', doc='Set to Manual or Auto adjustment '
+                          'mode')
+    pico_curr_adj_prog = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:CUR_ADJ_PROGRESS',
+                              kind='config', doc='Estimate of current adjustment '
+                              'progress')
+    pico_adj_done = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:ADJ_DONE',
+                         kind='config', doc='Is the auto adjustment done?')
 
     # Validation that Picoscale is working properly
-    pico_enable = Cpt(EpicsSignalRO, ':PS_ENABLE_RBV', kind='normal')
-    pico_stable = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:STABLE', kind='normal')
+    pico_enable = Cpt(EpicsSignalRO, ':PS_ENABLE_RBV', kind='normal',
+                      doc='Is the PicoScale enabled for this channel?')
+    pico_stable = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:STABLE', kind='normal',
+                       doc='Is system stable and ready for accurate msmts?')
 
     # Diagnostic information from PS controller
     pico_name = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:LOCATOR',
-                     kind='config')
+                     kind='config', doc='Name of the PicoScale')
     pico_wmin = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:WORKING_MIN',
-                     kind='config')
+                     kind='config', doc='Working distance minimum')
     pico_wmax = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:WORKING_MAX',
-                     kind='config')
+                     kind='config', doc='Working distance maximum')
 
-    def __init__(self, prefix='', *, ioc_base, **kwargs):
+    def __init__(self, prefix, *, ioc_base, **kwargs):
         self._ioc_base = ioc_base
         super().__init__(prefix, **kwargs)
 
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -1523,6 +1523,35 @@ class SmarActEncodedTipTilt(Device):
         super().__init__(prefix, **kwargs)
 
 
+class SmarActPicoscale(SmarAct):
+    """
+    Class for encoded SmarAct motors controller via the MCS2 controller
+    which use the Picoscale sensor heads for encoded positions.
+    Child of SmarAct device.
+    """
+    # configuration settings and readbacks
+    pico_present = Cpt(EpicsSignalRO, ':PS_PRESENT', kind='config')
+    pico_exists = Cpt(EpicsSignalRO, ':HAVE_PS', kind='config')
+    pico_valid = Cpt(EpicsSignalRO, ':PS_DATA_VALID', kind='config')
+    pico_sig_qual = Cpt(EpicsSignalRO, ':PS_SIG_QUALITY', kind='config')
+
+    # Validation that Picoscale is working properly
+    pico_enable = Cpt(EpicsSignalRO, ':PS_ENABLE_RBV', kind='normal')
+    pico_stable = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:STABLE', kind='normal')
+
+    # Diagnostic information from PS controller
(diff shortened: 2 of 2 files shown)


pcdsdevices/epics_motor.py
```diff
@@ -1523,6 +1523,35 @@ def __init__(self, prefix='', *, tip_pv, tilt_pv, **kwargs):
...
 
 
+class SmarActPicoscale(SmarAct):
+    """
+    Class for encoded SmarAct motors controller via the MCS2 controller
+    which use the Picoscale sensor heads for encoded positions.
+    Child of SmarAct device.
+    """
+    # configuration settings and readbacks
+    pico_present = Cpt(EpicsSignalRO, ':PS_PRESENT', kind='config')
```
tangkong, line 1533: * Could you add some `doc` keyword text describing these?  For example, I'm not immediately clear what the difference between present, exists, and valid is in a practical sense.
* Is the `pico_` variable prefix necessary here?  Or is it obvious enough from the class name?  It seems like the `SmarAct` parent class doesn't really share any of these names except for:
  * `enable`, which collides with the `.enable()`  (maybe `enabled`)?
  * `name`, which collides with the existing `name`  (why not `locator`?)

tangkong (commented):
Some docs/naming nitpicks.  The meat of the diff here looks good, FCpt is being used as the makers intended. 👍

docs/source/upcoming_release_notes/1238-Add_SmarActPicoscale.rst
```diff
@@ -0,0 +1,33 @@
...
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- Includes new PV RBVs for Picoscale at motor level:
+- pico_present, pico_exists, pico_sig_qual, pico_enable
+- Includes new PV RBVs for Picoscale at controller level:
+- pico_stable, pico_name, pico_wmin (working distance min), pico_wmax (working distance max)
```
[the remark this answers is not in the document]
aberges-SLAC: That's fair, I'll commit that away :]

pcdsdevices/epics_motor.py
```diff
@@ -1523,6 +1523,35 @@ def __init__(self, prefix='', *, tip_pv, tilt_pv, **kwargs):
         super().__init__(prefix, **kwargs)
 
 
+class SmarActPicoscale(SmarAct):
+    """
+    Class for encoded SmarAct motors controller via the MCS2 controller
+    which use the Picoscale sensor heads for encoded positions.
+    Child of SmarAct device.
```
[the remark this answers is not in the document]
aberges-SLAC: Also fair!

pcdsdevices/epics_motor.py
aberges-SLAC: PicoScale is weird and in order for the PicoScale controller to play nice with the MCS2, it needs a combination of states to satisfy the "MCS2 Sensor Enabled" flag that triggers `Channel <N> 0:1 ? (does, does not) have an encoder` which lets you do closed loop movement.

The channel has to be:

1. System stable - ready to produce accurate results for position, (pico_stable)
2. Channel must be enabled for picoscale data (pico_enable)
3. The data must be valid (pico_valid)
4. Beam must not be interrupted

I will gladly add the doc text in the `FCpt` and clean this up :]

I used the pico prefix 80% for ordering the indicators in the resulting detailed/engineering views, 20% for clarifying that it explicitly comes from the PicoScale.

`Locator` is the name of the PV itself so I wanted to keep apples to apples on indicator and associated PV. It's technically the PicoScale IP address, which is given by a node-name. Helpful because we have 3 of these in the FTL and each PicoScale only supports 3 channels on an MCS2. I could change this to `picoscale_name` if it is clearer

slactjohnson: @aberges-SLAC I think that leaving this as `picoscale_locator` would be preferred by me. That keeps us consistent with the SmarAct naming convention, and these picoscales are unusual enough that if you're using one you probably know what it means. I think that adding a short explanation in the DOC field (which shows up as a tool tip) would go a long way to helping clear up any confusion among those who are not that familiar with this device.

pcdsdevices/epics_motor.py
```diff
@@ -1523,6 +1523,35 @@ def __init__(self, prefix='', *, tip_pv, tilt_pv, **kwargs):
...
+
+    # Diagnostic information from PS controller
+    pico_name = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:LOCATOR',
+                     kind='config')
+    pico_wmin = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:WORKING_MIN',
+                     kind='config')
+    pico_wmax = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:WORKING_MAX',
+                     kind='config')
+
+    def __init__(self, prefix='', *, ioc_base, **kwargs):
```
slactjohnson, line 1550: When instantiation is more than just the standard  `device('<prefix>', name='<name>')`, I think it's helpful to provide an example instantiation in the class docstring.

pcdsdevices/epics_motor.py
```diff
@@ -1523,6 +1523,35 @@ def __init__(self, prefix='', *, tip_pv, tilt_pv, **kwargs):
...
+    pico_stable = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:STABLE', kind='normal')
+
+    # Diagnostic information from PS controller
+    pico_name = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:LOCATOR',
+                     kind='config')
+    pico_wmin = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:WORKING_MIN',
+                     kind='config')
+    pico_wmax = FCpt(EpicsSignalRO, '{self._ioc_base}:PS:WORKING_MAX',
+                     kind='config')
+
```
slactjohnson, line 1549: Do you want to add any of the PVs related to the picoscale adjustment? I think these would be nice to have to attempt auto-alignment for remote recovery.

pcdsdevices/epics_motor.py
ZLLentz, line 1: Partially off-topic based on the other thread here (I have no additional comments beyond Robert's review):

> I used the pico prefix 80% for ordering the indicators in the resulting detailed/engineering views

Is the alphabetic sorting in these screens disruptive? Typhos could also use natural ordering that matches the order the components were created, and it could also provide runtime re-sorting. Just a thought because I'm working on typhos this week.

aberges-SLAC: I'm not sure if it is "disruptive" per se, but I do think the SmarAct themselves could benefit from grouping _some_ components on the screen, e.g. CHAN/MOD Temp, Pico, open-loop control.

I don't want to rock the boat _too_ much on what's established for these motor screens as folks are already accustomed to the main flow of screen. 

As annoying of a request as it could be, how feasible would it be to enable a flag/option for exempting & grouping certain components of a Typhos device as a kwarg?

pcdsdevices/epics_motor.py
aberges-SLAC: I thought about pulling those through, but I haven't had a situation where I've actually seen those flags _change_. But that might just be the nature of what I'm staring at when debugging and nudging the stages that are being ornery. 

From what I can tell, `<IOC_BASE>:PS:CUR_ADJ_STATE` and `<IOC_BASE>:PS:REQ_ADJ_STATE` are always `disabled` in the FTL. Likewise, `CURR_ADJ_PROGRESS` and `ADJ_DONE` are always `0`.

Maybe I need to go mess around with it more with a `camonitor` on

slactjohnson: That is to be expected when you are not actively calibrating. 
`REQ_ADJ_STATE` allows you to set the Picoscale into "Manual" or "Auto" adjustment, same as the SmarAct GUI. 
`CUR_ADJ_STATE` is the current adjustment state. 
`CURR_ADJ_PROGRESS` is an estimate of the progress of the auto adjustment cycle. 
`ADJ_DONE` means... you're done! :)

aberges-SLAC: TIL I didn't have to connect via USB to the picoscale every time ...

I'll add these as config PVs then as it is very convenient to have.

Users probably _still_ want to be in the FTL for this, as you want to walk the stage to the front for dead path calc, but this is honestly great to have.

slactjohnson: I've seen it fail when the SmarAct software succeeds, so there are perhaps some bugs to sort out there. But It's better than nothing.

pcdsdevices/epics_motor.py
ZLLentz: It's not that hard to customize the arrangement behavior and even do wild things like add extra tabs. Maybe let's brainstorm a bit offline the kinds of options that will be helpful and come up with a way to flag them.

pcdsdevices/epics_motor.py
ZLLentz: Side note: I'd recommend against setting a default (incorrect) value for prefix on any ophyd device to avoid confusing others later.

pcdsdevices/epics_motor.py
aberges-SLAC: So I just tried this in OL-MDL 4-6, and it worked, but it caused some unexpected motor move timeout errors while doing so, which is interesting?
image

Possible bug or expected behavior from how `MoveStatus` works? Stages were _not_ moving and had already concluded their movements before the adjustment began. Position RBV was also frozen during this time until the adjustment finished and `pico_stable` and `pico_valid` switched back to their True states.

Error did not persist, btw, and I could move it just fine after adjustment concluded.

slactjohnson: @aberges-SLAC Can you explain a little more how this occurred? During the auto-adjustment, you should not move the motor _at all_.

aberges-SLAC: Exactly, no movement commands were processing _before_ auto-adjustment was triggered. But it threw these errors. Not sure if it just crashed the `MoveStatus` based on an associated flag?

slactjohnson: OK, re-read your comment more carefully. I think I understand now.
Were the errors thrown during adjustment? Or after? 
If during, then that _might_ make sense if the picoscale stops responding during auto-adjustment.

aberges-SLAC: During, which lines up with my other observations.  Glad to know I didn't bork it :]

pcdsdevices/epics_motor.py
tangkong: Did @aberges-SLAC decide to leave this one as is?

aberges-SLAC: Yes, I kept `pico_locator` as it but gave it a better tooltip

pcdsdevices/epics_motor.py
```diff
@@ -1523,6 +1523,58 @@ def __init__(self, prefix='', *, tip_pv, tilt_pv, **kwargs):
         super().__init__(prefix, **kwargs)
 
 
+class SmarActPicoscale(SmarAct):
+    """
+    Class for encoded SmarAct motors controller via the MCS2 controller
+    which use the PicoScale sensor heads for encoded positions.
+    Instantiating a device requires inclusion of the ioc_base in order to
```
ZLLentz, line 1530: Just making sure: ioc_base is something that would be obvious to someone who deploys smaract IOCs? I can think of two different things this can mean in abstract (the IOC admin prefix, a controller-based prefix)

aberges-SLAC: The latter not the former. IIRC the former has `IOC:<ioc_base>` format using my current meaning of `ioc_base`

ZLLentz: I'm trying to make sure someone will be able to follow the docs later, so my suggestion is to consider if it needs clarification, or if it does not. Otherwise the PR looks good to me!

slactjohnson:
@aberges-SLAC If the code is ready to be reviewed, can you confirm that you have completed all the pre-merge steps by checking the boxes?

pcdsdevices/epics_motor.py
aberges-SLAC: Definitely valid to ask! I'm actually _not_ entirely sure we're using this field as intended, but it certainly _is_ how we're using it in existing devices in MODS:
image

I tried to pick what already existing in the `happi` container

ZLLentz (approved):
Looks good to me!

slactjohnson (approved):
LGTM

aberges-SLAC:
> LGTM

_added to SLAC-speak_

tangkong (approved):
> LGTM added to SLAC-speak

Oh no this is an everywhere acronym.

ZLLentz:
I'm going to update the branch here and then merge after the tests pass again

PR #1240, opened 2024-06-04
Mfx spectrometer

vespos:
## Description
Class for the MFX 6 crystal spectrometer

## Motivation and Context
CLI control of the device

## How Has This Been Tested?
WIth mfx3

Docstring

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1240-mfx_6crystal_spec.rst
+++ b/docs/source/upcoming_release_notes/1240-mfx_6crystal_spec.rst
@@ -15,8 +15,8 @@ Device Updates
 
 New Devices
 -----------
-- VonHamos6Crystal: New MFX 6-crystals spectrometer running on plc-mfx-motion. Contains 6 crystals motions + 3 translation 1 rotation for the base
-- VonHamosCrystal_2: interface to motor stack of a single crystal in the sectrometer
+- VonHamos6Crystal: New MFX 6-crystals spectrometer running on plc-mfx-motion. Contains 6 crystals motions + 3 translations and 1 rotation for the base.
+- VonHamosCrystal_2: interface to the motor stack of a single crystal in the spectrometer.
 
 Bugfixes
 --------
--- a/docs/source/upcoming_release_notes/1240-mfx_6crystal_spec.rst
+++ b/docs/source/upcoming_release_notes/1240-mfx_6crystal_spec.rst
@@ -15,8 +15,8 @@ Device Updates
 
 New Devices
 -----------
-- VonHamos6Crystal
-- VonHamosCrystal_2
+- VonHamos6Crystal: New MFX 6-crystals spectrometer running on plc-mfx-motion. Contains 6 crystals motions + 3 translation 1 rotation for the base
+- VonHamosCrystal_2: interface to motor stack of a single crystal in the sectrometer
 
 Bugfixes
 --------
--- a/docs/source/upcoming_release_notes/1240-mfx_6crystal_spec.rst
+++ b/docs/source/upcoming_release_notes/1240-mfx_6crystal_spec.rst
@@ -1,0 +1,31 @@
+1240 mfx_6crystal_spec
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- VonHamos6Crystal
+- VonHamosCrystal_2
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- espov
--- a/pcdsdevices/spectrometer.py
+++ b/pcdsdevices/spectrometer.py
@@ -105,7 +105,10 @@ class Kmono(BaseInterface, GroupDevice, LightpathMixin):
 
 
 class VonHamosCrystal(BaseInterface, GroupDevice):
-    """Pitch, yaw, and translation motors for control of a single crystal."""
+    """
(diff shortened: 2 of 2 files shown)


vespos:
Let me know if you think we should change the names of the classes or the structure. The rest is pretty straight-forward I think.

pcdsdevices/spectrometer.py
```diff
@@ -218,6 +218,32 @@ def __init__(self, prefix, *, name, prefix_focus, prefix_energy, **kwargs):
                          prefix_energy=prefix_energy, **kwargs)
 
 
+class VonHamosCrystal_2(BaseInterface, GroupDevice):
+    """Translation, rotation and tilt motors for control of a single crystal."""
```
ZLLentz, line 222: This is the only name that's kind of wonky- an unexplained _2.
I think either the name or the docstrings needs to explain when to use it.

The rest of the classes here have similar names but hopefully they have descriptive enough docstrings...

docs/source/upcoming_release_notes/1240-mfx_6crystal_spec.rst
```diff
@@ -0,0 +1,31 @@
...
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- espov
```
tangkong, line 31:
```suggestion
- vespos
```

docs/source/upcoming_release_notes/1240-mfx_6crystal_spec.rst
```diff
@@ -0,0 +1,31 @@
...
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- VonHamos6Crystal
+- VonHamosCrystal_2
```
tangkong, line 19: I think along with disambiguating the two crystals in the code docstrings, we could use a brief explanation here

pcdsdevices/spectrometer.py
tangkong: This is probably to differentiate this from an existing VonHamosCrystal.  https://github.com/pcdshub/pcdsdevices/blob/781df8b23e111916e9f052fb43dd3a39ca5f6cd9/pcdsdevices/spectrometer.py#L109-L115

This seems to just be a different coordinate system.   They effectively do the same thing right?  I'm not volunteering, but this could be some fancy autodetecting pseudopositioner class that accepts either set of PVs?

(not for this PR of course)

vespos: Yeah I am not too happy with that name either. Another way would be to define the single crystal class inside the main spectrometer class, and then the names wouldn't be a probem. Not sure that's much more elegant though... Thoughts?

tangkong: For now I think just more descriptive names are sufficient.  Maybe:
- VonHamosCrystal_YPT  (Yaw-pitch-trans)
- VonHamosCrystal_Spherical (this is basically a spherical coordinate system, right?)

vespos: It's not quite spherical I think...
Also I doubt that we'll ever get to re-use the same crystal class for another spectrometer, so the more I think about the it, the more I kinda like the idea of embedding the crystal class inside the spectrometer class. But I don't feel very strongly about any names here...

ZLLentz: I don't mind if it doesn't get renamed, but the description should say "this is for the mfx spectrometer", for example.

vespos:
Added some docstring to clarify what spectrometer each single crystal control belongs to

pcdsdevices/spectrometer.py
```diff
@@ -218,6 +221,35 @@ def __init__(self, prefix, *, name, prefix_focus, prefix_energy, **kwargs):
...
+
+    c1 = Cpt(VonHamosCrystal_2, ':C1', kind='normal')
+    c2 = Cpt(VonHamosCrystal_2, ':C2', kind='normal')
+    c3 = Cpt(VonHamosCrystal_2, ':C3', kind='normal')
+    c4 = Cpt(VonHamosCrystal_2, ':C4', kind='normal')
+    c5 = Cpt(VonHamosCrystal_2, ':C5', kind='normal')
+    c6 = Cpt(VonHamosCrystal_2, ':C6', kind='normal')
+
+    rot = Cpt(BeckhoffAxis, ':ROT', kind='normal')
+    t1 = Cpt(BeckhoffAxis, ':T1', kind='normal')
```
tangkong, line 248: I feel like "t#" is pretty ambiguous here, since we've seen both "translation" and "tilt" in this context.  If scientists want this to be short, could we add something in the `doc` for these components letting us know what these are meant to control?  Since there's nothing in the docstring here either.  

Also it's entirely unclear where each of these axes lie. (Is one of these parallel to beam? Is it at some arbitrary angle?) Maybe that's not possible to know apriori, but if we can I think it warrants a bit more verbosity.

tangkong (commented):
Sorry I keep nitpicking the docs/naming, I just think back to times when I try to help with hutch-python stuff and get hit with the wall of letters I don't understand.

pcdsdevices/spectrometer.py
vespos: That's a good point. I have changed the name to match the names the scientists gave in the screen.

tangkong (approved):
I'm out of nitpicks here.  Thanks for humoring me 👍

ZLLentz (approved):

ZLLentz:
I'm going to update the branch on Adam's PR, then merge Adam's PR, then update the branch on this PR, then merge this PR

PR #1242, opened 2024-06-05
Add TprTrigger default timing mode

### LINKED ISSUE #1241: TprTrigger - Default timing_mode
## Current Behavior
The `TprTrigger` class forces you to specify whether you are using it for LCLS1 or LCLS2 timing through an enum, rather than a string or integer argument. There is no default. 

## Expected Behavior
Given that the TPR was designed for LCLS-2, the LCLS-2 mode should be the default. 
This enum is a little annoying and verbose. I feel that a string parameter would be easier to use. 

## Context / environment
I am trying to dynamically create device classes from a yaml config file. I can't think of a good, re-usable way to pass this enum to the class when I'm pulling all of my data from a text file. 

## Suggested Solution
Add a default state, and ideally change the enum to a simple string.

slactjohnson:
## Description
Make LCLS2 the default timing mode for the `TprTrigger` class. 

## Motivation and Context
closes #1241 

## How Has This Been Tested?
Tested interactively. 

This PR. 

image

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1241-Add_default_TprTrigger_timing_mode.rst
+++ b/docs/source/upcoming_release_notes/1241-Add_default_TprTrigger_timing_mode.rst
@@ -1,0 +1,30 @@
+1241 Add default TprTrigger timing mode
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- TprTrigger: Make LCLS2 timing the default timing_mode
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- slactjohnson
--- a/pcdsdevices/tpr.py
+++ b/pcdsdevices/tpr.py
@@ -59,7 +59,19 @@ class TprMotor(PVPositionerIsClose):
 
 
 class TprTrigger(BaseInterface, Device):
-    """Class for an individual TprTrigger."""
+    """
+    Class for an individual TprTrigger.
+
+    Parameters
+    ----------
+    timing_mode: int, str, or Enum
+        The timing mode the TPR is configured for. Can be an enum
+        (tpr.TimingMode.LCLS1 tpr.TimingMode.LCLS1) a str ("LCLS1" or "LCLS2")
+        or an int (1 or 2).
+
+    channel: int
+        The integer channel to be used (0 through 11).
+    """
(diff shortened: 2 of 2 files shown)


slactjohnson:
Hmm, looks like Happi may have similar trouble with this enum: 
image
image

tangkong (dismissed):
This seems fine to me.  👍

tangkong:
Yeah happi wouldn't know to interpret `"TimingMode.LCLS1"` as an enum it needs to import, so I guess we definitely need some type of string parsing here.  Maybe something like:

```python
def __init__(self, prefix, *, channel, name, timing_mode: TimingMode | str =TimingMode.LCLS2, **kwargs):
    if isinstance(timing_mode, str):
        try:
            timing_mode = getattr(TimingMode, timing_mode)
        except AttributeError as ex:
            raise ValueError(
                'Provided timing mode not recognized, please choose from: '
                f'{[mode.name for mode in TimingMode]}'
            )
    ...     
```

ZLLentz:
Enums have built-in tools for consuming strings and returning enum values:
```
In [1]: from pcdsdevices.tpr import TimingMode

In [2]: TimingMode["LCLS1"]
Out[2]: <TimingMode.LCLS1: 1>

In [3]: TimingMode["LCLS2"]
Out[3]: <TimingMode.LCLS2: 2>

In [4]: TimingMode(1)
Out[4]: <TimingMode.LCLS1: 1>

In [5]: TimingMode(2)
Out[5]: <TimingMode.LCLS2: 2>
```
This makes it very easy to convert str and int inputs into the enum members.

ZLLentz:
One more mini-tip: the call/int version also accepts the enums, reducing the number of cases to check:
```
In [6]: TimingMode(TimingMode.LCLS1)
Out[6]: <TimingMode.LCLS1: 1>

In [7]: TimingMode(TimingMode.LCLS2)
Out[7]: <TimingMode.LCLS2: 2>
```

slactjohnson:
My recent commit takes advantage of the tricks @ZLLentz mentioned, and now makes specifying `timing_mode` much more flexible while still erroring out on an invalid setting. 

image

image


I kept running into errors with the type hints in Robert's example and I wasn't type hint savvy enough to figure it out. 

image

tangkong:
Ah that's my bad, I kinda just pseudocoded and didn't actually run the code myself.  You probably have to use the explicit `typing.Union`

```
from typing import Union

...

def __init__(self, prefix, *, channel, name, timing_mode: Union[TimingMode, str]  = TimingMode.LCLS2, **kwargs):
```

Of course you could also leave it out as you have.

slactjohnson:
I think I'm happy enough with this as is.

tangkong (approved):

ZLLentz (approved):

PR #1245, opened 2024-06-12
FIX: drop tmo fzp att from 6 states to 4 to match plc update

ZLLentz:
## Description

See title
The fzp attenuator lost two targets due to space concerns and this new state count had not propagated down to the Python code. I made the required change here.

## Motivation and Context

There are some confusing errors if the state count doesn't match. This is probably also something to fix.

## How Has This Been Tested?

Ming Fu says he can move the device again

Will make brief pre-release notes

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1245-fix_fzp_att_state_count.rst
+++ b/docs/source/upcoming_release_notes/1245-fix_fzp_att_state_count.rst
@@ -1,0 +1,32 @@
+1245 fix_fzp_att_state_count
+############################
+
+API Breaks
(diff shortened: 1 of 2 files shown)


ZLLentz:
This is live in dev_conda to let tmo use their screen again

tongju12 (approved):
Thanks Zach to fix this. I should remember to do this after I comment two last states from PLC.

tangkong (approved):
Seems good and justifiable to me 👍

PR #1248, opened 2024-06-18
removing st3k4 auto error

tongju12:
remove st3k4 auto error so that it will not show error when st1k4 is not flowing st3k4 when st3k4 is OUT

remove error message lines
## Motivation and Context

Do not give error while st1k4 is not out when ST3K4 is OUT

## How Has This Been Tested?

Test it in tmo under dev_conda. It works as expected. 

Dev_conda

(https://jira.slac.stanford.edu/browse/ECS-5586)

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1248-DOC:ST1K4_decoupling_ST3K4_when_ST3K4_OUT.rst
+++ b/docs/source/upcoming_release_notes/1248-DOC:ST1K4_decoupling_ST3K4_when_ST3K4_OUT.rst
@@ -1,0 +1,30 @@
+1248 DOC:ST1K4 decoupling ST3K4 when ST3K4 OUT
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- Decoupling ST1K4 with ST3K4 when ST3K4 OUT
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- @tongju12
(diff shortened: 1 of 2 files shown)


pcdsdevices/sxr_test_absorber.py
```diff
@@ -34,10 +34,13 @@ def set(
...
+        """
         if self.st3k4_auto.get():
-            raise ST3K4AutoError(
-                "ST1K4 must follow ST3K4. Move rejected."
-            )
+           raise ST3K4AutoError(
+               "ST1K4 must follow ST3K4. Move rejected."
+           )
+        """
         return super().set(position, moved_cb=moved_cb, timeout=timeout)
```
ZLLentz, line 44: I think the correct fix here is to completely remove this `set` method override, starting from `def set` and ending at this line.

If we no longer need to prevent the moves here, we don't need this method at all, we can use the parent class's set instead.

docs/source/upcoming_release_notes/1248-DOC:ST1K4_decoupling_ST3K4_when_ST3K4_OUT.rst
```diff
@@ -0,0 +1,30 @@
+1248 DOC:ST1K4 decoupling ST3K4 when ST3K4 OUT
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- Decoupling ST1K4 with ST3K4 when ST3K4 OUT
```
ZLLentz, line 10: This isn't what the code change does here. The decoupling is implemented in the PLC code.
This also should be under `Device Updates`, not under `Features`.
Maybe something more like this?
```
- Allow ST1K4 to move regardless of auto mode.
  See https://github.com/pcdshub/lcls-plc-kfe-motion/pull/122,
  where we decouple ST1K4 from ST3K4 when ST3K4 is OUT.
```

ZLLentz:
Test suite failures are unrelated to Tong's PR, see #1250

ZLLentz (approved):
I'll bypass the failing CI and merge this

PR #1251, opened 2024-06-19
DEV/DOC: Adding PPM classes for power meter that uses responsivity in…

KaushikMalapati:
… calculations

## Description

Adding two classes - PPM_R_PowerMeter and PPM_R to be used with power meters that use new way of calculating pulse energies using responsivity and support automatic collection of background voltages.

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-875
New classes to support changes from this ticket

## How Has This Been Tested?

Not yet tested

New pre-release notes and docstrings on classes

## Pre-merge checklist


ZLLentz (changes requested):
Can you make this the normal PPM class instead of creating a new one?

Counterintuitively, creating a class bifurcation causes more backwards-compatibility problems rather than preventing them, because it requires us to update `happi` independently for each instance, and then anyone running an older version of pcdsdevices suddenly cannot `happi` load the updated PPM because the new class doesn't exist yet!

Rather than have some clients that can't use the PPMs at all, I'd prefer some distributed partial PV disconnects (provided they don't conflict with other PPM features)

I realize this is a fundamental issue with how we deploy `happi` devices and I'm open to suggestions on ways to do this better.

ZLLentz:
don't worry about the various weird test failures- there are some cascading issues related to the numpy update

docs/source/upcoming_release_notes/1249-Responsive_PPM.rst
ZLLentz, line 1: Looks like we have two copies of the pre-release notes in this pr

pcdsdevices/pim.py
```diff
@@ -420,6 +420,66 @@ class PPMPowerMeter(BaseInterface, Device):
...
+
+    responsivity = Cpt(PytmcSignal, ':RES', io='io', kind='normal', doc='Responsivity in  V/W, unique for every power meter.')
+
+    background_voltage = Cpt(PytmcSignal, ':BACK:VOLT', io='io', kind='normal', doc='Background voltage value used to calculate pulse energy.')
+
+    auto_background_reset = Cpt(PytmcSignal, ':BACK:RESET', io='io', kind='normal', doc='Set to reset auto background voltage collection.')
+    set_metadata(auto_background_reset, dict(variety='command-proc', value=1))
+
+    background_mode = Cpt(PytmcSignal, ':BACK:MODE', io='io', kind='normal', doc='Can be manual or passive. In manual mode, you can collect for a specified number of seconds. In passive mode, a buffer of automatically collected background voltages will be used to calculate the background voltage.')
+    set_metadata(background_mode, dict(variety='command-enum'))
```
ZLLentz, line 447: btw, `command-enum` makes the typhos screen display separate buttons for each option, arranged vertically. Just making sure you know, some people prefer the combobox but it's a matter of taste and I don't have a strong opinion.

ZLLentz (commented):
The class as-designed is really good

pcdsdevices/pim.py
KaushikMalapati: I was assuming that you needed command-enum for the combobox.

ZLLentz: These aren't documented super well, the only way to know is to flip through the decorators at https://github.com/pcdshub/typhos/blob/master/typhos/widgets.py

ZLLentz:
I think this looks great! Let's hold until we can try this on the real thing and then merge as-is if we're happy with it.

tangkong:
This also needs a master branch merge, but I agree with Zach's comments above

ZLLentz (approved):
I've confirmed that the screen is still totally functional for PPMs that haven't been upgraded yet :+1:

I'd like to see an example of it for an already updated device/ioc but this can wait until we have one.

ZLLentz:
I think this is still good as-is to merge but I need to click update branch first
To avoid clicking it twice I'm going to click it after the other PR is merged
And then we can merge this one after the tests re-pass again

PR #1252, opened 2024-06-19
FIX: workaround from ECS-5037

ZLLentz:
## Description

In Lcls2LaserTiming, handle cases where the done transition is shorter than the IOC scan rate by sleeping a bit.

## Motivation and Context

This workaround has allowed TMO and RIX to properly scan these devices.
It needs to be revisited in the future.

## How Has This Been Tested?

Live, with real data!

https://jira.slac.stanford.edu/browse/ECS-5037

## Pre-merge checklist


tangkong (commented):
I feel so powerless without the CI...

This smells like a classic case of "solving the problem in the wrong layer" to me.  If the IOC is busted, why are we fixing it in python and sweeping the real problem under the rug?  

I'm sure there's a reason, and I likely don't fully understand the issue here.  Is there an intent to fix this at the IOC level at all?

slactjohnson:
@tangkong I agree. Unfortunately the IOC is limited by the implementation of the ycpswAsyn driver. The ATCA common platform software is based on periodically polling registers, and does not have support for I/O interrupts, so we can't get the update from the register as soon as it updates. During very short timing adjustments, typically on the ~100 ps level, we don't catch the rising edge and that leads to timeouts. At this point I've increased the polling to 10Hz, which solves most of the issues, but there are occasionally timeouts when they scan small increments. 
I enquired with TID about making this possible via the ycpswAsyn driver, but it sounded like a significant effort for them with only one current use case (us).

slactjohnson (approved):
It's working in TMO/RIX and is effectively equivalent to the previous LXT, just with new PVs. LGTM

tangkong (approved):

ZLLentz:
Oh, if the polling is 10Hz, should I drop the sleep from 2s to 0.2s?

ZLLentz:
Actually, let's consider doing that separately- this is what's live and working, it can be refined.

PR #1253, opened 2024-06-19
MAINT/CI: drop pint in favor of sympy, various CI tweaks

ZLLentz:
## Description

Try to replace `pint` with `sympy` to see if we have more stable test suites.
Various other CI adjustments related to numpy 2.0

## Motivation and Context

We've historically had some compatibility issues with pint every time a major package has an API change.
For example: right now, numpy has updated to 2.0 and an optional dependency of pint is not importable with numpy 2 and therefore it busts the entire test suite.

Let's stop using it because we don't need all the features and this pint.compat module has been a constant thorn in our side.

Candidate 1: raw sympy (here)
Candidate 2: possibly unyt? (separate PR?)

## How Has This Been Tested?

The tests still pass for our existing unit conversions

Here only

## Pre-merge checklist


.github/workflows/standard.yml
```diff
@@ -20,6 +20,7 @@ jobs:
       # Extras to be installed only for conda-based testing:
       conda-testing-extras: ""
       # Extras to be installed only for pip-based testing:
-      pip-testing-extras: "PyQt5"
+      # Set numpy<2 for now, remove when xraylib updates
+      pip-testing-extras: "PyQt5 numpy<2"
```
tangkong, line 24: I thought it might be good to pin this in ci-helpers as well, but if we do that we'll never find all the other incompatibilities.

pcdsdevices/utils.py
```diff
@@ -154,15 +155,15 @@ def convert_unit(value, unit, new_unit):
     new_value : float
         The starting value, but converted to the new unit.
     """
+    unit = getattr(units, unit)
+    new_unit = getattr(units, new_unit)
 
-    global ureg
-    if ureg is None:
-        import pint
+    if unit == new_unit:
```
tangkong, line 161: I presume this is to catch cases like `convert_unit(1, 's', 'seconds')`?

I also like how this helps us avoid globals.

tangkong (approved):
I think this is totally sufficient and solves many of our problems. 

Unrelated to pcdsdevices: I do wonder if there are non-library consumers of pcds-conda  that use `pint`.  

I think I'm convinced the numpy2.0 conflicts end with xraylib, though you pinned it early in the commit history

pcdsdevices/utils.py
ZLLentz: Yes, and further this is because the results of the following conversion don't have a coefficient if you convert the same unit to itself, which was breaking the test suite in an earlier version.

.github/workflows/standard.yml
ZLLentz: Yeah, I think I'd rather learn about all the failures...

ZLLentz:
I'm going to get this in so the CI will start being useful in the other PRs, even though I think there are a couple small things to improve here.

PR #1259, opened 2024-07-19
ENH: Add MSTA Bits

slactjohnson:
## Description
Adds in an EPICS signal to bring the the motor MSTA bits. This signal is used to create a dictionary of interpreted values, as well as a "homed" property. 

## Motivation and Context
I would like to implement some automated homing on some finicky motor stacks. I need a way to check that the homing routine has completed.  

## How Has This Been Tested?
Tested live on one of the stages of interest (a SmarAct linear stage). 

This PR. 

## Screenshots:
Live session (standard motor record):
image

IMS Motor:

happy_ims

sad_ims

sad_ims_2

image

Newport Motor:

happy_newport

sad_newport

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1205-Add_MSTA_field_and_homed_status_to_EpicsMotorInterface.rst
+++ b/docs/source/upcoming_release_notes/1205-Add_MSTA_field_and_homed_status_to_EpicsMotorInterface.rst
@@ -8,13 +8,20 @@ API Breaks
 Features
 --------
 - MstaEnum: Enum describing the motor record MSTA bits
+- NewportMstaEnum: Enum describing the special Newport motor record MSTA bits
+- IMSMstaEnum: Enum describing the special IMS motor record MSTA bits
 
 Device Updates
 --------------
 - EpicsMotorInterface: Add a "raw" MSTA value, as well as the interpreted
                        values as a dictionary. Adds a "homed" property based
-                       based on this.
-
+                       on this. Uses a "generic" MstaEnum class.
+- Newport: Add a "raw" MSTA value, as well as the interpreted values as a
+           dictionary. Adds a "homed" property based on this. Uses the
+           "NewportMstaEnum" class.
+- IMS: Add a "raw" MSTA value, as well as the interpreted values as a
+       dictionary. Adds a "homed" property based on this. Uses the
+       "IMSMstaEnum" class.
 New Devices
 -----------
 - N/A
--- a/docs/source/upcoming_release_notes/1205-Add_MSTA_field_and_homed_status_to_EpicsMotorInterface.rst
+++ b/docs/source/upcoming_release_notes/1205-Add_MSTA_field_and_homed_status_to_EpicsMotorInterface.rst
@@ -1,0 +1,32 @@
+1205 Add MSTA field and homed status to EpicsMotorInterface
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- MstaEnum: Enum describing the motor record MSTA bits
+
+Device Updates
+--------------
+- EpicsMotorInterface: Add a "raw" MSTA value, as well as the interpreted
+                       values as a dictionary. Adds a "homed" property based
+                       based on this.
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- slactjohnson
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -6,6 +6,7 @@ import logging
 import shutil
 import subprocess
 import time
+from enum import Enum
 from typing import Callable, ClassVar, Optional
 
 import numpy as np
@@ -36,6 +37,30 @@ from .utils import get_status_float, get_status_value
 from .variety import set_metadata
 
 logger = logging.getLogger(__name__)
+
+
+class MstaEnum(Enum):
+    """
+    Enum for the EPICS motor record .MSTA field bits.
+    """
+    # Note: the motor record docs start bit numbering at 1, but the MSTA bits
+    # start at 0.
+    direction = 0       # last raw move direction (0: negative, 1: positive)
+    done = 1            # motion is complete
+    plus_ls = 2         # plus limit switch is hit
+    home_ls = 3         # state of the home limit switch
+    # Bit 4 is un-used
+    # closed-loop positioning enabled, called "position" in the docs
+    closed_loop = 5
+    slip_stall = 6      # slip stall is detected
+    home = 7            # at the home position
+    enc_present = 8     # encoder is present. Called "present" in the docs.
+    problem = 9         # driver stopped polling, or there's a hardware problem
+    moving = 10         # the motor has a non-zero velocity (it's moving)
+    gain_support = 11   # the motor supports closed loop control
+    comm_error = 12     # controller communication error
+    minus_ls = 13       # minus limit switch is hit
+    homed = 14          # the motor has been homed
 
 
 class EpicsMotorInterface(FltMvInterface, EpicsMotor):
@@ -112,6 +137,8 @@ class EpicsMotorInterface(FltMvInterface, EpicsMotor):
 
     velocity_base = Cpt(EpicsSignal, '.VBAS', kind='omitted')
     velocity_max = Cpt(EpicsSignal, '.VMAX', kind='config')
+
+    msta_raw = Cpt(EpicsSignalRO, '.MSTA', kind='omitted')
 
     _alarm_filter_installed: ClassVar[bool] = False
     _moved_in_session: bool
@@ -196,6 +223,26 @@ Limit Switch: {switch_limits}
             # Not initialized
             return (0, 0)
         return limits
+
+    @property
+    def msta(self):
+        """
+        Returns the msta fields as a dictionary.
+        """
+        # the MSTA field is a float for some reason...
+        val = int(self.msta_raw.get())
+        d = dict()
+        for bit in MstaEnum:
+            d[bit.name] = (val >> bit.value) & 0x1
+        return d
+
+    @property
+    def homed(self):
+        """
+        Get the home status of the motor as reported by the MSTA field.
+        """
+        msta = self.msta
+        return bool(msta[MstaEnum.homed.name])
 
     @limits.setter
     def limits(self, lims: tuple[float, float]):
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -63,6 +63,67 @@ class MstaEnum(Enum):
     homed = 14          # the motor has been homed
 
 
+class NewportMstaEnum(Enum):
+    """
+    Enum for the LCLS Newport XPS8 EPICS motor record .MSTA field bits.
+    """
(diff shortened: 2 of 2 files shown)


pcdsdevices/epics_motor.py
```diff
@@ -38,6 +39,30 @@
...
+    """
+    Enum for the EPICS motor record .MSTA field bits.
+    """
+    # Note: the motor record docs start bit numbering at 1, but the MSTA bits
+    # start at 0.
+    direction = 0       # last raw move direction (0: negative, 1: positive)
+    done = 1            # motion is complete
+    plus_ls = 2         # plus limit switch is hit
+    home_ls = 3         # state of the home limit switch
+    # Bit 4 is un-used
```
tangkong, line 52: gotta love EPICS.  Is this a bitmask thing?

slactjohnson: Yep. The field and bit descriptions are in a table here: https://epics.anl.gov/bcda/synApps/motor/R5-6/motorRecord.html#Fields_status

tangkong: That doc says bit 6 is also unused.  Is our motor record different?

slactjohnson: My bad. That's an old version of the documentation. Here's the latest, which I believe matches what I have here: https://epics.anl.gov/bcda/synApps/motor/motorRecord.html#Fields_status

pcdsdevices/epics_motor.py
```diff
@@ -197,6 +224,26 @@ def _get_epics_limits(self) -> tuple[float, float]:
...
+    def msta(self):
+        """
+        Returns the msta fields as a dictionary.
+        """
+        # the MSTA field is a float for some reason...
+        val = int(self.msta_raw.get())
+        d = dict()
+        for bit in MstaEnum:
+            d[bit.name] = (val >> bit.value) & 0x1
+        return d
```
tangkong, line 237:
```python
d = {bit.name: (val >> bit.value) & 0x1 for bit in MstaEnum}  
```
A fun pythonic dict-comprehension could make this a one-liner.

A more meaningful comment: I wonder if there's a clearer way to do these bit shifts.  It took my python mind a while to understand the simple bit-extraction you're doing here.  

Maybe `enum.IntFlag`?
```python
from enum import IntFlag, auto

class MstaEnum(IntFlag):
    """
    Enum for the EPICS motor record .MSTA field bits.
    """
    # Note: the motor record docs start bit numbering at 1, but the MSTA bits
    # start at 0.
    direction = auto()       # last raw move direction (0: negative, 1: positive)
    done = auto()            # motion is complete
    plus_ls = auto()         # plus limit switch is hit
    home_ls = auto()         # state of the home limit switch
    # Bit 4 is un-used
    closed_loop = auto()
    slip_stall = auto()      # slip stall is detected
    home = auto()            # at the home position
    enc_present = auto()     # encoder is present. Called "present" in the docs.
    problem = auto()         # driver stopped polling, or there's a hardware problem
    moving = auto()         # the motor has a non-zero velocity (it's moving)
    gain_support = auto()   # the motor supports closed loop control
    comm_error = auto()     # controller communication error
    minus_ls = auto()       # minus limit switch is hit
    homed = auto()          # the motor has been homed


print(MstaEnum(0))  # >> MstaEnum.0
print(MstaEnum(129))  # >> MstaEnum.enc_present|direction

print([flag.name for flag in MstaEnum if flag in MstaEnum(129)])  # >> ['direction', 'enc_present]
```

slactjohnson: That's a good one-liner. I forgot about dictionary comprehensions. Sometimes my python ends up a bit C-ish. :)

Does `auto()` work here? I'm not familiar with it but from the docs it sounds like it just generates sequential numbers. The raw MSTA value that we're getting from the IOC is based on the mapping in the motor record, so the values matter, and I think we'd get off by one after the missing 4th bit.

pcdsdevices/epics_motor.py
silkenelson: Please make sure this work as expected. I seem to recall ( but am not 100% positive) that our very special IMS2 record as redefined the MSTA fields.

silkenelson: Similarly, I'd suggest to also take a look at the Newport motor record.

slactjohnson: @silkenelson You're right. The IMS record and the Newport record does redefine some of these bits. I didn't see anything odd in the MCS2 motor record. I'll look into how to deal with this on Monday. 

Vanilla Motor Record: 

image

Newport Motor Record: 

image

Latest IMS motor record:

Screenshot from 2024-07-21 09-34-46

pcdsdevices/epics_motor.py
ZLLentz: I think to use auto (and an int flag in general) you'd have to include a placeholder bit

pcdsdevices/epics_motor.py
tangkong: I think the most natural way to deal with this is to make different `MstaEnum`'s for different motor records.  `EpicsMotorInterface` might use the default one, then `pcdsdevices.epics_motor.Newport` gets its own, etc.

This is assuming we motor subclasses for these records.

tangkong: It looks like we have the following `EpicsMotorInterface` subclasses to consider from the `pcdsdevices` side of things.  (Of course we could have motors that don't have their own subclass)

A quick (turned not-so-quick) survey of the `EpicsMotorInterface` family tree

* `pcdsdevices.ccm.CCMPico`
* `pcdsdevices.epics_motor.PCDSMotorBase`
  * `pcdsdevices.epics_motor.IMS`
  * `pcdsdevices.epics_motor.Newport`
  * `pcdsdevices.epics_motor.DelayNewport` 
    * `pcdsdevices.lxe.TimeToolDelay`
    * `pcdsdevices.lxe._ReversedTimeToolDelay`
  * `pcdsdevices.epics_motor.PMC100`
  * `pcdsdevices.epics_motor.MMC100`
* `pcdsdevices.epics_motor.BeckhoffAxis`
  * `pcdsdevices.epics_motor.BechoffAxisEPS`
  * `pcdsdevices.epics_motor.BechoffAxisEPSCustom`
  * `pcdsdevices.epics_motor.BechoffAxis_Pre140` 
  * `pcdsdevices.epics_motor.BechoffAxisNoOffset`
* `pcdsdevices.epics_motor.SmarAct`


An interesting mapping we have:
```
    motor_types = (('MMS', IMS),
                   ('CLZ', IMS),
                   ('CLF', IMS),
                   ('MMN', Newport),
                   ('MZM', PMC100),
                   ('MMC', MMC100),
                   ('MMB', BeckhoffAxis),
                   ('PIC', PCDSMotorBase),
                   ('MCS', SmarAct),
                   ('MCS2', SmarAct))
```

slactjohnson: > I think the most natural way to deal with this is to make different `MstaEnum`'s for different motor records. `EpicsMotorInterface` might use the default one, then `pcdsdevices.epics_motor.Newport` gets its own, etc.
> 
> This is assuming we motor subclasses for these records.

I agree. This is the direction I've taken in my dev branch, just waiting to test.

pcdsdevices/epics_motor.py
slactjohnson: Now that I see the IMS and Newport IOCs have members that are larger than one bit, I think `auto()` will not work, or at least not easily or cleanly. I think I'd rather stick with an explicit mapping rather than implicitly making it with `auto()`.

tangkong: I think `IntFlag` still works if you define the bits manually, but I agree `auto()` won't quite work here.
```python
class MstaEnum(IntFlag):
    """
    Enum for the EPICS motor record .MSTA field bits.
    """
    # Note: the motor record docs start bit numbering at 1, but the MSTA bits
    # start at 0.
    direction = 1       # last raw move direction (0: negative, 1: positive)
    done = 2            # motion is complete
    plus_ls = 4        # plus limit switch is hit
    home_ls = 8         #
    ...
```

slactjohnson:
@silkenelson I've now set up separate enums and decoding for the IMS and Newport motors. It appears to work as expected (see the images in the PR description). 

Unfortunately, I wasn't able to find a Newport or IMS motor that was reporting an actual error _code_. Is there anything you can think of that would force an error code on one of these stages? Is there anything else you would like to see in terms of verifying functionality?

ZLLentz:
> Is there anything else you would like to see in terms of verifying functionality?

I think the screenshots in your description are sufficient for functionality

I'm somewhat curious about the other various motor types in the tree and if they build off of the standard MSTA or off of one more like IMS/Newport. I'm eternally frustrated by these "special" record types, this seems to bite us surprisingly frequently.

docs/source/upcoming_release_notes/1205-Add_MSTA_field_and_homed_status_to_EpicsMotorInterface.rst
```diff
@@ -0,0 +1,32 @@
...
+
+Features
+--------
+- MstaEnum: Enum describing the motor record MSTA bits
+
+Device Updates
+--------------
+- EpicsMotorInterface: Add a "raw" MSTA value, as well as the interpreted
+                       values as a dictionary. Adds a "homed" property based
+                       based on this.
```
ZLLentz, line 16: I think the only thing missing is including that you also did this separately for IMS and Newport

pcdsdevices/epics_motor.py
```diff
@@ -197,6 +285,26 @@ def _get_epics_limits(self) -> tuple[float, float]:
             return (0, 0)
         return limits
 
+    @property
+    def msta(self):
+        """
+        Returns the msta fields as a dictionary.
+        """
+        # the MSTA field is a float for some reason...
+        val = int(self.msta_raw.get())
```
ZLLentz, line 294: This is so incredibly puzzling to me. The manual says that it's supposed to be a ULONG but it's clearly a DOUBLE if you check any of the fields...

ZLLentz (approved):
I'm full :+1: on this implementation

pcdsdevices/epics_motor.py
tangkong: This isn't something we did specific to our records right?...

tangkong (approved):
Looks good to me, once again we document EPICS weirdness better in python

pcdsdevices/epics_motor.py
ZLLentz: It seems to be universal between standard records and the weird stuff we created here. I can't rule out a strange custom motor record build I guess.

ZLLentz:
I'm going to click "Update branch" so hopefully it's ready to merge when we look again tomorrow

ZLLentz:
CI failed on some network issue- running again before I sign off

pcdsdevices/epics_motor.py
slactjohnson: Looking at our latest motor record (R6.10.1-2.3.3), it's a DBF_ULONG in motorRecord.dbd as well as in e.g. ims.dbd, so I'm not sure why it's a float. 
image
image

slactjohnson:
Going to wait on merging for a little bit for @silkenelson to comment on the implementation or any other tests that should be done.

slactjohnson:
I'm going to go ahead and merge this because some of my current work is waiting on this update. We can revise later if needed.

PR #1260, opened 2024-07-24
deploying at2k4 gui

tongju12:
New axis in beam line called at2k4 GUI needs to be deploy

at2k4_tmo.py is added into pcdsdevices
## Motivation and Context

Scientists add one more axis at2k4(sample calibration foil) into beamline

## How Has This Been Tested?

Make a GUI and it looks good. 

python dev env test

https://jira.slac.stanford.edu/browse/ECS-4824
'

Screenshot 2024-07-24 at 2 43 32 PM

## Pre-merge checklist


pcdsdevices/at2k4_tmo.py
ZLLentz, line 1: What kind of device is this? Is it unique enough to sensibly give it it's own module? I assume this is an attenuator given the `at` in the name `at2k4`? https://confluence.slac.stanford.edu/display/PCDS/Photon+Systems+Naming+Conventions#PhotonSystemsNamingConventions-FunctionalComponent(FFFFF)

pcdsdevices/at2k4_tmo.py
```diff
@@ -0,0 +1,55 @@
...
+    name : str
+        Alias for the device
+    """
+    # UI Representation
+    _icon = 'fa.minus-square'
+    tab_component_names = True
+
+    # Sample calibration Y axe
+    at2k4 = Cpt(CalibrationAxe, 'AT2K4:IP1:STATE', add_prefix=(), kind='normal')
+    at2k4_y = Cpt(BeckhoffAxis, ':MMS', doc="Y-axis of sample caliration at2k4", kind='normal')
```
ZLLentz, line 41: We shouldn't mix styles here- either everything should be hard-coded or everything should flow from one or more common prefixes. This mixing only gives possibility for mistakes later when using the device.

pcdsdevices/at2k4_tmo.py
```diff
@@ -0,0 +1,55 @@
...
+
+    name : str
+        Alias for the device
+    """
+    # UI Representation
+    _icon = 'fa.minus-square'
+    tab_component_names = True
+
+    # Sample calibration Y axe
+    at2k4 = Cpt(CalibrationAxe, 'AT2K4:IP1:STATE', add_prefix=(), kind='normal')
```
ZLLentz, line 40: The subcomponent names are confusing to me. `at2k4` is a component of `at2k4`?

pcdsdevices/at2k4_tmo.py
```diff
@@ -0,0 +1,55 @@
...
+    # dummy signal, state is always the same
+    lightpath_cpts = ['at2k4.user_readback']
+
+    def calc_lightpath_state(self, **kwargs) -> LightpathState:
+        # TODO: get real logic here, instead of legacy hard-coding
+        return LightpathState(
+            inserted=True,
+            removed=False,
+            output={self.output_branches[0]: 1}
+        )
```
ZLLentz, line 55: Perhaps it's simple enough to include real lightpath handling on this device instead of "always in"?

pcdsdevices/at2k4_tmo.py
```diff
@@ -0,0 +1,55 @@
...
+from ophyd.device import Component as Cpt
+
+from .device import GroupDevice
+from .device import UpdateComponent as UpCpt
+from .epics_motor import BeckhoffAxis
+from .interface import BaseInterface, LightpathMixin
+from .pmps import TwinCATStatePMPS
+
+
+class CalibrationAxe(TwinCATStatePMPS):
```
tangkong, line 12: Do we mean `CalibrationAxis` here?

pcdsdevices/at2k4_tmo.py
```diff
@@ -0,0 +1,55 @@
...
+    at2k4 = Cpt(CalibrationAxe, 'AT2K4:IP1:STATE', add_prefix=(), kind='normal')
+    at2k4_y = Cpt(BeckhoffAxis, ':MMS', doc="Y-axis of sample caliration at2k4", kind='normal')
+    removed = False
+    transmission = 1
+    SUB_STATE = 'state'
+
+    # dummy signal, state is always the same
+    lightpath_cpts = ['at2k4.user_readback']
+
+    def calc_lightpath_state(self, **kwargs) -> LightpathState:
```
tangkong, line 49: We should make sure this is doing the right thing for lightpath.  If there is indeed an attenuator then we probably want to keep track of the blades to determine transmission and insertion state.  Maybe something similar to [AttenuatorSXR_Ladder](https://github.com/pcdshub/pcdsdevices/blob/c1bfa9cd54b4d875d043d12fa30ed3096d16cc25/pcdsdevices/attenuator.py#L1071)?

pcdsdevices/at2k4_tmo.py
tongju12: No, this is not as At1K4 like other attenuator, but due to the naming convention rule, TMO wants to call it at2k4 to make it normal.

pcdsdevices/at2k4_tmo.py
tongju12: let me change the name here. I will talk to scientists about the name.

pcdsdevices/at2k4_tmo.py
tongju12: Thanks Robert. No, this is not real attenuator like at1k4. no blades. Just a state mover with 5 targets on beamline with PMPS.

pcdsdevices/at2k4_tmo.py
ZLLentz: Hmm, that seems wrong to me (using `at` if this is not primarily used to attenuate the beam). Perhaps there's a better name to use, or a new category needs to be added... I think it's really bad to abuse the naming system and I don't think it should get this far that I'm confused about a device in a UI PR.

pcdsdevices/at2k4_tmo.py
ZLLentz: The main reason to change these is so the states row doesn't say `at2k4 at2k4` when we add the device to happi.

pcdsdevices/tmo_ip1.py
```diff
@@ -0,0 +1,50 @@
...
+    name : str
+        Alias for the device
+    """
+    # UI Representation
+    _icon = 'fa.minus-square'
+    tab_component_names = True
+
+    # Sample calibration Y axe
+    cali_foil = Cpt(CalibrationAxis, 'AT2K4:IP1:STATE', add_prefix=(), kind='normal')
+    cali_foil_y = Cpt(BeckhoffAxis, ':MMS', doc="Y-axis of sample caliration at2k4", kind='normal')
```
ZLLentz, line 37: My old comment about this is auto-marked as outdated now, but it's still relevant.

I don't like how one of these is hard-coded and the other is not.
In this case, I would suggest that they should both be hard-coded.

If only one is hard-coded, it will be possible to pick a wrong prefix and get a half-working device.

tongju12:
I will close this PR due to the new name approved by naming committee which is not at2k4 anymore. Will start another new one when I am ready. Thanks Zach and Robert help here!

PR #1267, opened 2024-08-01
ENH: Update Lcls2LaserTiming to have smaller timeout

### LINKED ISSUE #1266: Lcls2LaserTiming - Reduce Timeout
## Current Behavior
Currently ```Lcls2LaserTiming``` waits 2 seconds for a falling edge on the "moving" signal from the laser locker before marking movement done: https://github.com/pcdshub/pcdsdevices/blob/665c1f40c68894e28ae611bb7708b8f18cea403e/pcdsdevices/lxe.py#L485-L487

This is a bit of a kludge to deal with short time shifts. Since the ATCA common platform IOC only polls registers, and can't do I/O interrupts, we are limited to the polling rate of the PV. If the move takes less than this amount of time to complete, we may miss the edge entirely, hence the kludge.

## Context / environment
This has resulted in unnecessarily slow scans when moving small time steps. The beamline staff would like to speed this up. 
Example Jira: https://jira.slac.stanford.edu/browse/ECS-5056

## Suggested Solution
The original poll rate was 1Hz, which meant we waited 2 seconds to guarantee that we would see an edge if the IOC was able to catch it. I've update the polling rate to 10Hz, which means we can drop this down to 0.2 seconds wait.

slactjohnson:
## Description
Reduces the sleep timeout on ```Lcls2LaserTiming``` to be more appropriate for the current SCAN time. 

## Motivation and Context
closes #1266 

## How Has This Been Tested?
Tested very simply with a hutch python function using the bay 2 laser in the laser hall. The laser locker was setup for a phase shift velocity of 2 ns/s. This means a 1ns move should take ~0.5 seconds, a 0.1ns move ~0.05 seconds, and a 0.01ns move 0.005 seconds. 

Using the update in this PR, the 1ns move took ~0.5 seconds as expected, and the 0.1ns and 0.01ns moves timed out with 0.2 seconds, as expected. 

image

This PR and the associated Jira tickets. 

## Pre-merge checklist


ZLLentz:
I think this is a good idea :+1:

tangkong (approved):
Comments from the issue make this seem appropriate to me

ZLLentz (approved):

PR #1273, opened 2024-08-19
adding PA1K4-PF device in TMO  beamline

tongju12:
## Adding a new pcdsdevices in TMO beamline located at PA1K4 as photon filter

Adding a new device: PA1K4-PF
## Motivation and Context

Scientists require this device to work as calibrating photon energy

## How Has This Been Tested?

Run command to show the GUI. 

https://jira.slac.stanford.edu/browse/ECS-4824

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1273-adding_pa1k4-pf_device.rst
+++ b/docs/source/upcoming_release_notes/1273-adding_pa1k4-pf_device.rst
@@ -27,4 +27,4 @@ Maintenance
 
 Contributors
 ------------
-- @tongju12
+- tongju12
--- a/docs/source/upcoming_release_notes/1273-adding_pa1k4-pf_device.rst
+++ b/docs/source/upcoming_release_notes/1273-adding_pa1k4-pf_device.rst
@@ -1,0 +1,30 @@
+1273 adding pa1k4-pf device
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- N/A
+
+Device Updates
+--------------
+- N/A
+
+New Devices
+-----------
+- adding new device PA1K4-PF
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- @tongju12
--- a/pcdsdevices/tmo_ip1.py
+++ b/pcdsdevices/tmo_ip1.py
@@ -1,4 +1,4 @@
-"TMO sample calibration foil to work with spectrometer"
+"""TMO sample calibration foil to work with spectrometer"""
 from lightpath import LightpathState
 from ophyd.device import Component as Cpt
 
@@ -32,9 +32,9 @@ class SCaFoil(BaseInterface, GroupDevice, LightpathMixin):
     _icon = 'fa.minus-square'
     tab_component_names = True
 
-    # Sample calibration Y axe
+    # Sample calibration Y axis
 
-    pf = Cpt(CalibrationAxis, ':STATE', kind='normal')
+    pf_state = Cpt(CalibrationAxis, ':STATE', kind='normal')
     pf_y = Cpt(BeckhoffAxis, ':MMS:Y', doc="Y-axis of photo filter pa1k4", kind='normal')
     removed = False
     transmission = 1
--- a/pcdsdevices/tmo_ip1.py
+++ b/pcdsdevices/tmo_ip1.py
@@ -34,7 +34,7 @@ class SCaFoil(BaseInterface, GroupDevice, LightpathMixin):
 
     # Sample calibration Y axe
 
-    pf = Cpt(CalibrationAxis, 'PA1K4:PF:STATE', add_prefix=(), kind='normal')
+    pf = Cpt(CalibrationAxis, ':STATE', kind='normal')
     pf_y = Cpt(BeckhoffAxis, ':MMS:Y', doc="Y-axis of photo filter pa1k4", kind='normal')
     removed = False
     transmission = 1
--- a/pcdsdevices/tmo_ip1.py
+++ b/pcdsdevices/tmo_ip1.py
@@ -1,0 +1,51 @@
+"TMO sample calibration foil to work with spectrometer"
+from lightpath import LightpathState
+from ophyd.device import Component as Cpt
+
+from .device import GroupDevice
+from .device import UpdateComponent as UpCpt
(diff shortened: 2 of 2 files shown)


tongju12:
For the PV name, I try to hardcode pf_y from ":MMS:Y" to "PA1K4:PF:MMS:Y", however, after I run typhos command , it comes out as "PA1K4:PF:PA1K4:PF:MMS:Y", it automatically add the prefix"PF1K4:PF". Same thing, when I try to remove hardcode in pf from "PA1K4:PF:STATE" to ":STATE", after running command, it comes out as "STATE:GET_RBV", then I have to hardcode for pf. I am very confused how to fix the conflict here. So I have to use one line code is hard coded, the other line is not. If there is  a way to fix it, please let me know. thanks.

docs/source/upcoming_release_notes/1273-adding_pa1k4-pf_device.rst
```diff
@@ -0,0 +1,30 @@
...
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- @tongju12
```
tangkong, line 30:
```suggestion
- tongju12
```

tangkong:
Just to confirm what you're trying to do here with the prefixes, are you trying to have 
* `pf_y` always correspond to `PA1K4:PF:MMS:Y`, ignoring the prefix
* `pf` take the device prefix and become `<DEVICE:PREFIX>:STATE`?

Why don't we want both to take the device prefix?

pcdsdevices/tmo_ip1.py
```diff
@@ -0,0 +1,51 @@
...
+    name : str
+        Alias for the device
+    """
+    # UI Representation
+    _icon = 'fa.minus-square'
+    tab_component_names = True
+
+    # Sample calibration Y axe
+
+    pf = Cpt(CalibrationAxis, 'PA1K4:PF:STATE', add_prefix=(), kind='normal')
```
tangkong, line 37: You may know this, but by setting `add_prefix=()` here you stop the device prefix from being pre-pended to this component's PV.

Should this also be named something like `state`?  What does `pf` stand for here?

tongju12:
Yes, pf_y automatically takes prefix "PA1K4:PF" while pf did not take the prefix. Why we need to take prefix "PA1K4:PF"?  Because related PVs of this device are all involved "PA1K4:PF". Or can we put prefix as empty ""? I am still confused here.

pcdsdevices/tmo_ip1.py
tongju12: thanks Robert. I will push new code then. pf here is for photon filter,  PA1K4 is the location. It is the name committee decision:PA1K4:PF.

tangkong:
When you create an ophyd device, you normally provide a prefix.  This is because we assume that devices will be re-used with different prefixes.  
```python
my_foil = SCaFoil("MY:PREFIX:", name="my_foil")
your_foil = SCaFoil("OTHER:PREFIX:", name="your_foil")
```

This prefix gets pre-pended to the components automatically, creating the full PV.  As written in the code, `pf_y` takes this prefix as usual, but you have explicitly told `pf` to not do so.  If we set the prefix to "MY:PREFIX:" like I did in the previous snippet, we'll get
```
my_foil.pf --> "PA1K4:PF:STATE"
my_foil.pf_y --> "MY:PREFIX::MMS:Y"
```

If `SCaFoil.pf` will **ALWAYS** have the same PV (PA1K4:PF:STATE), this is fine.  But if you might ever make another foil and want to reuse this device, we should let `SCaFoil.pf` take the prefix as normal.

tongju12:
Thank you very much, Robert. I think this should be the last one to add because this device is right after the first mirror in TMO hutch, it should calibrate the beam for almost all of the devices after this is installed. I will keep the prefix here so far. If they will add another one later, I will come back to modify later.  Thank you very much for your suggestions though.

pcdsdevices/tmo_ip1.py
```diff
@@ -0,0 +1,51 @@
+"TMO sample calibration foil to work with spectrometer"
```
tangkong, line 1:
```suggestion
"""TMO sample calibration foil to work with spectrometer"""
```
Quick nitpick, docstrings at the top of a module use triple quotes.

pcdsdevices/tmo_ip1.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # dummy signal, state is always the same
+    lightpath_cpts = ['cali_foil.user_readback']
+
+    def calc_lightpath_state(self, **kwargs) -> LightpathState:
+        return LightpathState(
+            inserted=True,
+            removed=False,
+            output={self.output_branches[0]: 1}
+        )
```
tangkong, line 51: Right now the lightpath logic here says that the foil is always inserted, but fully transmissive.  This might be ok for now, but do we want to hash this out here?  It should probably be a transmission of 0 when it's inserted, and 1 when removed right?

Also as a side note, when we make the happi device for this we need to make sure the output branches are set.  Otherwise `self.output_branches` won't be assigned

pcdsdevices/tmo_ip1.py
```diff
@@ -0,0 +1,51 @@
...
+    prefix : str
+        Base PV for the motion system
+    name : str
+        Alias for the device
+    """
+    # UI Representation
+    _icon = 'fa.minus-square'
+    tab_component_names = True
+
+    # Sample calibration Y axe
```
tangkong, line 35:
```suggestion
    # Sample calibration Y axis
```
needless nitpicking here from me, sorry

pcdsdevices/tmo_ip1.py
```diff
@@ -0,0 +1,51 @@
...
+
+    pf = Cpt(CalibrationAxis, ':STATE', kind='normal')
```
tangkong, line 37: It's fine if this makes sense for you, but this might be better named as `pf_state` and `pf_y`.

tangkong (commented):
I think this is more or less ok to go.  I do have some nitpicks, as well as a question about our intention with the lightpath state

tongju12:
Thanks a lot, Robert! As we discussed, I edit several spots you mentioned here and leave light path as current status to deliver the GUI.

tangkong (approved):
Opting to get this in and follow up on lightpath behavior later.  Tagging will happen soon so it's best to not leave this hanging

PR #1276, opened 2024-08-21
BLD: use bluesky-base instead of bluesky(all)

tangkong:
## Description
Use bluesky-base in conda recipe instead of bluesky to avoid qt6-main

## Motivation and Context
NOTE: this will require a change in the conda-feedstock.  Without this other packages will fail by installing qt6 dependencies

## How Has This Been Tested?
N/A

This PR. 

## Pre-merge checklist

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -21,7 +21,7 @@ requirements:
     - setuptools_scm
   run:
     - python >=3.9
-    - bluesky >=1.6.4
+    - bluesky-base >=1.6.4
     - happi
     - jsonschema
     - lightpath >=1.0.0
@@ -45,7 +45,7 @@ test:
   requires:
     - pytest
     - pytest-timeout
-    - matplotlib
+    - matplotlib <3.9  # avoid installing qt6
     - typhos
   imports:
     - {{ import_name }}
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -45,7 +45,6 @@ test:
   requires:
     - pytest
     - pytest-timeout
-    - matplotlib <3.9  # avoid installing qt6
     - typhos
   imports:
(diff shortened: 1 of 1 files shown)


tangkong:
Confirming this prevents the installation of `qt6-main`, as compared to other CI runs (https://github.com/pcdshub/pcdsdevices/actions/runs/10475669888/job/29012841610#step:15:368)

conda-recipe/meta.yaml
```diff
@@ -45,7 +45,7 @@ test:
   requires:
     - pytest
     - pytest-timeout
-    - matplotlib
+    - matplotlib <3.9  # avoid installing qt6
```
ZLLentz, line 48: I don't this is appropriate as a CI test dependency, surely it still runs without it?
```suggestion
```

tangkong: There's this one pesky test... https://github.com/pcdshub/pcdsdevices/blob/75d68e72a1d4bcdedef6295c1ba4adec7474f25e/pcdsdevices/tests/test_interface.py#L65-L68

I can try removing it but I think this loads a backend

ZLLentz: oh sorry, I didn't realize this- but in this case, we need to specify pyqt5 in the test requires if it doesn't work, rather than pinning back matplotlib

ZLLentz (approved):

conda-recipe/meta.yaml
tangkong: uh it works.  I don't understand but I'm not going to fight the machine.  Maybe because lightpath pulls in qt dependencies or something i dunno

ZLLentz: it's worth noting that we need to render in offline/offscreen mode anyway for a CI build- it's not clear that qt would technically be needed with some configuration effort

PR #1281, opened 2024-09-03
BLD: depend on lightpath-base to avoid ui dependencies in builds

tangkong:
## Description
- Swaps `lightpath` -> `lightpath-base` in conda recipe
- Specifies >1.0.5 for both, to ensure pypi builds pull from version with GUI extras separated
  - Conda builds will fail until the feedstock updates, but I figured it's better to keep the version numbers the same

## Motivation and Context
Stop depending on things we don't need!

## How Has This Been Tested?
go CI go

This PR

## Pre-merge checklist

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -24,7 +24,7 @@ requirements:
     - bluesky-base >=1.6.4
     - happi
     - jsonschema
-    - lightpath >=1.0.0
+    - lightpath-base >=1.0.5
     - matplotlib-base
     - numpy
     - ophyd >=1.7.0
--- a/docs/source/upcoming_release_notes/1281-bld_lightpath_base.rst
+++ b/docs/source/upcoming_release_notes/1281-bld_lightpath_base.rst
@@ -1,0 +1,31 @@
+1281 bld_lightpath_base
+#######################
+
+API Breaks
+----------
+- N/A
+
+Library Features
+----------------
+- N/A
+
+Device Features
+---------------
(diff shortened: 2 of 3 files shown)


ZLLentz:
fwiw lightpath-base is currently only up to 1.0.4 on conda-forge so maybe this will need to wait until the tag flows through the CI chain

tangkong:
I believe the dependency to look for here is typhos.  (As in typhos should not be installed, as it is a gui-only dependency of lightpath)

tangkong:
A local pip-install successfully pulls in lightpath without gui dependencies, and the conda build here pulls in lightpath-base

I was sort of hoping to get pcdsdevices to build without qt as a dependency, but as long as typhos is a test-dependency we won't see it in CI.  I think it's probably easiest to test this locally after this is merged.  

I still think the changes here are correct, regardless of whether or not qt is fully extricated.

ZLLentz:
Confirmed that the changes look good
Confirmed that typhos is actually used in the test suite and can't be removed without stripping out some tests

Does this need a pre-release notes stub?

ZLLentz (approved):

ZLLentz:
fwiw the main thing typhos does in the test suite is verify that it can see our entrypoints, if you want to think a bit about how valuable or forgettable that might be.

tangkong:
I think the test suite requiring typhos is fine, I just wanted to make sure that normal installs don't pull in more dependencies than needed.  The CI just isn't the place to do that I think, and that's ok

PR #1282, opened 2024-09-03
FIX: ensure np.interp gets an increasing sequence in LUTs

ZLLentz:
## Description

- Fix an issue where the `LookupTablePositioner` would fail silently if the lookup table was not strictly increasing in both axes.
- Lookup tables that are strictly decreasing in either axis will now be supported.
- Lookup tables that have inconsistent ordering in either axis will not log a warning when the errant conversion is done.
- Refactor the `LookupTablePosition` class slightly to avoid code duplication

## Motivation and Context

`lxe` in XCS didn't work on master because the look up table was decreasing instead of increasing.
`np.interp` expects the first array to be strictly increasing.

https://numpy.org/doc/stable/reference/generated/numpy.interp.html

## How Has This Been Tested?

- Interactively in XCS
- Added unit tests

Pre-release notes and docstrings only

## Pre-merge checklist


pcdsdevices/pseudopos.py
```diff
@@ -1013,10 +1013,18 @@ def forward(self, pseudo_pos: tuple) -> tuple:
         pseudo_field, = self.PseudoPosition._fields
         real_field, = self.RealPosition._fields
 
+        xp = self._table_data_by_name[pseudo_field]
+        fp = self._table_data_by_name[real_field]
+
+        # xp must be increasing
+        if not xp[1] > xp[0]:
```
tangkong, line 1020: This is very intolerant of noise in the signal, are we confident we won't jitter enough to trigger this by accident?

ZLLentz: These come from fixed values in a text file that can look something like this:
```
50        71
55        41.5
60        18.8
65        6
70        1.3
75        0.5
```
Since they are fixed values there won't be any jitter, but I think it's worth adding a more thorough check so we can throw a warning if the values aren't monotonic

pcdsdevices/tests/test_pseudopos.py
```diff
@@ -216,21 +229,34 @@ class MyLUTPositioner(LookupTablePositioner):
...
-    np.testing.assert_allclose(lut.inverse(7)[0], 200)
-    np.testing.assert_allclose(lut.inverse(1.5)[0], 55)
-
-    lut.move(100, wait=True)
-    np.testing.assert_allclose(lut.pseudo.position, 100)
-    np.testing.assert_allclose(lut.real.position, 6)
-
-    assert lut.real.limits == (0, 9)
-    assert lut.pseudo.limits == (40, 400)
+    np.testing.assert_allclose(lut.forward(60 * ps)[0], 2 * rs)
```
tangkong, line 236: how did I not know `np.testing` existed...

pcdsdevices/pseudopos.py
```diff
@@ -1037,16 +1033,83 @@ def inverse(self, real_pos: tuple) -> tuple:
...
+
+        return xp, fp
+
+
+def is_strictly_increasing(arr: np.ndarray) -> bool:
+    """
+    Returns True if axis 1 of arr is strictly increasing and False otherwise.
+    """
+    # see numpy.interp docs https://numpy.org/doc/stable/reference/generated/numpy.interp.html
+    return np.all(np.diff(arr) > 0)
```
tangkong, line 1111: I guess monotonically increasing does mean strictly increasing (never decreasing or staying the same).  We just had the wrong idea of "monotonically increasing" before.

tangkong (approved):
this seems right to me.  Nice pseudopositioner tests.  I looked through the commit history and the spirit of stayed true throughout the various iterations

pcdsdevices/pseudopos.py
ZLLentz: I actually re-read the docs and found that they specified _strictly_ increasing instead of _monotonically_ increasing, the difference being that a monotonic sequence is allowed to repeat adjacent numbers or have a zero slope.

pcdsdevices/tests/test_pseudopos.py
ZLLentz: You know, I didn't actually _read_ the functions I was adding the sign swaps to, I was only looking for numbers and figuring out whether they were the real positions or the pseudo positions, so I also didn't realize `np.testing` existed.

My test here is just using the pre-existing test from Ken but adding the cases where either side is decreasing instead of increasing.

PR #1284, opened 2024-09-13
Add classes for cooling sensors

nrwslac:
## Description

- added `PhotonCollimatorFDQ` to pc.py
- added `BeckhoffPneumaticFDQ` to pneumatic.py
- add an instance of `FDQ` to `TMOSpectrometer` - happi prefix here was TMO:SPEC, so ended up making a pull request in tmo-motion to change the PV name: https://github.com/pcdshub/lcls-plc-tmo-motion/pull/132

## Motivation and Context

installed flow meters in TMO for ST4K4, SP1K4, and PC5K4.

## How Has This Been Tested?

- I loaded local typhos screens, all checked out good. TMO SPEC discrepancy resolved.  

https://jira.slac.stanford.edu/browse/ECS-5568

## Pre-merge checklist


pcdsdevices/pneumatic.py
```diff
@@ -96,3 +97,12 @@ def calc_lightpath_state(self, limit_switch_in=None, limit_switch_out=None):
             output={self.output_branches[0]: trans}
         )
         return status
+
+
+class BeckhoffPneumaticFDQ(BeckhoffPneumatic):
+    """
+    Beckhoff Pneumatics with a flow meter for cooling readback.
+
```
tangkong, line 105:
```suggestion
```

tangkong:
This needs some pre-release notes,

pcdsdevices/pc.py
```diff
@@ -4,12 +4,21 @@
...
 from ophyd import Device
 
+from .analog_signals import FDQ
 from .digital_signals import J120K
 
 
 class PhotonCollimator(Device):
     """
-    Photon Collimator Stub for Cooling Readback
+    Photon Collimator Stub for Cooling Switch Readback
```
tangkong, line 13: Is a stub referring to the type of device?  Or is it saying that this python class is a stub device (not really standard terminology, but it made me think of a class that's meant to be subclassed). 

Also these class names imply to me that `PhotonCollimatorFDQ` is a child of `PhotonCollimator`.  The `PhotonCollimatorFDQ` device doesn't seem to have any components related to a collimator at all, just the FDQ.  is there a better name fore this device?  `PCWFlowMeter`?

tangkong (commented):
This needs pre-release notes, then I had some questions about the naming of some of these device classes

tangkong:
Whoa my bad I did not mean to close this

nrwslac:
> This needs some pre-release notes,
> 
> Edit: this wayward comment must have closed the PR 🤔

Whoops, this PR fell behind, didn't notice the missing docs: good catch.

pcdsdevices/pc.py
nrwslac: The reason I used the word stub is because I've created this photon collimator class with nothing in it other than the flow sensors readout; it's superfluous. I will delete it. I agree that it sounds inherited, but its not. I can think of one good reason here to actually inherit in FDQ; if later someone wants to add functionality, they can update the base `PhotonCollimator` class only.

tangkong (approved):
I think this makes more sense now 👍  Thanks for putting up with my nitpicks

PR #1292, opened 2024-10-10
recover sp1k4-att target from 3 to 5

tongju12:
SP1K4-ATT targets change from three targets to five targets so I need to update the devices

Change state count from 4 to 6
## Motivation and Context

Scientists add target 4 and 5 back after swapping hardware

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1292-recoving_sp1k4-att_targets.rst
+++ b/docs/source/upcoming_release_notes/1292-recoving_sp1k4-att_targets.rst
@@ -1,0 +1,30 @@
+1292 recoving sp1k4-att targets
+#################
+
+API Breaks
+----------
+- N/A
+
+Library Features
+----------------
+- N/A
+
+Device Features
+---------------
+- recover target number from 3 to 5
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- tongju12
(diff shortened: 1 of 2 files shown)


tangkong (approved):
Seems simple enough.  Just to verify, this is the configuration that's currently in the beamline now right?  

Is this something we expect to change often?  If so we can investigate a more flexible way to set these states.

tongju12:
Currently, this is a temporary paddle. The real one may come back later(6 months at least?), so let's wait to see how it goes.

docs/source/upcoming_release_notes/1292-recoving_sp1k4-att_targets.rst
```diff
@@ -0,0 +1,30 @@
...
+----------
+- N/A
+
+Library Features
+----------------
+- N/A
+
+Device Features
+---------------
+- recover target number from 3 to 5
```
ZLLentz, line 14: I think this doc shouldn't pass review, it will be missing context when combined with the other release notes. I'll try to remember that this one comes from sp1k4 when we do the pcdsdevices tag.

tongju12: Thanks Zach. I can upload again for more details" recover sp1k4-att target number from 3 to 5".

tangkong: whoops sorry, I was focused on the code changes.  That's my bad

PR #1293, opened 2024-10-10
ENH/FIX: apply TMO's mpod apalis hotfixes, adding features and fixing a major bug

ZLLentz:
## Description

Add the following components to `MPODApalisChannel`:

- `desc` signal: Description of the mpod channel, used to configure what gets showed on the screen and to check the description during experiments.
- `last_voltage_set` signal: The setpoint of the MPOD channel, again. When we `put` to the `voltage` signal it also uses the setpoint PV, but having this as a separate signal lets you quickly see what the last value set was, and compare it to the apparent readback voltage of the channel. This is aliased to `voltage_setpoint` for some dev-time backwards compatibility for tmo. The new name is more clear.
- `is_trip` signal: `True` if the channel is tripped, which is helpful for some recovery automation and error reporting.
- Add long docstrings in various places to explain how signals are used.

Fix a bug:
- Voltage set bounding was implemented assuming voltages were always positive.
- One thing this did was allow unbounded voltage sets in the negative direction.
- Another thing this did was cause any voltage sets to negative-polarity channels to be immediately jumped to the most-negative possible value.
- Now, the voltage and current signals will have proper limits for error reporting and the `set_voltage` and `set_current` utilities will be properly clamped to the valid range.

## Motivation and Context

TMO's hutch-python has these updates in `mpod_apalis.py`.
It's not clear who originally implemented these, but they are used by tmo.

Some time recently, these were accidentally removed.
So, let's formally introduce these into pcdsdevices proper.

One of these is a major bugfix and the others are niceties.

I added some additional changes on top of these as well as associated documentation and unit test updates.

## How Has This Been Tested?

The original hotfix was tested interactively at TMO.
The updated hotfix was tested interactively by me and the new unit tests are now passing.

Here only, in the code and in the pre-release notes.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/mpod_apalis.py
+++ b/pcdsdevices/mpod_apalis.py
@@ -45,8 +45,16 @@ class MPODApalisChannel(BaseInterface, Device):
     desc = Cpt(EpicsSignal, ':VoltageMeasure.DESC', kind='normal',
                doc='MPOD Channel Description')
 
-    voltage_setpoint = Cpt(EpicsSignalRO, ':VoltageSet', kind='normal',
-                           doc='MPOD Channel Voltage Setpoint [V]')
+    last_voltage_set = Cpt(
+        EpicsSignalRO, ':VoltageSet', kind='normal',
+        doc=(
+            'Readback to verify the MPOD Channel Voltage Setpoint [V]. '
+            'This is used to compare the measured readback voltage with '
+            'the last value we set to the channel. '
+            'To change the voltage, use the voltage signal or the '
+            'set_voltage helper method.'
+        )
+    )
 
     is_trip = Cpt(EpicsSignalRO, ':isTrip', kind='omitted',
                   doc='True if MPOD channel is tripped.')
@@ -54,6 +62,11 @@ class MPODApalisChannel(BaseInterface, Device):
     tab_component_names = True
     tab_whitelist = ['on', 'off',
                      'set_voltage', 'set_current']
+
+    @property
+    def voltage_setpoint(self) -> EpicsSignalRO:
+        """Name alias for backwards compatibility."""
+        return self.last_voltage_set
 
     def on(self):
         """Set mpod channel On."""
--- a/pcdsdevices/mpod_apalis.py
+++ b/pcdsdevices/mpod_apalis.py
@@ -6,6 +6,7 @@ from ophyd.device import Device
 from ophyd.signal import EpicsSignal, EpicsSignalRO
 
 from pcdsdevices.interface import BaseInterface
+from pcdsdevices.signal import EpicsSignalEditMD
 
 from .device import GroupDevice
 
@@ -18,22 +19,35 @@ class MPODApalisChannel(BaseInterface, Device):
 
     Parameters
     ----------
-    channel_prefix : str
+    prefix : str
         The EPICS base of the MPOD Channel, e.g. 'TMO:MPOD:01:M6:C6'.
-    name : str
+    name : str, keyword-only
         A name to refer to the device.
-
     """
-    voltage = Cpt(EpicsSignal, ':VoltageMeasure',
-                  write_pv=':VoltageSet', kind='normal',
-                  doc='MPOD Channel Voltage Measurement [V]')
+    voltage = Cpt(
+        EpicsSignalEditMD, ':VoltageMeasure',
+        write_pv=':VoltageSet', kind='normal', limits=True,
+        doc=(
+            "MPOD Channel Voltage Measurement [V]. "
+            "The value of this signal is the actual measured voltage "
+            "of the channel. If you put to this signal it will change "
+            "the channel's voltage setpoint."
+        ),
+    )
 
     max_voltage = Cpt(EpicsSignalRO, ':VoltageNominal', kind='normal',
                       doc='MPOD Channel Maximum Voltage [V]')
 
-    current = Cpt(EpicsSignal, ':CurrentMeasure',
-                  write_pv=':CurrentSet', kind='normal',
-                  doc='MPOD Channel Current Measure')
+    current = Cpt(
+        EpicsSignalEditMD, ':CurrentMeasure',
+        write_pv=':CurrentSet', kind='normal', limits=True,
+        doc=(
+            "MPOD Channel Current Measurement [A]. "
+            "The value of this signal is the actual measured current "
+            "of the channel. If you put to this signal it will change "
+            "the channel's current setpoint."
+        ),
+    )
 
     max_current = Cpt(EpicsSignalRO, ':CurrentNominal', kind='normal',
                       doc='MPOD Channel Current Maximum')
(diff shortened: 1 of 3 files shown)


pcdsdevices/mpod_apalis.py
```diff
@@ -64,7 +73,7 @@ def set_voltage(self, voltage_number):
         """
         max_voltage = self.max_voltage.get()
 
-        if voltage_number <= max_voltage:
+        if abs(voltage_number) <= abs(max_voltage):
             self.voltage.put(voltage_number)
         else:
             self.voltage.put(max_voltage)
```
ZLLentz, line 79: Self review: I notice now that this fallback case might not be quite what we want.
I will double-check what `max_voltage` looks like for the TMO devices, hopefully it's negative for channels with negative polarity. Otherwise, this will need to be revised.

ZLLentz: I've confirmed that `max_voltage` is negative for the negative-polarity channels. So this is probably OK but still feels a bit wonky (e.g. how are we supposed to interpret positive numbers put to negative channels?)

slactjohnson: Should we be bounding the set voltage in the IOC, rather than in python? The DRVH and DRVL fields seem like the right place to do this. But, I'm not familiar with that IOC, so perhaps it's more complicated than that.

ZLLentz: I have the same opinion, it's much better to do these in the IOC. I am also unfamiliar with the IOC though.

ZLLentz: I think I'm going to make a jira ticket targeting the IOC feature, referencing this PR. Then, I'm going to complete this PR and ask for reviews.

ZLLentz: https://jira.slac.stanford.edu/browse/ECS-6495

ZLLentz: I thought about this some more- maybe this isn't about clamping (maybe the IOC does this already?), but perhaps it's about the case where:
- `max_voltage` is -3000
- call `set_voltage(-100)`
- -100 is not less than -3000
- therefore, put max_voltage of -3000

And with the revised logic:
- `max_voltage` is -3000
- call `set_voltage(-100)`
- 100 is less than 3000
- therefore, put 100

ZLLentz: It's not immediately obvious from the IOC since the control limit fields are not set, but it's still possible that these are implemented internally in some roundabout way. I'll include a note on the jira.

tangkong (commented):
As I understand it, this is the change that is deployed, and this change is accepted by the users?

This is a little strange to me since there's only a max voltage, but no min voltage.  Are we assured that `voltage_number` will always be the same sign as `max_voltage`?  

The PV being named "nominal voltage" is misleading, and the implementation in this device class is also strange.  Zach and I are hashing out our confusion offline, and I think we need to talk with the owners of this IOC to make sure we're not setting improper voltages

pcdsdevices/mpod_apalis.py
```diff
@@ -64,7 +73,7 @@ def set_voltage(self, voltage_number):
         """
         max_voltage = self.max_voltage.get()
 
-        if voltage_number <= max_voltage:
+        if abs(voltage_number) <= abs(max_voltage):
             self.voltage.put(voltage_number)
```
slactjohnson, line 77: This code isn't from you, but I am very confused here. Based on the PV, the `voltage` is a measured value. Shouldn't this be putting a value to `voltage_setpoint` ... which is `EpicsSignalRO` for some reason?

ZLLentz:
I am OK with not merging this soon or with any sort of urgency at all. It's most important to document the change and understand if we're doing the right thing to avoid redoing anything.

pcdsdevices/mpod_apalis.py
slactjohnson: Re-reading the PR description makes the last part more clear, but the PV naming in this IOC seems very odd.

pcdsdevices/mpod_apalis.py
```diff
@@ -42,6 +42,15 @@ class MPODApalisChannel(BaseInterface, Device):
                 kind='normal', string=True,
                 doc='MPOD Channel State [Off/On]')
 
+    desc = Cpt(EpicsSignal, ':VoltageMeasure.DESC', kind='normal',
+               doc='MPOD Channel Description')
+
+    voltage_setpoint = Cpt(EpicsSignalRO, ':VoltageSet', kind='normal',
```
slactjohnson, line 48: Since this is the last written setpoint, maybe we could rename this to `last_voltage_setpoint` or something to indicate that it's not the actual control value.

ZLLentz:
The IOC is https://github.com/pcdshub/ioc-common-isegPower

The TMO IOCs are in the children folder there
Everyone who has wrote code for this IOC has either left SLAC or is semi-retired

pcdsdevices/mpod_apalis.py
ZLLentz: I'll check tmo's scripts, and if they don't use this by name I will rename it.
If they do use it by name, I will update the description field instead.

ZLLentz: I'm going to rename this as suggested but include a backcompatibility alias for tmo's scripts to keep working
```
tmo/tmo_mpods.py:                mpod_dict["Voltage Set (V):"].append(mpod.voltage_setpoint.get())
tmo/tmo_cvmi_utility.py:                self.voltage_settings[key] = mpod.voltage_setpoint.get()
tmo/tmo_cvmi_utility.py:            while ( tmo_mpod.MCP_Back.voltage.get() < (0.9 * tmo_mpod.MCP_Back.voltage_setpoint.get() ) ):
tmo/tmo_cvmi_utility.py:            while ( tmo_mpod.Phosphor.voltage.get() < (0.9 * tmo_mpod.Phosphor.voltage_setpoint.get() ) ):
```

pcdsdevices/mpod_apalis.py
ZLLentz: I think what I'll end up doing here is:
- add ctrl limits to the signal (not present in the IOC) to give a clear error when someone tries to put a voltage outside the range to the signal directly
- clamp values in a more straightforward way in `set_voltage` and have this give a warning message
- allow an IOC investigation to be done as future work

ZLLentz:
Ok I renamed the signal, revised the logic, expended the docs, and added unit test cases- it no longer looks so much like the original hotfix but it should hopefully be less confusing and just as (or more) functional.

ZLLentz:
Before merge I still need to check it on a real IOC, I'll do this on psbuild and run through the logic until I hit write access errors

slactjohnson (dismissed):
LGTM. I like this implementation.

tangkong (dismissed):
This certainly goes above and beyond the original assignment, good work.

ZLLentz:
Works as expected with a real IOC (with write permission errors in the cases it should be writing)

Remaining:
- Fix minor typos in docstrings (re-request review)
- Update original PR text
- Merge
- Weekend

PR #1297, opened 2024-10-11
Add triggered signal

mseaberg:
## Description

Subclassed AvgSignal in order to allow averaging of a signal within a bluesky scan.

## Motivation and Context

Sometimes there are cases where one wants to average a signal before moving to the next scan point.

## How Has This Been Tested?

Tested in xcs3 using 'XCS:SND:DIO:AMPL_8' wave8 signal, with the following test:
```
d1 = AvgSignalTriggered(EpicsSignalRO('XCS:SND:DIO:AMPL_8'),120,1,name='d1')
RE(scan([d1],sim.fast_motor1,0,120,num=11))

Transient Scan ID: 1     Time: 2024-10-11 12:23:24
Persistent Unique Scan ID: '567b8720-03ad-4ed6-a837-5effa5ac3058'
New stream: 'primary'
+-----------+------------+-------------+------------+
|   seq_num |       time | fast_motor1 |         d1 |
+-----------+------------+-------------+------------+
|         1 | 12:23:25.8 |           0 |      0.833 |
|         2 | 12:23:26.8 |          12 |       0.21 |
|         3 | 12:23:27.8 |          24 |     -0.371 |
|         4 | 12:23:28.8 |          36 |     0.0688 |
|         5 | 12:23:29.8 |          48 |      0.279 |
|         6 | 12:23:30.8 |          60 |     -0.285 |
|         7 | 12:23:31.8 |          72 |      -0.29 |
|         8 | 12:23:32.8 |          84 |      0.131 |
|         9 | 12:23:33.9 |          96 |     -0.152 |
|        10 | 12:23:34.9 |         108 |      0.635 |
|        11 | 12:23:35.9 |         120 |       -0.2 |
+-----------+------------+-------------+------------+
generator scan ['567b8720'] (scan num: 1)
```
## Where Has This Been Documented?

AvgSignalTriggered docstring

## Pre-merge checklist
- [ ] Code works interactively
- [ ] Code contains descriptive docstrings, including context and API
- [ ] New/changed functions and methods are covered in the test suite where possible
- [ ] Test suite passes locally
- [ ] Test suite passes on GitHub Actions
- [ ] Ran docs/pre-release-notes.sh and created a pre-release documentation page
- [ ] Pre-release docs include context, functional descriptions, and contributors as appropriate

PR #1298, opened 2024-10-11
ENH: merge AvgSignalTriggered into AvgSignal

ZLLentz:
## Description

- Take the features implemented by @mseaberg in #1297 and pull them into the parent class.
- Refactor the original class slightly to make `reset_buffer` usable directly.
- Fix some old documentation issues and add type annotations
- Add unit tests

## Motivation and Context

- The features in #1297 are useful and fit seamlessly into the main class, this consolidates the similar features in one place.
- Avoids the small oddity of setting the average count back to itself in order to reset the buffer
- I like type annotations
- Unit tests protect us against regressions

## How Has This Been Tested?

- I added unit tests
- I redid the tests Matt did in xcs in #1297 and had similar results.

- I added release notes

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/signal.py
+++ b/pcdsdevices/signal.py
@@ -27,7 +27,7 @@ from ophyd.device import Device
 from ophyd.signal import (DEFAULT_WRITE_TIMEOUT, DerivedSignal, EpicsSignal,
                           EpicsSignalBase, EpicsSignalRO, Signal, SignalRO)
 from ophyd.sim import FakeEpicsSignal, FakeEpicsSignalRO, fake_device_cache
-from ophyd.status import StatusBase
+from ophyd.status import Status
 from ophyd.utils import ReadOnlyError
 from pytmc.pragmas import normalize_io
 
@@ -869,13 +869,13 @@ class AvgSignal(Signal):
             # This takes a mean, skipping nan values.
             self.put(np.nanmean(self.values))
 
-    def trigger(self):
+    def trigger(self) -> Status:
         if self.duration is None:
             return super().trigger()
         # reset the averaging buffer
         self.reset_buffer()
         # set status to complete after duration
-        status = StatusBase(settle_time=self.duration)
+        status = Status(obj=self, settle_time=self.duration)
         status.set_finished()
         return status
 
--- a/pcdsdevices/signal.py
+++ b/pcdsdevices/signal.py
@@ -21,6 +21,7 @@ from typing import Any, Generator, Mapping, Optional, Union
 
 import numpy as np
 import ophyd
+from ophyd.device import Device
 from ophyd.signal import (DEFAULT_WRITE_TIMEOUT, DerivedSignal, EpicsSignal,
                           EpicsSignalBase, EpicsSignalRO, Signal, SignalRO)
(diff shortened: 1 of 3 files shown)


pcdsdevices/signal.py
```diff
@@ -794,97 +802,80 @@ class AvgSignal(Signal):
...
         self.raw_sig.subscribe(self._update_avg)
 
     @property
-    def connected(self):
+    def connected(self) -> bool:
         return self.raw_sig.connected
 
     @property
-    def averages(self):
+    def averages(self) -> int:
```
tangkong, line 845: I dislike this name but I dislike breaking people's functionality more

pcdsdevices/tests/test_signal.py
```diff
@@ -60,6 +60,34 @@ def test_avg_signal():
     assert cb.called
 
 
+def test_avg_signal_with_duration():
+    logger.debug("test_avg_signal_with_duration")
+    sig = Signal(name='raw')
+    avg = AvgSignal(sig, 2, name='avg', duration=2)
```
tangkong, line 66: It took me squinting at this and re-reading the code to realize that the duration only does anything when calling `.trigger()`.  (this could be obvious to most people, but I'm glad there were tests to scrutinize.

pcdsdevices/tests/test_signal.py
```diff
@@ -60,6 +60,34 @@ def test_avg_signal():
...
+    sig.put(1)
+    sig.put(3)
+    assert avg.get() == 2
+
+    # trigger should clear the buffer and not take the full duration to return
+    start_time = time.monotonic()
+    st = avg.trigger()
+    end_time = time.monotonic()
+    assert end_time - start_time < avg.duration
+    assert not st.done
```
tangkong, line 79: Is there ever a concern that the runner hardware is so bad that the trigger finishes before the puts are done?
2 seconds seems sufficient, but I'm left wondering if there's a better way to ensure it's never a problem.

I can't think of anything better after about a minute of navel gazing

tangkong (approved):
I nitpicked the tests, but I couldn't come up with a constructive suggestion so this gets my green ✅

pcdsdevices/signal.py
ZLLentz: In retrospect, I also really dislike this old name!

pcdsdevices/tests/test_signal.py
ZLLentz: I think maybe this isn't so obvious in general. If you ponder for a bit you're realize that the class is continually accumulating values, so a time duration must only apply in some specific circumstances. I tried to document it appropriately.

pcdsdevices/tests/test_signal.py
ZLLentz: I think there will always be some fundamental problem wrt unit tests when coding something with a duration.
Maybe some alternate formulation would wait until the buffer is full instead of a fixed time interval, and then this is much easier to make consistent in the test suite.

I could make the duration longer but then we're hitting tradeoffs with test suite speed.

ZLLentz: Also to clarify: duration is used in this test to:
- Verify that the trigger returns quickly, instead of waiting
- Verify that the final status after waiting the duration doesn't raise

PR #1301, opened 2024-10-17
ENH: las timing configurables (cxi3 hotfix)

ZLLentz:
## Description

- Add cxi3's `invert` argument for `LaserTiming` that allows them to un-invert their laser timing device.
- Add init-time configuration of the `LaserTiming` limits so cxi3 doesn't need to override the class limits
- Add the above to `Lcls2LaserTiming` too while I'm here
- Ignore cxi3's other hotfix- it isn't actually needed to achieve these

## Motivation and Context

- Apply cxi3's hotfix so they don't have to run out of dev in the future
- Add limit handling for the same reason

## How Has This Been Tested?

- I made unit tests

- I made pre-release notes

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1301-enh_config_las_timing.rst
+++ b/docs/source/upcoming_release_notes/1301-enh_config_las_timing.rst
@@ -1,0 +1,31 @@
+1301 enh_config_las_timing
+##########################
+
+API Breaks
+----------
+- N/A
+
+Library Features
+----------------
+- N/A
+
+Device Features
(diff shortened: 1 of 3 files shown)


slactjohnson (approved):
This looks good to me. The only thing I can think of is doing some checking to make sure `limits` is properly formed, but I'm not sure if we do that kind of checking with any other motors or pseudomotors.

tangkong (approved):
LGTM

PR #1303, opened 2024-10-21
PPM Screen Changes

KaushikMalapati:
## Description

Changing responsivity io to be input only so it displays without a connection error
Renaming manual pvs to be less similar to auto background pvs
Adding microjoule mirror of calibrated_mj
Added manual collection in progress indicator
Replaced references to passive mode with auto mode

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-875

## How Has This Been Tested?

Interactively

upcoming-release notes

image

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/pim.py
+++ b/pcdsdevices/pim.py
@@ -374,8 +374,6 @@ class PPMPowerMeter(BaseInterface, Device):
 
     `calibrated_mj` = (Signal - Background) / (Responsivity * Beam_Rate)
     """
-    def __init__(self, prefix, *, name, **kwargs):
-        super().__init__(prefix, name=name, **kwargs)
 
     tab_component_names = True
 
--- a/pcdsdevices/pim.py
+++ b/pcdsdevices/pim.py
@@ -400,6 +400,7 @@ class PPMPowerMeter(BaseInterface, Device):
 
     manual_in_progress = Cpt(PytmcSignal, ':BACK:MANUAL_COLLECTING', io='i', kind='normal',
                              doc='Manual collection currntly in progress')
+    set_metadata(manual_in_progress, dict(variety='command'))
 
     manual_collect_time = Cpt(PytmcSignal, ':BACK:TIME', io='io', kind='normal',
                               doc='Time to collect background voltages for.')
--- a/pcdsdevices/pim.py
+++ b/pcdsdevices/pim.py
@@ -397,10 +397,10 @@ class PPMPowerMeter(BaseInterface, Device):
 
     background_mode = Cpt(PytmcSignal, ':BACK:MODE', io='io', kind='normal', doc='Can be manual or auto In manual mode, you can collect for a specified number of seconds. In auto mode, a buffer of automatically collected background voltages will be used to calculate the background voltage.')
 
-    background_collect = Cpt(PytmcSignal, ':BACK:COLL', io='io', kind='normal', doc='Start collecting background voltages for specified time.')
-    set_metadata(background_collect, dict(variety='command-proc', value=1))
-
-    background_collect_time = Cpt(PytmcSignal, ':BACK:TIME', io='io', kind='normal', doc='Time to collect background voltages for.')
+    manual_collect = Cpt(PytmcSignal, ':BACK:COLL', io='io', kind='normal', doc='Start collecting background voltages for specified time.')
+    set_metadata(manual_collect, dict(variety='command-proc', value=1))
+    manual_in_progress = Cpt(PytmcSignal, 'BACK:MANUAL_COLLECTING', io='i', kind='normal', doc='Manual collection currntly in progress')
+    manual_collect_time = Cpt(PytmcSignal, ':BACK:TIME', io='io', kind='normal', doc='Time to collect background voltages for.')
 
     raw_voltage = Cpt(PytmcSignal, ':VOLT', io='i', kind='normal',
                       doc='Raw readback from the power meter.')
--- a/pcdsdevices/pim.py
+++ b/pcdsdevices/pim.py
@@ -388,19 +388,30 @@ class PPMPowerMeter(BaseInterface, Device):
 
     tab_component_names = True
 
-    responsivity = Cpt(PytmcSignal, ':RES', io='i', kind='normal', doc='Responsivity in  V/W, unique for every power meter.')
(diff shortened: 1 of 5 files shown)


pcdsdevices/pim.py
```diff
@@ -372,32 +372,43 @@ class PPMPowerMeter(BaseInterface, Device):
...
     is used to calculate the actual energy in the following way:
 
-    `calibrated_mj` = 1000 * (Signal - Background) / (Responsivity * Beam_Rate)
-    """
+    `calibrated_mj` = (Signal - Background) / (Responsivity * Beam_Rate)
+    """
+    def __init__(self, prefix, *, name, **kwargs):
+        import sympy.physics.units as units
+        from sympy.physics.units.prefixes import micro, milli
+        from sympy.physics.units.quantities import Quantity
```
tangkong, line 380: We should avoid importing modules inside of class instance methods, unless there's a very specific reason not to.  This is a base class that gets inherited in many later devices, and we'll be re-importing these modules each time one of these classes is instantiated

pcdsdevices/pim.py
```diff
@@ -372,32 +372,43 @@ class PPMPowerMeter(BaseInterface, Device):
...
+    responsivity = Cpt(PytmcSignal, ':RES', io='i', kind='normal', doc='Responsivity in  V/W, unique for every power meter.')
 
     background_voltage = Cpt(PytmcSignal, ':BACK:VOLT', io='io', kind='normal', doc='Background voltage value used to calculate pulse energy.')
 
     auto_background_reset = Cpt(PytmcSignal, ':BACK:RESET', io='io', kind='normal', doc='Set to reset auto background voltage collection.')
     set_metadata(auto_background_reset, dict(variety='command-proc', value=1))
 
-    background_mode = Cpt(PytmcSignal, ':BACK:MODE', io='io', kind='normal', doc='Can be manual or passive. In manual mode, you can collect for a specified number of seconds. In passive mode, a buffer of automatically collected background voltages will be used to calculate the background voltage.')
+    background_mode = Cpt(PytmcSignal, ':BACK:MODE', io='io', kind='normal', doc='Can be manual or auto In manual mode, you can collect for a specified number of seconds. In auto mode, a buffer of automatically collected background voltages will be used to calculate the background voltage.')
```
tangkong, line 398: Can we break up some of these lines, they are very long

pcdsdevices/pim.py
```diff
@@ -372,32 +372,43 @@ class PPMPowerMeter(BaseInterface, Device):
...
+    """
+    def __init__(self, prefix, *, name, **kwargs):
+        import sympy.physics.units as units
+        from sympy.physics.units.prefixes import micro, milli
+        from sympy.physics.units.quantities import Quantity
+        millijoule = mJ = Quantity("millijoule", abbrev='mJ')
+        millijoule.set_global_relative_scale_factor(milli, units.joule)
+        microjoule = uJ = Quantity("microjoule", abbrev='uJ')
+        microjoule.set_global_relative_scale_factor(micro, units.joule)
+        setattr(units, 'mJ', mJ)
```
tangkong, line 385: So what we're doing here is manually adding a custom unit to the `sympy.physics.units` module?  This is almost certainly now how sympy is meant to be used, and import order can make this inconsistent.

The definition of these should probably go into `pcdsdevices.utils`, with maybe a modification of how we acquire units (First look for native sympy units, then look for our custom defined units)

pcdsdevices/pim.py
KaushikMalapati: I removed all of this from the class

KaushikMalapati:
@tangkong, do you know if there's a way to remove the lineedit by calibrated_uj and the two circle indicators by the command buttons for auto_background_reset and manual_collect? calibrated_uj already has write_access=False and the two command indicators are not useful because they are reset to 0 on the cycle when they are set to one, so the screen never gets updates the circles. These would be purely cosmetic changes.

tangkong:
Not to my knowledge, as far as I know the autogenerated screens don't let you pick and choose which parts of the widget get displayed.  You could make a custom template for your screens, but that does seem like overkill

tangkong (commented):
I like how this turned out, I think the implementation of our custom units turned out quite clean.

Do you feel up to writing a quick unit test for the new `convert_unit` function?  You could make a new file (`test_units.py`) and just write a test that converts between mJ, uJ, and J.  We don't have a test for this yet and it'll be good to make sure this function stays functional

pcdsdevices/pim.py
```diff
@@ -372,24 +372,38 @@ class PPMPowerMeter(BaseInterface, Device):
     calibrated for each power meter in units of volts per watt
     is used to calculate the actual energy in the following way:
 
-    `calibrated_mj` = 1000 * (Signal - Background) / (Responsivity * Beam_Rate)
+    `calibrated_mj` = (Signal - Background) / (Responsivity * Beam_Rate)
     """
+    def __init__(self, prefix, *, name, **kwargs):
+        super().__init__(prefix, name=name, **kwargs)
```
tangkong, line 378: Just realizing this is here, we don't really need this init do we?

KaushikMalapati: No

KaushikMalapati:
Changed the unit test to use the long unit names and added a shorter one for sympy units

tangkong (approved):

KaushikMalapati:
@tangkong Can I git pull origin master in /cds/group/pcds/pyps/apps/dev/pcdsdevices to update pcdsdevices for dev_conda?

KaushikMalapati:
> > a way to remove the lineedit by calibrated_uj and the two circle indicators by the command buttons for auto_background_reset and manual_collect? calibrated_uj already has write_access=False and the two command indicators are not useful because they are reset to 0 on the cycle when they are set to one, so the screen never gets updates the circles. These would be purely cosmetic changes.
> 
> Would you find it useful for us to add a way to do this using https://github.com/pcdshub/pcdsdevices/blob/master/pcdsdevices/variety.py and https://github.com/pcdshub/typhos/blob/master/typhos/variety.py ?
> 
> Hypothetically we can come up with some variety tag as a message for typhos that indicates that we don't want circle indicators on our commands etc.

I don't think it's worth doing since it'd be pretty specific. I could always make a custom ui for it but I think leaving the indicator there is fine.

ZLLentz:
Ok, no worries. I figured this might be a more common ask.

Issue #1308, opened 2024-11-18
Add acquire rate control on PDCSAreaDetectorTyphos open_viewer method


aberges-SLAC:
Currently, opening cam viewer on the LUCID screen for Basler gigecams runs the following method:
```
def open_viewer(self):
        """
        Launch the python camera viewer for this camera.
        """
        arglist = ['/reg/g/pcds/pyps/apps/camviewer/latest/run_viewer.sh',
                   '--instrument',
                   '{}.format(get_hutch_name())',
                   '--oneline',
                   'GE:16,{0}:IMAGE1;{0},,{0}'.format(self.prefix[0:-1])]
```
Which results in the default 5 Hz rate being run for camViewer

It would be nice if we could _not_ hard code the acquire rate and pull it forward in `BaslerBase` for the user to define the acquisition rate. There are certain scenarios in alignment where being able to run a single cam at 10-20 Hz is much more important than running many cameras at 5 Hz.

aberges-SLAC:
Probably an arg you pass to the screen so it can change on the fly and restart your acquisition, as the demand can be dynamic. For example, during fine alignment you may realistically want 10-20 FPS for decent feedback to your moves, but for monitoring afterwards for drift, 5 Hz would be sufficient. I think if we ran all cameras at max resolution, bit depth, and acquisition rate then many hosts would riot, so I don't want to be an init that gets left running for very long times.

My users tend to use `pcdsdevices.areadetector.detectors.LasBasler` as the device_class with either of the 2 models we use often in lasers: Basler_acA2500-20gm or Basler_acA1440-73gm.

If you want a live one to play with and test your changes, I'll need to find something not in use in the hutches. Alternatively, I could always plug something into the LRL and coordinate with Nishant on having some test equipment.

PR #1309, opened 2024-11-20
Added configurable rate for  open_viewer

KaushikMalapati:
## Description

Added a rate signal that appears as a typhos slider on the screen which is passed as an argument to run_viewer which opens camviewer in open_viewer. Default is five hertz, which is also the default used by run_viewer. 

## Motivation and Context

#1308 

## How Has This Been Tested?

Interactively
Left side is at one hertz, right is at twenty. You can see the left camviewer update about once a second and right update much faster, although maybe not quite at twenty because of lag.
https://github.com/user-attachments/assets/c3537c01-8e48-4b55-8583-9d82793fa1ff

upcoming-release notes

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1309-PCDSAreaDetectorTyphos_camviewer_rate.rst
+++ b/docs/source/upcoming_release_notes/1309-PCDSAreaDetectorTyphos_camviewer_rate.rst
@@ -1,0 +1,32 @@
+1309 PCDSAreaDetectorTyphos_camviewer_rate
+#################
+
+API Breaks
+----------
+- N/A
+
+Library Features
+----------------
+- N/A
+
+Device Features
+---------------
+- Added rate signal to PCDSAreaDetectorTyphos
+- Value is passed in open_viewer to open camViewer with the selected rate
+- Default is five, but can be changed from gui with a TyphosSlider
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- KaushikMalapati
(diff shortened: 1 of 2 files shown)


tangkong:
Strange, I'm not sure.  I haven't used the typhos slider much, and would have to take a deeper look.  I'm not immediately sure if the other typhos screens that use this slider have a similar problem.

I tried changing the signal to a caproto-backed PV, on the off chance that there was some repetitive caput loop that kept snapping the slider.  (no change was seen)

ZLLentz:
This might just be related to https://github.com/slaclab/pydm/releases/tag/v1.25.0 since typhos relies on the pydm internals quite a bit

tangkong:
How are you invoking typhos/opening the screen?  perhaps that's stuck looking at `pcds_conda`?  (It looks like you're setting yourself up correctly)

aberges-SLAC:
Just adding a +1 to having a rate that is just an `int` being typed by the user. I think the sliders will drive the users in Lasers mad.

We probably _should_ use some smart bounds since acquisition rates vary wildly by camera. I don't remember if there is already a record in the gigecam ioc that describes that, but you can absolutely read that property from the genicam tool on a per-model basis (at least for aravis flavors)

ZLLentz:
> Actually, I was thinking about this more - is there any reason why this shouldn't be an option in the camviewer gui?

I literally wrote the code for this a few hours ago, with the intention of making this part of the next camviewer release.
Sneak preview very WIP (the cli argument will choose the starting value for the setting):

image

aberges-SLAC:
> If this is being added to camviewer, I think this pr is redundant. @aberges-SLAC, can you think of any reason why we would want to set the rate in typhos before opening in camviewer as opposed to dynamically selecting the rate in camviewer? If not, I will close this. I think a small benefit of the latter is to prevent people from repeatedly using a high framerate for lucid screens that are open for long periods of time.

I think that even if we opted for the latter, the users that would take advantage of this feature would have the cams open for very long times. I think keeping the acquisition rate inside the camviewer screen is fine. If the users really want it in the typhos screen afterwards, we can reopen the issue.

Thank you for the prompt response on all of this btw!

ZLLentz:
(if you want to speed up the cam viewer rate options PR, I'm looking for beta tester opinions in https://jira.slac.stanford.edu/browse/ECS-5976)

PR #1307, opened 2024-11-13
Making ExitSlits subclass with FDQ component, only for SL1K2 for now

KaushikMalapati:
## Description

ExitSlitsFDQ has an FDQ component so we can see flow meter readings on the typhos screen for SL1K2. I did not name this SL1K2 in case an ExitSlits+FDQ class is ever wanted in the future

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-6745

## Pre-merge checklist
- [ ] Code works interactively
- [ ] Code contains descriptive docstrings, including context and API
- [ ] New/changed functions and methods are covered in the test suite where possible
- [ ] Test suite passes locally
- [ ] Test suite passes on GitHub Actions
- [ ] Ran docs/pre-release-notes.sh and created a pre-release documentation page
- [ ] Pre-release docs include context, functional descriptions, and contributors as appropriate

PR #1311, opened 2024-12-05
TST: reproduce and fix pypi failures

ZLLentz:
## Description

Fix an issue where one of the attenuator tests started failing in the pypi builds.
This is done by waiting a bit, so the background threads have time to update all the metadata etc. or whatever they're doing. The fake epics signals do similar background thread work to what the real epics signals do which causes issues sometimes that can be solved by waiting.

## Motivation and Context

https://github.com/pcdshub/pcdsdevices/pull/1307#issuecomment-2518532901

## How Has This Been Tested?

The tests pass now.
I made a little utility function because it was simple to do and it may be used again.

Docstrings only

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1311-tst_failing_on_pip.rst
+++ b/docs/source/upcoming_release_notes/1311-tst_failing_on_pip.rst
@@ -1,0 +1,31 @@
+1311 tst_failing_on_pip
+#######################
+
+API Breaks
+----------
+- N/A
+
+Library Features
+----------------
+- N/A
+
+Device Features
+---------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- Make test_at2l0_clear_errors pass more consistently on CI.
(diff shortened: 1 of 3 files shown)


ZLLentz:
I agree it is best to use the event-driven tools in general.

Using `Status` for this requires any one of the following things that I wanted to avoid:
- Changing the source code of the attenuator class to make this method return a status object
- Referencing the "private" `_status` attribute
- Figuring out if `FakeEpicsSignal` has a useful status or not (for normal epics signals, it's only useful if put completion is used and supported by the IOC)

I think some of these reasons contribute to why ophyd internally polls in its set_and_wait routines.

tangkong (approved):
This is a reasonable first shout, though it's not as robust as we may like.  We discussed building something like `QtBot.waitUntil`, but we can circle back if this becomes a more widely used utility (and after blocked PRs are merged)

ZLLentz:
I thought about the event-driven idea again and realized that another way would have been to call `subscribe` and have the callback function set a `threading.Event` that we wait on for some timeout.

Maybe this is a worthwhile refactor to do later.

PR #1312, opened 2024-12-07
FIX: override velocity_base in MMC100, it doesn't have one

ZLLentz:
## Description

WIP

## Pre-merge checklist


tangkong (approved):
we should probably crystallize / record the different types of EPICS motor records we have.  We've been caught by this quite a few times, or at least enough times that we should add a test.

PR #1314, opened 2024-12-19
Improve Qmini spectrometer GUI

### LINKED ISSUE #1313: Improve QminiSpectrometer.Embedded.ui
## Current Behavior
The embedded UI gives the plots that users desire, but it is missing some key features that would improve its utilization in MODS and beam time.

## Expected Behavior
The screen could use the following PVs:

1. `START_EXPOSURE.PROC` (as a simple button) for manual collection of single spectra
2. `GET`/`SET_EXPOSURE_TIME` for controlling the integration time
3. `FIT_ON` for getting/setting fit params
4. `SENSITIVITY_CAL` and `CORRECT_PRNU` for correcting spectral sensitivity & photo-response non-uniformity respectively. 

Due to the way the `PROCESSING_STEPS_RBV` PV works with other records, it may be nice to have a small array of LED indicators for the `mbboDirect` record to indicate to the user what they are requesting of the spectrometer.

The screen could also use a way to save a single spectra to a `.CSV` or `.TXT` file that the user specifies the name/file location of. This is often very useful during experiments/beam time.

## Context / environment
Several folks from the laser science team have been just unplugging the qmini spectrometers from the TILEs/diagnostic towers to plug directly into their laptop so they can run the vendor software and get the calibrated (nW/nm) spectrum for their uses. This defeats the whole point of integrating the controls, but it is due to our GUI not providing the desired user experience. Adding these features should satisfy the majority of use cases in lasers. 

## Suggested Solution
For saving the spectra, this can probably be wrapped in a relatively painless method that:
1. Checks to make sure the spectrometer is not continuously streaming images (or maybe not in the future, if you want file bloat?)
2. Prompt the user for file storage
3. `caget -t SOME_SPECTROMETER_ALIAS:WAVELENGTHS` and `SOME_SPECTROMETER_ALIAS:SPECTRUM` and append that to said file.

I think you need to ignore the first element since it always pasts some part of a date in there, even in terse mode.

aberges-SLAC:
## Description
Gave a face-lift to the Qmini embedded & (new) detailed screen(s) to better suit the needs of SRD in MODS.

## Motivation and Context
Closes #1313 

Adds a shiny new save-spectrum feature so that scientists aren't tempted to disconnect the spectrometers and plug it into their laptops ;]

## How Has This Been Tested?
I've tested this in the pcds-5.9.1 environment and pretty extensively:
https://github.com/user-attachments/assets/fb2ba094-2165-40a2-9e9b-bd6075e3bb4e

In this PR and in the code!

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/lasers/qmini.py
+++ b/pcdsdevices/lasers/qmini.py
@@ -1,8 +1,12 @@
+import json
 import logging
+import os
+import time
 
 from ophyd import Component as Cpt
 from ophyd import Device, EpicsSignal, EpicsSignalRO
 from ophyd import FormattedComponent as FCpt
+from ophyd.signal import AttributeSignal, Signal
 
 from pcdsdevices.variety import set_metadata
 
@@ -82,6 +86,47 @@ class QminiSpectrometer(Device):
     fit_amplitude = Cpt(EpicsSignalRO, ':AMPLITUDE', kind='config')
     fit_stdev = Cpt(EpicsSignalRO, ':STDEV', kind='config')
     fit_chisq = Cpt(EpicsSignalRO, ':CHISQ', kind='config')
+
+    # Save spectra functions
+    def save_data(self):
+        """
+        Save the wavelength and spectrum PVs to a text file
+        """
+        if len(self.file_dest.get().strip()) == 0:
+            # set a default destination for the file if you're lazy or give it whitespace
+            _file = os.getcwd() + '/' + self.name + time.strftime("_%Y-%m-%d_%H%M%S") + '.txt'
+        else:
+            _file = self.file_dest.get()
+        self.log.info('Saving spectrum to disk...')
+        # Let's format to JSON for the science folk with sinful f-string mangling
+        _settings = ['sensitivity_cal', 'correct_prnu', 'correct_nonlinearity',
+                     'normalize_exposure', 'adjust_offset', 'subtract_dark',
+                     'remove_bad_pixels', 'remove_temp_bad_pixels']
+        _data = {'timestamp': time.strftime("%Y-%m-%d %H:%M:%S"),
+                 'exposure (us)': str(self.exposure.get()),
+                 'averages': str(self.exposures_to_average.get()),
+                 # Lets do some sneaky conversion to bool from int
+                 'settings': {f"{sig}": str(eval(f"{self}.{sig}.get() > 0")).lower()
+                              for sig in _settings},
+                 'wavelength (nm)': [str(x) for x in self.wavelengths.get()],
+                 'intensity (a.u.)': [str(y) for y in self.spectrum.get()]
+                 }
+        # and let's assume you have permission to save your file where you want to
+        with open(_file, 'w') as _f:
+            _f.write(json.dumps(_data))
+
+    save_spectrum = Cpt(AttributeSignal, attr='_save_spectrum', kind='omitted')
+    file_dest = Cpt(Signal, value='', kind='omitted')
+    set_metadata(save_spectrum, dict(variety='command-proc', value=1))
+
+    @property
+    def _save_spectrum(self):
+        return 0
+
+    # Setter will just save the data
+    @_save_spectrum.setter
+    def _save_spectrum(self, value):
+        self.save_data()
 
 
 class QminiWithEvr(QminiSpectrometer):
--- a/pcdsdevices/lasers/qmini.py
+++ b/pcdsdevices/lasers/qmini.py
@@ -88,25 +88,32 @@ class QminiSpectrometer(Device):
     fit_chisq = Cpt(EpicsSignalRO, ':CHISQ', kind='config')
 
     # Save spectra functions
-    def save_data(self):
+    def save_data(self, file_dest: str = ''):
         """
         Save the wavelength and spectrum PVs to a text file
         """
-        if not self.file_dest.get().strip():
-            # set a default destination for the file if you're lazy or give it whitespace
-            _file = os.getcwd() + '/' + self.name + time.strftime("_%Y-%m-%d_%H%M%S") + '.txt'
+        # Let's check to see if we set this in a non-gui context
+        if not file_dest.strip():
+            # let's try to use the signal instead
+            if not self.file_dest.get().strip():
+                # set a default destination for the file we didn't set it
+                _file = (os.getcwd() + '/' + self.name
+                         + time.strftime("_%Y-%m-%d_%H%M%S") + '.txt')
+            # otherwise just use it, silly
+            else:
+                _file = self.file_dest.get()
         else:
-            _file = self.file_dest.get()
+            _file = file_dest
         self.log.info('Saving spectrum to disk...')
         # Let's format to JSON for the science folk with sinful f-string mangling
         _settings = ['sensitivity_cal', 'correct_prnu', 'correct_nonlinearity',
                      'normalize_exposure', 'adjust_offset', 'subtract_dark',
                      'remove_bad_pixels', 'remove_temp_bad_pixels']
         _data = {'timestamp': time.strftime("%Y-%m-%d %H:%M:%S"),
-                 'exposure (us)': str(self.exposure.get()),
-                 'averages': str(self.exposures_to_average.get()),
+                 'exposure (us)': self.exposure.get(),
+                 'averages': self.exposures_to_average.get(),
                  # Lets do some sneaky conversion to bool from int
-                 'settings': {f"{sig}": str(eval(f"{self}.{sig}.get() > 0")).lower()
+                 'settings': {f"{sig}": bool(getattr(self, sig).get())
                               for sig in _settings},
                  'wavelength (nm)': [str(x) for x in self.wavelengths.get()],
                  'intensity (a.u.)': [str(y) for y in self.spectrum.get()]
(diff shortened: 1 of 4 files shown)


pcdsdevices/lasers/qmini.py
```diff
@@ -83,6 +87,47 @@ class QminiSpectrometer(Device):
     fit_stdev = Cpt(EpicsSignalRO, ':STDEV', kind='config')
     fit_chisq = Cpt(EpicsSignalRO, ':CHISQ', kind='config')
 
+    # Save spectra functions
+    def save_data(self):
```
ZLLentz, line 91: Nitpick, can ignore:
Is there any utility in making this more useful in non-gui contexts?
```
def save_data(self, filename: str = ""):
    if not filename:
        filename = self.file_dest.get().strip()
```

pcdsdevices/lasers/qmini.py
```diff
@@ -83,6 +87,47 @@ class QminiSpectrometer(Device):
...
+        """
+        Save the wavelength and spectrum PVs to a text file
+        """
+        if len(self.file_dest.get().strip()) == 0:
+            # set a default destination for the file if you're lazy or give it whitespace
+            _file = os.getcwd() + '/' + self.name + time.strftime("_%Y-%m-%d_%H%M%S") + '.txt'
```
ZLLentz, line 97: Question: I noticed that you prepend a lot of function-scoped variables with a leading underscore. I'd previously only seen this as a convention for semi-private class members and function names that are intended not to be exported via import as per https://peps.python.org/pep-0008/.

Are there other reasons to name variables like this?

pcdsdevices/lasers/qmini.py
```diff
@@ -83,6 +87,47 @@ class QminiSpectrometer(Device):
...
+        else:
+            _file = self.file_dest.get()
```
ZLLentz, line 99: A few lines above you strip the text contents from this signal. Maybe you should strip here too to help users avoid having leading/trailing spaces in their filenames.

pcdsdevices/lasers/qmini.py
```diff
@@ -83,6 +87,47 @@ class QminiSpectrometer(Device):
     fit_stdev = Cpt(EpicsSignalRO, ':STDEV', kind='config')
     fit_chisq = Cpt(EpicsSignalRO, ':CHISQ', kind='config')
 
+    # Save spectra functions
+    def save_data(self):
+        """
+        Save the wavelength and spectrum PVs to a text file
+        """
+        if len(self.file_dest.get().strip()) == 0:
```
ZLLentz, line 95: Nitpick: empty strings evaluate to False without needing to get the length.
```suggestion
        if not self.file_dest.get().strip():
```

pcdsdevices/lasers/qmini.py
```diff
@@ -83,6 +87,47 @@ class QminiSpectrometer(Device):
...
+        self.log.info('Saving spectrum to disk...')
+        # Let's format to JSON for the science folk with sinful f-string mangling
+        _settings = ['sensitivity_cal', 'correct_prnu', 'correct_nonlinearity',
+                     'normalize_exposure', 'adjust_offset', 'subtract_dark',
+                     'remove_bad_pixels', 'remove_temp_bad_pixels']
+        _data = {'timestamp': time.strftime("%Y-%m-%d %H:%M:%S"),
+                 'exposure (us)': str(self.exposure.get()),
+                 'averages': str(self.exposures_to_average.get()),
+                 # Lets do some sneaky conversion to bool from int
+                 'settings': {f"{sig}": str(eval(f"{self}.{sig}.get() > 0")).lower()
```
ZLLentz, line 109: Nitpick: maybe simpler without an `eval` so the intent is clearer and the next maintainer doesn't get confused.
I probably botched this because I also got confused. I left it as an int because ints are allowed in json.
```suggestion
                 'settings': {f"{sig}": int(getattr(self, sig).get())
```

pcdsdevices/lasers/qmini.py
```diff
@@ -83,6 +87,47 @@ class QminiSpectrometer(Device):
...
+            _file = os.getcwd() + '/' + self.name + time.strftime("_%Y-%m-%d_%H%M%S") + '.txt'
+        else:
+            _file = self.file_dest.get()
+        self.log.info('Saving spectrum to disk...')
+        # Let's format to JSON for the science folk with sinful f-string mangling
+        _settings = ['sensitivity_cal', 'correct_prnu', 'correct_nonlinearity',
+                     'normalize_exposure', 'adjust_offset', 'subtract_dark',
+                     'remove_bad_pixels', 'remove_temp_bad_pixels']
+        _data = {'timestamp': time.strftime("%Y-%m-%d %H:%M:%S"),
+                 'exposure (us)': str(self.exposure.get()),
```
ZLLentz, line 106: I'm not sure I understand why these need to be cast to strings.

ZLLentz (commented):
I like what you've done here in the python file and I mostly have nitpicks and questions
I'll peek at the ui files next

pcdsdevices/lasers/qmini.py
aberges-SLAC: What did you have in mind?

pcdsdevices/ui/QminiSpectrometer.detailed.ui
ZLLentz, line 1: UI files look good!

Note that there's one small discrepancy: the embedded has a button with the text "Start Single Exposure" and the detailed has the same button with the text "Force Single Exposure". I don't have any preference but maybe the text should be the same if they do the same thing!

ZLLentz:
I think this is mergeable with consideration or rejection of my comments + a pre-release docs page

pcdsdevices/lasers/qmini.py
aberges-SLAC: I'm doing for (paranoid) privacy reasons on the calls. I'm probably being overly cautious, honestly, but it's a habit I've kept. I don't think there's going to be any conflicts (aside from the signal attribute name shenanigans that need it — as I found out the hard way)

pcdsdevices/lasers/qmini.py
aberges-SLAC: Huh! I didn't realize that, and it is _much_ cleaner

pcdsdevices/lasers/qmini.py
aberges-SLAC: I neglect `getattr` a little too often for my own good, that's much cleaner. `int` are allowed but I preferred `true` and `false` outputs for the states to match their respective PVs as `bo` records. 

This is maybe just convenience for the consumer of the dump, since it saves them a comparison when checking settings in the object.

I might just be enforcing my own preferences, tbh, so it isn't strictly necessary. Maybe I'll ask the scientists what they prefer 🤷

pcdsdevices/lasers/qmini.py
aberges-SLAC: `json.dumps` was being weird for a second and I think I cast everything to a `str` when sorting out the problem and left that in there. These can probably live as the `int`s that they are

pcdsdevices/ui/QminiSpectrometer.detailed.ui
aberges-SLAC: Oooooh very good catch! I meant for these to both be "Force" to be consistent with the typhos default detailed screen that the folks are used to.

pcdsdevices/lasers/qmini.py
ZLLentz: Can you expand a bit on privacy reasons? My question is mostly to learn

pcdsdevices/lasers/qmini.py
ZLLentz: Gotcha, I guess since the point is for the end-user to read there could a lot of ways to do it!

I you can also put boolean literals into json and it looks like `true` or `false` without the quotes.

pcdsdevices/lasers/qmini.py
aberges-SLAC: I think, strictly speaking, `_var` is functionally a stylistic convention for saying "this `var` is private-ish, but I'm not gonna enforce it. Remember that you need to be janky to get here" vs. `__var` getting the built-in string mangling. 

Mostly useful for me in contexts where there are pretty commonly recycled var names (i.e. `result` , `data`, `output`, etc.) and I want to be explicit about which one I am messing with. That way I can have a parent function return a `result` dependent on an inner function call with output `_result` without getting too weird. Idk maybe I inherited some weird jank convention from `C` that I haven't unlearned.

pcdsdevices/lasers/qmini.py
aberges-SLAC: I guess I could've just cast to `bool` anyway, right? Not sure why I compared to get there in the first place lol. Very good point, I like that much better and will do that

pcdsdevices/lasers/qmini.py
aberges-SLAC: Ah, but the wavelength & spectra _must_ be cast to str since `Float32` is not `json` compat, but that's on the consumer to handle turning back into a number.

pcdsdevices/lasers/qmini.py
ZLLentz: Gotcha, I do the same but only for data attributes e.g. `self._var`. I can see how it'd also be nice to indicate which variables are meant to persist through the function and which are e.g. loop-only. Thanks for humoring me.

ZLLentz (approved):
I think we should bring this in!

PR #1316, opened 2025-01-07
add XCS ladm.py file and add tab whitelist/alias to lens.py

carogee:
## Description

Adding ladm.py file and tab whitelists/aliases names in lens.py

## Motivation and Context

XCS ladm.py file has been copied manually into pcdsdevices upgrades, but it would be helpful to have it as part of the git directory. 
Also tab whitelists/aliases used in XCS in the lens.py file are helpful.

## How Has This Been Tested?

Has been running in XCS pcdsdevices.

 Here

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/ladm.py
+++ b/pcdsdevices/ladm.py
@@ -90,6 +90,11 @@ def ThetaToMotors_print(theta):
 class LADM(BaseInterface, GroupDevice):
     """
     Class to control the LADM. Includes Single motors and beamstops.
+    
+    The LADM is the Large Angle Detector Mover in XCS.
+    It's the large device that sweeps across the instrument's floor in a wide angle.
+    It can be used to position and reposition a detector at various angles relative
+    to the sample position.
 
     Parameters
     ----------
--- a/pcdsdevices/ladm.py
+++ b/pcdsdevices/ladm.py
@@ -19,6 +19,9 @@ alpha_cos = np.cos(27 * np.pi/180)
 
 
 def ThetaToMotors(theta, samz_offset=0):
+    """
+    Calculates lengths x1, x2, and dz
+    """
     theta = theta * np.pi/180  # change to radian
     x1 = (r) * np.sin(theta)/(np.sin(alpha_r) * np.sin(alpha_r+theta))
     x2 = (R) * np.sin(theta)/(np.sin(alpha_r) * np.sin(alpha_r+theta))
@@ -29,18 +32,27 @@ def ThetaToMotors(theta, samz_offset=0):
 
 
 def y1ToGamma(y1):
+    """
+    Calculates angle gamma=arctan(y1/r) in radians
+    """
     gamma = np.arctan((y1)/(r))
     gamma = gamma * 180/np.pi
     return gamma
 
 
 def y2ToGamma(y2):
+    """
+    Calculates angle gamma=arctan(y2/r) in radians
+    """
     gamma = np.arctan((y2)/(R))
     gamma = gamma * 180/np.pi
     return gamma
 
 
 def GammaToMotors(gamma):
+    """
+    Calculates lengths 'y1', 'y2', their difference 'dy'
+    """
     gamma = gamma * np.pi/180
     y1 = (r) * np.tan(gamma)
     y2 = (R) * np.tan(gamma)
@@ -49,6 +61,9 @@ def GammaToMotors(gamma):
 
 
 def x1ToTheta(x1, samz_offset=0):
+    """
+    Calculates angle theta in radians using length x1 and samz_offset
+    """
     theta = np.arctan(x1 * (np.sin(alpha_r))**2 /
                       (r-samz_offset * alpha_cos-x1 * np.sin(2 * alpha_r)/2))
     theta = theta * 180/np.pi
@@ -56,6 +71,9 @@ def x1ToTheta(x1, samz_offset=0):
 
 
 def x2ToTheta(x2, samz_offset=0):
+    """
+    Calculates angle theta in radians using length x2 and samz_offset
+    """
     theta = np.arctan(x2 * (np.sin(alpha_r))**2 /
                       (R-samz_offset * alpha_cos-x2 * np.sin(2 * alpha_r)/2))
     theta = theta * 180/np.pi
@@ -63,16 +81,25 @@ def x2ToTheta(x2, samz_offset=0):
 
 
 def xTox12(x):
+    """
+    Calculates 'x12' from 'x' and the rail angle alpha_r
+    """
     x12 = x/np.sin(alpha_r)
     return x12
 
 
 def xToz(x):
+    """
+    Calculates 'z' from 'x' and the rail angle alpha_r
+    """
     z = x/np.tan(alpha_r)
     return z
 
 
 def MotorsTox(x1, x2, z):
+    """
+    Calculates x_x1, x_x2, and x_z from x1, x2, and z trigonomic relationships with rail angle alpha_r
+    """
     x_x1 = x1 * np.sin(alpha_r)
     x_x2 = x2 * np.sin(alpha_r)
     x_z = z * np.tan(alpha_r)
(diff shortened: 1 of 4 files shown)


ZLLentz (commented):
Thank you for bringing this into the main pcdsdevices!

I made a lot of comments here. I marked each one to help note the relative importance of each.

Change requests are things that definitely need to be done before I can approve the PR.

Questions and confusions really need a response so we can figure out together if something needs to be done or not.

Suggestions are things that I highly recommend doing but I won't hold back the PR for.

You can ignore or defer follow-up ideas and nitpicks if you don't have time, even though I think they are at least somewhat useful.

pcdsdevices/ladm.py
```diff
@@ -0,0 +1,423 @@
...
+from pcdsdevices.epics_motor import IMS
+
+from .device import GroupDevice
+from .interface import BaseInterface
+
+logger = logging.getLogger(__name__)
+
+alpha_r = 63 * np.pi/180  # rail angle
+r = 2960.0  # first rail distance from sample to rail rod at 27deg
+R = 6735.0
```
tangkong, line 19: What is this?  Surely there's a more descriptive name than captial R (to avoid confusion with lowercase r)

pcdsdevices/ladm.py
```diff
@@ -0,0 +1,423 @@
...
+    theta = theta * np.pi/180  # change to radian
+    x1 = (r) * np.sin(theta)/(np.sin(alpha_r) * np.sin(alpha_r+theta))
+    x2 = (R) * np.sin(theta)/(np.sin(alpha_r) * np.sin(alpha_r+theta))
+    dz = (r)/np.sin(alpha_r)-x1 * np.sin(alpha_r)/np.sin(theta)
+    if theta == 0:
+        dz = 0
+    return x1, x2, dz
+
+
+def y1ToGamma(y1):
```
tangkong, line 33: Why are these defined outside the LADM class?  They seem to be rather specific to the geometry of the LADM, with coordinate names that only someone familiar with the LADM would understand.  

Generally I dislike such opaque naming.  Without a diagram it's impossible to verify that this math is doing the right thing

pcdsdevices/ladm.py
```diff
@@ -0,0 +1,423 @@
...
+                self.x1.wait()
+                self.x2.wait()
+                print("Done")
+                print(f"Moving z to {z_var}")
+                self.z.mvr(z_var - z_now)
+                print("Done")
+        except KeyboardInterrupt:
+            self.stop()
+
+    def mvrV(self, y):
```
tangkong, line 347: Enhancement / Improvement idea / Musing: 

It looks like we're most interested in derived quantities (V, H, X, Theta), rather than the motors that have been defined in this class (x1, y1, z, etc)

Should we be wrapping these quantities in pseudo positioners so that we can use normal manipulate them with normal `FltMvInterface` methods?

pcdsdevices/ladm.py
```diff
@@ -0,0 +1,423 @@
...
+        return self.__hilimX
+
+    def stop(self):
+        self.x1.stop()
+        self.x2.stop()
+        self.y1.stop()
+        self.y2.stop()
+        self.z.stop()
+
+    def __call__(self):
```
tangkong, line 422: Nitpick: We probably want to use the standard `format_status_info` method that displays device information in ipython.  That way you don't actually have to write the parentheses `()` when you want to see the status info.

More concisely, I think if we define `LADM.format_status_info` we should just be able to type LADM in the ipython terminal to see the status info

tangkong (commented):
Some additional thoughts from me.  I'm happy this is being upstreamed, and this is a good opportunity to clean some of this up

pcdsdevices/ladm.py
carogee: I'm guessing it is another rail distance, but I'm not sure who wrote this code...

pcdsdevices/ladm.py
```diff
@@ -0,0 +1,423 @@
...
+                     'ca_theta', 'stop'
+                     ]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.theta = None  # VirtualMotor
+        self.XT = None  # VirtualMotor
+        self.gamma = None
+        self.__lowlimX = None
+        self.__hilimX = None
```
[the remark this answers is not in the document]
carogee: Need to ask scientists if they use __lowlimX and __hilimX

pcdsdevices/lens.py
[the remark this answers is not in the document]
carogee: Added

pcdsdevices/ladm.py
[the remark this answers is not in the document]
tangkong: Somehow during init `LADM.theta` and `LADM.XT` become `FunPositioner`'s, I'm still unclear where this happens

```python
In [16]: ladm.theta
Out[16]:

ladmTheta
---------
position: nan

In [17]: type(ladm.theta)
Out[17]: pcdsdevices.positioner.FuncPositioner
```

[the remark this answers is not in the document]
tangkong: I see now.  These get set in the `beamline.py` file:

```python
    from pcdsdevices.ladm import LADM
    from pcdsdevices.positioner import FuncPositioner
    ladm = LADM('XCS:LAM', name='ladm')

    ladm.__lowlimX=ladm._set_lowlimX(-10)
    ladm.__hilimX=ladm._set_hilimX(2000)

    ladmTheta = FuncPositioner(name='ladmTheta', move=ladm.moveTheta, get_pos=ladm.wmTheta, set_pos=ladm.setTheta, stop=ladm.stop)
    ladm.theta = ladmTheta

    ladm.XT = FuncPositioner(name='ladmXT', move=ladm.moveX, get_pos=ladm.wmX, set_pos=ladm._setX, egu='mm', limits=(ladm._get_lowlimX, ladm._get_hilimX))
```

This is sort of a dangerous/inconsistent way to go about loading these devices.  This appears in multiple beamline*.py files in slightly different ways, with and without explicit limits being set.  I don't see why we can't initialize this all at once within the ladm class. (adding `ladm.{theta, XT}`)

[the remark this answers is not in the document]
tangkong: `ladm.gamma` remains unset and unused as far as I can tell (beamline.py's included)

pcdsdevices/ladm.py
```diff
@@ -29,18 +32,27 @@ def ThetaToMotors(theta, samz_offset=0):
...
+    Calculates angle gamma=arctan(y1/r) in radians
+    """
     gamma = np.arctan((y1)/(r))
     gamma = gamma * 180/np.pi
     return gamma
 
 
 def y2ToGamma(y2):
+    """
+    Calculates angle gamma=arctan(y2/r) in radians
```
tangkong, line 45: Thanks for adding these docstrings! I think they're pretty good, just lacking a bit of context around what x/y/gamma are.  Do you have that information handy?

carogee: Added descriptions of the variables

pcdsdevices/lxe.py
```diff
@@ -268,7 +268,7 @@ class LaserTiming(FltMvInterface, PVPositioner):
                    original_units='ns',
                    kind='hinted',
                    doc='Setpoint which handles the timing conversion.',
-                   limits=(-100e-6, 100e-6),
+                   limits=(-2e-3, 2e-3),
```
ZLLentz, line 271: Hang on, I think I missed this the first time I reviewed here.

Is this an xcs-specific limits change?

carogee: The XCS scientists wanted to be able to have a wider laser virtual motor range they could scan over. I copied the new limits from XPP's code, though I'm not sure if the other hutches want it.

pcdsdevices/ladm.py
carogee: Added comment in code
r=distance from the gon to the first (shorter) LADM rail at 27 degrees, and R = distance from the gon to the second (longer) LADM rail at 27 degrees

pcdsdevices/lxe.py
tangkong: This should definitely be changed in XCS's device loading procedure, either in happi or in their hutch-python startup files.  If you change it here in pcdsdevices it will affect ever user of the `LaserTiming` class

PR #1055, opened 2022-09-19
TST: skip race condition test for now, must revisit later

ZLLentz:
## Description

Mark test_presets as xfail for now

## Motivation and Context

We're not actively working on the presets system and the test has a race condition that is giving false failures on pull request builds. See #1050 for details.

I don't immediately have time to fix the test so I thought this was a prudent band-aid.

## How Has This Been Tested?

Via test suite

## Where Has This Been Documented?

I'll add a release notes blurb but hopefully we'll remove it before the next tag

## Pre-merge checklist
- [x] Code works interactively
- [x] Code contains descriptive docstrings, including context and API
- [x] New/changed functions and methods are covered in the test suite where possible
- [x] Test suite passes locally
- [x] Test suite passes on travis
- [x] Ran docs/pre-release-notes.sh and created a pre-release documentation page
- [x] Pre-release docs include context, functional descriptions, and contributors as appropriate

PR #1317, opened 2025-01-11
PERF: delay preset sync until tab completion / attribute access, track file desync

tangkong:
## Description
Delays preset file reading until tab-completion has been attempted.  Stashes this information on the device instance itself.  This is not the default behavior, and you must opt into this by supplying the optional argument `defer_loading` to `setup_presets_paths`

Also tracks the modification time of the preset file on sync, making the `Preset` instance the source of truth to see if the presets need to be synced.

## Motivation and Context
Presets have taken a very long time to load recently, and the first place to look is often file i/o.  This delays this as long as possible while attempting to preserve the behavior hutch scientists use: tab completion.

This has also been made optional.  Both `setup_presets_paths` and `Presets.sync()` now take an optional `defer_loading` argument, that if true will skip the preset data loading step

In a nutshell:
* the device's Preset instance tracks whether the preset is out of sync with the files.  This is done by tracking file modification times

* The device checks if a sync is necessary, and if so syncs with the files when:
  * `__dir__` is called (during tab-completion)
  * `__getattribute__` is called with a preset-related prefix (`wm_`, `umv_`, `mv_`)

`Presets.sync()` is also called on device init, since it's part of `Presets.__init__`.  This happens both on device-creation and `setup_presets_path`, though on device-creation we don't do any file loading.  This is a ~0.013s double-hit, split between the device load and presets load sections for `FltMvInterface` devices

## How Has This Been Tested?
Tests pass

Tested interactively through hutch-python, loading some subset of 559 xcs presets
* this PR: 7.53s (0.01345s / preset)
* current master: 16.21s (0.0289s / preset)

There's some variation but I'm not going to do a battery of tests here

### Why isn't this just 0 s?
There's still some path access going on in hutch-python, it's just stopping short of actually opening the presets file.

This PR, a stray comment

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -752,14 +752,14 @@ class FltMvInterface(MvInterface):
 
     def __dir__(self):
         # Initialize presets if tab-completion requested.
-        if self.presets.sync_needed:
+        if self.presets.sync_needed():
             self.presets.sync()
 
         return super().__dir__()
 
     def __getattribute__(self, name: str):
         if (any((name.startswith(prefix)) for prefix in ['mv_', 'wm_', 'umv_'])
-                and self.presets.sync_needed):
+                and self.presets.sync_needed()):
             self.presets.sync()
 
         return super().__getattribute__(name)
@@ -1328,7 +1328,6 @@ class Presets:
                     closest = diff
         return state
 
-    @property
     def sync_needed(self) -> bool:
         """True if this preset has fallen out of sync with backing files"""
         curr_mtimes = {}
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -227,8 +227,13 @@ class BaseInterface:
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._tab = self._class_tab.new_instance(self)
+        self._tab_initialized = False
 
     def __dir__(self):
+        if not self._tab_initialized and hasattr(self, 'presets'):
+            self._tab_initialized = True
+            self.presets.sync()
+
         return self._tab.get_dir()
 
     def __repr__(self):
@@ -1085,16 +1090,19 @@ class Presets:
         self._remove_methods()
         self._cache = {}
         logger.debug('filling %s cache', self.name)
-        for preset_type in self._paths.keys():
-            path = self._path(preset_type)
-            if path.exists():
-                try:
-                    self._cache[preset_type] = self._read(preset_type)
-                except BlockingIOError:
-                    self._log_flock_error()
-            else:
-                logger.debug('No %s preset file for %s',
-                             preset_type, self._device.name)
+
+        # only consult files if tab-completion has been attempted
+        if self._device._tab_initialized:
+            for preset_type in self._paths.keys():
+                path = self._path(preset_type)
+                if path.exists():
+                    try:
+                        self._cache[preset_type] = self._read(preset_type)
+                    except BlockingIOError:
+                        self._log_flock_error()
+                else:
+                    logger.debug('No %s preset file for %s',
+                                 preset_type, self._device.name)
         self._create_methods()
 
     def _log_flock_error(self):
@@ -1140,6 +1148,7 @@ class Presets:
         self._methods.append((obj, method_name))
         setattr(obj, method_name, MethodType(method, obj))
         if hasattr(obj, '_tab'):
+            # obj._tab: TabCompletionHelperInstance
             obj._tab.add(method_name)
 
     def _make_add(self, preset_type):
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -904,7 +904,7 @@ class FltMvInterface(MvInterface):
         self.set_current_position(position)
 
 
-def setup_preset_paths(**kwargs):
+def setup_preset_paths(defer_loading: bool = False, **paths):
     """
     Prepare the :class:`Presets` class.
 
@@ -920,8 +920,6 @@ def setup_preset_paths(**kwargs):
         (Optional) "defer_loading": bool, whether or not to defer the loading
         of preset files until the first tab completion
     """
-    defer_loading = kwargs.pop("defer_loading", False)
-    paths = kwargs
     Presets._paths = {}
     for k, v in paths.items():
         Presets._paths[k] = Path(v)
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -904,7 +904,7 @@ class FltMvInterface(MvInterface):
         self.set_current_position(position)
 
 
-def setup_preset_paths(**paths):
+def setup_preset_paths(**kwargs):
     """
     Prepare the :class:`Presets` class.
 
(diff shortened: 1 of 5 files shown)


tangkong:
I did some more digging (manual profiling because I'm dumb) as to what causes the remaining add-methods to take so long, and I think it boils down to `stat()` calls having awful performance.  I'll do some more digging but I do think this is a contribution to be considered separately

pcdsdevices/interface.py
```diff
@@ -899,25 +904,29 @@ def set_position(self, position):
         self.set_current_position(position)
 
 
-def setup_preset_paths(**paths):
+def setup_preset_paths(**kwargs):
```
ZLLentz, line 907: Nitpick: marginally simpler
```suggestion
def setup_preset_paths(defer_loading: bool = False, **paths):
```

pcdsdevices/interface.py
```diff
@@ -1140,6 +1152,7 @@ def _register_method(self, obj, method_name, method):
         self._methods.append((obj, method_name))
         setattr(obj, method_name, MethodType(method, obj))
         if hasattr(obj, '_tab'):
+            # obj._tab: TabCompletionHelperInstance
```
ZLLentz, line 1155: Nitpick: lingering commented code line

pcdsdevices/interface.py
```diff
@@ -899,25 +904,29 @@ def set_position(self, position):
...
+        of preset files until the first tab completion
+    """
+    defer_loading = kwargs.pop("defer_loading", False)
+    paths = kwargs
     Presets._paths = {}
     for k, v in paths.items():
         Presets._paths[k] = Path(v)
     for preset in Presets._registry:
-        preset.sync()
+        preset.sync(defer_loading=defer_loading)
```
ZLLentz, line 929: Suggestion: maybe it's simpler to just not call sync at all if `defer_loading` is set to `True`, as long as this doesn't cause any unexpected issues. Then, `sync` doesn't need any code changes.

pcdsdevices/interface.py
```diff
@@ -227,8 +227,13 @@ def __init_subclass__(cls, **kwargs):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._tab = self._class_tab.new_instance(self)
+        self._tab_initialized = False
 
     def __dir__(self):
+        if not self._tab_initialized and hasattr(self, 'presets'):
+            self._tab_initialized = True
+            self.presets.sync()
```
ZLLentz, line 235: Change/addition request: we need to identify other times where presets may need a late sync. For example, if someone is expecting presets to be available in a hutch-python function call they need a clear way to trigger this without tabbing (or we need to load them automatically).

pcdsdevices/interface.py
ZLLentz, line 1: Side-side thought: maybe this is an opportunity to look at https://jira.slac.stanford.edu/browse/ECS-923. (In short, since presets are only loaded once, they desync quickly between different sessions)

ZLLentz (commented):
I like this and have a lot of opinions about presets in general.


> stat() calls having awful performance

I agree this is a problem on NFS/WEKA.
Probably the main way for us to mitigate this would be reworking presets to use fewer files, but that has its own annoyances.

pcdsdevices/interface.py
tangkong: I kinda thought that would be useful for future us as a sort of pseudo type-hint

pcdsdevices/interface.py
tangkong: The other half of `sync` sets up the `add_*` methods, which I thought would be nice to have in all cases.  But if we're ok removing those methods then I could get behind it.

ZLLentz: I forgot about that. I'm not immediately sure how I feel about those methods in the context of having not loaded the presets at all yet.

pcdsdevices/interface.py
tangkong: The `device.presets.sync()` call is still available, and seems to be how we suggest people sync presets across sections.  I imagine we haven't advertised this as thoroughly as necessary.

tangkong: I guess we could overwrite `__getattribute__` and catch attempts to access `mv_*`, `umv_*`, and `wm_*` methods, but are there any other access modes we can think of?

ZLLentz: Is it possible that doing this on `__getattribute__` is actually most correct to make sure the loaded values are always up-to-date?

tangkong: My one qualm with this is that it does break the already fragile composition abstraction barrier we've set up with the Presets module.  Maybe there's also a way around this.  

Back to the mines!

pcdsdevices/interface.py
tangkong: Indeed, I think now is probably the time.  A quick investigation shows that file modification queries are indeed faster than the entire sync (perhaps obvious now, but it answers the question in ECS-923)

```python
In [10]: %%time
    ...: print(list(os.path.getmtime(p) for p in xcs_ccm.x.presets._paths.values()))
    ...: 
    ...: 
[1736549297.5101147, 1736549297.5232627]
CPU times: user 0 ns, sys: 399 µs, total: 399 µs
Wall time: 2.19 ms

In [11]: %%time
    ...: xcs_ccm.x.presets.sync()
    ...: 
    ...: 
CPU times: user 15.7 ms, sys: 1.31 ms, total: 17 ms
Wall time: 65.7 ms
```

It should be simple enough to stash these values and re-sync if they fall out of date

tangkong: I've given this a shot.  It was a bit squirrely to test, but I think I ironed it out

tangkong:
Updated with some rather significant changes, now we store the last file modification times with every `sync()` call, and use that to decide whether or not to update.  Also added some tests to codify this

ZLLentz:
I think at least one of your commits has not been pushed

tangkong:
I blame github

tangkong:
Fingers crossed that we don't see race conditions like in #1050 , and the resulting #1055

pcdsdevices/tests/test_interface.py
```diff
@@ -191,6 +201,55 @@ def test_presets_type(presets, fast_motor):
...
+    # original object needs a sync, but second does not
+    assert fast_motor.presets.sync_needed
+    assert not fast_motor2.presets.sync_needed
+
+
+def test_presets_tab_init(fast_motor: FastMotor, deferred_fast_motor_presets):
+    # deferred_fast_motor_preset must come last,
+    # to clear cache after motor is created (and sync-ed at init)
+    assert fast_motor.presets.sync_needed
+    fast_motor.__dir__()  # mimic tab completion request
```
ZLLentz, line 233: Nitpick: is `dir(fast_motor)` equivalent?

ZLLentz (approved):
I like how this is implemented and I like how it speeds up the startup preset loading by 2x

pcdsdevices/tests/test_interface.py
tangkong: I believe so, do we prefer that to the dunder-access?  I used `__dir__` just to match the code usage.

janeliu-slac:
First time seeing updates to pcdsdevices, so probably don't have good feedback this time. Are presets for device settings? Why does preset file reading need to be delayed until tab completion has been attempted?

tangkong:
Presets are basically position shortcuts that users can save/load.  They're dynamically added to the devices attributes, and are backed by simple yaml files.
https://pcdshub.github.io/pcdsdevices/v8.7.0/presets.html

Because each device is backed by a file, we end up opening and reading a lot of files in some hutch-python sessions.  So instead of loading them all at once, we defer loading until the user actually needs it

pcdsdevices/interface.py
```diff
@@ -1304,6 +1328,17 @@ def state(self):
                     closest = diff
         return state
 
+    @property
+    def sync_needed(self) -> bool:
```
ZLLentz, line 1332: One last tiny tiny nitpick: this is a property that touches the filesystem, usually I prefer things that may take time to be function calls. Usually we expect attribute access to be fast, so if they are not fast we can accidentally write code with performance issues.

ZLLentz (approved):

pcdsdevices/interface.py
tangkong: A good point.  I'll do that

ZLLentz (approved):
I'm excited for the speedup

PR #1318, opened 2025-01-24
Changing EpicsSignal to PytmcSignal for plc value

KaushikMalapati:
## Description

Made reset_cmd in TwinCATStatePositioner a PytmcSignal instead of an EpicsSignal so that the indicator uses the _RBV pv

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-7110
https://jira.slac.stanford.edu/browse/ECS-6741

## How Has This Been Tested?

Not tested yet. 

upcoming release notes

image

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1318-TwinCATStatePositionerResetIndicator.rst
+++ b/docs/source/upcoming_release_notes/1318-TwinCATStatePositionerResetIndicator.rst
@@ -19,7 +19,7 @@ New Devices
 
 Bugfixes
 --------
-- reset_cmd is now a PytmcSignal with io='io' instead of an EpicsSignal
+- TwinCATStatePositioner.reset_cmd is now a PytmcSignal with io='io' instead of an EpicsSignal to properly display the RBV readback PV in typhos.
 
 Maintenance
 -----------


NSLentz (approved):
This looks good, however, Tong told me she already tested the change to the sequence mover and that it was working and we applied that change to the PLC so perhaps something else is happening if it still has issues updating on the display?

ZLLentz:
I suspect that you'll never catch this ui element updating unless there's an issue with the PLC code that gets it stuck in the `TRUE` state, since the PLC clears the indicator very quickly.

ZLLentz:
I think this is mergeable with a pre-release docs added (no need to make it details, just a quick one-liner would suffice)

docs/source/upcoming_release_notes/1318-TwinCATStatePositionerResetIndicator.rst
```diff
@@ -0,0 +1,30 @@
...
+---------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- reset_cmd is now a PytmcSignal with io='io' instead of an EpicsSignal
```
ZLLentz, line 22: Sorry, maybe I'm at least a little picky with respect to context. This won't make any sense in the combined release notes. I'm expanding it slightly so that I understand where it came from later.
```suggestion
- TwinCATStatePositioner.reset_cmd is now a PytmcSignal with io='io' instead of an EpicsSignal to properly display the RBV readback PV in typhos.
```

ZLLentz (approved):

ZLLentz:
> I wonder if it makes sense to not show the indicator for things like this

It's possible that the indicator is more distracting than useful. I think this tweak is better than what we had before because it will help us find PLC bugs, maybe we can follow-up by thinking more broadly about what would be most useful to show for these sorts of "set and then they get reset" sorts of signals.

If we can think of a most-correct thing to show in the readback indicator we can add it as a standard typhos variety and it will absolutely get some use in various places.

Some examples of possible variants, to give an idea of the decision space:
- remove the circle indicator
- have the circle indicator light up briefly if the button's caput was successful (I think this one would be neat)
- have two circle indicators, one for setpoint and one for readback

tangkong:
Is "getting stuck in the `True` state" something that happens frequently?  Aside from that one error state, it's hard to imagine the indicator providing much value.

I 100% agree that this is an obvious improvement over the status quo though, I have no objections to merging this.

ZLLentz:
> Is "getting stuck in the True state" something that happens frequently?

It happened once in TMO as part of a tough to track down bug, otherwise I haven't seen it. I suspect it only could happen as a bug.

PR #1319, opened 2025-01-24
BUG: make Presets.status more tolerant of unknown/uninitialized positions

### LINKED ISSUE #1236: BUG: preset formatting should be more error tolerant
## Current Behavior
https://github.com/pcdshub/pcdsdevices/blob/e1be5bfee941ea0999ba18c2cc22b4e25bdcd966/pcdsdevices/epics_motor.py#L178
this should refer to the "ERROR" preset and display the status instead of exploding

Also in this case the state repr fails:
```python
In [2]: lib_y.presets.state()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [2], in <cell line: 1>()
----> 1 lib_y.presets.state()

File /cds/group/pcds/pyps/apps/hutch-python/xcs/dev/devpath/pcdsdevices/interface.py:1301, in Presets.state(self)
   1299 state_name = method_name.replace('wm_', '', 1)
   1300 wm_state = getattr(device, method_name)
-> 1301 diff = abs(wm_state())
   1302 if diff < closest:
   1303     state = state_name

File /cds/group/pcds/pyps/apps/hutch-python/xcs/dev/devpath/pcdsdevices/interface.py:1263, in Presets._make_wm_pre.<locals>.wm_pre(self)
   1251 """
   1252 Check the offset from the {} preset position.
   1253
   (...)
   1259     is in the positive direction from us.
   1260 """
   1262 pos = self.presets._cache[preset_type][name]['value']
-> 1263 return pos - self.wm()

TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'
```

## Expected Behavior
Error presets don't crash everything

## Context / environment
From debugging some issues in XCS

[slack thread](https://slac.slack.com/archives/C2PNS4Q74/p1716935256463299?thread_ts=1716932102.900739&cid=C2PNS4Q74)
1. Initialize an IMS motor
2. Somehow don't let its readback callback complete (such that self.position == None)
3. `motor.presets.state()`

## Suggested Solution
Fix it?

tangkong:
## Description
Makes Preset `Preset.device.wm_*` and `Presets.state` methods more tolerant of missing or uninitialized device positions.  

This changes the signature of the `wm_*` methods slightly to also return "Unknown" if the position cannot be resolved.  

## Motivation and Context
closes #1236 

As these are autogenerated methods, it's a little harder to confirm that this return type change won't break downstream consumers.  At least within `pcdsdevices`, the only method that calls a `device.wm_*` method is `Presets.state`.  It's possible that some script very strictly requires the output of a preset to be a numeric, but if the position was uninitialized, those scripts would throw exceptions anyway (no return would be reached)

## How Has This Been Tested?
Checking that CI passes. 

By manually setting `device.postition` to `None` and seeing the output not explode
```python
In [1]: xcs_dg3_pim.y.position
Out[1]: -51.99951171875

In [2]: xcs_dg3_pim.y.wm_ccm()
Out[2]: 59.49951171875

In [3]: xcs_dg3_pim.y.presets.state()
Out[3]: 'Unknown'

In [4]: xcs_dg3_pim.y._position = None

In [5]: xcs_dg3_pim.y.wm_ccm()
Out[5]: 'Unknown'

In [6]: xcs_dg3_pim.y.presets.state()
Out[6]: 'Unknown'
```

This PR

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -1331,7 +1331,7 @@ class Presets:
                 state_name = method_name.replace('wm_', '', 1)
                 wm_state = getattr(device, method_name)
                 state_val = wm_state()
-                if not isinstance(state_val, (int, float)):
+                if not isinstance(state_val, numbers.Real):
                     continue
                 diff = abs(state_val)
                 if diff < closest:
--- a/pcdsdevices/interface.py
+++ b/pcdsdevices/interface.py
@@ -1277,14 +1277,22 @@ class Presets:
 
             Returns
             -------
-            offset : float
+            offset : Optional[float, str]
                 How far we are from the preset position. If this is near zero,
                 we are at the position. If this positive, the preset position
                 is in the positive direction from us.
+                If the current position is unknown, return "Unknown".
             """
+            preset_pos = self.presets._cache[preset_type][name]['value']
+            # here we expect self: FltMvInterface
+            curr_pos = self.wm()
+            try:
+                pos = preset_pos - curr_pos
+            except Exception:
+                # current position is not known or unset
+                pos = "Unknown"
 
-            pos = self.presets._cache[preset_type][name]['value']
-            return pos - self.wm()
+            return pos
 
         wm_pre.__doc__ = wm_pre.__doc__.format(name)
         return wm_pre
@@ -1322,7 +1330,10 @@ class Presets:
             if method_name.startswith('wm_'):
                 state_name = method_name.replace('wm_', '', 1)
                 wm_state = getattr(device, method_name)
-                diff = abs(wm_state())
+                state_val = wm_state()
+                if not isinstance(state_val, (int, float)):
+                    continue
+                diff = abs(state_val)
                 if diff < closest:
                     state = state_name
                     closest = diff
--- a/docs/source/upcoming_release_notes/1319-bug_preset_error.rst
+++ b/docs/source/upcoming_release_notes/1319-bug_preset_error.rst
@@ -1,0 +1,30 @@
+1319 bug_preset_error
+#####################
+
+API Breaks
+----------
+- N/A
+
+Library Features
+----------------
+- N/A
+
+Device Features
+---------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- Improves error handling for presets when the position is unknown or uninitialized.
+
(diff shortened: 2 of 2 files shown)


pcdsdevices/interface.py
```diff
@@ -1322,7 +1330,10 @@ def state(self):
             if method_name.startswith('wm_'):
                 state_name = method_name.replace('wm_', '', 1)
                 wm_state = getattr(device, method_name)
-                diff = abs(wm_state())
+                state_val = wm_state()
+                if not isinstance(state_val, (int, float)):
```
ZLLentz, line 1334: We don't ever get numpy floats in these, right?

pcdsdevices/interface.py
```diff
@@ -1277,14 +1277,22 @@ def wm_pre(self):
...
+                If the current position is unknown, return "Unknown".
             """
+            preset_pos = self.presets._cache[preset_type][name]['value']
+            # here we expect self: FltMvInterface
+            curr_pos = self.wm()
+            try:
+                pos = preset_pos - curr_pos
+            except Exception:
+                # current position is not known or unset
+                pos = "Unknown"
```
ZLLentz, line 1293: This approach is definitely better than the status quo, but I have to wonder if a cleaner approach would have been to make a custom error message.

It's not clear to me.

At any rate, these are intended more for interactive behavior at beamlines, so eschewing an exception for a str output will often be appreciated.

pcdsdevices/interface.py
tangkong: Unsure, but it's probably prudent to future-proof ourselves.  I thought about this for half a second then forgot about it.  `numbers.Number` should work.

```python
In [5]: for n in [1, 1.2, np.int32(3.3), np.float64(3.2)]:
   ...:     print(isinstance(n, numbers.Number))
   ...: 
True
True
True
True
```

pcdsdevices/interface.py
tangkong: That was my reasoning here as well.  Thinking about the actual issue that spawned this bug report (overloaded callback queue delaying position update), an uninitialized position could be caused by a multitude of causes.  I think it might be hard to provide anything more useful than the "position is unknown" in an error message

ZLLentz (approved):

Issue #1325, opened 2025-03-04
Add embedded screens for SmarActEncodedTipTilts


aberges-SLAC:
SmarAct encoded tip-tilts bundle both the `open-loop` and `closed-loop` component signals and needs a screen that handles these nuances correctly. 

The idea would be to have a closed-loop tab for motor tweaks, accompanying expert screen buttons, and single-buttons for doing the `home` and `calibration` sequences unique to these stages.

A second tab that is essentially `SmarActTipTilt.embedded` would be more than sufficient.

PR #1326, opened 2025-03-04
SmarAct Encoded Tip Tilt Embedded Screens

### LINKED ISSUE #1268: Custom "Home" button and sequence for SmarAct Encoded Tip Tilt
## Current Behavior
Due to the peculiar "referencing" sequence involved in "homing" SmarAct encoded tip-tilt stages, it would be very useful to capture this [procedure](https://confluence.slac.stanford.edu/display/PCDS/Calibration%2C+Referencing%2C+And+Position+Holding) in a single "home" button that will prevent future confusion/errors.

## Expected Behavior
We can ideally overwrite the `home_forward` and `home_reverse` signals that the SmarAct class inherits and replace it with a single button that does the following:
**Axis 1**
1. `home` in the "reverse" direction, towards the end stop
2. step forward in open-loop mode (so that you are away from the mechanical end stop)
3. `home_reverse` again
**Axis 2**
4. `home` in the "reverse" direction towards the end stop
5. step forward in open-loop mode
6. `home_reverse` again
7. Repeat steps 1-3 on **Axis 1** again. 

This could also be similarly executed for `calibration` sequences

## Context / environment
The referencing & calibration procedure can be cumbersome for SmarAct encoded tip-tilt devices. Since the mechanical end stop is the only _valid_ reference specified by the vendor, there is no need for a `home_forward` and `home_reverse` in the usual context. This sequence for calibration and `home` can be substituted instead.

Tyler has something similar cooking for the btms gui which prompted this issue being raised.

aberges-SLAC:
## Description
Adds:
* SmarActEncodedTipTilt.embedded.ui
* SmarActEncodedTipTilt.embedded.py

New features beyond typical motor motion:
* Homing sequence for SmarAct STT##.#iv stages (ref A -> ref B -> ref A)
* Calibration sequence for SmarAct STT##.#iv stages (calib A -> ref A -> calib B -> ref B -> calib A -> ref A)

Couple caveats:
Don't love that I'm reusing classes and code from `SmarActTipTilt` but the alternatives for importing it as a module are hairy at best. 

## Motivation and Context
As per issue #1325, we need an embedded screen that successfully bundles the open & closed loop motions for these stages, and I have done just that! It also closes #1268

## How Has This Been Tested?
I've tested this screen and code out on an STT50.8iv stage in the LRL. See video of the demo below!

If you want to take it for a spin, you can use this script: `/cds/home/a/aberges/bin/test_tip_tilt`

Documented in the code, release notes, and this PR!

https://github.com/user-attachments/assets/2853471e-bb38-4fbe-9e76-78c95aa327a4

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1325-SmarActEncodedTipTilt.rst
+++ b/docs/source/upcoming_release_notes/1325-SmarActEncodedTipTilt.rst
@@ -12,8 +12,9 @@ Library Features
 Device Features
 ---------------
 - SmarActEncodedTipTilt: added embedded screens
-- - bundles open-loop and closed-loop motion
-- - Wraps the home and calibrate sequence into single methods
+- SmarActEncodedTipTilt: bundles open-loop and closed-loop motion
+- SmarActEncodedTipTilt: wraps the home and calibrate sequence into single methods
+- SmarActEncodedTipTilt: add channel_state_raw, calibrating, and referencing signals
 
 New Devices
 -----------
--- a/docs/source/upcoming_release_notes/1325-SmarActEncodedTipTilt.rst
+++ b/docs/source/upcoming_release_notes/1325-SmarActEncodedTipTilt.rst
@@ -11,9 +11,9 @@ Library Features
 
 Device Features
 ---------------
-- Add the SmarActEncodedTipTilt.embedded screens
-- bundles open-loop and closed-loop motion
-- Wraps the home and calibrate sequence into single methods
+- SmarActEncodedTipTilt: added embedded screens
+- - bundles open-loop and closed-loop motion
+- - Wraps the home and calibrate sequence into single methods
 
 New Devices
 -----------
--- a/docs/source/upcoming_release_notes/1325-SmarActEncodedTipTilt.rst
+++ b/docs/source/upcoming_release_notes/1325-SmarActEncodedTipTilt.rst
@@ -1,0 +1,32 @@
+1325 SmarActEncodedTipTilt
+#################
+
+API Breaks
+----------
+- N/A
+
+Library Features
+----------------
+- N/A
+
+Device Features
+---------------
+- Add the SmarActEncodedTipTilt.embedded screens
+- bundles open-loop and closed-loop motion
+- Wraps the home and calibrate sequence into single methods
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- aberges-SLAC
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -1663,11 +1663,6 @@ class SmarAct(EpicsMotorInterface):
                       doc='Temperature of the MCS2 Module in the rack')
     channel_state_raw = Cpt(EpicsSignalRO, ':STATE_RBV', kind='omitted',
                             doc='Channel state bitmask represented as raw int')
-    # Only valid if using common IOC R1.0.25 or greater, else use channel_state_raw
-    calibrating = Cpt(EpicsSignalRO, ':STATE_RBV.B2', kind='omitted',
-                      doc='Channel is busy calibrating')
-    referencing = Cpt(EpicsSignalRO, ':STATE_RBV.B3', kind='omitted',
-                      doc='Channel is busy referencing')
 
     # These PVs will probably not be needed for most encoded motors, but can be
     # useful
@@ -1707,8 +1702,6 @@ class SmarAct(EpicsMotorInterface):
         self.dist_code_inv.long_name = 'Distance Code Inversion'
         self.channel_temp.long_name = 'Channel Temp. (°C)'
         self.module_temp.long_name = 'Module Temp. (°C)'
-        self.calibrating.long_name = 'Calibrating'
-        self.referencing.long_name = 'Referencing'
         self.channel_state_raw.long_name = 'Channel State'
 
 
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -1661,6 +1661,13 @@ class SmarAct(EpicsMotorInterface):
                        doc="Temperature at the channel's amplifier")
     module_temp = Cpt(EpicsSignalRO, ':MODTEMP', kind='normal',
                       doc='Temperature of the MCS2 Module in the rack')
+    channel_state_raw = Cpt(EpicsSignalRO, ':STATE_RBV', kind='omitted',
+                            doc='Channel state bitmask represented as raw int')
+    # Only valid if using common IOC R1.0.25 or greater, else use channel_state_raw
+    calibrating = Cpt(EpicsSignalRO, ':STATE_RBV.B2', kind='omitted',
+                      doc='Channel is busy calibrating')
+    referencing = Cpt(EpicsSignalRO, ':STATE_RBV.B3', kind='omitted',
+                      doc='Channel is busy referencing')
 
     # These PVs will probably not be needed for most encoded motors, but can be
     # useful
@@ -1700,6 +1707,9 @@ class SmarAct(EpicsMotorInterface):
         self.dist_code_inv.long_name = 'Distance Code Inversion'
         self.channel_temp.long_name = 'Channel Temp. (°C)'
         self.module_temp.long_name = 'Module Temp. (°C)'
+        self.calibrating.long_name = 'Calibrating'
+        self.referencing.long_name = 'Referencing'
+        self.channel_state_raw.long_name = 'Channel State'
 
 
 class SmarActEncodedTipTilt(Device):
--- a/pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
+++ b/pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
@@ -277,8 +277,8 @@ class SmarActEncodedTipTiltWidget(Display, utils.TyphosBase):
             _reverse.set_channel('')
 
         _prefix = getattr(self.device, axis).prefix
-        _forward = getattr(self.ui, f'{axis}_fwd')
-        _reverse = getattr(self.ui, f'{axis}_rev')
+        _forward = getattr(self.ui, f'{axis}_jog_fwd')
+        _reverse = getattr(self.ui, f'{axis}_jog_rev')
 
         if invert:
             clear_channels()
--- a/pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
+++ b/pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
@@ -1,0 +1,376 @@
+from __future__ import annotations
+
+import logging
+from time import sleep
+
+import pydm
+from ophyd.status import MoveStatus
+from pydm import Display
+from pydm.utilities import IconFont
+from qtpy import QtCore, QtGui, QtWidgets
+from typhos import utils
+from typhos.related_display import TyphosRelatedSuiteButton
+
+logger = logging.getLogger(__name__)
+
+
+class _SmarActEncodedTipTiltEmbeddedUI(QtWidgets.QWidget):
+    """Annotations helper for SmarActEncodedTipTilt.embedded.ui. Do not instantiate."""
+    # Open-loop
+    dpad_open_loop_label: QtWidgets.QLabel
+    tip_jog_fwd: pydm.widgets.pushbutton.PyDMPushButton
+    tip_jog_rev: pydm.widgets.pushbutton.PyDMPushButton
+    tip_step_count: pydm.widgets.label.PyDMLabel
+    tilt_jog_fwd: pydm.widgets.pushbutton.PyDMPushButton
+    tilt_jog_rev: pydm.widgets.pushbutton.PyDMPushButton
+    tilt_step_count: pydm.widgets.label.PyDMLabel
+    tip_calibrated_led: pydm.widgets.byte.PyDMByteIndicator
+    tip_homed_led: pydm.widgets.byte.PyDMByteIndicator
+    # Closed-loop
+    dpad_closed_loop_label: QtWidgets.QLabel
+    tip_tweak_fwd: pydm.widgets.pushbutton.PyDMPushButton
+    tip_tweak_rev: pydm.widgets.pushbutton.PyDMPushButton
+    tip_tweak_value: QtWidgets.QLineEdit
+    tilt_tweak_fwd: pydm.widgets.pushbutton.PyDMPushButton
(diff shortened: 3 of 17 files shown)


pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
```diff
@@ -0,0 +1,607 @@
...
+    def run(self):
+        stage = self._motor
+        sequence = [stage.tip, stage.tilt, stage.tip]
+        progress = 0
+
+        for axis in sequence:
+            # Calibrate first
+            self._status = axis.do_calib.put(1)
+            # Have to manually sleep :[
+            sleep(5)
```
tangkong, line 121: Can we get an explanation of why you need to sleep? It'd be nice if we could instead wait on some signal to let us know that calibration is done, because as is this is a brittle, hard-coded 5s sleep

pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
```diff
@@ -0,0 +1,607 @@
...
+        """
+
+        if self.device is None:
+            print('No device set!')
+            return
+
+        def set_open_loop(self, axis: str):
+            """
+            A wrapper to set the open-loop widget channels.
+            Ironically more lines than just hard coding it.
```
tangkong, line 199: But easier to extend! Maybe :)

pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
```diff
@@ -0,0 +1,607 @@
...
+
+    @QtCore.Slot()
+    def _tip_tweak_fwd(self):
+        """Tweak positive by the amount listed in ``ui.tip_tweak_set``"""
+        try:
+            self.tweak_setpoint(self.device.tip,
+                                float(self.ui.tip_tweak_value.text()))
+        except Exception:
+            logger.exception('Positive tweak failed on tip')
+            return
```
tangkong, line 317: You technically don't need these, if you have no returns by the end of the function, it will return None for you

tangkong (commented):
All the duplication is annoying, as you point out.  I see a couple of ways to approach this, none of which need be implemented in this PR:
* Do some funky python-y import stuff by calling out the file name specifically
* Move all of these into some shared widget library, maybe `typhos` itself or possibly reviving [`pcdswidgets`](https://github.com/pcdshub/pcdswidgets).
  * You'd then import those widgets from these files, just assuming that pcdswidgets is installed.  (we don't list it as a dependency, similar to your use of typhos/pydm here)

All that aside, I think this looks pretty reasonable at a first readthrough.  I'll take it for a spin and see how it looks

pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
aberges-SLAC: Ah, very true. I think I had this in there for what seems like superstitious reasons at this point. I'll clear it

aberges-SLAC:
Oh, lol, my commits from my other PR snuck in. We should delete those from this PR

tangkong (approved):
Took another look after taking it for a spin, I'm still on board with this once some of the nits are resolved.

Looking back I think they're not enough to really hold this back, so I'll leave my check before I leave.  

If you end up vendoring any of the positioner row logic again, we should look into a way to make those utilities more reusable

pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
```diff
@@ -0,0 +1,607 @@
...
+    tilt_rbv: pydm.widgets.label.PyDMLabel
+    tilt_calibrated_led: pydm.widgets.byte.PyDMByteIndicator
+    tilt_homed_led: pydm.widgets.byte.PyDMByteIndicator
+    # General
+    home_button: QtWidgets.QPushButton
+    calibrate_button: QtWidgets.QPushButton
+    sequence_progress_bar: QtWidgets.QProgressBar
+    tip_expert_button: TyphosRelatedSuiteButton
+    tilt_expert_button: TyphosRelatedSuiteButton
+    devices: list
```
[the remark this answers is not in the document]
aberges-SLAC: Ah, this is probably left in from a previous iteration of the class. I can probably delete it because that context help is meaningless now.

pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
```diff
@@ -0,0 +1,607 @@
...
+        stage = self._motor
+        sequence = [stage.tip, stage.tilt, stage.tip]
+        progress = 0
+
+        for axis in sequence:
+            # Calibrate first
+            self._status = axis.do_calib.put(1)
+            # Have to manually sleep :[
+            sleep(5)
+            progress += 1
```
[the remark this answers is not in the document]
aberges-SLAC: Right! Much less elegant than what you were doing in BTMS-gui, but it works and I don't have to have separate threads

docs/source/upcoming_release_notes/1325-SmarActEncodedTipTilt.rst
```diff
@@ -0,0 +1,32 @@
...
+----------
+- N/A
+
+Library Features
+----------------
+- N/A
+
+Device Features
+---------------
+- Add the SmarActEncodedTipTilt.embedded screens
```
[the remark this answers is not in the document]
aberges-SLAC: What do you mean?

No devices were technically modified, just added `.ui` and `.py` files. Typhos backend will just prioritized these screens if the screen name == device class (i.e. `pcdsdevices.epics_motor.SmarActEncodedTipTilt`).

pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
aberges-SLAC: You know, what I _could_ do (and is probably smarter) is just add some `calibrated` signal to the `SmarAct` closed loop device class that reads the `:STATE_RBV` pv and checks the bit for "is calibrated" like I do manually on the LEDs.

Caveat is that it breaks on older common IOC versions because it was originally an `mbbi` record and not `mbbidirect`.

docs/source/upcoming_release_notes/1325-SmarActEncodedTipTilt.rst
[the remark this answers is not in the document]
slactjohnson: Yeah, it's a really nitty nit. Generally we just put the class(es) related to the PR in front of the description, e.g. 
```SmarActEncodedTipTilt: Added embedded screens```

File: pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py (hunk 1 above)
Comment (aberges-SLAC) [reply pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py:121]: I guess I could build one into a func for it?

But this method is actually flawed too because if you attempt to re-calibrate a stage you'll get a false-positive.

Also, example of the `motor_is_moving` signal not doing its job:

https://github.com/user-attachments/assets/00db17e2-65a4-4dc0-8f91-ce5f1109bae4

Comment (aberges-SLAC) [reply pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py:121]: Ah, but if I hardcode an initial `sleep(1)` into `wait_on_calib` to wait for `motor_is_moving` to change, then it works.

_However_ the first `motor_is_moving` for an axis gets missed, apparently. So if it is the first time calling `wait_on_calib` for that axis, it'll just see `motor_is_moving == 0` which is really annoying.

Regardless, I wasn't too far off with my hardcoded estimate. Looks like it takes on average 3.8-4 seconds.

File: pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py [hunk 7]
```diff
@@ -0,0 +1,607 @@
...
+
+        We want to watch for this resize and set the scroll area width such
+        that there's enough room to see the widget at its minimum size.
+        --------------------------------------------------------------------
+        Also shamelessly stolen from TyphosPositionerRow
+        """
+        if (self.tip_panel.minimumWidth() <= self.tip_panel.original_panel_min_width or
+                self.tilt_panel.minimumWidth() <= self.tilt_panel.original_panel_min_width):
+            # No change
+            self.resize_timer.start()
```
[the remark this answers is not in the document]
tangkong: QT is a demon sometimes

aberges-SLAC:
Okay, new logic for waiting on calibration sequence. Slightly less brittle, but maybe still silly.

pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
```diff
@@ -0,0 +1,627 @@
...
+    """
+    def run(self):
+        def is_calibrating(device: any):
+            """
+            Check the 2nd bit in the channel state to see if the stage is
+            currently in the calibration sequence.
+            """
+            _state_raw = device.channel_state_raw.get()
+            # use this if the IOC is > R1.0.25
+            if device.calibrating.connected:
```
aberges-SLAC, line 119: Turns out this breaks if `device.calibrating` fails to connect, because then the component device itself breaks and throws `ophyd.utils.errors.DisconnectedError`.

I suppose I can just _only_ do it the manual way so that it is backwards compatible?

pcdsdevices/epics_motor.py
```diff
@@ -1661,6 +1661,13 @@ class SmarAct(EpicsMotorInterface):
                        doc="Temperature at the channel's amplifier")
     module_temp = Cpt(EpicsSignalRO, ':MODTEMP', kind='normal',
                       doc='Temperature of the MCS2 Module in the rack')
+    channel_state_raw = Cpt(EpicsSignalRO, ':STATE_RBV', kind='omitted',
+                            doc='Channel state bitmask represented as raw int')
+    # Only valid if using common IOC R1.0.25 or greater, else use channel_state_raw
+    calibrating = Cpt(EpicsSignalRO, ':STATE_RBV.B2', kind='omitted',
```
aberges-SLAC, line 1667: These seem to be the source of the `ophyd.utils.errors.DisconnectedError` for older versions of the common IOC. Perhaps I should just omit them entirely?

tangkong: > Notes
    -----
    This should be considered for internal use only, and not for
    user-facing device components.  If you use this in your new device,
    there is a good chance we will reject your PR.

I'm inclined to trust the warnings of our forefathers here.  This signal provides a signal regardless of whether or not the backing PV is connected, which will be nothing but confusing for users and me during debugging

Is the implication that some iocs have this PV but some don't?  I think that normally we try to set up different devices if the backing IOC's are different.

slactjohnson: Makes sense, I hadn't considered that the signal would still exist regardless of whether or not the PV exists.

aberges-SLAC:
To avoid a separate PR for the same request, I add the invert axis logic to and checkboxes to this screen too

aberges-SLAC:
https://github.com/slaclab/pydm/issues/1186

This issue plagues this screen when the `expert` screens are opened :/

aberges-SLAC:
Added the invert jog workaround from #1331 and made the same UI changes to reflect user requests. This is _finally_ done in my opinion

aberges-SLAC:
@tangkong Ready for review [:

pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
```diff
@@ -0,0 +1,682 @@
...
+    in the parent.
+    """
+    def run(self):
+        stage = self._motor
+        sequence = [stage.tip, stage.tilt, stage.tip]
+        progress = 0
+
+        for axis in sequence:
+            self._status = axis.home('reverse', wait=True)
+            self._status.wait()
```
slactjohnson, line 99: Is a timeout on this status object worth considering here?

pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
```diff
@@ -0,0 +1,682 @@
...
+            return (_state_raw & (1 << 2)) > 0
+
+        def wait_on_calib(device):
+            """
+            Wait for the calibration sequence to finish.
+            Needs to initially sleep for a second to let records update.
+            """
+            sleep(1)
+            while is_calibrating(device):
+                sleep(0.2)
```
slactjohnson, line 132: Similar to above, should we timeout at some point?

pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
```diff
@@ -0,0 +1,682 @@
...
+        """
+        Spawn a confirmation-dialog box similar to PyDMPushButtons.
+        """
+        button = self.ui.home_button
+        offset = button.mapToGlobal(QtCore.QPoint(0, 0))
+
+        _msg = QtWidgets.QMessageBox()
+        _msg.setWindowTitle('Warning')
+        _msg.setText('Are you sure you want to proceed?')
+        _msg.setInformativeText('This will move both axes to their mechanical end-stop.')
```
slactjohnson, line 454: This seems incorrect when used in `confirm_calibrate`, right? You could make the informative text a parameter you pass to this to make things a littler more customizable for the different situations you have here.

pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
aberges-SLAC: The wait _has_ to be called, otherwise the thread will try to home the other axis while the first is still moving, which is dangerous for the tip-tilt stages.

pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
aberges-SLAC: I'm not sure — if the signals are disconnected in the first place, this thread fails and doesn't get trapped in the while loop. Or rather, you just can't even call the method in the first place.

I suppose it _is_ possible for something to happen to the device immediately after the call and you could possibly get trapped in the while loop, but I thought `is_calibrating` would throw an exception and break the loop in the first place. 

Calibrating takes roughly 3-5 secs, so a 5 sec timeout would probably work?

pcdsdevices/ui/SmarActEncodedTipTilt.embedded.py
aberges-SLAC: The `calibrate` sequence necessarily homes the stages, so the same warning is valid. As per vendor instructions you must:
1) Calibrate A
2) Home reverse A
3) Calibrate B
4) Home reverse B
5) Calibrate A (again)
6) Home reverse B (again)

slactjohnson (approved):
I think all my comments have been addressed. LGTM. Nice work!

PR #1329, opened 2025-03-05
SmarActTipTilt updates

aberges-SLAC:
## Description
Add the wrapper approach from SmarActEncodedTipTilt when adding PVs to layout objects. Technically more verbose, but more scalable?

Also adding line-edits for `step_size` signals without having to open the pop-up screen.

## Motivation and Context
Laser folks pointed out a good point about the tip-tilts:
* The tilt axis affects vertical, and while the vendor defines positive = piezo pushes away from the rear mechanical end stop, this necessarily steers the beam down
* Users want to be able to dictate if the "up" button corresponds to the motion they expect on some detector that is monitoring the steering
* Periscopes make simple inversion of `STEP_FORWARD` and `STEP_REVERSE` complicated, and we don't want to be too granular.

Solution:
Added the `step_size` signals to the embedded screen so users could invert the axes at will.

Since `STEP_FORWARD` and `STEP_REVERSE` in the ioc already handle the sign of the open-loop step, giving a negative `step_size` has the same effect as inverting the axis. This also prevents us from having to obfuscate the `pcdsdevices` class and the `TOTAL_STEP_COUNT` pv's previous behavior is maintained.

## How Has This Been Tested?
As always, tested this on my labrat STT50.8iv in the LRL. Behaves as expected.

In the PR and in the code!

Looks like this now:
image

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1329-SmarActTipTilt_update.rst
+++ b/docs/source/upcoming_release_notes/1329-SmarActTipTilt_update.rst
@@ -1,0 +1,30 @@
+1329 SmarActTipTilt_update
+#################
+
+API Breaks
+----------
+- N/A
+
+Library Features
+----------------
+- N/A
+
+Device Features
+---------------
+- SmarActTipTilt: added step_size to embedded screen
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- aberges-SLAC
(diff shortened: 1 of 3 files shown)


tangkong (dismissed):
Seems generally agreeable.  It may not be the most important thing to do, but if you like this better then go for it haha.

The old way is arguably more readable, but not by too much I think

tangkong:
Tests are failing for no real reason... I'm taking a look

aberges-SLAC:
> Seems generally agreeable. It may not be the most important thing to do, but if you like this better then go for it haha.
> 
> The old way is arguably more readable, but not by too much I think

My biggest pet peeve with it currently is that the directional pad is not centered with the pydm label/line edits now. I _might_ just have to get rid of the labels on the left hand side to fix that? Couldn't get the `pydm.LineEdits` to actually respect size hints on the widths beyond expanding.

slactjohnson (approved):
No complaints from me

tangkong:
Hah procrastination wins again.  I brainstormed a few ideas for how to deal with the tests but punted the actual effort to the next day.  And they work again! 😭

tangkong (approved):

slactjohnson (approved):

PR #1331, opened 2025-03-10
SmarActTipTilt Invert Open-loop Axes

aberges-SLAC:
## Description
1) Got rid of some text labels for the embedded screen to reduce footprint.
2) Added checkboxes for inverting the `tip` or `tilt` axes
3) Add logic to change which channel is assigned to which `pushbutton` for jogging the each stage. This lets the user functionally, and in real time, invert the directional pad to suit their needs/alignment/design.

Found a _neat_ ~~bug~~ feature in `pydm` when using `set_channel` sequentially on objects who may or may not have those channels set already. If you are re-assigning the channels on a pydm widget, explicitly set the channels to blank strings before proceeding or you _will_ have callback bugs.

## Motivation and Context
Unfortunately #1329 did not fix the issue the Laser team presented, due to how the protocol file works in the common IOC. Turns out `asyn` doesn't do math, who'd've thunk 🙃 

The real solution I was dragging my feet on was adding a user-widget to invert the axis by swapping what channels correspond to which button. 

## How Has This Been Tested?
Extensively tested using the STT50.8iv work horse in the laser lab. See below.

In the code and in this PR!

https://github.com/user-attachments/assets/8014e783-ba49-4d89-a118-a1c9b8450efd

UPDATE: 
New layout based on user request
image

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/ui/SmarActTipTilt.embedded.py
+++ b/pcdsdevices/ui/SmarActTipTilt.embedded.py
@@ -19,9 +19,11 @@ class _SmarActTipTiltEmbeddedUI(QtWidgets.QWidget):
     tip_forward: pydm.widgets.pushbutton.PyDMPushButton
     tip_reverse: pydm.widgets.pushbutton.PyDMPushButton
     tip_step_count: pydm.widgets.label.PyDMLabel
+    invert_tip: QtWidgets.QCheckBox
     tilt_forward: pydm.widgets.pushbutton.PyDMPushButton
     tilt_reverse: pydm.widgets.pushbutton.PyDMPushButton
     tilt_step_count: pydm.widgets.label.PyDMLabel
+    invert_tilt: QtWidgets.QCheckBox
     settings_button: pydm.widgets.pushbutton.PyDMPushButton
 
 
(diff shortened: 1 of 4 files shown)


pcdsdevices/ui/SmarActTipTilt.embedded.py
```diff
@@ -119,6 +121,60 @@ def set_open_loop(self, axis: str):
...
+    def invert_axis(self, axis: str, invert: bool = False):
+        """
+        Invert the jog pushbuttons
+
+        Parameters
+        -----------
+            axis: str
+                Name of axis
+            invert: bool
+                Whether or not to invert. Default is False.
```
tangkong, line 136:
```suggestion
        axis: str
            Name of axis
        invert: bool
            Whether or not to invert. Default is False.
```
Giga-nit: I think these are supposed to be de-dented one level

pcdsdevices/ui/SmarActTipTilt.embedded.py
```diff
@@ -119,6 +121,60 @@ def set_open_loop(self, axis: str):
...
+        """
+
+        def clear_channels():
```
tangkong, line 139: python scoping is weird, but I guess this works.  I'd probably have defined this after the non-local variables that we're grabbing, but it's not an issue

tangkong (approved):
From the trenches of this bugfix I also approve this. 👍 Hard fought

slactjohnson (approved):
This is cool. LGTM. Nice work!

slactjohnson:
Is the PyDM feature worthy of submitting an issue?

pcdsdevices/ui/SmarActTipTilt.embedded.py
aberges-SLAC: Valid, I'll gladly accept

pcdsdevices/ui/SmarActTipTilt.embedded.py
aberges-SLAC: Yeah, it confuses me often but it worked well enough 🤷 I have a bad habit of always defining the subfunction before doing anything so that you don't have to go searching for it.

I think subfunction scoping only checks the parent `func` vars so it _typically_ is not an issue in the handful of cases I do it.

aberges-SLAC:
> Is the PyDM feature worthy of submitting an issue?

[Done](https://github.com/slaclab/pydm/issues/1186)

aberges-SLAC:
@tangkong Reminder: save this merge into `dev_conda` for next week's PAMM so it doesn't rustle any jimmies

pcdsdevices/ui/SmarActTipTilt.embedded.py
```diff
@@ -104,13 +106,18 @@ def set_open_loop(self, axis: str):
...
+            _open_loop_dict = {'forward': '_jog_fwd',
+                               'reverse': '_jog_rev',
                                'step_count': ':TOTAL_STEP_COUNT',
                                'step_size': ':STEP_COUNT'}
             for obj, _suffix in _open_loop_dict.items():
                 _widget = getattr(self.ui, f'{axis}_{obj}')
-                _widget.set_channel(f'ca://{_prefix}{_suffix}')
+                if ':' in _suffix:
+                    _widget.set_channel(f'ca://{_prefix}{_suffix}')
+                elif type(_widget) is pydm.widgets.pushbutton.PyDMPushButton:
```
tangkong, line 117: python has `isinstance` for this purpose.

Also at the point where you're parsing your `suffix` to see if it has `:`, I think it makes more sense to just enumerate the connections and channels explicitly.

tangkong (dismissed):
Took it for a quick spin and confirmed the buttons jog the way they're supposed to.  You're right that at this point we should just be using QPushButtons, but this works just fine.

pcdsdevices/ui/SmarActTipTilt.embedded.py
aberges-SLAC: Idk why I'm always averse to `isInstance` in these situations, it's literally the right approach.

I suppose the right way to do this would've just been to check the `_widget` for being a `PyDMPushButton` in this context, then just defaulting to setting the channel otherwise. No real need to parse `suffix` in reality.

aberges-SLAC: Also, `PyDMPushButtons` still have one advantage over `QPushButton`: you get to save like 4-6 lines of code for setting and coloring your icons 🙃

tangkong (approved):
LGTM

slactjohnson (approved):

PR #1332, opened 2025-03-12
ENH: set mpod voltage limits based on parent module

tangkong:
## Description
Sets channel limits based on the value obtained from `.max_voltage` and the model number from the parent module.

## Motivation and Context
In response to TMO's recent troubles setting the voltage limits on the mpod channels.  Our previous application of the hotfix (#1293) only assumed positive definite voltage limits, and failed to cover the negative or bipolar channels.

According to James, the simplest way to grab this information is from the model number of the parent module.  All channels in a given module have the same polarity.

Also, it seems our approach to imposing limits on the MPOD channels is to monitor the max voltage PV and set the limits in python-land.  This was a bit misleading, and I'd like to document this more thoroughly (TODO)

## How Has This Been Tested?
Interactively, though tests will be added I swear.

This PR

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/mpod_apalis.py
+++ b/pcdsdevices/mpod_apalis.py
@@ -1,6 +1,5 @@
 import logging
 import time
-from enum import Enum, auto
 
 import numpy as np
 from ophyd.device import Component as Cpt
@@ -139,19 +138,8 @@ class MPODApalisChannel(BaseInterface, Device):
 
         If not, defaults to setting limits to positive definite
         """
-        polarity = getattr(self, "biological_parent.polarity", Polarity.POSITIVE)
         limit_scales = getattr(self, "biological_parent.limit_scales", (0, 100))
         limits = [lim * self._max_voltage / 100 for lim in limit_scales]
-
-        if polarity is Polarity.BIPOLAR:
-            if 0 in limits:
-                logger.debug(f"Bipolar limits expected, got: ({limits})")
-        elif polarity is Polarity.NEGATIVE:
-            if all(lim >= 0 for lim in limits):
-                logger.debug(f"Negative limits expected, got: ({limits})")
-        elif polarity is Polarity.POSITIVE:
-            if all(lim <= 0 for lim in limits):
-                logger.debug(f"Positive limits expected, got: ({limits})")
 
         self.voltage._override_metadata(
             lower_ctrl_limit=min(limits),
@@ -191,12 +179,6 @@ def _put_clamped(signal: EpicsSignal, value: float) -> None:
         value = high_val
 
     signal.put(value)
-
-
-class Polarity(Enum):
-    POSITIVE = auto()
-    NEGATIVE = auto()
-    BIPOLAR = auto()
 
 
 class MPODApalisModule(BaseInterface, GroupDevice):
@@ -248,9 +230,6 @@ class MPODApalisModule(BaseInterface, GroupDevice):
                  write_pv=':Control:doClear',
                  kind='normal', string=True,
                  doc='Clears all MPOD module faults')
-
-    model = Cpt(EpicsSignalRO, ':Model', kind='omitted', string=True,
-                doc="Model number")
 
     limit_pos = Cpt(EpicsSignalRO, ':VoltageLimit', kind='omitted',
                     doc='Positive voltage limit as a % of the max voltage')
--- a/pcdsdevices/mpod_apalis.py
+++ b/pcdsdevices/mpod_apalis.py
@@ -132,14 +132,12 @@ class MPODApalisChannel(BaseInterface, Device):
 
     def _update_max_voltage_limits(self) -> None:
         """
-        Set explicit symmetric limits on voltage for better error reporting.
+        Update limits on voltage, referring to the parent module if it exists
 
-        Refers to the parent module if it exists, to determine the limit scaling
-
-        If not, defaults to setting limits to positive definite
+        If parent module does not exist, defaults limits to (0, 100)
         """
-        limit_scales = getattr(self, "biological_parent.limit_scales", (0, 100))
-        limits = [lim * self._max_voltage / 100 for lim in limit_scales]
+        limit_pct = getattr(self, "biological_parent.limit_percents", (0, 100))
+        limits = [lim * self._max_voltage / 100 for lim in limit_pct]
 
         self.voltage._override_metadata(
             lower_ctrl_limit=min(limits),
@@ -281,13 +279,13 @@ class MPODApalisModule(BaseInterface, GroupDevice):
         self.current_ramp_speed.put(ramp_speed)
 
     @property
-    def limit_scales(self) -> tuple[float, float]:
+    def limit_percents(self) -> tuple[float, float]:
         """
-        Returns limit scaling tuple.  When multiplied by the channel's max_voltage
-        signal (:VoltageNominal) will produce the actual limits for the channel
+        Returns limit percentage tuple.  Represents the % of the channel's max_voltage
+        signal (:VoltageNominal) to be used as the actual limits for the channel
 
         Returns
-        -------
+        -------p
         tuple[float, float]
             The voltage limits as a proportion of the max_voltage
         """
--- a/pcdsdevices/mpod_apalis.py
+++ b/pcdsdevices/mpod_apalis.py
@@ -2,6 +2,7 @@ import logging
 import time
 from enum import Enum, auto
 
+import numpy as np
 from ophyd.device import Component as Cpt
 from ophyd.device import Device
 from ophyd.signal import EpicsSignal, EpicsSignalRO
@@ -265,7 +266,10 @@ class MPODApalisModule(BaseInterface, GroupDevice):
         super().__init__(*args, **kwargs)
         self._limit_pos = 0.0
         self._limit_neg = 0.0
-        self._model_str = ''
+        # tolerances for voltage limit tolerances (expected 0-100)
+        # .5% relative change (of original value), and .5% absolute
+        self._limit_rtol = 0.005
+        self._limit_atol = 0.5
 
     def clear_faults(self):
         """Clears all module faults"""
@@ -297,21 +301,6 @@ class MPODApalisModule(BaseInterface, GroupDevice):
         """
         self.current_ramp_speed.put(ramp_speed)
 
-    @model.sub_value
-    def _stash_model(self, value, **kwargs) -> None:
-        self._model_str = self.model.get()
-
-    @property
-    def polarity(self) -> Polarity:
-        if self._model_str.endswith("x"):
-            return Polarity.BIPOLAR
-        elif self._model_str.endswith("p"):
-            return Polarity.POSITIVE
-        elif self._model_str.endswith("n"):
-            return Polarity.NEGATIVE
-
-        raise ValueError(f"Unable to determine module polarity: {self._model_str}")
-
     @property
     def limit_scales(self) -> tuple[float, float]:
         """
@@ -337,11 +326,17 @@ class MPODApalisModule(BaseInterface, GroupDevice):
 
     @limit_pos.sub_value
     def _update_limit_pos(self, value, **kwargs):
+        if np.isclose(self._limit_pos, value,
+                      rtol=self._limit_rtol, atol=self._limit_atol):
+            return
         self._limit_pos = value
         self._update_channel_limits()
 
     @limit_neg.sub_value
     def _update_limit_neg(self, value, **kwargs):
+        if np.isclose(self._limit_neg, value,
+                      rtol=self._limit_rtol, atol=self._limit_atol):
+            return
         self._limit_neg = value
         self._update_channel_limits()
 
--- a/pcdsdevices/mpod_apalis.py
+++ b/pcdsdevices/mpod_apalis.py
@@ -1,5 +1,6 @@
 import logging
 import time
+from enum import Enum, auto
 
 from ophyd.device import Component as Cpt
 from ophyd.device import Device
@@ -120,8 +121,22 @@ class MPODApalisChannel(BaseInterface, Device):
 
     @max_voltage.sub_value
     def _new_max_voltage(self, value: float, **kwargs) -> None:
-        """Set explicit limits on voltage for better error reporting."""
-        bounds = (0, value)
+        """
+        Set explicit symmetric limits on voltage for better error reporting.
+
+        Refers to the parent module if it exists, to determine the polarity
+
+        If not, defaults to setting limits to positive definite
+        """
+        polarity = getattr(self.parent, "polarity", Polarity.POSITIVE)
+
+        if polarity is Polarity.POSITIVE:
+            bounds = (0, value)
+        elif polarity is Polarity.NEGATIVE:
+            bounds = (-abs(value), 0)
+        elif polarity is Polarity.BIPOLAR:
+            bounds = (-abs(value), abs(value))
+
         self.voltage._override_metadata(
             lower_ctrl_limit=min(bounds),
             upper_ctrl_limit=max(bounds),
@@ -160,6 +175,12 @@ def _put_clamped(signal: EpicsSignal, value: float) -> None:
         value = high_val
 
     signal.put(value)
+
+
+class Polarity(Enum):
+    POSITIVE = auto()
+    NEGATIVE = auto()
+    BIPOLAR = auto()
 
 
 class MPODApalisModule(BaseInterface, GroupDevice):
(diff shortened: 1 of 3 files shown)


tangkong:
I'm going to go with the schema Zach has posted here, I want this to be as transparent / similar to the IOC as possible

ZLLentz:
>  one half-initialized ioc away from another lodcm problem here

In the LODCM, the specific bad thing we do is get values from a second IOC during a callback for the first IOC. This can cause a timeout during a callback which has unfortunate cascading implications.

Note that in some epics frameworks this choice actually crashes the program (not allowed to do epics operations in a callback)! But it doesn't crash ophyd, since the epics callback is simply used to submit a job to a python thread, so we leave that context promptly.

We can avoid these issues entirely by never doing a `caget` or equivalent inside a callback.
I'll keep an eye out for this during review.

ZLLentz:
Small note: the PR description is slightly out of date now

pcdsdevices/mpod_apalis.py
```diff
@@ -120,11 +124,38 @@ def set_current(self, current_number: float) -> None:
...
+
+        If not, defaults to setting limits to positive definite
+        """
+        # Hacky, we need to be careful not to call this without a value from
+        # inside a callback, probably
+        if value is None:
+            value = self.max_voltage.get()
+
+        polarity = getattr(self, "biological_parent.polarity", Polarity.POSITIVE)
+        limit_scales = getattr(self, "biological_parent.limit_scales", (0, 100))
```
ZLLentz, line 143: Properties are usually safe to reference in a callback, but in this case the properties might get from epics io, so it could cause a callback threading timeout issue.

pcdsdevices/mpod_apalis.py
```diff
@@ -246,6 +292,50 @@ def set_current_ramp_speed(self, ramp_speed):
         """
         self.current_ramp_speed.put(ramp_speed)
 
+    @property
+    def polarity(self) -> Polarity:
+        model_str = self.model.get()
```
ZLLentz, line 297: Is it possible to store this via monitor callback and use the cached value instead of getting io in a property?

pcdsdevices/mpod_apalis.py
```diff
@@ -246,6 +292,50 @@ def set_current_ramp_speed(self, ramp_speed):
...
+        Returns limit scaling tuple.  When multiplied by the channel's max_voltage
+        signal (:VoltageNominal) will produce the actual limits for the channel
+
+        Returns
+        -------
+        tuple[float, float]
+            The voltage limits as a proportion of the max_voltage
+        """
+        pos_limit = self.limit_pos.get()
+        neg_limit = self.limit_neg.get()  # positive % here means negative limit
```
ZLLentz, line 319: Similar to above, if we cache these (and raise if missing) we can't hit the same sort of queue-blocking issues that the LODCM has.

pcdsdevices/mpod_apalis.py
```diff
@@ -120,11 +124,38 @@ def set_current(self, current_number: float) -> None:
...
+        Set explicit symmetric limits on voltage for better error reporting.
+
+        Refers to the parent module if it exists, to determine the limit scaling
+
+        If not, defaults to setting limits to positive definite
+        """
+        # Hacky, we need to be careful not to call this without a value from
+        # inside a callback, probably
+        if value is None:
+            value = self.max_voltage.get()
```
ZLLentz, line 140: Probably it'd be best to avoid codepaths that require this.

Broken record, but the safest pattern is to cache all the values we need using monitor callbacks and only call the lowest callback if all the values are ready.

ZLLentz (commented):
Food for thought about avoiding callback issues
Approval withheld only because of missing pre-release docs

tangkong:
Thank you for being a broken record, I'd adjusted these to cache values and refer to those

pcdsdevices/mpod_apalis.py
```diff
@@ -246,6 +297,54 @@ def set_current_ramp_speed(self, ramp_speed):
         """
         self.current_ramp_speed.put(ramp_speed)
 
+    @model.sub_value
+    def _stash_model(self, value, **kwargs) -> None:
+        self._model_str = self.model.get()
```
tangkong, line 302: A bit of ignorance, are these run once on connection?  If not then I probably need to initialize these values to something more real (but potentially incorrect)

ZLLentz (commented):
This looks like it works and is pretty clean
To finish review I'll clone and point it to a few of the real tmo channels

Review still gated on pre-release notes

pcdsdevices/mpod_apalis.py
ZLLentz, line 1: Real device gives a spammy error message:
```
In [5]: from pcdsdevices.mpod_apalis import MPODApalisModule24Channel
In [6]: mp0 = MPODApalisModule24Channel("TMO:MPOD:01:M0", name="mp0")
```

```
Subscription value callback exception (EpicsSignalRO(read_pv='TMO:MPOD:01:M0:VoltageLimitNegative', name='mp0_limit_neg', value=101.923, timestamp=1742598239.3643, auto_monitor=False, string=False))
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.9.1/lib/python3.9/site-packages/ophyd/ophydobj.py", line 492, in inner
    cb(*args, **kwargs)
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 346, in _update_limit_neg
    self._update_channel_limits()
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 336, in _update_channel_limits
    cpt._update_max_voltage_limits()
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 141, in _update_max_voltage_limits
    polarity = getattr(self, "biological_parent.polarity", Polarity.POSITIVE)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.9.1/lib/python3.9/site-packages/ophyd/device.py", line 1325, in __getattr__
    return operator.attrgetter(name)(self)
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 313, in polarity
    raise ValueError(f"Unable to determine module polarity: {self._model_str}")
ValueError: Unable to determine module polarity:
Subscription value callback exception (EpicsSignalRO(read_pv='TMO:MPOD:01:M0:VoltageLimit', name='mp0_limit_pos', value=100.028, timestamp=1742598244.0573, auto_monitor=False, string=False))
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.9.1/lib/python3.9/site-packages/ophyd/ophydobj.py", line 492, in inner
    cb(*args, **kwargs)
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 341, in _update_limit_pos
    self._update_channel_limits()
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 336, in _update_channel_limits
    cpt._update_max_voltage_limits()
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 141, in _update_max_voltage_limits
    polarity = getattr(self, "biological_parent.polarity", Polarity.POSITIVE)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.9.1/lib/python3.9/site-packages/ophyd/device.py", line 1325, in __getattr__
    return operator.attrgetter(name)(self)
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 313, in polarity
    raise ValueError(f"Unable to determine module polarity: {self._model_str}")
ValueError: Unable to determine module polarity:
Subscription value callback exception (EpicsSignalRO(read_pv='TMO:MPOD:01:M0:VoltageLimitNegative', name='mp0_limit_neg', value=101.924, timestamp=1742598244.0573, auto_monitor=False, string=False))
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.9.1/lib/python3.9/site-packages/ophyd/ophydobj.py", line 492, in inner
    cb(*args, **kwargs)
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 346, in _update_limit_neg
    self._update_channel_limits()
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 336, in _update_channel_limits
    cpt._update_max_voltage_limits()
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 141, in _update_max_voltage_limits
    polarity = getattr(self, "biological_parent.polarity", Polarity.POSITIVE)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.9.1/lib/python3.9/site-packages/ophyd/device.py", line 1325, in __getattr__
    return operator.attrgetter(name)(self)
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 313, in polarity
    raise ValueError(f"Unable to determine module polarity: {self._model_str}")
ValueError: Unable to determine module polarity:
Subscription value callback exception (EpicsSignalRO(read_pv='TMO:MPOD:01:M0:VoltageLimit', name='mp0_limit_pos', value=100.03, timestamp=1742598248.3786, auto_monitor=False, string=False))
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.9.1/lib/python3.9/site-packages/ophyd/ophydobj.py", line 492, in inner
    cb(*args, **kwargs)
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 341, in _update_limit_pos
    self._update_channel_limits()
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 336, in _update_channel_limits
    cpt._update_max_voltage_limits()
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 141, in _update_max_voltage_limits
    polarity = getattr(self, "biological_parent.polarity", Polarity.POSITIVE)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.9.1/lib/python3.9/site-packages/ophyd/device.py", line 1325, in __getattr__
    return operator.attrgetter(name)(self)
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 313, in polarity
    raise ValueError(f"Unable to determine module polarity: {self._model_str}")
ValueError: Unable to determine module polarity:
Subscription value callback exception (EpicsSignalRO(read_pv='TMO:MPOD:01:M0:VoltageLimitNegative', name='mp0_limit_neg', value=101.925, timestamp=1742598248.3786, auto_monitor=False, string=False))
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.9.1/lib/python3.9/site-packages/ophyd/ophydobj.py", line 492, in inner
    cb(*args, **kwargs)
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 346, in _update_limit_neg
    self._update_channel_limits()
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 336, in _update_channel_limits
    cpt._update_max_voltage_limits()
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 141, in _update_max_voltage_limits
    polarity = getattr(self, "biological_parent.polarity", Polarity.POSITIVE)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.9.1/lib/python3.9/site-packages/ophyd/device.py", line 1325, in __getattr__
    return operator.attrgetter(name)(self)
  File "/cds/home/z/zlentz/github/pcdsdevices/pcdsdevices/mpod_apalis.py", line 313, in polarity
    raise ValueError(f"Unable to determine module polarity: {self._model_str}")
ValueError: Unable to determine module polarity:
```

pcdsdevices/mpod_apalis.py
ZLLentz, line 302: avoiding a callback get
```suggestion
        self._model_str = value
```

pcdsdevices/mpod_apalis.py
```diff
@@ -212,10 +248,25 @@ class MPODApalisModule(BaseInterface, GroupDevice):
...
                  doc='Clears all MPOD module faults')
 
+    model = Cpt(EpicsSignalRO, ':Model', kind='omitted', string=True,
+                doc="Model number")
+
+    limit_pos = Cpt(EpicsSignalRO, ':VoltageLimit', kind='omitted',
+                    doc='Positive voltage limit as a % of the max voltage')
+
+    limit_neg = Cpt(EpicsSignalRO, ':VoltageLimitNegative', kind='omitted',
+                    doc='Negative voltage limit as a % of the max voltage')
```
ZLLentz, line 258: Apparently these update often (+/- a fraction of a %)

pcdsdevices/mpod_apalis.py
```diff
@@ -246,6 +297,54 @@ def set_current_ramp_speed(self, ramp_speed):
...
+    @property
+    def polarity(self) -> Polarity:
+        if self._model_str.endswith("x"):
+            return Polarity.BIPOLAR
+        elif self._model_str.endswith("p"):
+            return Polarity.POSITIVE
+        elif self._model_str.endswith("n"):
+            return Polarity.NEGATIVE
+
+        raise ValueError(f"Unable to determine module polarity: {self._model_str}")
```
ZLLentz, line 313: Even without the spam bug, we might be at risk to see this at startup depending on the callback order, since the default value is not valid.

pcdsdevices/mpod_apalis.py
tangkong: This got lost whoops

pcdsdevices/mpod_apalis.py
tangkong: ohno

pcdsdevices/mpod_apalis.py
tangkong: Are these not static settings? If they're being calculated/generated then maybe this is also not the right setting to pull

pcdsdevices/mpod_apalis.py
tangkong: Ah a fair point.  I wanted to preserve this check but maybe it's not worth the complexity.

pcdsdevices/mpod_apalis.py
ZLLentz: Possibly so, if you monitor them they fluctuate slightly

tangkong:
So after some investigation, VoltageLimit is indeed a modification of VoltageNominal, but it's set by a physical screw on a potentiometer.  The manual does specify that this is the maximum with some error, due to digitization. 
image

We don't expect this to fluctuate too far outside the set range, so maybe we read it less frequently.  It maybe just on put?  (I'm less happy with that)

ZLLentz:
get exactly once in a thread, maybe?

tangkong:
Maybe the fact that someone has to physically turn the screw makes me more ok with getting this value only once.  I can leave a big loud comment about it, if we're ok with that solution

ZLLentz:
yeah I think that's OK
another way to approach it would be to monitor, but only do the updates if the number has changed by "a lot"

tangkong:
My annual "avoid threads if I can" moment is here!

pcdsdevices/mpod_apalis.py
tangkong: The whole polarity code path has been removed

pcdsdevices/mpod_apalis.py
tangkong: removed this polarity logic entirely

tangkong:
This should have the the logic we need for setting the voltage limits now.  Requested reviews from Tong and Basil as people familiar with the MPODs, as a way to promote visibility

ZLLentz (approved):
This looks ready to go to me
The code is clean and I spot-checked a few tmo channels successfully
```
In [1]: mp0 = MPODApalisModule24Channel("TMO:MPOD:01:M0", name="mp0")
In [5]: mp0.c0.voltage.limits
Out[5]: (-509.625, 500.16)
In [6]: mp0.c1.voltage.limits
Out[6]: (-509.625, 500.16)
In [16]: mp2 = MPODApalisModule24Channel("TMO:MPOD:01:M2", name="mp0")
In [17]: mp2.c0.voltage.limits
Out[17]: (-3057.75, 3057.78)
In [18]: mp4 = MPODApalisModule24Channel("TMO:MPOD:01:M4", name="mp0")
In [19]: mp4.c0.voltage.limits
Out[19]: (-4998.258, 0.0)
```
In each case, even then the numbers seem odd, they always matched the nominal voltage multiplied by the percent settings.

PR #1334, opened 2025-03-17
Elliptec updates for Sliders

aberges-SLAC:
## Description
Changed `channel` to be a mandatory arg in `EllBase` constructor for consistency in API usage. All our devices in the wild (as far as lasers are concerned) call the channel prop from their `happi` entry anyway. In `hutch_python` sessions this will now be more obvious to the user what information they need before device instantiation.

Also overloaded `done_comparator` to just use `readback` and `setpoint` methods if using `Ell6` or `Ell9`. No need to call a `numpy.isclose()` on `enums`.

## Motivation and Context
Couple of JIRA tickets on user frustration and terminal spam when opening LUCID with `Ell6` or `Ell9` implemented. Easy fix for root causes. See ECS-7383 and ECS-7390.

## How Has This Been Tested?
Tested with the same sliders specified in the above tickets using the CRIX_MODS toolbar and LUCID. 2 birds, 1 stone: no terminal spam now and clearer API for the class.

This PR and the aforementioned tickets.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/lasers/elliptec.py
+++ b/pcdsdevices/lasers/elliptec.py
@@ -77,6 +77,11 @@ class Ell6(EllBase):
     --------
     ell6 = Ell6('LM1K4:COM_DP1_TF1_SL1:ELL', port=0, channel=1, name='ell6')
     """
+    # Since the record for 2 and 4-position sliders are stored as an enum,
+    # we need a "str comparator" overload.
+    def done_comparator(self, readback, setpoint):
+        return readback == setpoint
+
     # Names for slider positions
     name_0 = FCpt(EpicsSignal, '{prefix}:M{self._channel}:NAME0',
                   kind='config')
--- a/pcdsdevices/lasers/elliptec.py
+++ b/pcdsdevices/lasers/elliptec.py
@@ -54,7 +54,7 @@ class EllBase(PVPositionerIsClose):
     user_readback = FCpt(EpicsSignal, '{prefix}:M{self._channel}:CURPOS',
                          kind='omitted')
 
-    def __init__(self, prefix, channel, port=0, **kwargs):
+    def __init__(self, prefix, *, channel, port=0, **kwargs):
         self._port = port
         self._channel = channel
         super().__init__(prefix, **kwargs)
--- a/pcdsdevices/lasers/elliptec.py
+++ b/pcdsdevices/lasers/elliptec.py
@@ -54,7 +54,7 @@ class EllBase(PVPositionerIsClose):
     user_readback = FCpt(EpicsSignal, '{prefix}:M{self._channel}:CURPOS',
                          kind='omitted')
 
-    def __init__(self, prefix, port=0, channel=1, **kwargs):
+    def __init__(self, prefix, channel, port=0, **kwargs):
         self._port = port
         self._channel = channel
         super().__init__(prefix, **kwargs)
--- a/docs/source/upcoming_release_notes/1334-Elliptec_slider_update.rst
+++ b/docs/source/upcoming_release_notes/1334-Elliptec_slider_update.rst
@@ -1,0 +1,30 @@
+1334 Elliptec_slider_update
+#################
+
+API Breaks
+----------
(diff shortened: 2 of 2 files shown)


christina-pino (approved):
I trust your test for the sliders. Changes seem pretty straight forward

aberges-SLAC:
> LGTM. Something to keep in mind is that this may break the HAPPI database instantiation because the channel kwarg is now an arg.

I thought this would just get treated the way `tip_pv` and `tilt_pv` do for SmarActs — it's still a named arg, just not an optional arg.

tangkong:
Well spotted by Tyler.  It's more a difference between an argument with a default and one without.  If the happi entries ever expected the default value to be there, the device would no longer load.  

From a quick `happi search device_class=*Ell*`, it seems like all our existing devices specify `channel` as a kwarg, so we should be ok.  (If I'm not missing anything)

pcdsdevices/lasers/elliptec.py
```diff
@@ -77,6 +77,11 @@ class Ell6(EllBase):
     --------
     ell6 = Ell6('LM1K4:COM_DP1_TF1_SL1:ELL', port=0, channel=1, name='ell6')
     """
+    # Since the record for 2 and 4-position sliders are stored as an enum,
+    # we need a "str comparator" overload.
+    def done_comparator(self, readback, setpoint):
```
ZLLentz, line 82: Since you're writing your own `done_comparitor`, perhaps this class should now inherit directly from `PVPositionerComparator` instead of `PvPositionerIsClose` (which itself only exists to define a `done_comparitor`.

aberges-SLAC: Yeah this was the whole root cause for the terminal spam that kept crashing `las-toolbar` btw, the `np.isclose` kept failing because it was comparing `[(float, str). (float, str)]` and it cannot cope. 

So since `Ell9` subclasses `Ell6`, I opted to overload `done_comparator` here to keep all else equal. Granted, all subclasses of `PVPositionerComparator` are supposed to define their own `done_comparator`, I thought

aberges-SLAC:
> The only instantiations this new signature will break is code that supplied `channel` and `port` as positional arguments, since their positional order swapped (and of course code that neglected to provide channel at all).
> 
> Both the signature in the diff
> 
> ```
> def __init__(self, prefix, channel, port=0, **kwargs):
> ```
> 
> and your proposed
> 
> ```
> def __init__(self, prefix, *, channel, port=0, **kwargs):
> ```
> 
> Will accept channel as a kwarg.
> 
> The second syntax here makes `channel` (and `port`) a keyword-only argument (rejects positional argument passing), which may be independently valuable to you, or might break positional arg passers (which you might also want).

Good point — I think I want to opt for the `channel` as `kwarg` only. If you're gonna use the class, you need to know the channel and the IOC's configuration. I don't want folks getting into accidents again by riding on the defaults which might incidentally work on controllers with only one stage defined.

pcdsdevices/lasers/elliptec.py
tangkong: These setpoints return (float, str)?...

aberges-SLAC:
> I expect that this works with all existing Ell devices in `happi` and will prevent silent failures in the future. (soft approval)
> 
> Maybe before merge, you can identify some Ell6, 9 devices in `happi` (and some EllLinear) that have their IOCs up? These are pretty easy to verify working (all the ones I tried were disconnected)
> 
> ```
> cd pcdsdevices
> git fetch adam
> git checkout Elliptec
> export PYTHONPATH=$(pwd)
> happi load <something>
> <something>
> ```

I tested this running `~/bin/test_mods` which is functionally doing the same thing. I get no errors and HAPPI loads normally in the CRIX_MODS toolbar. I'd assume `lucid` would crash out if I broke it otherwise

ZLLentz:
Good enough for me! Thank you for confirming!

ZLLentz (approved):

pcdsdevices/lasers/elliptec.py
tangkong: Sorry now that we have a working ioc, I see they're enums

PR #1345, opened 2025-04-11
WIP: Atm delay limit protection

jozamudi:
## Description

Added soft limits to DelayBase. When PseudoPositioner position is outside of limits, an exception is thrown.

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-6621

## How Has This Been Tested?

I created a test that mimics RIX LXTTTC device and shift_t0 function in rix hutch python lxt.py. I commanded moves that would cause the limits to trip.

## Pre-merge checklist


pcdsdevices/pseudopos.py
```diff
@@ -746,6 +746,7 @@ class DelayBase(FltMvInterface, PseudoPositioner):
     user_offset = Cpt(NotepadLinkedSignal, ':OphydOffset',
                       notepad_metadata={'record': 'ao', 'default_value': 0.0})
     motor = None
+    limits = None
```
tangkong, line 749: So these devices already have limits via [`pcdsdevices.pseudopos.PseudoPositioner`](https://github.com/pcdshub/pcdsdevices/blob/02a0bcefe4101d5dc8c91cd744a6f3ca59c5d1ac/pcdsdevices/pseudopos.py#L114) -> [`ophyd.pseudopos.PseudoPositioner`](https://github.com/bluesky/ophyd/blob/1c2b121e63de747f9d3780f6fc208c8912f89d49/ophyd/pseudopos.py#L659-L662)

By overwriting the property here, we're disabling all of that inherited behavior.

tangkong:
```python
In [4]: DelayBase.__mro__
Out[4]: 
(pcdsdevices.pseudopos.DelayBase,
 pcdsdevices.interface.FltMvInterface,
 pcdsdevices.interface.MvInterface,
 pcdsdevices.interface.BaseInterface,
 pcdsdevices.pseudopos.PseudoPositioner,
 ophyd.pseudopos.PseudoPositioner,
 ophyd.device.Device,
 ophyd.device.BlueskyInterface,
 ophyd.positioner.SoftPositioner,
 ophyd.positioner.PositionerBase,
 ophyd.ophydobj.OphydObject,
 object)
```

https://github.com/bluesky/ophyd/blob/1c2b121e63de747f9d3780f6fc208c8912f89d49/ophyd/positioner.py#L284-L286

tangkong:
Josue and I looked at this offline, and while I struggled to remember how pseudopositioners and DelayBase worked, I think we came to some amount of clarity:

* `DelayBase` holds onto a "real" motor (real space) that we convert into a delay via pseudo positioners
* The core of the request here is to add limits to that real-space motor
  * If you provide limits to the constructor of the `DelayBase` subclass, those limits are applied in delay-space
  * There is no way to instantiate the real-motor component with limits through the `delay_motor_factory` that is typically used to create a Delay motor
* Only some motor classes allow you to set the limits dynamically, most `motor.limits` calls are read-only properties

But critically:
* Our `EpicsMotorInterface` base class grabs the limits from the motor record itself

After all this, I think the proper way to address this is to:
* make sure the motor record has valid limits
* Ensure the motor class being used in the delay motor properly inherits this information from `EpicsMotorInterface`

tangkong:
The short follow up to this is that I think that this change isn't correct, and that the only change we might consider making is adjusting the `delay_*_factory` functions to allow passing motor limits to the motor component constructors.

ZLLentz:
> allow passing motor limits to the motor component constructors

This could potentially write to EPICS PVs at `__init__`. We need to be careful to avoid this kind of behavior.

tangkong:
A good point.  I think the idea we're settling on here is that our policy is to populate limits from EPICS, and these limits shouldn't be changed without explicit user request (aka on the creation of a device)

ZLLentz:
Yes, though it is permissible for a python motor class to enforce stricter limits than the epics limits! (at the risk of confusing people)

Issue #1342, opened 2025-03-31
Add error checking to Lcls2LaserTiming Class


slactjohnson:
The LCLS-II laser locker will fail to execute a move of timing if certain conditions are not met, such as having a valid calibration. 

The IOC will report an error to the IOC shell when an move is attempted in this state, and the HLA enabled state is reported on the screen, but there is no feedback to the python class. Apparently the laser locker lost its calibration in TMO recently and the users were trying to move timing from python, but were unaware of the loss of calibration. There was some beam time loss associated with this. 

We should look into how to best communicate this with the user from the python session. 

First idea is to add the HLA enabled PV to the class and if we are not enabled and a move is requested, just `raise` an error? 

Related Jira ticket: https://jira.slac.stanford.edu/browse/TSS-453

PR #1353, opened 2025-05-31
Adding hla check for lxt

KaushikMalapati:
## Description

Added the hla_enabled component to Lcls2LaserTiming
Added a decorator called check_hla which raises an exception hla_enabled equals 0 (meaning HLA is disabled)
Added the decorator to _setup_move

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-8017
Closes https://github.com/pcdshub/pcdsdevices/issues/1342

## How Has This Been Tested?

I only checked that new signal works since I don't want to try moving any lxts right now. 

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/1357-mnt_presets_sync_2.rst
+++ b/docs/source/upcoming_release_notes/1357-mnt_presets_sync_2.rst
@@ -1,0 +1,30 @@
+1357 mnt_presets_sync_2
+#######################
+
+API Breaks
+----------
+- N/A
+
+Library Features
+----------------
+- N/A
+
+Device Features
+---------------
+- N/A
+
+New Devices
+-----------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- Sync presets when Presets.positions is accessed.
+
+Contributors
+------------
+- tangkong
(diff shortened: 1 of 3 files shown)


slactjohnson:
@KaushikMalapati testing this would have to be scheduled with one of the hutches. I think it would be best to reach out to Mat Britton (TMO) or Matt Bain (RIX) to schedule this, since they'll be aware of both the laser schedule and the hutch operations schedule.

slactjohnson:
The calibration can be invalidated by unlocking the system. The timing team can help; you can reach out to Ryan Lodico for that.

pcdsdevices/lxe.py
```diff
@@ -482,6 +486,15 @@ def _offset_changed(self, value, **kwargs):
         """
         self.setpoint.user_offset = value
 
+    def check_hla(func):
+        def wrapper(self, *args, **kwargs):
```
tangkong, line 490: Consider decorating this with functools.wraps to keep the name/docstring of the original function
https://docs.python.org/3/library/functools.html#functools.wraps

tangkong:
Should you be calling `._setup_move` directly?  Did you check that this worked when calling `.move`f/

tangkong:
Right, but the intended user interface is `.move`, so I'd want to see that the changes here don't have any unintended consequences with the logic included from that method.

tangkong:
I'm mostly concerned about `.move()`.  If zone 3 is closed then we can skip the check for now.  I'm mostly confident this won't cause issues but I wanted to be diligent

PR #1360, opened 2025-06-19
add SL2K4 device class

baljamal:
## Description
SL2K4 is a BeckhoffSlits device with new looking GUI and PVA based camera view which looking at beam after SL2K4 slit

## Motivation and Context
James cryan asked for a separate screen for SL2K4 which could be used to look at beam while moving slit to remove scatter around the beam.

## How Has This Been Tested?
unit test written: pcdsdevices/tests/test_sl2k4_slits.py , also tested manually using TMO conda env

<!--
image

## Pre-merge checklist


baljamal:
screenshot:
image

ZLLentz:
I have not read the code yet but I audibly made a "oooh" sound at the screenshot, it looks good. I like the slit graphic arrangement much better than how it was done for older slits.

pcdsdevices/slits.py
```diff
@@ -602,6 +602,18 @@ def calc_lightpath_state(
                                             ywidth=ywidth_readback)
 
 
+class SL2K4Slits(BeckhoffSlits):
+    def __init__(self, prefix, cam, data_source, *, name, **kwargs):
+        self._started_move = False
+        self._top_done = False
+        self._bottom_done = False
+        self._north_done = False
+        self._south_done = False
```
ZLLentz, line 611: Question/confusion: Aren't these all already ran in the `super().__init__` call? Are they needed again here?

pcdsdevices/slits.py
```diff
@@ -602,6 +602,18 @@ def calc_lightpath_state(
...
+        self._cam = cam
+        self._data_source = data_source
```
ZLLentz, line 613: Suggestion: this class needs some brief documentation about what `cam` and `data_source` are, how one would get them, and what we use them for.

pcdsdevices/tests/test_sl2k4_slits.py
```diff
@@ -0,0 +1,21 @@
...
+    return FakeSlits(prefix='SL2K4:SCATTER', cam='IM6K4:PPM:CAM',
+                     data_source='IMAGE2',
+                     name='sl2k4_slits')
+
+
+def test_slits_initialization(fake_slits):
+    sp = 4.0
+    fake_slits.xwidth.setpoint.put(sp)
+    xwidth_sp = fake_slits.xwidth.setpoint.get()
+    assert pytest.approx(xwidth_sp, 0.1) == sp
```
ZLLentz, line 21: Comment: I'm not sure I understand this test, you're essentially putting to and reading back from a known fake epics signal. None of the code you included is tested here (it's usually not possible to write a meaningful unit test for a device like this unless there are specific coding things added).

I think a more meaningful test would be checking how `cam` and `data_source` are used, but maybe I'm wrong because I don't understand what they are used for.

Lastly, from a test organization perspective, everything from `slits.py` should be tested in `test_slits.py`, we should not make new mini test modules for every new device.

ZLLentz (commented):
Partial review: just docs/coding stuff
I'll look more closely at the ui next

pcdsdevices/ui/SL2K4.detailed.ui
ZLLentz, line 1: How exactly are you testing and loading this ui file? It seems like it cannot be done using typhos, yet it is here in the typhos stuff folder. It does look very nice though.

baljamal: If there is a better way I'm willing to follow it

pcdsdevices/slits.py
ZLLentz: I don't immediately see where they are used, but I was more concerned about documenting what they are for our fellow team members who see this in the future.

pcdsdevices/ui/SL2K4.detailed.ui
ZLLentz: I'm not asking because I necessarily suspect there is a better way, I'm acting because running the screen is part of code review and I couldn't figure out how to do it.

ZLLentz, line 1: Do you want this to run in `typhos`? If so, we'll need to make some changes. If not, maybe this belongs in https://github.com/pcdshub/lcls-pydm-tmo instead of in `pcdsdevices`.

As a standalone PyDM screen this is 100% great (with some minor sizing issues if you try to make the window bigger).

but as a typhos screen it has the wrong filename and it doesn't have a way to load the image in.

baljamal: I need to have it as Typhos because they want it to appear as a dot on Lucid grid. if that is not possible, I will use it as a PyDM screen

ZLLentz: It is possible but we need to make some small tweaks:

- Rename screen file or rename class so that they match exactly
- Decide how to load the macros for the image via typhos

I'll look into the second one to remind myself what the options are for loading extra macros in typhos

ZLLentz: Ok I read through the code etc.

The general thing to note is that `typhos` reads the `md` metadata attached to the device class (typically from happi) to pick macros:

- [link to typhos code for macro loading](https://github.com/pcdshub/typhos/blob/ca90f51519196281773c13df5f5adb058253440c/typhos/display.py#L1491-L1495) 
- [link to happi code for setting up md](https://github.com/pcdshub/happi/blob/1b9e37251ffa45567cccc090ee14ea8d13da8081/happi/loader.py#L204-L210)

One simple thing to do is to add a function to your device called `post_happi_md`. `happi` will call `device.post_happi_md` if it exists which you can use to inject extra metadata.

So my recommendation (untested) is something like:

```
def post_happi_md(self):
    self.md.extraneous["cam"] = self._cam
    self.md.extraneous["data_source"] = self._data_source
```

And then we can include the cam/data_source kwargs in happi when we get to prod or pass them manually for testing:

```
typhos "pcdsdevices.slits.SL2K4Slits[{'prefix':'SL2K4:SCATTER','cam':'IM6K4:PPM:CAM','data_source':'IMAGE2'}]"
```

And after that dance, typhos should pick them up and put them into the screen

baljamal:
@ZLLentz I not sure where the best place to put MetaDataDict class, please advice

ZLLentz:
> simple script to test typhos screen

I'll also clone your branch and load from the typhos CLI as a review

tangkong:
I'll also echo that this screen looks great!

ZLLentz:
RE: the missing title, it apparently disappears if the typhos screen doesn't have a layout applied (???)
I think this might be a typhos bug? I'm not sure

Here's what the screen looks like if I edit it to have a vertical layout
image

Notice that the title section has returned!
If you think this looks good I can commit and PR it into your fork

pcdsdevices/slits.py
```diff
@@ -602,6 +602,119 @@ def calc_lightpath_state(
...
+        Area detector stream name. e.g. IMAGE2
+    """
+    def __init__(self, prefix, cam, data_source, *, name, **kwargs):
+        self._cam = cam
+        self._data_source = data_source
+        super().__init__(prefix, name=name, **kwargs)
+        self.md = MetaDataDict({"cam": cam,
+                                "data_source": data_source,
+                                "input_branches": None,
+                                "output_branches": None})
```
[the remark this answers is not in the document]
ZLLentz: Or, maybe it's advisable for Basil to use `HappiItem` directly for maximum compatibility?
```
self.md = HappiItem(name=name, cam=cam, data_source=data_source, input_branches=None, output_branches=None)
```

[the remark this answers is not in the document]
tangkong: That seems more cumbersome to me, but I guess it is the most technically consistent typing-wise

I wasn't even sure if happi was a dependency of pcdsdevices here, but it turns out we do need it for the container definitions.

ZLLentz:
> what do you think is wrong with QFrame?

I spent fewer than 5 seconds nitpicking the spacing within the frame (I'm not sure if I cut off the top edge of the image or not)

baljamal:
@ZLLentz I tested your changes and it look good

ZLLentz (approved):
I'm going to merge this now before another PR gets merged first!
Let's get this into happi tomorrow

PR #1364, opened 2025-07-03
improve smarpod class

baljamal:
## Description
- use PVPositionerIsClose instead of raw records
- add mutli axis move
- add checking reachability

## Motivation and Context
James Cryan asked for improving smarpod class as listed at description

## How Has This Been Tested?

> from pcdsdevices.smarpod import SmarPod
> spd = SmarPod(prefix="TMO:SPD:01", name="tmo_spd1_01")
> spd.z.move(-2.8) # works
> spd.z.move(-7) # valueError raised with "Position is unreachable !!!" message
> spd.multi_axis_move(x_sp=1.0, y_sp=1.0, z_sp=1.0) # only x,y,z will move
> spd.multi_axis_move() # try to move to the current setpoints

TMO pcsd dev-conda env 

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/smarpod.py
+++ b/pcdsdevices/smarpod.py
@@ -15,7 +15,6 @@ from __future__ import annotations
 
 import enum
 import logging
-import time
 
 from ophyd.device import Component as Cpt
 from ophyd.device import Device
@@ -28,7 +27,7 @@ from pcdsdevices.pv_positioner import PVPositionerIsClose
 logger = logging.getLogger(__name__)
 
 
-class Axis(str, enum.Enum):
+class SmarPodAxisEnum(str, enum.Enum):
     """
     An enumeration representing axes available in tri-sphere table stage.
 
@@ -98,9 +97,9 @@ class SmarPodAxis(PVPositionerIsClose):
     atol = 1E-5  # 10 nm
     csy_cmd = Cpt(EpicsSignal, ':CSY:CMD:{axis}', kind='normal')
     csy_rb = Cpt(EpicsSignalRO, ':CSY:{axis}', kind='normal')
-    group_motion : bool = False
+    defer_motion : bool = False
 
-    def __init__(self, prefix, axis: Axis, **kwargs):
+    def __init__(self, prefix, axis: SmarPodAxisEnum, **kwargs):
         self.axis = axis.value
         return super().__init__(prefix, **kwargs)
 
@@ -109,16 +108,14 @@ class SmarPodAxis(PVPositionerIsClose):
         Before moving check the position is reachable.'''
 
         logger.info('Checking if position is reachable')
-        self.setpoint.put(position)
-        self.cmd_reachable.put(1)
-        time.sleep(0.1)
+        self.setpoint.put(position, wait=True)
         is_reachable = self.reachable.get()
 
         logger.info(f"is_reachable : {is_reachable}")
         if not is_reachable:
             raise ValueError('Position is unreachable !!!')
         self.setpoint.put(position, wait=True)
-        if self.actuate and not self.group_motion:
+        if self.actuate and not self.defer_motion:
             logger.debug('%s.actuate=%s', self.name, self.actuate_value)
             self.actuate.put(self.actuate_value, wait=False)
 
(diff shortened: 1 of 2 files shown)


pcdsdevices/smarpod.py
```diff
@@ -151,3 +212,61 @@ class SmarPod(Device):
     cmd_calibrate = Cpt(EpicsSignal, ':CMD:CALIBRATE', kind='normal')
     cmd_read_ver = Cpt(EpicsSignal, ':CMD:READ_VER', kind='normal')
     ver_sys = Cpt(EpicsSignalRO, ':VER:SYS', kind='omitted')
+    # coordinate offset
+    csy_set = Cpt(EpicsSignal, ':CSY:SET', kind='normal', doc="set csy offset")
+    csy_zero = Cpt(EpicsSignal, ':CMD:SPZ', kind='normal', doc="set csy to zero")
+
+    def multi_axis_move(
```
tangkong, line 219: Question: So if no positions are specified, we'll just be moving the same place we're currently in?

pcdsdevices/smarpod.py
```diff
@@ -53,6 +86,43 @@ class SmarPodStatus(Device):
...
+    actuate = Cpt(EpicsSignal, ':CMD:MOVE')
+    actuate_value = 0
+    stop_signal = Cpt(EpicsSignal, ':CMD:STOP')
+    stop_value = 0
+    cmd_reachable = Cpt(EpicsSignal, ':CMD:REACHABLE', kind='normal')
+    reachable = Cpt(EpicsSignalRO, ':REACHABLE')
+    atol = 1E-5  # 10 nm
+    csy_cmd = Cpt(EpicsSignal, ':CSY:CMD:{axis}', kind='normal')
+    csy_rb = Cpt(EpicsSignalRO, ':CSY:{axis}', kind='normal')
+    group_motion : bool = False
```
tangkong, line 101: Question / Nitpick: Is this just meant to signal whether moves are allowed to happen immediately?  An individual axis doesn't seem to have any notion of what group it's part of, so it's a little weird to name it group motion here.

tangkong (commented):
I like how this looks on the whole, I think the organization here makes a good amount of sense.  Have you tested the group motion methods?  (your PR description only mentions a single axis move)

pcdsdevices/smarpod.py
baljamal: currently, yes. sometimes user try to move to the same positions multiple time to reach the exact setpoints

pcdsdevices/smarpod.py
baljamal: maybe group_motion is not a good name. what it means, just setpoint but don't move, because move command will be issued independently. Do you think "defer_motion" is a better name?

pcdsdevices/smarpod.py
```diff
@@ -10,9 +10,42 @@
...
+from ophyd.device import FormattedComponent as FCpt
 from ophyd.signal import EpicsSignal, EpicsSignalRO
+from ophyd.status import wait as status_wait
+
+from pcdsdevices.pv_positioner import PVPositionerIsClose
+
+logger = logging.getLogger(__name__)
+
+
+class Axis(str, enum.Enum):
```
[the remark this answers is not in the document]
baljamal: good idea, I will rename it

pcdsdevices/smarpod.py
ZLLentz: Naming aside, maybe the more important thing to fix is to document what it does in a docstring here.

pcdsdevices/smarpod.py
```diff
@@ -151,3 +213,61 @@ class SmarPod(Device):
...
+        self.rx.setpoint.put(rx_sp)
+        self.ry.setpoint.put(ry_sp)
+        self.rz.setpoint.put(rz_sp)
+
+        x_status = self.x.move(x_sp, wait=False, timeout=timeout)
+        y_status = self.y.move(y_sp, wait=False, timeout=timeout)
+        z_status = self.z.move(z_sp, wait=False, timeout=timeout)
+        rx_status = self.rx.move(rx_sp, wait=False, timeout=timeout)
+        ry_status = self.ry.move(ry_sp, wait=False, timeout=timeout)
+        rz_status = self.rz.move(rz_sp, wait=False, timeout=timeout)
```
ZLLentz, line 257: Question: what's the purpose of putting to the setpoints manually and also calling move?

pcdsdevices/smarpod.py
```diff
@@ -53,6 +85,41 @@ class SmarPodStatus(Device):
...
+        Before moving check the position is reachable.'''
+
+        logger.info('Checking if position is reachable')
+        self.setpoint.put(position, wait=True)
+        is_reachable = self.reachable.get()
+
+        logger.info(f"is_reachable : {is_reachable}")
+        if not is_reachable:
+            raise ValueError('Position is unreachable !!!')
+        self.setpoint.put(position, wait=True)
```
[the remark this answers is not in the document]
baljamal: the IOC has one move command that moves six axes to setpoint, so you have to set the setpoint for each axes based on arguments, then call move method for each axes to get its status object, or all status objects and then wait if wait=True

tangkong (approved):
I think this looks good, any remaining qualms I have are with the ioc I guess

PR #1370, opened 2025-07-29
Adding PDU typhos screen support

TrumanPar:
## Description

I've added typhos screens for PDUs, essentially coping over the current EDM screens into pydm. I've broken this out into two pcdsdevices, one for Tripplite PDUs, and another for Leviton / Sentry4. The reason being that Tripplite IOCs are too different than Leviton/Sentry4 to justify (in my mind) keeping them together.

Each class has its own happi container and .py file for typhos to execute when the detailed screen is opened. 

Some features include:
- Channel status widgets will change color when a minor (yellow) or major (red) alarm state is active.
- Channels cannot be accidentally cycled by hovering the mouse key over the enum button while scrolling. You have to click to change state,

Some personal reservations include:
- I had to source pcds-6.0.0 during testing in order for typhos to execute my custom .py files when launching the detailed screens.
- Old Sentry IOCs will not work for my screen, only newer Sentry4 IOCs.

## Motivation and Context

This effort is motivated mainly by the fact that PDUs are not supported in pydm. Furthermore, this coincides with cleaning up the Lucid buttons on some of the more cluttered hutches by utilizing happi/typhos instead. 

## How Has This Been Tested?

Interactively with test PDUs in the makerspace. Testing included ensuring channels cannot be easily powered down by accident and that all channels are controllable in the same fashion to the EDM screens.

Once this PR is merged, I will add to the confluence pages on how to set up Tripplite & Sentry4/Leviton PDUs

Screenshot 2025-07-28 165921

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/happi/containers.py
+++ b/pcdsdevices/happi/containers.py
@@ -679,3 +679,17 @@ class Leviton(LCLSItem):
     kwargs.default['elevations'] = "{{elevations}}"
     elevations = EntryInfo(doc='List of elevation numbers for rack',
                            optional=False, enforce=list)
+
+class PDU(LCLSItem):
+    """
+    Container for non-Tripplite PDUs
+    """
+    device_class = copy(LCLSItem.device_class)
+    device_class.default = 'pcdsdevices.pdu.PDU'
+
+class TripplitePDU(LCLSItem):
+    """
+    Container for Tripplite PDUs
+    """
+    device_class = copy(LCLSItem.device_class)
+    device_class.default = 'pcdsdevices.pdu.TripplitePDU'
--- a/pcdsdevices/happi/containers.py
+++ b/pcdsdevices/happi/containers.py
@@ -702,12 +702,14 @@ class Leviton(LCLSItem):
     elevations = EntryInfo(doc='List of elevation numbers for rack',
                            optional=False, enforce=list)
 
+
 class PDU(LCLSItem):
     """
     Container for non-Tripplite PDUs
     """
     device_class = copy(LCLSItem.device_class)
     device_class.default = 'pcdsdevices.pdu.PDU'
+
 
 class TripplitePDU(LCLSItem):
     """
--- a/pcdsdevices/pdu.py
+++ b/pcdsdevices/pdu.py
@@ -1,0 +1,212 @@
+"""
+Standard classes for L2SI DC power devices.
+"""
+import logging
+
+from ophyd import Device, EpicsSignal, EpicsSignalRO
+from ophyd import FormattedComponent as FCpt
+from ophyd import Component as Cpt
+
+from .device import GroupDevice
+from .interface import BaseInterface
+
+logger = logging.getLogger(__name__)
+
+
+class PDUChannel(Device):
+    """
+    Class to define a particular channel of the PDU.
+
+    Parameters
+    ---------
+    prefix : str
+        The base PV of the relevant PDU.
+
+    channel : str
+        The output channel on the pPDU, e.g. '1', '2', etc.
+
+    pdu_type : str
+        The pdu type, e.g TrippLite, Leviton, or Sentry
+    """
+
+    ch_index = FCpt(EpicsSignalRO, '{self.prefix}:Outlet:{self._ch}:{self._ch_index}',
+                    kind='hinted', string=True)
+    ch_name = FCpt(EpicsSignal, '{self.prefix}:Outlet:{self._ch}:{self._ch_name}',
+                    kind='hinted', string=True)
+    ch_status = FCpt(EpicsSignalRO, '{self.prefix}:Outlet:{self._ch}:{self._ch_status}',
+                    kind='hinted', string=True)
+    ch_ctrl_state = FCpt(EpicsSignalRO, '{self.prefix}:Outlet:{self._ch}:{self._ch_ctrl_state}',
+                    kind='hinted', string=True)
+    ch_ctrl_command_on = FCpt(EpicsSignal, '{self.prefix}:Outlet:{self._ch}:{self._ch_command_on}',
+                    kind='hinted', string=True)
+    ch_ctrl_command_off = FCpt(EpicsSignal, '{self.prefix}:Outlet:{self._ch}:{self._ch_command_off}',
+                    kind='hinted', string=True)
+    ch_ctrl_command_reboot = FCpt(EpicsSignal, '{self.prefix}:Outlet:{self._ch}:{self._ch_command_reboot}',
+                    kind='hinted', string=True)
+
+    def __init__(self, prefix, *, name=None, parent=None, channel=None, **kwargs):
+        self._ch = f'{channel}'
+        self._ch_index = 'GetID'
+        self._ch_status = 'GetStatus'
+        self._ch_name = 'SetName'
+        self._ch_ctrl_state = 'GetControlState'
+        self._ch_command_on = 'TurnOn'
+        self._ch_command_off = 'TurnOff'
+        self._ch_command_reboot = 'CycleOffOn'
+
+        super().__init__(prefix, name=name, **kwargs)
+
+class TripplitePDUChannel(Device):
+    """
+    Class to define a particular channel of a Tripplite PDU.
+
+    Parameters
+    ---------
+    prefix : str
+        The base PV of the relevant PDU.
+
+    channel : str
+        The output channel on the pPDU, e.g. '1', '2', etc.
+
+    pdu_type : str
+        The pdu type, e.g TrippLite, Leviton, or Sentry
+    """
+    ch_index = FCpt(EpicsSignalRO, '{self.prefix}:Outlet:{self._ch}:{self._ch_index}',
+                    kind='hinted', string=True)
+    ch_name = FCpt(EpicsSignal, '{self.prefix}:Outlet:{self._ch}:{self._ch_name}',
+                    kind='hinted', string=True)
+    ch_status = FCpt(EpicsSignalRO, '{self.prefix}:Outlet:{self._ch}:{self._ch_status}',
+                    kind='hinted', string=True)
+    ch_ctrl_state = FCpt(EpicsSignalRO, '{self.prefix}:Outlet:{self._ch}:{self._ch_ctrl_state}',
+                    kind='hinted', string=True)
+    ch_ctrl_command = FCpt(EpicsSignal, '{self.prefix}:Outlet:{self._ch}:{self._ch_command}',
+                    kind='hinted', string=True)
+
+    def __init__(self, prefix, *, name=None, parent=None, channel=None, **kwargs):
+        self._ch = f'{channel}'
+        self._ch_index = 'getPduOutletIndex'
+        self._ch_status = 'getPduOutletState'
+        self._ch_name = 'setPduOutletName'
+        self._ch_ctrl_state = 'getPduOutletControllable'
+        self._ch_command = 'setPduOutletCommand'
+
+        super().__init__(prefix, name=name, **kwargs)
+
+
+class PDU(Device):
+    """
+    Class to define a non-Tripplite PDU.
+
+    Parameters
+    ---------
+    prefix : str
+        The base PV of the relevant PDU.
+
+    num_channels : int
+        The output channels for the PDU.
+
+    """
+
+    pdu_name = FCpt(EpicsSignal, '{prefix}{self._tower}:SetTowerName', kind='hinted')
+    pdu_location = Cpt(EpicsSignal, ':SetSystemLocation', kind='hinted')
+    pdu_status = FCpt(EpicsSignalRO, '{prefix}{self._tower}:GetTowerStatus', kind='hinted')
+    pdu_output_c_max = FCpt(EpicsSignalRO, '{prefix}{self._tower}:SetInfeedLoadHighThresh', kind='hinted')
+    pdu_output_c = FCpt(EpicsSignalRO, '{prefix}{self._tower}:GetInfeedLoadValue', kind='hinted')
+    pdu_output_c_status = FCpt(EpicsSignalRO, '{prefix}{self._tower}:GetInfeedLoadStatus', kind='hinted')
+    pdu_output_p = FCpt(EpicsSignalRO, '{prefix}{self._tower}:GetTowerActivePower', kind='hinted')
+
+
+    pdu_temperature_lo = FCpt(EpicsSignal, '{prefix}{self._tower}:Sensor:All:SetTempLowThresh', kind='hinted')
+    pdu_humidity_lo = FCpt(EpicsSignal, '{prefix}{self._tower}:Sensor:All:SetHumidLowThresh', kind='hinted')
+    pdu_temperature_hi = FCpt(EpicsSignal, '{prefix}{self._tower}:Sensor:All:SetTempHighThresh', kind='hinted')
+    pdu_humidity_hi = FCpt(EpicsSignal, '{prefix}{self._tower}:Sensor:All:SetHumidHighThresh', kind='hinted')
+
+    pdu_sensor1_id = Cpt(EpicsSignal, ':Sensor:1:GetID', kind='hinted')
+    pdu_sensor1_status = Cpt(EpicsSignal, ':Sensor:1:GetStatus', kind='hinted')
+    pdu_sensor1_temperature = Cpt(EpicsSignal, ':Sensor:1:GetTempStatus', kind='hinted')
+    pdu_sensor1_humidity = Cpt(EpicsSignal, ':Sensor:1:GetHumidStatus', kind='hinted')
+
+    pdu_sensor2_id = Cpt(EpicsSignal, ':Sensor:2:GetID', kind='hinted')
+    pdu_sensor2_status = Cpt(EpicsSignal, ':Sensor:2:GetStatus', kind='hinted')
+    pdu_sensor2_temperature = Cpt(EpicsSignal, ':Sensor:2:GetTempStatus', kind='hinted')
+    pdu_sensor2_humidity = Cpt(EpicsSignal, ':Sensor:2:GetHumidStatus', kind='hinted')
+
+    channel_ctrl = FCpt(EpicsSignal, '{prefix}{self._ctrl}:SetControlAction', kind='hinted')
+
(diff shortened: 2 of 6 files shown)


tangkong:
There are also some pre-commit checks that are failing.  You can install pre-commit in your local dev area so that it actually checks your changes before you commit them.  It's rather painless, and I find it helps a lot!

```
pre-commit install
```

> I had to source pcds-6.0.0 during testing in order for typhos to execute my custom .py files when launching the detailed screens.

Could you expand more on this?  Were the older pcds-conda environments not working?  You'd need to set up a conda environment to use typhos/pydm/etc, that's not a negotiable step.

pcdsdevices/ui/PDU.detailed.py
```diff
@@ -0,0 +1,288 @@
...
+        Up some of the signals and maybe add new widgets to the display.
+        Add any other init-esque shenanigans you need here.
+        """
+        self.fix_pvs()
+        self.add_channels()
+
+    def fix_pvs(self):
+        """
+        Reconnect PyDM widgets to the actual PVs from the device object.
+        This is necessary because the macros aren't expanded during UI parsing.
```
ZLLentz, line 45: Is this true? If so, it's not intentional.

pcdsdevices/ui/PDU.detailed.py
```diff
@@ -0,0 +1,288 @@
...
+from typhos import utils
+
+
+class PDUDetailedWidget(Display, utils.TyphosBase):
+    """
+    Custom widget for managing the pdu detailed screen
+    """
+
+    def __init__(self, parent=None, ui_filename='PDU.detailed.ui', **kwargs):
+        super().__init__(parent=parent, ui_filename=ui_filename)
```
ZLLentz, line 18: I think the broken macros come from this init override which disables macros by absorbing them into kwargs and then not passing them on to the parent class.

See Display class at https://github.com/slaclab/pydm/blob/f4df2d419a782aa300f792181819e251c59f21a0/pydm/display.py#L332

ZLLentz: This suspicion might not be correct according to Robert, sorry if it's a red herring. We're looking into typhos a bit.

aberges-SLAC: I thought I tried this before and if I didn't have an explicitly defined `macros` in my class and defined in my `__init__`, it was complaining loudly. Now that I think about it, I probably needed to just make it an optional kwarg in the `__init__` that gets passed the magic after the typhos hook.

I think I originally went this route with the SmarActs too because `Display` was not inheriting the child devices before making the signals, so a lot of things would just break or fail to expand, and needed a small tweak. Can't say I've tried it without explicitly catching the orphaned macros

pcdsdevices/happi/containers.py
```diff
@@ -701,3 +701,19 @@ class Leviton(LCLSItem):
...
+
+class PDU(LCLSItem):
+    """
+    Container for non-Tripplite PDUs
+    """
+    device_class = copy(LCLSItem.device_class)
+    device_class.default = 'pcdsdevices.pdu.PDU'
+
+
+class TripplitePDU(LCLSItem):
```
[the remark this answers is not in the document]
TrumanPar: You're correct that we don't need this. I was naively under the impression that I needed a happi container in order for typhos to recognize the device class, but this is obviously not the case. I removed the container and I am still able to get the device screen up.

TrumanPar:
> There are also some pre-commit checks that are failing. You can install pre-commit in your local dev area so that it actually checks your changes before you commit them. It's rather painless, and I find it helps a lot!
> 
> ```
> pre-commit install
> ```
> 
> > I had to source pcds-6.0.0 during testing in order for typhos to execute my custom .py files when launching the detailed screens.
> 
> Could you expand more on this? Were the older pcds-conda environments not working? You'd need to set up a conda environment to use typhos/pydm/etc, that's not a negotiable step.

I'll run pre-commit to clean this up.
For the conda env issues, I noticed that the older pcds-conda was not working with this. I couldn't tell you why. The new release works, however.

pcdsdevices/ui/PDU.detailed.py
TrumanPar: @tangkong some of my signals are formatted components, like `tower_name = FCpt(EpicsSignal, '{prefix}{self._tower}:SetTowerName', kind='hinted')` For this example, would I use `sig://${name}.tower_name` to connect it to the widget?

pcdsdevices/pdu.py
```diff
@@ -0,0 +1,212 @@
...
+    channels_reboot = Cpt(EpicsSignal, ':setPduControlPduReboot', kind='hinted')
+
+    tab_component_names = True
+
+    def __init__(self, prefix, num_channels, name='PDU', **kwargs):
+
+        # Dynamically create channels here
+        for i in range(1, num_channels + 1):
+            ch = TripplitePDUChannel(f"{prefix}", name=f"channel{i}", channel=i)
+            setattr(self, f'channel{i}', ch)
```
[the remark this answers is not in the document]
tangkong: That may be your intention, but the fact that it is accessible by typhos/happi means it is also accessible from our other applications (e.g. hutch-python).  

I'd not classify this as a "design decision", but more as misuse of the tools ophyd gives us.  

How many different channel configurations do you expect.  Is it unreasonable to have a different subclass for each channel count?  ([a la MPod](https://github.com/KaushikMalapati/pcdsdevices/blob/741c9763f7a53063bbec5baaaaf91cb7a26f8a08/pcdsdevices/mpod_apalis.py#L329-L345))

tangkong:
> For the conda env issues, I noticed that the older pcds-conda was not working with this. I couldn't tell you why. 

I can get the ui file to load on 5.9.1.  I didn't think anything here required our new environment.  Was there a traceback to scrutinize?

pcdsdevices/ui/PDU.detailed.py
TrumanPar: Yes, we're on the same page, I'm just not being precise. I'm asking for the above example, should I put sig://${name}.tower_name in the channel property in designer to connect the widget to the component?

tangkong: Ah yes, I believe that's correct.  Sorry I misunderstood

TrumanPar: @tangkong I'm seeing this error:
`ERROR - Unable to find signal for <PyDMChannel (sig://$(name).tower_name)> in signal registry.Use typhos.plugins.register_signal()`
What is the standard way to register my signals?

tangkong: I think we need curly braces here, not parentheses

TrumanPar: Ah let me try that

TrumanPar: The macro expanded with the curly brace but I still get the error:

`ERROR - Unable to find signal for <PyDMChannel (sig://rix_pdu_r2.tower_name)> in signal registry.Use typhos.plugins.register_signal()`

pcdsdevices/ui/TripplitePDU.detailed.py
```diff
@@ -0,0 +1,312 @@
...
+            add_channel_row(right_layout, ch)
+
+
+    def monitor_pv_to_label(self, label, pvname, scale=1.0, precision=1):
+        """
+        Subscribe the PV, as saved in the ophyd component, to an EPICS
+        monitor that executes a function to scale its value and assign
+        it to the cooresponding widget's text. This works around using
+        the channel property in these widgets, which to the best of
+        my knowledge does not support scalling.
```
aberges-SLAC, line 249: I could've sworn `pyDM` had an option to declare precision. 

Maybe [this utility](https://slaclab.github.io/pydm/widgets/utilities.html#pydm.widgets.display_format.parse_value_for_display) is something you want to use instead of creating extra overhead with additional `PV` classes to monitor?

It looks like precision is handled at label creation: https://slaclab.github.io/pydm/_modules/pydm/widgets/label.html#PyDMLabel

But you're right that this does not dynamically update precision. Not sure it really warrants it for the PDU display, but I can say I am interested in it for motors.

pcdsdevices/ui/TripplitePDU.detailed.py
```diff
@@ -0,0 +1,312 @@
...
+        camonitor(pvname, callback=on_change)
+
+    def extract_outlet_number(self, pvname):
+        """
+        Helper function for sorting the channel dictionaries
+        """
+        match = re.search(r':Outlet:(\d+):', pvname)
+        return int(match.group(1)) if match else float('inf')
+
+    def update_color(self, label, pvname):
```
aberges-SLAC, line 289: Is there any reason why you can't use `pyDM`'s built in severity? https://slaclab.github.io/pydm/widgets/PyDMWidget.html#pydm.widgets.base.PyDMWidget.alarmSeverityChanged

pcdsdevices/ui/PDU.detailed.py
tangkong: Truman and I took a look at this offline and found that the happi entry wasn't properly setting the name of the ophyd device.  This was resulting in the ophyd device getting a default name (`PDU.tower_name`), while typhos macro expansion expected the name as passed from happi metadata.

This should be solved by updating the happi entry.

This bug was also hidden from me at first because I was calling typhos with a fully formed device definition, which set the name:
```
typhos 'pcdsdevices.pdu.PDU[{"prefix":"RIX:PWR:R02:01", "num_channels":24, "name": "rix_pdu_r2", "pdu_type": "sentry4"}]'
```

Whereas Truman was having typhos refer to happi
```
typhos rix_pdu_r2 --display-type detailed
```

pcdsdevices/ui/TripplitePDU.detailed.py
TrumanPar: I think I need to override that function with my color change logic. I'll see if I can get it working.

pcdsdevices/ui/TripplitePDU.detailed.py
TrumanPar:
```
def alarm_color_handler(self, widget):
"""Reusable alarm handler for any widget"""
original_method = widget.alarmSeverityChanged

python

    def alarm_handler(severity):
        # Call the original function
        original_method(severity)
        # Now call our custom function
        self.update_color(widget, severity)

    widget.alarmSeverityChanged = alarm_handler
```

This is what I've tried, but the colors don't update anymore...

@tangkong what do you think? Should I keep my original strategy, despite the extra overhead, or spend time to get it to work.

TrumanPar: yeah this is just to change background colors

aberges-SLAC: Ah, I see. I misunderstood then. Yeah, then setting up `PV` objects and handling the subscriptions is probably what I'd do

pcdsdevices/ui/TripplitePDU.detailed.py
aberges-SLAC: OH, I totally misunderstood. Duh, this is for scaling raw readbacks.

However, I noticed you're just scaling everything with the same factor and precision. Any reason why this isn't just a flat fix applied to the attributes? (I guess it _could_ reasonably change across PDU models)

pcdsdevices/ui/TripplitePDU.detailed.py
tangkong: The pvs for these temp/humidity sensors seem to be reporting undefined alarm states to me, and the actual values are 0

```
(pcds-6.0.1)roberttk@psbuild-rocky9-01:~$ caget RIX:PWR:R02:01:1:Sensor:All:SetTempLowThresh
RIX:PWR:R02:01:1:Sensor:All:SetTempLowThresh 0
(pcds-6.0.1)roberttk@psbuild-rocky9-01:~$ caget -d 34 RIX:PWR:R02:01:1:Sensor:All:SetTempLowThresh
RIX:PWR:R02:01:1:Sensor:All:SetTempLowThresh
    Native data type: DBF_DOUBLE
    Request type:     DBR_CTRL_DOUBLE
    Element count:    1
    Value:            0
    Status:           UDF
    Severity:         INVALID
    Units:            C
    Precision:        0
```

tangkong: They also show up in designer with that border if you put the raw PV into the channel field

pcdsdevices/ui/TripplitePDU.detailed.py
```diff
@@ -0,0 +1,312 @@
...
+        if off_btn:
+            off_btn.channel = f"ca://{self.device.channels_off.pvname}"
+            off_btn.pressValue = 1
+        if reboot_btn:
+            reboot_btn.channel = f"ca://{self.device.channels_reboot.pvname}"
+            reboot_btn.pressValue = 1
+
+
+        # Set PDU data
+        if name_widget:
```
aberges-SLAC, line 96: For these big boilerplate blocks, you can do:
````
pdu_attr = ['name', 'status', 'location', 'num_inputs', 'num_outputs', 'num_outlets', 'num_groups', 'num_phases', 'num_circuits', 'num_breakers', 'num_heatsinks']

for _attr in pdu_attr:
    _widget = getattr(self, _attr)
    if _widget:
        _widget.channel = f"ca://{self.device.pdu_{_attr}.pvname"
        
Although maybe this all isn't necessary anymore given that the macros should work when we pass it to the `ui` file correctly.

**Review by tangkong — COMMENTED** [review]
Mostly nitpick notes from me, I think this is looking a lot cleaner 🎉

File: pcdsdevices/ui/PDU.detailed.py [hunk 8]
```diff
@@ -0,0 +1,206 @@
...
+        Find all attributes of the PDU that start with the name 'channel'
+        Append them and their signals to a dictionary, then call the
+        function responsible for adding them to the layout
+        """
+        self.channel_signal_dict = {}
+
+        for attr in dir(self.device):
+            if attr.startswith("channel"):
+                ch_obj = getattr(self.device, attr, None)
+                if ch_obj is None or not all(hasattr(ch_obj, k) for k in ('ch_index', 'ch_name', 'ch_status', 'ch_ctrl_state')):
```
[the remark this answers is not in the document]
TrumanPar: I need to do this slightly different since `ComponentWalk` objects are not device class instances. They are structured like like this (Tripplite shown):
`ComponentWalk(ancestors=(<class 'pcdsdevices.pdu.TripplitePDU8'>,), dotted_name='channel1', item=Component(TripplitePDUChannel, channel=1, kind='normal'))`

So I check `if cpt.item.cls is TripplitePDUChannel and len(cpt.ancestors) == 1:` to get the top level channel object, and not the channel attributes themselves. This seems to work.

pcdsdevices/ui/PDU.detailed.ui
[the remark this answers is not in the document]
TrumanPar: Yeah I spent a ton of time to prevent the widgets from stretching, which can look odd in my opinion. Should I not do this moving forward?

TrumanPar:
One last nitpik of my own that I'd like feedback on @tangkong. Some of the tripplite PVs need scaling, as you already know. I can't use macro expansion for these because I need to scale the PV in my helper function, and then write that value to the text attribute in the PydmLabel widgets. I personally dislike that I have to do this differently, but I can't think of a way to get it working. I tried using a calc:// with the sig:// protocol but this seems to break. Any thoughts?

pcdsdevices/ui/PDU.detailed.py
```diff
@@ -0,0 +1,206 @@
...
+        """
+        self.channel_signal_dict_sorted = dict(sorted(self.channel_signal_dict.items(), key=lambda item: self.order_channels(item[1]['ch_index'])))
+        self.generate_rows()
+
+    def generate_rows(self):
+        """
+        Format each ophyd pdu channel component into a row and append to the layout.
+        Split channels into two collumns to keep the screen from growing length-wise
+        """
+        def add_channel_row(layout, ch_info, ch_obj):
```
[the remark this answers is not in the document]
TrumanPar: I'll do this moving forward.

pcdsdevices/ui/PDU.detailed.ui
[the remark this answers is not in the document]
tangkong: Certainly some of the stretching you've prevented is reasonable, but I was thinking widgets like the channel rows and "Temperature" label could use more horizontal space. 

image

It's not a hard and fast rule here, we're entering a somewhat subjective realm here.  If it works for the people using the screen, that's what matters.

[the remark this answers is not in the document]
TrumanPar: That does look a bit squished. My display looks a bit different. I'm not sure why this is the case.
image
image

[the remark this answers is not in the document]
tangkong: huh.  Well if it shows up like that on the hutch machines then I have no real issues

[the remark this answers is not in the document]
TrumanPar: If it's okay there I'll assume it's okay in other hutches

tangkong:
Are you looking for something like this?  I'm able to get macro substitution and the `sig://` plugin working with the `calc://` plugin

image

pcdsdevices/ui/PDU.detailed.ui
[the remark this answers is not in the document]
TrumanPar: Looks good in RIX

TrumanPar:
This was a classic case of me being goofy, kinda silly. I just didn't format the calc correctly. Thanks Robert!

TrumanPar:
@tangkong I think we're good to merge, unless you have any final comments/nitpicks.

pcdsdevices/ui/PDU.detailed.py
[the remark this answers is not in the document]
tangkong: On second thought a grid layout might be better.  But as before, no pressing need to change this.

tangkong (approved):
I think this is good to go.  The UI looks good!  (just not on my setup I guess).

The only other nitpick I might have is that some of the line lengths are a tad long.  But don't let that hold this up any longer.

PR #1377, opened 2025-09-19
Dream diagnostic paddle stage grouping

TrumanPar:
## Description

Similar to Tong's recent PR, TMO wants to group motors that are part of the same subsystem together into one screen. This is only for the 4 diagnostic sample paddle stages on DREAM. I still need to double check with scientists that removing the individual happi entries and replacing it will interfere with any scripts.

## Motivation and Context

TMO wants to clean up their LUCID screen and to also make it more convenient when using the sample paddle motors.

## How Has This Been Tested?

Interactively.

This is a poor image since the stages are all disconnected at the moment, but you can see that they match. I will be renaming the happi entry to not include the word "test"
image

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/8007-Grouping_DREAM_diagnostic_paddle_stages.rst
+++ b/docs/source/upcoming_release_notes/8007-Grouping_DREAM_diagnostic_paddle_stages.rst
@@ -1,0 +1,30 @@
+8007 Grouping DREAM diagnostic paddle stages
+#################
+
+API Breaks
+----------
+- N/A
+
+Library Features
+----------------
+- N/A
+
+Device Features
+---------------
+- Groups 4 stages into one GUI
+
+New Devices
+-----------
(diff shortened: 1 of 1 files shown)


pcdsdevices/dream_motion.py
```diff
@@ -44,3 +44,25 @@ class DREAM_SL3K4(BaseInterface, GroupDevice):
...
+
+class DREAM_Sample_Paddle(BaseInterface, GroupDevice):
+    """
+    DREAM Motion Class
+    This class controls sample paddle X,Y, Z, and Ret motors fixed to the DREAM Motion system for the
+    DREAM endstation in TMO.
+    Parameters
+    ----------
+    prefix : str
+        TMO:DREAM:MCS2:01
```
ZLLentz, line 57: Does this prefix example make sense?

Code looks good otherwise, just missing pre-release docs

PR #1390, opened 2025-11-17
Initial pass for he satt

KaushikMalapati:
## Description

Adding new classess/screens for he solid attenuator

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECSENG-1523
https://jira.slac.stanford.edu/browse/ECSENG-1467
https://jira.slac.stanford.edu/browse/ECS-10423

image
image

## Pre-merge checklist


vespos:
Let's update a couple things:

- Running HE_SATT(0.001) should actually sets the transmission to the value, not just the desired setpoint
- The output of the status is just overwhelming, let's trim it down to reporting the current transmission (both main and 3rd harmonic), the desired setpoint, the current safe_power and the state.
- There is some weird things happening with the auto-complete: 
image

- The auto-complete should be more useful and contain access to the holders' axis for example.
image

KaushikMalapati:
@vespos can you try it again? 
image

vespos:
Tested again, functionality looks good.

Do you know what these warnings are about?

```
In [19]: satt(0.01)

In [20]: WARNING  set_thread at1l2_commands_reset at1l2_commands_reset: _set_and_wait(value=1, timeout=5.0, atol=None, rtol=None, kwargs={})
WARNING  set_thread at1l2_commands_request_trans at1l2_commands_request_trans: _set_and_wait(value=1, timeout=5.0, atol=None, rtol=None, kwargs={})
```

ZLLentz:
Maybe that's this timeout warning from ophyd:
https://github.com/bluesky/ophyd/blob/ed7d4bb73ef0dd870ec63228664dc727c85541cb/ophyd/signal.py#L404

I think this would happen when we put a value and then don't see the result happen in the readback for 5.0 seconds?

KaushikMalapati:
@NSLentz can you please review this?

pcdsdevices/ui/HE_SATT.py
```diff
@@ -0,0 +1,56 @@
...
+        for cpt in self.device.walk_components():
+            if cpt.item.cls is HE_SATT_Commands and len(cpt.ancestors) == 1:
+                self.commands = getattr(self.device, cpt.dotted_name)
+
+    def reset_and_request_return_callback(self):
+        self.reset_and_request_timer.singleShot(500, self.reset_and_request_timer_callback)
+
+    def reset_and_request_timer_callback(self):
+        self.commands.reset.put(1)
+        time.sleep(1)
```
janeliu-slac: `time.sleep(1)`  might not work well here. In a pydm GUI this will block the main Qt event loop thread, so Qt is frozen for a second and can't can't process any user clicks or update PV values. `QTimer.singleShot(1000, self.request_transmission)` is probably better. It delays by 1 second without blocking the Qt main thread.

pcdsdevices/ui/HE_SATT.py
```diff
@@ -0,0 +1,56 @@
...
+        self.commands.request_trans.put(1)
+
+    def turn_big_error_red(self, text):
+        if abs(float(text)) > 30:
```
janeliu-slac, line 46: Is `text` guaranteed to always have a value? If `text` is empty, `float(text)` can raise a ValueError.

pcdsdevices/ui/HE_SATT.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self, parent=None, ui_filename="HE_SATT.ui", macros=None, **kwargs):
+        super().__init__(parent=parent, ui_filename=ui_filename, macros=macros, **kwargs)
+
+        self.reset_and_request_timer = QTimer()
+        self.ui.req_trans_edit_pydm.returnPressed.connect(self.reset_and_request_return_callback)
+        self.ui.predicted_error_label_pydm.textChanged.connect(self.turn_big_error_red)
+        self.ui.state_label_pydm.currentTextChanged.connect(self.turn_state_error_red)
+
+    @property
+    def device(self) -> Device:  # type: ignore
```
janeliu-slac: `Device | None` might be a better return type because of the try-except.

vespos (dismissed):

pcdsdevices/ui/HE_SATT.py
KaushikMalapati: The value is set via a calc expression that is always numerical, so text will always have a value that can be cast into a float

pcdsdevices/ui/HE_SATT.py
janeliu-slac: This doesn't seem to have been resolved but was marked as resolved? Not a dealbreaker but this would prevent potential issues where the main Qt thread doesn't update PVs correctly or a user clicks on something but the changes never go through.

KaushikMalapati: I made the change but I can't retest it until later today after which I'll push the commit and request another review

KaushikMalapati:
@janeliu-slac can you please re-review now?

janeliu-slac (dismissed):

ZLLentz:
Let's merge this one next, then we'll merge the ci fixes etc. into the other one

.pre-commit-config.yaml
```diff
@@ -2,7 +2,7 @@
 # See https://pre-commit.com/hooks.html for more hooks
 exclude: |
   (?x)^(
-    pcdsdevices/_version.py|
+    pcdsdevices/_version.py|pcdsdevices/tests/test_lens_sets/test
```
ZLLentz, line 5: Thank you for including this

pcdsdevices/attenuator.py
```diff
@@ -1885,3 +1887,150 @@ def render_ascii_att(blade_states, *, start_index=0):
...
+    holder3 = Cpt(HE_SATT_Wedge_Holder, ':Holder:03', kind='normal')
+    holder3_axis = Cpt(BeckhoffAxis, ':MMS:03', kind='normal')
+    holder3_velo = Cpt(PytmcSignal, ':MMSOOP:03:fVelocity', io='io', kind='normal')
+    holder4 = Cpt(HE_SATT_Wedge_Holder, ':Holder:04', kind='normal')
+    holder4_axis = Cpt(BeckhoffAxis, ':MMS:01', kind='normal')
+    holder4_velo = Cpt(PytmcSignal, ':MMSOOP:01:fVelocity', io='io', kind='normal')
+
+    def __call__(self, transmission):
+        self.commands.reset.put(1)
+        time.sleep(1)
```
ZLLentz, line 2017: Thoughts for the future: maybe a future version of this wouldn't need magic sleeps

docs/source/upcoming_release_notes/1390-HE_SATT.rst
ZLLentz, line 1: We're not using these any more- probably best to delete this page

ZLLentz (dismissed):
Approving but we should remove the pre-release page at some point

ZLLentz (approved):

PR #1409, opened 2026-04-02
Lxt update position

carogee:
## Description

Added code so lxt position will update.

## Motivation and Context

lxt motor was not updating position.

## How Has This Been Tested?

Tested in XCS.

## Pre-merge checklist


ZLLentz:
This seems reasonable, but I have to wonder: what makes lxe special such that it needs an extra call here, where other devices with notepad signals don't need this?

carogee:
I thought that motor PVs are only supposed to update when there is a position change? I think with this motor they are just setting the current position to be zero without moving the motor, but they want the associated plots to immediately show the same position as zero.

ZLLentz:
I understand now- I agree with the fix as-is. I think we merge with just the docs update in the checklist.

One thing I want to check related to this is if there are other special motors where we forgot to do this in `set_current_position` etc.

ZLLentz:
Yes, this affects the other base class here too, same bug:

https://github.com/pcdshub/pcdsdevices/blob/e3d6e458deddff61ee705e0542b4ff1c71669ba3/pcdsdevices/pseudopos.py#L209-L230

We can fix that separately (not your scope) but I wanted to find and note other cases

ZLLentz (approved):
I'll edit the text slightly when we make a release and switch your name for your github username, but for expedience let's merge as-is

PR #1419, opened 2026-05-15
Updating hla disabled error message

KaushikMalapati:
## Description

Changing the exception message from `HLA is not enabled` to 'Laser Locker Timing: Calibration Required' when you attempt to move an lxt while hla is not enabled.

## Motivation and Context

https://slac.slack.com/archives/C03D4JT7U15/p1778806386537109
This exception happened correctly yesterday but the exception (and what to do about it) was unclear. The new message comes from the timing team

## How Has This Been Tested?

Not tested because I only changed a string

## Pre-merge checklist


yannS2016 (approved):
LGTM! Perhaps a second approval is needed here.

ipotanin (approved):

PR #1420, opened 2026-05-21
ENH: add undulator pointing classes from mfx

ZLLentz:
## Description

- Import the commits from mfx related to undulator pointing over to this repo
- Modify slightly to clean up docs and un-mfx it
- Add tests

This needs some additional work to figure out why the tests fail entirely sometimes

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-8835
https://slac.slack.com/archives/C03D4JT7U15/p1779313409663469

## How Has This Been Tested?

I made some unit tests to make sure the sim devices have consistent behavior but they seem strangely flaky? This needs some additional attention.

Here only

## Pre-merge checklist


pcdsdevices/undpoint.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+def _put_2d_delta(mds: MultiDerivedSignal, value: tuple[float, float]) -> SignalToValue:
+    # Note on the polarity:
+    # Setting 50 to the delta PV causes the real value to move by -50
+    # So, when you ask for 50 we'll put -50 to the delta PV instead, to move the real value by 50
+    return {mds.signals[0]: -value[0], mds.signals[1]: -value[1]}
+
+
+def coerce_input_to_tuple(
```
KaushikMalapati, line 52: Is it just for convenience that the functions can take either two floats or a tuple with two floats?

pcdsdevices/undpoint.py
```diff
@@ -0,0 +1,439 @@
...
+    position: Union[tuple[float, float], float], ypos: Optional[float]
+) -> tuple[float, float]:
+    """
+    Utility to coerce user input like move(3, 4) to move((3, 4))
+    """
+    if isinstance(position, tuple):
+        if len(position) == 2:
+            return position
+        raise ValueError(f"Expected tuple of length 2, got {position}")
+    elif ypos is None:
```
KaushikMalapati, line 62: Is it worth checking if position is not None like ypos? At this point we only know it's not a tuple

pcdsdevices/undpoint.py
```diff
@@ -0,0 +1,439 @@
...
+        done_pvname: str = DEFAULT_DONE_MOVE_PV,
+        x_abs_pvname: str = "BPMS:UNDH:4690:XOFF.D",
+        y_abs_pvname: str = "BPMS:UNDH:4690:YOFF.D",
+        **kwargs,
+    ):
+        self.done_pvname = done_pvname
+        self.x_abs_pvname = x_abs_pvname
+        self.y_abs_pvname = y_abs_pvname
+        self._xpos_cache = None
+        self._ypos_cache = None
```
KaushikMalapati, line 186: Why do we cache these values when we can get them from the signals?

pcdsdevices/undpoint.py
```diff
@@ -0,0 +1,439 @@
...
+            )
+
+        # for segmented moves, we require synchronous execution
+        if not wait:
+            raise ValueError("Safe chunked move requires wait=True.")
+
+        # determine number of equal segments so each per-axis step magnitude <= max_step
+        n_x = (
+            int(math.ceil(abs(dx_total) / max_step))
+            if max_step and abs(dx_total) > 0
```
KaushikMalapati, line 320: I think we don't need to check if max_step because if it was None we would have returned in the non-segmentation case

KaushikMalapati:
Had a few nitpicks, but mainly I think we need to update the branch, run pre-commit, and get the tests passing.

pcdsdevices/undpoint.py
ZLLentz: Yes, the intent is to be lenient for:
`undp.move(1, 2)`
`undp.move((1, 2))`

Passing around tuples is a bit nicer in code but the extra braces are easy to miss in interactive use

pcdsdevices/undpoint.py
ZLLentz: Yeah I think this isn't implemented as fluidly as it could be- suggestions welcome.

pcdsdevices/undpoint.py
ZLLentz: This is so `undp.position` always returns immediately like an attribute should rather than possibly incurring an EPICS timeout.

pcdsdevices/undpoint.py
ZLLentz: Full disclosure: I didn't write the "safe" class at all- someone added it later (see imported commit history). It's not obvious to me that this should be a separate class at all, for example. (opinions welcome)

Putting on my reviewer hat, I also thought some of the code here was redundant but I didn't want to move it around until I had a unit test in place, and then I spent my afternoon getting upset at the unit test being flaky... maybe I revisit it today.

ZLLentz:
I want to fix the flakiness of the tests and give the safe pointing another look- thanks for the review.
The tests pass offline some % of the time.

pcdsdevices/undpoint.py
ZLLentz: Rereading this function now, I see it was coded in a very defensive way- lots of codepaths that probably don't matter unless someone else mucks with the rest of the code.

I think since I didn't write it and don't have a feel for how it has been tested in prod at MFX, I'm going to err on the side of leaving it alone for now.

ZLLentz:
The tests should pass every time now, I think?

KaushikMalapati (approved):

PR #1421, opened 2026-05-24
Add TwinCATMREAxis for motor-record emulation support

yannS2016:
## Summary
Adds a BeckhoffAxis configurable PLC suffix and `TwinCATMREAxis`, a motor-record edition of `TwinCATAxis`. The TwinCAT firmware now exposes a motor-record-compatible presentation; this keeps both layouts available. Also unifies the TwinCAT axis prefix convention and consolidates signal components onto `TwinCATMotorInterface`.

## What changed
- BeckhoffAxis PLC sub-device suffix was hardcoded to ':PLC:'. It's now a per-instance plc_suffix argument (default ':PLC:'), so newer layouts can pass plc_suffix=':' (new motion-library PVs) without forking the class. The default keeps legacy instantiations unaffected.
- New `TwinCATMREAxis(TwinCATAxis)`. Components rebind to tcmotor record fields (`.VAL`, `.RBV`, `.DMOV`, `.STOP`, `.VELO`/`.VBAS`/`.VMAX`, `.ACCS`, `.MOVN`, `.EGU`, `.SPMG`, `.MSTA`, `.LLM`/`.HLM`, `.HOMF`/`.HOMR`).
- Limit switches read `.HLS`/`.LLS` directly (no InvertedBoolEpicsSignal); `fwd_enabled`/`bwd_enabled` set to `None` so they drop off the Typhos screen.
- New `EnabledDisabledSignal` for soft-limit enable longin records lacking ONAM/ZNAM.
- Two latent bugs fixed in TwinCATMREAxis: `reset()` targets `reset_signal`; `is_homed` reads the `bHomed` signal.

## Prefix convention change (BREAKING)
`TwinCATMotorInterface` and `TwinCATAxis` migrated to a colonless prefix convention: component suffixes lead with ':' and plc uses ':'. Both classes are now instantiated with a colonless prefix (e.g. 'TST:MOTION:M1' instead of 'TST:MOTION:M1:'). Existing Typhos users must drop the trailing colon from their launch strings; they will be notified. This unification lets TwinCATMREAxis inherit the PLC-tag components instead of redefining ~12 of them just to prepend a colon.

## Consolidation
- `homed`, `home_mode`, and the soft-limit enables moved up to `TwinCATMotorInterface`; it now owns all signal components.
- `low/high_limit_enable` upgraded to `EnabledDisabledSignal` so the ONAM/ZNAM-less records display Enabled/Disabled for both classes.
- `TwinCATAxis` now holds only behavior (methods) and the plc sub-device. `TwinCATMREAxis` redefines only genuine record-field differences plus MRE-only fields (velocity_base/max, spmg, msta, home_cmd_fwd/rev).

## What did not change
`TwinCATAxisEPS` is unchanged.

## Testing
Instantiated via Typhos against a motor-record emulation IOC (`TST:MOTION:M1`). Verified motion config, limits, homing fields, and that PLC motion errors propagate to per-field alarm borders.
image

Related work: https://github.com/pcdshub/ioc-common-ads-ioc/pull/119

## Known follow-ups
- tcmotor output fields (.HLM/.LLM/.VAL) show a UDF/INVALID alarm border in Typhos until first processed; to be addressed IOC-side (PINI / MS flag review on INP_HLM/INP_LLM).
- EnabledDisabledSignal registered as FakeEpicsSignalRO; it's writable, so consider FakeEpicsSignal when adding tests.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -1296,14 +1296,18 @@ class BeckhoffAxis(EpicsMotorInterface):
     __doc__ += basic_positioner_init
     tab_whitelist = ['clear_error']
 
-    plc = Cpt(BeckhoffAxisPLC, ':PLC:', kind='normal',
-              doc='PLC error handling and aux functions.')
+    plc = FCpt(BeckhoffAxisPLC, '{prefix}{_plc_suffix}', kind='normal',
+               doc='PLC error handling and aux functions.')
     motor_spmg = Cpt(EpicsSignal, '.SPMG', kind='config',
                      doc='Stop, Pause, Move, Go')
 
     # Clear the normal homing PVs that don't really work here
     home_forward = None
     home_reverse = None
+
+    def __init__(self, prefix, *, plc_suffix=':PLC:', **kwargs):
+        self._plc_suffix = plc_suffix
+        super().__init__(prefix, **kwargs)
 
     def subscribe(
         self,
--- a/pcdsdevices/twincat_motor.py
+++ b/pcdsdevices/twincat_motor.py
@@ -581,7 +581,7 @@ class TwinCATAxis(TwinCATMotorInterface):
         - A `home()` method exposing all PLC-configured homing routines with
           proper mode setting, actuation, status, and timing.
         - Robust session-tracking and logging logic for post-move diagnostics.
-   Note
+    Note
    ----
     Some methods in this class are identical or adapted from BeckhoffAxis,
     to ensure the error-handling, staging, homing, and value checking logic
--- a/pcdsdevices/twincat_motor.py
+++ b/pcdsdevices/twincat_motor.py
@@ -73,7 +73,7 @@ class TwinCATMotorInterface(FltMvInterface, PVPositioner):
     docstring for duplication/adaptation notes.
 
     Notes
-   --
+    -----
     The PCDS interface preferences and design patterns implemented here are:
 
     1. Uses the `FltMvInterface` mixin for name aliases and advanced functionality, such as improved
@@ -209,7 +209,7 @@ class TwinCATMotorInterface(FltMvInterface, PVPositioner):
         """The current position of the motor in its engineering units
 
         Returns
-       ----
+        ----
         position : float
         """
         return self.readback.get()
@@ -325,7 +325,7 @@ class TwinCATMotorInterface(FltMvInterface, PVPositioner):
         Set the low limit. Adapted from EpicsMotorInterface.
 
         Parameters
-       ----
+        ----
         value : float
             Limit of travel in the negative direction.
 
@@ -355,7 +355,7 @@ class TwinCATMotorInterface(FltMvInterface, PVPositioner):
         Limit of travel in the positive direction. Adapted from EpicsMotorInterface.
 
         Parameters
-       ----
+        ----
         value : float
             High limit value to be set.
 
@@ -398,7 +398,7 @@ class TwinCATMotorInterface(FltMvInterface, PVPositioner):
         Duplicated logic from EpicsMotorInterface for API consistency.
 
         Returns
-       -
+        ----
         moving : bool
         """
         return bool(self.motor_is_moving.get(use_monitor=False))
@@ -421,12 +421,12 @@ class TwinCATMotorInterface(FltMvInterface, PVPositioner):
     def enabled(self):
         """
         Returns
-       ----
+        ----
         bool
             True if power is enabled **and** either negative or positive direction is enabled.
 
         Explanation
-       --------
+        --------
         - Returns True if:
             - Power is on (`power_is_enabled`)
             - At least one direction (`negative_dir_enabled` OR `positive_dir_enabled`) is enabled
@@ -578,7 +578,7 @@ class TwinCATAxis(TwinCATMotorInterface):
           proper mode setting, actuation, status, and timing.
         - Robust session-tracking and logging logic for post-move diagnostics.
    Note
-   ---
+   ----
     Some methods in this class are identical or adapted from BeckhoffAxis,
     to ensure the error-handling, staging, homing, and value checking logic
     behaves identically.
@@ -771,7 +771,7 @@ class TwinCATMREAxis(TwinCATAxis):
     :meth:`home` and :meth:`check_value`).
 
     Notes
-   --
+    -----
     The PLC sub-device (``plc``) and the deceleration/jerk/enable-mode/
     direction-status tags have no ``tcmotor`` record equivalent, so they
     continue to use :class:`PytmcSignal` against the PLC PVs.
--- a/pcdsdevices/twincat_motor.py
+++ b/pcdsdevices/twincat_motor.py
@@ -53,8 +53,9 @@ class EnabledDisabledSignal(EpicsSignal):
             value = self._disable_map.get(value, value)
         super().put(value, **kwargs)
 
+
 """
-EnabledDisabledSignal & InvertedBoolEpicsSignal  registered as FakeEpicsSignalRO; 
+EnabledDisabledSignal & InvertedBoolEpicsSignal  registered as FakeEpicsSignalRO;
 it's writable, so consider FakeEpicsSignal when adding tests.
 """
 fake_device_cache[InvertedBoolEpicsSignal] = FakeEpicsSignalRO
--- a/pcdsdevices/twincat_motor.py
+++ b/pcdsdevices/twincat_motor.py
@@ -3,9 +3,10 @@ import logging
 from typing import Callable, ClassVar, Optional
 
 from ophyd.device import Component as Cpt
+from ophyd.device import FormattedComponent as FCpt
 from ophyd.device import required_for_connection
 from ophyd.pv_positioner import PVPositioner
-from ophyd.signal import DerivedSignal, EpicsSignal
+from ophyd.signal import DerivedSignal, EpicsSignal, EpicsSignalRO
 from ophyd.sim import FakeEpicsSignalRO, fake_device_cache
 from ophyd.status import MoveStatus
 from ophyd.status import wait as status_wait
@@ -29,7 +30,33 @@ class InvertedBoolEpicsSignal(DerivedSignal):
         return not value
 
 
+class EnabledDisabledSignal(EpicsSignal):
+    """
+    EpicsSignal for longin enable/disable records that lack ONAM/ZNAM.
+
+    Converts 0 -> 'Disabled', 1 -> 'Enabled' for display.
+    Accepts 'Enable'/'Disable' or 0/1 for writes.
+
+    Used by the motor-record (MRE) TwinCAT axis variants whose soft-limit
+    enable PVs are plain longin records without string field definitions.
+    """
+    _enable_map = {0: 'Disabled', 1: 'Enabled'}
+    _disable_map = {'Enable': 1, 'Disable': 0,
+                    'Enabled': 1, 'Disabled': 0,
+                    1: 1, 0: 0}
+
+    def get(self, **kwargs):
+        val = super().get(**kwargs)
+        return self._enable_map.get(int(val), str(val))
+
+    def put(self, value, **kwargs):
+        if isinstance(value, str):
+            value = self._disable_map.get(value, value)
+        super().put(value, **kwargs)
+
+
 fake_device_cache[InvertedBoolEpicsSignal] = FakeEpicsSignalRO
+fake_device_cache[EnabledDisabledSignal] = FakeEpicsSignalRO
 
 
 class TwinCATMotorInterface(FltMvInterface, PVPositioner):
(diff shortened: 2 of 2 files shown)


pcdsdevices/epics_motor.py
```diff
@@ -1296,15 +1296,19 @@ class BeckhoffAxis(EpicsMotorInterface):
     __doc__ += basic_positioner_init
     tab_whitelist = ['clear_error']
 
-    plc = Cpt(BeckhoffAxisPLC, ':PLC:', kind='normal',
-              doc='PLC error handling and aux functions.')
+    plc = FCpt(BeckhoffAxisPLC, '{prefix}{_plc_suffix}', kind='normal',
```
ZLLentz, line 1299: Question: can you give one example where this is either used or planned to be used? I read the description but I don't understand why this change is needed.

pcdsdevices/twincat_motor.py
```diff
@@ -47,7 +73,7 @@ class TwinCATMotorInterface(FltMvInterface, PVPositioner):
     docstring for duplication/adaptation notes.
 
     Notes
-    -----
+   --
```
ZLLentz: Change request: please revert these formatting changes to all of the `-----` etc. lines in the docstrings (there are dozens of these in the diff). This likely breaks the docstring parsing for sphinx docs generation.

Do you know how these changes got included?

pcdsdevices/twincat_motor.py
```diff
@@ -29,7 +29,33 @@ def inverse(self, value: bool):
         return not value
 
 
+class EnabledDisabledSignal(EpicsSignal):
```
ZLLentz, line 32: Side-comment: this is maybe a bit confusing to me vs modifying signal metadata but it seems reasonably simple so I don't have any complaints

pcdsdevices/twincat_motor.py
```diff
@@ -724,6 +751,173 @@ def check_value(self, pos: float) -> None:
...
+        """
+        Pulse the reset signal.
+
+        Fixes the legacy ``reset`` which called ``self.reset.put`` (the bound
+        method) instead of the ``reset_signal`` component.
+        """
+        self.reset_signal.put(1, wait=False)
+
+    @raise_if_disconnected
+    def home(self, wait: bool = True, **kwargs):
```
ZLLentz, line 887: Question: did you test these functions? I didn't see any mention of testing these in the PR description.

pcdsdevices/twincat_motor.py
```diff
@@ -29,7 +29,33 @@ def inverse(self, value: bool):
...
+        return self._enable_map.get(int(val), str(val))
+
+    def put(self, value, **kwargs):
+        if isinstance(value, str):
+            value = self._disable_map.get(value, value)
+        super().put(value, **kwargs)
+
+
 fake_device_cache[InvertedBoolEpicsSignal] = FakeEpicsSignalRO
+fake_device_cache[EnabledDisabledSignal] = FakeEpicsSignalRO
```
ZLLentz, line 62: > EnabledDisabledSignal registered as FakeEpicsSignalRO; it's writable, so consider FakeEpicsSignal when adding tests.

Suggestion: do this now? It's not a monumental change, but it's in your known follow-ups section.

ZLLentz (commented):
Seems mostly good, see comments

pcdsdevices/twincat_motor.py
yannS2016: sure.

pcdsdevices/epics_motor.py
yannS2016: This mainly involves backward compatibility changes, allowing the new motion library PV name format to be used with older EPICS motor screens by providing the ':' suffix instead of ':PLC:'. If using the current motion library, the change is transparent.

pcdsdevices/twincat_motor.py
yannS2016: I must have omitted that. But yes, the function works; the MRE will force the home mode to the expected one, whether the user sets it or not.

ZLLentz: Right, but please update the description accordingly then. You tested doing a move too, right?

pcdsdevices/twincat_motor.py
ZLLentz: > Do you know how these changes got included?

pcdsdevices/epics_motor.py
ZLLentz: If there is no example, I would request a reversion?

yannS2016: I believe making the library code compatible with existing screens is a valuable improvement. This is also a good way to show that our MRE is functioning and transparent to the user. The change itself does not break BeckhAxis. So it's transparent in both directions.

pcdsdevices/twincat_motor.py
yannS2016: I had code stored in different scripts using various formats that I bundled together while testing these changes.

pcdsdevices/twincat_motor.py
yannS2016: By "move," do you mean absolute move? If so, yes, all features that were in TwinCATAxis work as expected in TwinCATMREAxis.

pcdsdevices/epics_motor.py
ZLLentz: My perspective as a library maintainer is that all code that isn't intended to be used is bad for maintenance. I asked for an example because I wanted to try to understand how and where this code will be used. I can make any number of transparent changes that don't break anything that are not good for code health. So please help me understand who and what this feature is for.

pcdsdevices/twincat_motor.py
ZLLentz: Right, but there (is? was?) no documented test of any of this, and then there's a commit post-review that you fixed a basic move action, right?

pcdsdevices/twincat_motor.py
ZLLentz: How does that result in changing (breaking) many of the docstrings though, in a way where it made it into the PR review? I'm trying to understand how we can avoid these sorts of regressions without needing to discuss them.

pcdsdevices/epics_motor.py
yannS2016: I do not consider this a maintenance headache; the change is light and transparent. The new motion stack hasn’t been deployed yet, so it’s difficult to provide a use case right now. This change bridges a gap for those who want a smooth migration and prefer to use their existing GUI.

pcdsdevices/twincat_motor.py
yannS2016: Well,  it's on me, though. I should have given it a 'git diff' review before pushing.

pcdsdevices/epics_motor.py
ZLLentz: Right, but I don't understand this change at all. When I see:

> making the library code compatible with existing screens

That doesn't mean anything to me. Why would any specific thing not be compatible without this change?
And then you say, it's difficult to provide an example- so, why is it here?

I want a concrete example and I shouldn't have to beg for it if this change is well-motivated.

ZLLentz (changes requested):
Putting the changes requested mark on at least until I understand all the inclusions in the PR

pcdsdevices/epics_motor.py
yannS2016: I honestly don't understand why there is any contention here. I mentioned that this change is intended to bridge a gap so that the PVs from the new motion library can be used with the existing Epic's motor-compatible screens. Why wouldn't that be a sufficient reason?

ZLLentz: It's not sufficient because I don't understand why PVs from the new motion library cannot be used with the existing screens.

ZLLentz: I'm harping on this not to be mean, but to make sure we're sufficiently documenting what features exist and why they exist. I assumed one example would be simple to get and enough for me (and others) to understand.

ZLLentz: I can't stress enough how extremely frustrating this thread is to me. Please just give me a usage example like I requested hours ago, this doesn't have to be like pulling teeth. When you ignore my question/request I'm not just going to say "ahh let's move on", I'm going to get confused further. One example and this whole thread would probably have been prevented.

ZLLentz: Even the example of how you tested this feature would be enough

yannS2016: I don't quite understand what you're expecting from me. I'm not asking for a free pass, but rather a legitimate reason for the review to hold on this. I have provided the context and reasoning for this change. Everything in this pull request has been thoroughly tested. I’ve already discussed the differences between the new motion library and the old one in previous conversations/meetings. Why not simply pull these changes, load the BeckhoffAxis from Typhos with an IOC build using the new motion library, and observe the changes for yourself?

ZLLentz: Show me how I would use this kwarg, one example. That's all I want.

pcdsdevices/twincat_motor.py
yannS2016: The recent change pushed here alone was not the source of the issue when considered in isolation. However, when integrated with the TCMotor emulation internal logic, it created a conflict. The most reasonable and obvious solution was to implement the fix here.

yannS2016:
I'm not planning on writing unit tests for twincat_motor in this run, but it's part of the upcoming effort for this work.

pcdsdevices/twincat_motor.py
ZLLentz: What I mean is, why put off using the correct class for the fake device cache? Not why not do the tests now? I understand that suddenly creating a bunch of useful unit tests is nontrivial, but removing `RO` is.

pcdsdevices/epics_motor.py
yannS2016: New motion Library invocation ( use` : `prepended to BeckhoffPLC PVs)
typhos "pcdsdevices.epics_motor.BeckhoffAxis[{'prefix':'TST:MOTION:M1', 'name':'epics:motor:mr-emulatio
n', 'plc_suffix':':'}]" &
Current motion library invocation ( uses `:PLC:` prepended to BeckhoffPLC PVs) - invocation stays as is, no changes required.
typhos "pcdsdevices.epics_motor.BeckhoffAxis[{'prefix':'TST:MOTION:M1', 'name':'epics:motor'}]" &

pcdsdevices/twincat_motor.py
ZLLentz: Right, but my request is documentation: your PR description doesn't mention the full scope of the testing you did, right?

pcdsdevices/epics_motor.py
ZLLentz: Thank you, this helps my understanding: I now understand why the code exists and what would happen if it wasn't here, and I can ponder alternatives as a reviewer instead of just being confused.

Previously what I was missing was that the old `BeckhoffAxis` class could be and was being used to control new motion library axes (that is: you'd done such a transparent job in the translation layer that it could work in-place). So, I could not understand why it was being edited at all: I thought it had already been discarded. But now I see the full picture that I was missing.

pcdsdevices/twincat_motor.py
```diff
@@ -294,12 +329,12 @@ def set_low_limit(self, value):
...
 
         Parameters
-        -------
+        ----
         value : float
             Limit of travel in the negative direction.
 
         Raises
-        ---
+
```
ZLLentz: Last thing before approve: I don't want to go through and point all of these out, but please revert all the `----` changes in the docstrings. Some of them are benign, but others like this one definitely break the sphinx autodocs which needs to follow the numpy style.

pcdsdevices/twincat_motor.py
yannS2016: It would be too lengthy to list every single feature test in the description, so having a test suite would be helpful to ensure everyone is on the same page. Most of these features have already been tested in the field via TwinCATAxis. I am currently testing the TwinCAT MRE axis alongside the motor record emulation, and I haven't seen anything that suggests it does not respond accordingly.

pcdsdevices/twincat_motor.py
```diff
@@ -547,13 +582,13 @@ class TwinCATAxis(TwinCATMotorInterface):
           proper mode setting, actuation, status, and timing.
         - Robust session-tracking and logging logic for post-move diagnostics.
    Note
```
ZLLentz: Unlike the other docstring stuff, this one was wrong beforehand- under-indented by one space. Again this likely breaks the sphinx docs if we care about this. Maybe this one being misaligned is why the following lines got pulled back too?
```suggestion
    Note
```
I can come through and clean these in another PR- but the ones that broke in this PR should be reverted first

pcdsdevices/twincat_motor.py
ZLLentz: I'm giving you advice for future PRs where you didn't have time for unit tests: any time you don't say you tested something, you'll be asked if you tested it- anything less would be an incomplete review. Please document at minimum "I tested the homing and move commands", for example, and this process gets a lot easier.

pcdsdevices/epics_motor.py
yannS2016: Based on the description and context I provided, it’s not that the BeckhAxis class is currently being used with the new motion library; rather, it has the potential to be used with it. The changes aim to bridge that gap. Effective changes are typically incremental, and creating this bridge will facilitate smoother adoption of the new motion stack.

ZLLentz: Right, but the description doesn't explain this fully, hence my confusion. The reader is expected to know your plans.

yannS2016: The discussion shifted from why the change was made to how to use it. Indeed, the description should have included the invocation calls. Had this been the initial request, I would have provided the missing information.

pcdsdevices/twincat_motor.py
yannS2016: actively cleaning up the doc strings. We do want to preserve our documentation.

pcdsdevices/epics_motor.py
ZLLentz: Right, but I didn't understand what was missing until I saw the example- hence my question being for the example, because I expected it to help me bridge the understanding gap.

pcdsdevices/twincat_motor.py
yannS2016: I appreciate the suggestion, but I want to flag that this PR already has a Testing section documenting exactly that. It states the change was instantiated via Typhos against a motor-record emulation IOC (TST:MOTION:M1), and that I verified motion config, limits, homing fields, and that PLC motion errors propagate to per-field alarm borders. That's more specific than "I tested the homing and move commands", it names the harness, the target IOC, and the concrete behaviors checked

pcdsdevices/epics_motor.py
ZLLentz: Thinking about this one more time I'm confused again (I'm sorry...)

Why would someone use the old class in this way? Is this very narrowly scoped to typhos screens? Naively I'd assume all new uses would use the new class in all forms.

yannS2016:
I'll fix the doc string issues. and push a commit tomorrow.

pcdsdevices/twincat_motor.py
yannS2016: Doc string formats have been restored.

pcdsdevices/epics_motor.py
ZLLentz: Putting the lid on this thread:
- I disagree that this should be included in the code
- I'm not going to ask for a removal

pcdsdevices/twincat_motor.py
ZLLentz: There's nothing in there that implies you called `home` or `move` a single time on the class here. I make these suggestions after having read the PR description and contents thoroughly.

ZLLentz:
The diff is clean here now. My only remaining suggestions were related to the PR description in the testing section (missing details) and confusion over why a class change on the fake epics signal class is being deferred (a two line-change). If you're sure that you don't want to do these let me know.

ZLLentz:
We talked offline- cleared up confusion. Lesson-learned: talking via text on github in depth is impossible.

We're going to merge this PR and not worry about if there will be a follow-up or not.

ZLLentz (approved):

PR #1423, opened 2026-06-05
add compact diffractometer class

vespos:
## Description
Add class for XPP compact diffractometer device

## Motivation and Context
Device left behind at the end of LCLS-1 era. Will be used again in run 27.

FIx description of GBS as well.

## How Has This Been Tested?
Used in operation

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/gon.py
+++ b/pcdsdevices/gon.py
@@ -21,6 +21,29 @@ from .utils import get_status_float, get_status_value
 logger = logging.getLogger(__name__)
 
 
+class CompactDiffractometer(BaseInterface, Device):
+    """
+    Class for diffractometer.
+    Parameters
+    ----------
+    name : str
+        A name to refer to the device
+    prefix_base : str
+        The EPICS base PV of the diffractometer stages
+    """
+
+    base_h = Cpt(BeckhoffAxis, 'BASE_H', kind='normal')
+    base_v = Cpt(BeckhoffAxis, 'BASE_V', kind='normal')
+    th = Cpt(BeckhoffAxis, 'TH', kind='normal')
+    tth = Cpt(BeckhoffAxis, '2TH', kind='normal')
+    chi = Cpt(BeckhoffAxis, 'CHI', kind='normal')
+
+    tab_component_names = True
+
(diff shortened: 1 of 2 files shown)


pcdsdevices/gon.py
```diff
@@ -21,6 +21,29 @@
...
+    base_h = Cpt(BeckhoffAxis, 'BASE_H', kind='normal')
+    base_v = Cpt(BeckhoffAxis, 'BASE_V', kind='normal')
+    th = Cpt(BeckhoffAxis, 'TH', kind='normal')
+    tth = Cpt(BeckhoffAxis, '2TH', kind='normal')
+    chi = Cpt(BeckhoffAxis, 'CHI', kind='normal')
+
+    tab_component_names = True
+
+    def __init__(self, prefix, name, **kwargs):
+        super().__init__(prefix, name=name, **kwargs)
```
KaushikMalapati, line 44: No need for __init__

pcdsdevices/gon.py
```diff
@@ -21,6 +21,29 @@
 logger = logging.getLogger(__name__)
 
 
+class CompactDiffractometer(BaseInterface, Device):
```
KaushikMalapati, line 24: This class looks identical to HxrDiffractometer except the tth component uses 2TH instead of TTH for the pv. What is the relationship between these devices? Maybe HxrDiffractometer doesn't exist anymore and we just update the pv or something

pcdsdevices/gon.py
```diff
@@ -21,6 +21,29 @@
...
+    """
+    Class for diffractometer.
```
ZLLentz, line 26: Change request: can you expand this docstring with more context on what this device is? I think people will stumble upon this as the very first item in gon.py and be confused as to what device or devices this controls.

pcdsdevices/gon.py
vespos: I think it is the same device... Thanks for pointing it out!
I suspect that there was some version lingering in dev at XPP and did not notice that this had already been done. I think I'll close this PR!

PR #1427, opened 2026-06-22
CI/TST/BLD: fix various incompatibilities

ZLLentz:
## Description

The CI was broken in pcdsdevices. I made the following changes:
- Edit the broken test for the updated rules in the changelog for pytest 9.1.0, which now unpacks single-element tuples as one would expect in normal python syntax, which ends up breaking the parameterize here. Also update a clearly malformed test element.
  - Note: this test doesn't really do anything and should be a candidate for removal or revision. Even with extremely bad inputs the test passes.
- Drop setuptools_scm back to before version 10 because it seems to fix a strange build issue
- Rearrange the ophyd device generator to not require whatrecord on import so we can remove whatrecord (and therefore, epicsmacrolib) from the CI requirements, which is currently broken even on the versions that build for other packages

Possible Follow-up tasks:

- [ ] Learn precisely which version of setuptools_scm breaks us and why
- [ ] Figure out if we can get a new environment built that has whatrecord included at all anywhere
- [ ] Consider scrutinizing the test suite for bad/useless tests

## Motivation and Context

CI has been failing on this repo for at least a week

## How Has This Been Tested?

The CI passes now.

Here only

## Pre-merge checklist

### DIFF
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -1,7 +1,6 @@
 docs-versions-menu
 sphinx >=3.3.1
 sphinx_rtd_theme
-whatrecord
 ipython
 # Required for compatibility with Sphinx 6.0+:
 sphinxcontrib-jquery
--- a/pcdsdevices/make_ophyd_device.py
+++ b/pcdsdevices/make_ophyd_device.py
@@ -142,8 +142,9 @@ def write_file(lines, name):
 
 
 if __name__ == '__main__':
-    from whatrecord import db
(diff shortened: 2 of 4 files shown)


KaushikMalapati (approved):

ZLLentz:
One more food for thought: I'm considering switching our pypi builds from the setuptools backend to the uv backend because it's much faster, this could theoretically sidestep problems with setuptools.

PR #1428, opened 2026-06-24
CI: GHA SHA test in advance of enterprise requirements

ZLLentz:
## Description

Update the pcds-ci-helpers reference to use the latest SHA
Add a dependabot config to keep the SHA periodically synchronized

## Motivation and Context

At the start of July this will be required or else all GHA builds will no longer run
https://jira.slac.stanford.edu/browse/ECS-10557

I want to get this working manually on a few repositories before trying to automate the update everywhere
In pcdsdevices, I will check that setting to the latest SHA "just works" and that dependabot won't have anything to do yet.

I'm picking a monthly update so we don't get crazy spam PRs from dependabot too often, but we still stay up-to-date.
A repo that needs an immediate update can force dependabot to run early via the github ui.

## How Has This Been Tested?

I tested dependabot lightly over in pcds-ci-helpers
I'm planning to use the results of this CI run as a green light that this is the correct way to do an update

Here only

## Pre-merge checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    cooldown:
+      default-days: 7
+      exclude:
+      - pcdshub/pcds-ci-helpers/*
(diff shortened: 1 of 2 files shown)


ZLLentz:
I am again reminded that I need to add someone else to the admin team to review these
I will request someone from python-reviewers instead for now

KaushikMalapati (approved):

PR #1432, opened 2026-07-09
FIX: Update for PseudoPositioner to handle PseudoSingle's delegated moves correctly

janeliu-slac:
## Description
This fixes a bug where a PseudoPositioner has a PseudoSingle as one of its axes and waits indefinitely for the move status, which never arrives. This is because PseudoSingles delegate the move to their parent PseudoPositioner. The parent reports the move as done using a callback, but the current PseudoPositioner ignores the parent and waits for the PseudoSingle child to report the move as done. This never happens, so the PseudoPositioner's status hangs and blocks other operations.

The solution is to have the PseudoPositioner object track the axis it asked to move, instead of trusting the callback's obj identity.

## Motivation and Context
https://jira.slac.stanford.edu/browse/ECS-4381

## How Has This Been Tested?
Need to test this in a hutch.

Documentation is in the code.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/pseudopos.py
+++ b/pcdsdevices/pseudopos.py
@@ -206,6 +206,22 @@ class PseudoPositioner(ophyd.pseudopos.PseudoPositioner):
         self._update_notepad_ioc(position, 'notepad_setpoint')
         return status
 
+    def _concurrent_move(self, real_pos, **kwargs):
+        """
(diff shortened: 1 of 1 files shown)


janeliu-slac:
This PR was opened on the wrong branch.

PR #1433, opened 2026-07-09
ENH: add new-format TwinCAT MAL state positioner classes and screen

yannS2016:
## Description
Add state positioners for the new `lcls-twincat-motion-abstraction` state PV format, where **motion parameters are not part of the state**,  velocity/acceleration/deceleration are owned by the drive layer (`FB_MotionStageNC` via `MotionCmd`), not `ST_PositionState`.

**`state.py`** : `TwinCATMalStatePositioner`, `TwinCATMalStateConfigOne`/`Dynamic`, implemented as thin *extend + turn-off* subclasses of `TwinCATStatePositioner`:
- drop the aggregated `state_velo` signal and the per-state `:VELO` field;
- expose the per-state setpoint as read-only (`:SETPOINT_RBV`);
- keep the `:GET_RBV`/`:SET` interface, error/reset PVs, and `:NAME` enum sourcing unchanged.

**`pmps.py`** : `TwinCATMalStatePMPS`, the new-format counterpart of `TwinCATStatePMPS`. 
- inherits the new-format turn-off (no `state_velo`, no-velo config) from `TwinCATMalStatePositioner`;
- keeps the PMPS PVs (`arb_enable`, `maint_mode`) and in/out semantics from `TwinCATStatePMPS`;
- `config = UpCpt(state_count=2)` re-pins the in/out default (2 states) onto the no-velo config.

**UI templates:**
- `TwinCATMalStatePositioner.detailed.ui`: The standard state screen with the Configuration tab **removed** (that tab only ever showed the now-removed `state_velo` summary; the plain state has no configurable content).
- `TwinCATMalStatePMPS.detailed.ui`: keeps the Configuration tab, since the PMPS variant has real config content (`arb_enable`, `maint_mode`).

## Motivation and Context
The new TwinCAT state format decouples motion setup from the position state. The legacy `TwinCATStatePositioner`/`TwinCATStatePMPS` can't fully connect to it (they assume a `:VELO` field and a writable setpoint). This adds compatible positioners, plain and PMPS without disturbing the widely-used legacy classes.

## How Has This Been Tested?
- Live against a **TST PLC / ads-ioc running the latest ADS IOC** (`TST:MOTION:STATE`): verified connection, state readback / per-state `NAME_RBV`/`SETPOINT_RBV`/`MOVE_OK_RBV` reads, and state moves 
- `pytest pcdsdevices/tests/test_state.py` → 12 passed (no regressions).
- Typhos screen verified; results in the screenshots below.

Class docstrings in `pcdsdevices/state.py` (`TwinCATMalStatePositioner`, `TwinCATMalStateConfigOne`).

**_Not too sold on the new names, open to suggestions._**

## Screenshots
**State Mover**
image
image
image

**PMPS**
image
image
image

## Follow-up
- Add pytest coverage for the new state mover (`TwinCATMalStatePositioner` / `TwinCATMalStateND`) 

## Pre-merge checklist


ZLLentz (dismissed):
pre-commit is unhappy, but no objections from me: it looks like you have taken the old classes and modified/documented them for the new library in exactly the ways that they differ.

(I looked pretty closely at the implementations and compared them to what we were running before and made sure I understood the decisions)

yannS2016:
> pre-commit is unhappy, but no objections from me: it looks like you have taken the old classes and modified/documented them for the new library in exactly the ways that they differ.
> 
> (I looked pretty closely at the implementations and compared them to what we were running before and made sure I understood the decisions)

Thanks for the feedback, strange that pre-commit is complaining i'll check it out.

ZLLentz:
@yannS2016 side question: are these force pushes a necessary/helpful part of your workflow? From a reviewer point of view it obfuscates my tracking of what I did and did not review so far, I can't filter by "new commits" for example.

ZLLentz (dismissed):
Looks like the only thing that changed is the pre-commit fix, I think? It's not obvious if the ui changed or not.

pcdsdevices/pmps.py
```diff
@@ -22,3 +24,16 @@ class TwinCATStatePMPS(TwinCATInOutPositioner):
...
+    """
+    Combines the ``lcls-twincat-motion-abstraction`` new-format state behavior of
+    `TwinCATMalStatePositioner` (motion parameters live on the drive layer, so
+    there is no per-state ``:VELO`` field or ``state_velo`` summary) with the
+    PMPS protection PVs and in/out semantics of `TwinCATStatePMPS`.
+
+    Unlike the plain `TwinCATMalStatePositioner`, this still has meaningful
+    Configuration content: the ``arb_enable`` and ``maint_mode`` PMPS PVs.
+    """
+    config = UpCpt(state_count=2)
```
KaushikMalapati, line 39: Doesn't this already have state_count=2 inherited from TwinCATStatePMPS inherited from TwinCATInOutPositioner?

ZLLentz: I thought it did too on my first review, but the multi-inheritance tree is super confusing and I think it actually ends up with TWINCAT_MAX_STATES as the state count

pcdsdevices/ui/TwinCATMalStatePMPS.detailed.ui
KaushikMalapati, line 1: This is identical to TwinCATStatePositioner.detailed.ui. Doesn't TwinCATMalStatePMPS inherit from that class and already use its ui file?

pcdsdevices/state.py
```diff
@@ -515,6 +515,30 @@ class TwinCATStateConfigOne(Device):
...
+
+    - removes the ``:VELO`` field, and
+    - makes the setpoint read-only (``:SETPOINT_RBV``).
+
+    ``state_name`` and ``move_ok`` are inherited unchanged.
+    """
+    # Motion parameters are owned by the drive layer, not the state.
+    velo = None
+    # Setpoint is read-only in the new format (only :SETPOINT_RBV exists).
+    setpoint = Cpt(PytmcSignal, ':SETPOINT', io='i', kind='config',
```
KaushikMalapati: I think UpCpt also could have been used here to change io

yannS2016:
> @yannS2016 side question: are these force pushes a necessary/helpful part of your workflow? From a reviewer point of view it obfuscates my tracking of what I did and did not review so far, I can't filter by "new commits" for example.

Just to avoid having to storm the PR with commits. but i get it that it can be cumbersome for the reviewer.

pcdsdevices/pmps.py
yannS2016: @ZLLentz this is what i was seeing in my test as well.

pcdsdevices/ui/TwinCATMalStatePMPS.detailed.ui
yannS2016: The difference here would that the plain state positioner doe not have a config component anymore while the pmps one still has one minus the motion parameters components

yannS2016:
> > @yannS2016 side question: are these force pushes a necessary/helpful part of your workflow? From a reviewer point of view it obfuscates my tracking of what I did and did not review so far, I can't filter by "new commits" for example.
> 
> Just to avoid having to storm the PR with commits. but i get it that it can be cumbersome for the reviewer.

pcdsdevices/ui/TwinCATMalStatePMPS.detailed.ui
KaushikMalapati: https://github.com/pcdshub/pcdsdevices/blob/38f09dbeed94639bf36893e1e05161b4c61a5c9a/pcdsdevices/ui/TwinCATStatePositioner.detailed.ui has the config tab. It has an identical ui file

yannS2016: If TwinCATMalStatePMPS.detailed.ui didn't exist, typhos would  fall through to TwinCATMalStatePositioner.detailed.ui and render the plain screen with no config tab which is wrong for PMPS. but instead of a copy here, i could jsut create a sym link.

KaushikMalapati: Forgot about TwinCATMalStatePositioner.detailed.ui. I think a symlink is a good idea

ZLLentz (approved):

KaushikMalapati (approved):

ZLLentz:
> Just to avoid having to storm the PR with commits. but i get it that it can be cumbersome for the reviewer.

It's not a huge deal- mostly was curious (I definitely should do this more, my PR commit streams are messy as heck)

There's probably a good balance to hit between squashing some commits together pre-review and having more restraint after reviews flow in

PR #1434, opened 2026-07-09
ENH: Integrated Grating Beam Splitter (GBS) into LODCM energy calculation

jcotter814:
## Description
Added functionality to gbs.py to return the beam splitter grating period for a given target. Used values for motor positions provided by a hutch scientist. Modified he_lodcm.py to read current beam splitter grating period when calculating position and energy. Modified test_he_lodcm.py to account for changes to gbs.py and updated test results to reflect a 650nm period instead of 600nm, which was incorrect. Verified behavior with hutch scientist.

## Motivation and Context

This change is required so users do not need to manually enter the GBS grating period when calculating or moving the LODCM.

https://jira.slac.stanford.edu/browse/ECS-10600

## How Has This Been Tested?

Verified behavior with hutch scientist. Updated test values based on feedback. Updated test_he_lodcm.py to account for changes to GBS. Ran standard PCDS test suite.

Tested using pytest in pcds_conda environment on psbuild-rocky9.

Comments in gbs.py explain new functionality.

## Pre-merge checklist


pcdsdevices/gbs.py
```diff
@@ -40,3 +41,63 @@ class GratingBeamSplitterTarget(BaseInterface, GroupDevice,
 
     rtd1 = Cpt(TwinCATTempSensor, ':RTD:01', kind='normal', doc='First RTD.')
     rtd2 = Cpt(TwinCATTempSensor, ':RTD:02', kind='normal', doc='Second RTD.')
+
+    def get_current_grating_period(self):
```
ZLLentz: Suggestion: type annotation for return type
```suggestion
    def get_current_grating_period(self) -> float | None:
```

pcdsdevices/gbs.py
```diff
@@ -40,3 +41,63 @@ class GratingBeamSplitterTarget(BaseInterface, GroupDevice,
...
+
+        Returns
+        -------
+        str
+            current target grating
+        """
+
+        return self.get_grating_period(self.target.position)
+
+    def get_grating_period(self, target_pos):
```
ZLLentz: Suggestion: type annotations for inputs and returns
```suggestion
    def get_grating_period(self, target_pos: str) -> float | None:
```

pcdsdevices/gbs.py
```diff
@@ -40,3 +41,63 @@ class GratingBeamSplitterTarget(BaseInterface, GroupDevice,
...
     rtd2 = Cpt(TwinCATTempSensor, ':RTD:02', kind='normal', doc='Second RTD.')
+
+    def get_current_grating_period(self):
+        """
+        Get the grating period based on current grating target position.
+
+        Returns
+        -------
+        str
+            current target grating
```
ZLLentz: Suggestion: revise because this is incorrect- we return a float or None, right? Not a str?

pcdsdevices/gbs.py
```diff
@@ -40,3 +41,63 @@ class GratingBeamSplitterTarget(BaseInterface, GroupDevice,
...
+        return self.get_grating_period(self.target.position)
+
+    def get_grating_period(self, target_pos):
+        """
+        Get the grating period associated with a given target position based on hardcoded
+        table provided by Haoyuan.
+
+        Returns
+        str
+            current target grating period
```
ZLLentz: Suggestion: fix formatting (missing ---- line), correct return type, and clarify that this isn't necessarily the "current" target grating period (it's the grating period from the inputs)

pcdsdevices/gbs.py
```diff
@@ -40,3 +41,63 @@ class GratingBeamSplitterTarget(BaseInterface, GroupDevice,
...
+            state_obj = getattr(self.target.config, state)
+            target_name = state_obj.state_name.get()
+            setpoint = state_obj.setpoint.get()
+            target_map[target_name] = np.abs(setpoint)
+
+        # check if target exists
+        if target_pos not in target_map.keys():
+            raise ValueError(f"{target_pos} not in list, possible targets are: {list(target_map.keys())}")
+
+        grating_y_position = target_map[target_pos]
```
ZLLentz, line 86: Optional suggestion (possible future task?): to me, this seems pretty indirect, we're starting from the state str, converting it to a position, and then comparing against hard-coded position values.

I think we ought to pick either working in state-name space or position-value space to make it easier to keep this updated if the state positions change, or else this function will start returning `None` if everything shifts by a few mm (or worse, we might get states mixed up with each other)

My suggested options are:

- Switch the input to the y position, get the current y position from `self.y_motor.position` as needed in the other function, compare the y position directly against the hard-coded values. This simplifies the implementation.
- Or, switch the `positions_650` and `positions_980` arrays to be lists of state strings instead, and compare to the state string directly.

pcdsdevices/tests/test_he_lodcm.py
ZLLentz, line 1: Comment: the tests seem reasonable to me, it's maybe notable that only the 980 states are exercised but the code is similar in the implementation so it probably doesn't matter.

pcdsdevices/tests/test_he_lodcm.py
```diff
@@ -68,35 +74,43 @@ def test_crystal_invalid_t1(fake_he_lodcmenergy):
...
     motor_setup(energy.t1ty)
     motor_setup(energy.t2ty1)
     motor_setup(energy.t2ty2)
     energy.t1ty.user_readback.sim_put(5)
     energy.t2ty1.user_readback.sim_put(5)
     energy.t2ty2.user_readback.sim_put(5)
+    energy.beam_splitter.target._late_state_init(enum_strs=target_states)
+    energy.beam_splitter.target.state.sim_put(2)
+    energy.beam_splitter.target.config.state03.state_name.sim_put('TARGET1a')
+    energy.beam_splitter.target.config.state03.setpoint.sim_put(33.105)
```
ZLLentz: Optional suggestion: you also had the option of including these setup steps inside the `fake_he_lodcmenergy` fixture, instead of redoing them the same way in each test.

ZLLentz (commented):
Code seems fine :+1:, good job getting the tests to pass again. I have some suggestions for future edits for maintainability but the main hangup for merging is the factual errors in the docstrings.

pcdsdevices/he_lodcm.py
```diff
@@ -50,11 +52,12 @@ class HE_LODCMEnergy(FltMvInterface, PseudoPositioner):
     t2ty1 = Cpt(BeckhoffAxis, ':MMS:T2Ty', kind='normal', doc='Tower 2 translation Y 1')
     t2ty2 = Cpt(BeckhoffAxis, ':MMS:T2Ty2', kind='normal', doc='Tower 2 translation Y 2')
 
+    beam_splitter = FCpt(GratingBeamSplitterTarget, 'PF2L0:GBS')
```
KaushikMalapati, line 55: Why is this a formatted component?

pcdsdevices/gbs.py
jcotter814: Yes, good catch I'll fix that. It's left over from from original way I wrote it which was to return the target name.

pcdsdevices/tests/test_he_lodcm.py
jcotter814: Haoyuan said they only use the 980nm states because the transmission on the 650nm are poor, but maybe it's better to test both for completeness sake? I wasn't sure if there was a preferred way to do this - testing each period for each crystal forward and reverse seemed a bit overkill, but mixing and matching arbitrarily seemed unprincipled. Perhaps I am overthinking it.

pcdsdevices/tests/test_he_lodcm.py
jcotter814: Yes, I think that's a better way to do it.

pcdsdevices/tests/test_he_lodcm.py
ZLLentz: I think it's probably fine as-is

One way to do this is to create a table of inputs/outputs and use `pytest.parametrize` to run through all the combinations

jcotter814: I ended up modifying the tests to test the 650nm and 980nm cases separately. Doing so revealed that I had made a mistake - I had simput the state name of TARGET1a to the TARGET1b state, so
```
    energy.beam_splitter.target.state.sim_put(2)
    energy.beam_splitter.target.config.state03.state_name.sim_put('TARGET1a')
    energy.beam_splitter.target.config.state03.setpoint.sim_put(33.105)
```
was testing 650nm values for a target name that should have actually been 980nm. I split the 650nm and 980nm cases into two fixtures to hopefully make things more explicit and less prone to this kind of error.

pcdsdevices/he_lodcm.py
jcotter814: I am unfortunately blanking on why I specifically chose to use a formatted component here... @ZLLentz I think you may have advised me to use FCpt, but I could be misremembering.

KaushikMalapati: Actually using a FCpt here is correct since the gbs is its own device and doesn't share its prefix with the lodcm which I forgot when I left the comment. Using an FCpt will stop the component from inheriting the prefix from the parent and allow you to use a prefix for a different device. If it was a normal component, you would be instatiating GratingBeamSplitterTarget with the prefix `{HE_LODCMEnergyObject.prefix}PF2L0:GBS` instead of just PF2L0:GBS

pcdsdevices/gbs.py
jcotter814: Yes, I agree. I think doing it with state strings is more in line with the original intention, which was to map between the PLC state and the setpoint associated with that state, so that if the motor setpoint changes, the mapping between target and motor position is still correct.

pcdsdevices/he_lodcm.py
ZLLentz: Yes, exactly- the other way to do this is `Cpt(..., suffix=())` (or something similar, memory is hard) which is even less intuitive

ZLLentz (approved):
I'm :+1: on this now (was only holding it back for the outdated docstrings)

Some future work is identified, but I don't see any reason to not merge the PR

PR #1435, opened 2026-07-10
FIX: Update for PseudoPositioner to handle PseudoSingle's delegated moves correctly

janeliu-slac:
## Description
This update fixes a bug where a PseudoPositioner has a PseudoSingle as one of its axes and waits indefinitely for the move status.  

**Issue:** When a nested PseudoSingle receives a move request from the PseudoPositioner, it does not perform the move itself. Instead, it delegates the move to its parent PseudoPositioner. The parent reports that the move is done using a callback, but the current PseudoPositioner is waiting for the PseudoSingle child to report that it is done. Since the PseudoSingle didn't perform the move directly, that callback never arrives. The current PseudoPositioner's status hangs and blocks other operations.

**Solution:** Have the PseudoPositioner track the axis it asked to move, instead of trusting which object the callback reports as done.

## Motivation and Context
https://jira.slac.stanford.edu/browse/ECS-4381

## How Has This Been Tested?
(Need to test this in a hutch)

Documentation is in the method `_concurrent_move()`.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/pseudopos.py
+++ b/pcdsdevices/pseudopos.py
@@ -206,6 +206,35 @@ class PseudoPositioner(ophyd.pseudopos.PseudoPositioner):
         self._update_notepad_ioc(position, 'notepad_setpoint')
         return status
 
+    def _concurrent_move(self, real_pos, **kwargs):
+        """
+        Move all real positioners in parallel.
+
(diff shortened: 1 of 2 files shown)


janeliu-slac:
Closing this PR because its history became cluttered during a rebase, making the changes difficult to review. Will open a clean replacement PR.

PR #1436, opened 2026-07-22
`btms_confg` ENH: add new destinations, update text for current ones

aberges-SLAC:
## Description
Updates for `btms_config` based on requests from Laser Team during Summer Down Time:
* Add new `Laser Destinations` (LD):
    * 3RIX (LD2 just renamed)
    * Laser Lab 2 (new line)
    * TXI (new line)
* Update hard-coded PV references to alignment cameras.
   * Should we be doing this here?
*  Renamed `Laser Sources` (LS) to use standard `LS#` name, since everything is now 1 micron or a singleton source (e.g. Zone 1)

## Motivation and Context
Requested by Laser Systems Engineering team during the Summer Down Time

## How Has This Been Tested?
Currently untested, but is a bugfix on incorrect PVs anyway and just docstring updates

So far just this PR

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/lasers/btms_config.py
+++ b/pcdsdevices/lasers/btms_config.py
@@ -165,12 +165,12 @@ class SourcePosition(str, enum.Enum):
         str
         """
         return {
-            SourcePosition.ls1: "Bay 1",
-            SourcePosition.ls3: "Bay 2 1um",
-            SourcePosition.ls4: "Bay 2 800nm",
-            SourcePosition.ls5: "Bay 3 800nm",
-            SourcePosition.ls6: "Bay 3 1um",
-            SourcePosition.ls8: "Bay 4",
+            SourcePosition.ls1: "Bay 1 LS1",
+            SourcePosition.ls3: "Bay 2 LS3",
+            SourcePosition.ls4: "Bay 2 LS4",
+            SourcePosition.ls5: "Bay 3 LS5",
+            SourcePosition.ls6: "Bay 3 LS6",
+            SourcePosition.ls8: "Bay 4 LS8",
         }.get(self, "Unknown")
 
     def is_above(self, other: SourcePosition) -> bool:
@@ -207,10 +207,10 @@ class SourcePosition(str, enum.Enum):
         """
         return {
             SourcePosition.ls1: "las_lhn_bay1_cam_nf",
-            SourcePosition.ls3: "las_lhn_bay2_amphos_nf",
-            SourcePosition.ls4: "las_lhn_bay2_amphos_nf",
-            SourcePosition.ls5: "las_lhn_bay3_amphos_nf",
-            SourcePosition.ls6: "las_lhn_bay3_amphos_nf",
+            SourcePosition.ls3: "las_lhn_bay2_ls3_nf",
+            SourcePosition.ls4: "las_lhn_bay2_ls4_nf",
+            SourcePosition.ls5: "las_lhn_bay3_ls5_nf",
+            SourcePosition.ls6: "las_lhn_bay3_ls6_nf",
             SourcePosition.ls8: "las_lhn_bay4_cam_nf",
         }.get(self, None)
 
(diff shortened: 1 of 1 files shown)


pcdsdevices/lasers/btms_config.py
KaushikMalapati: Should this be called RIX 3RIX?

pcdsdevices/lasers/btms_config.py
```diff
@@ -304,8 +304,10 @@ def description(self) -> str:
             DestinationPosition.ld4: "RIX ChemRIXS",
             DestinationPosition.ld6: "RIX QRIXS",
             DestinationPosition.ld8: "TMO IP1",
-            DestinationPosition.ld9: "Laser Lab",
+            DestinationPosition.ld9: "Laser Lab 1",
             DestinationPosition.ld10: "TMO IP2",
```
KaushikMalapati: Should this say DREAM somewhere?

KaushikMalapati:
> * Update hard-coded PV references to alignment cameras.
>   * Should we be doing this here?

Maybe, maybe not. If this is only used in btms-ui, maybe you let the pvs be inputs in class and pass them there. If you use it multiple locations, you might find it convenient that everything is standardized to use the same pvs (for a given version of pcdsdevices). You could also store a configuration somewhere else, or pick a standard way of naming alignment cameras for each laser source and change the camera ioc pvs/aliases to match. Leaving it here is probably fine, but maybe not that convenient to update. Other laser teamers might have an opinion about whether it needs to be changed

pcdsdevices/lasers/btms_config.py
slactjohnson: Yeah, that's a better name.

pcdsdevices/lasers/btms_config.py
slactjohnson: I've heard DREAM and IP2 used fairly interchangeably in TMO, but perhaps we can check with Mat Britton and see if there is a preference.

slactjohnson:
> > * Update hard-coded PV references to alignment cameras.
> >   
> >   * Should we be doing this here?
> 
> Maybe, maybe not. If this is only used in btms-ui, maybe you let the pvs be inputs in class and pass them there. If you use it multiple locations, you might find it convenient that everything is standardized to use the same pvs (for a given version of pcdsdevices). You could also store a configuration somewhere else, or pick a standard way of naming alignment cameras for each laser source and change the camera ioc pvs/aliases to match. Leaving it here is probably fine, but maybe not that convenient to update. Other laser teamers might have an opinion about whether it needs to be changed

The hard coded PVs are annoying. It would be better to load these from e.g. the HAPPI database. But I don't hate it enough to say that this _must_ be changed. We touch this system once a year or less. I think it's fine as is. I think there are more impactful things to do.

ZLLentz:
We've got bigger fish to fry, but this entire file feels out of place in `pcdsdevices`. Hardcoded or not, given some extra cycles it would make sense to move this elsewhere.

ZLLentz (dismissed):
Given that this is already here, I see nothing objectionable in the diff.

I'm curious how the application that uses this is deployed and if we'd want to migrate it to using the pixi envs or not.

Please don't merge until someone from the laser groups signs off.

slactjohnson (approved):
The only changes I'd like to see are the small changes regarding 3rd RIXS and TMO IP2/DREAM. Everything else looks fine.

slactjohnson (changes requested):
Sorry, meant to request changes.

pcdsdevices/lasers/btms_config.py
KaushikMalapati: It would also match the edited comment on line 388

pcdsdevices/lasers/btms_config.py
KaushikMalapati: On line 393, Adam already changed the comment for DestinationPosition.ld10 from "# TMO IP2" to "# TMO IP2 (DREAM)" so I thought this should match

pcdsdevices/lasers/btms_config.py
aberges-SLAC: Ah! Thank you, I thought I did this but missed it in my commit.

aberges-SLAC:
> We've got bigger fish to fry, but this entire file feels out of place in `pcdsdevices`. Hardcoded or not, given some extra cycles it would make sense to move this elsewhere.

I agree and think this honestly should just live in the `btms_ui` repo where it is being utilized in the first place. I agree with Tyler in that I touch it infrequently enough to not feel strongly about doing it right now, but it _is_ weird to wait for a `pcdsdevices` release tag for something that is arguably unrelated and for configuration changes. Maybe if time permits later in the SDT I will extract this from this repo.

slactjohnson (approved):

KaushikMalapati (approved):

PR #1439, opened 2026-08-10
Enh smaract ethercat

yannS2016:
## Description
Adds device support and a typhos screen for SmarAct MCS2 stages driven through a Beckhoff EtherCAT DS402 PLC (as opposed to the serial MCS2 IOC).

- New `SmarActEtherCAT` (combines `SmarAct` config/diagnostics with `BeckhoffAxis` motor record, PLC error handling, and homing) and `SmarActEtherCATOpenLoop`.
- Adds DS402 config/diagnostic components : diagnostic closed-loop frequency (max/avg), safe direction, referencing/calibration options, sensor mode, hold time, step scaling, operation-mode request, motor load, decoded channel-state flags (`SmarActChannelState`), etc.
- Removes components with no DS402 backing (module temperature, default range, need-calibration readback) and re-points the rest to their pytmc record names.
- Adapts `SmarAct.detailed.ui`/`SmarAct.detailed.py`: re-points renamed pytmc PVs and hides widgets whose objects have no DS402 backing (orphaned TTZV threshold label and chanState RBV widgets) so they do not sit disconnected.

## Motivation and Context
The DREAM EtherCAT SmarAct stages run through a TwinCAT DS402 PLC, which exposes a reduced, pytmc-named object set that the existing serial `SmarAct` device and its detailed screen do not match. Without this, components connect to the wrong PV names (or write-only PVs with no readback) and the screen shows disconnected widgets.
https://jira.slac.stanford.edu/browse/ECS-10975

## How Has This Been Tested?
Loaded `SmarActEtherCAT` and its typhos detailed screen interactively against alive MCS2 EtherCAT IOC (`DREAM:DGPD:MMT:RET`). Confirmed the DS402 config/diagnostic PVs read back and that unsupported widgets are hidden rather than disconnected. Also perfomed full motion control on hardware: homing, closed-loop moves, and open-loop step operations.
image
image

### Default serial interface screens test
Also verified the shared serial path using the default serial interfa **SmarAct device class**. Built and launched a test IOC from the common SmarAct IOC and drove a real SmarAct stage over the serial interface. Focused on the calibrate command and confirmed with @aberges-SLAC  it fires exactly once per request without re-triggering. Every operation ran through the default SmarAct screens, including our changes. All motion tests passed.
image

Component-level `doc=` strings and class docstrings on `SmarActEtherCAT` / `SmarActEtherCATOpenLoop`, plus explanatory comments in `maybe_fix_ethercat_pvs`.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -1922,7 +1922,7 @@ class SmarActEtherCAT(SmarAct, BeckhoffAxis):
                      doc='Motor load')
     channel_type = Cpt(PytmcSignal, ':CHAN_TYPE', io='io', kind='config',
                        doc='Channel type')
-    op_mode_req = Cpt(EpicsSignal, ':OPMODE_REQ', kind='config',
+    op_mode_req = Cpt(PytmcSignal, ':OPMODE_REQ', io='io', kind='config',
                       doc='DS402 operation mode request')
     home_mode = Cpt(PytmcSignal, ':HOME_MODE', io='i', kind='config',
                     doc='Vendor homing mode')
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -1908,13 +1908,16 @@ class SmarActEtherCAT(SmarAct, BeckhoffAxis):
                           doc='Referencing options')
     reference_type = Cpt(PytmcSignal, ':REF_TYPE', io='i', kind='config',
                          doc='Reference type')
-    safe_direction = Cpt(PytmcSignal, ':SAFE_DIR', io='io', kind='config',
+    safe_direction = Cpt(PytmcSignal, ':SAFE_DIR', io='i', kind='config',
                          doc='Safe direction (0x200C). For positioners '
                              'referenced by a mechanical end stop, homing and '
                              'calibration ignore the requested start direction '
                              'and use this plus the logical scale inversion. '
-                             'FORWARD=0, BACKWARD=1. Stored in NVRAM; '
-                             're-calibrate after changing.')
+                             'FORWARD=0, BACKWARD=1. Read-only here: this is an '
+                             'R(W)* object, writable only in the Pre-Op ESM '
+                             'state, so configure it via the CoE startup list '
+                             'or vendor software. Stored in NVRAM; re-calibrate '
+                             'after changing.')
     motor_load = Cpt(PytmcSignal, ':MOTOR_LOAD', io='i', kind='normal',
                      doc='Motor load')
     channel_type = Cpt(PytmcSignal, ':CHAN_TYPE', io='io', kind='config',
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -1908,16 +1908,16 @@ class SmarActEtherCAT(SmarAct, BeckhoffAxis):
                           doc='Referencing options')
     reference_type = Cpt(PytmcSignal, ':REF_TYPE', io='i', kind='config',
                          doc='Reference type')
-    safe_direction = Cpt(PytmcSignal, ':SAFE_DIR', io='i', kind='config',
+    safe_direction = Cpt(PytmcSignal, ':SAFE_DIR', io='io', kind='config',
                          doc='Safe direction (0x200C). For positioners '
                              'referenced by a mechanical end stop, homing and '
                              'calibration ignore the requested start direction '
                              'and use this plus the logical scale inversion. '
-                             'FORWARD=0, BACKWARD=1. Read-only here: this is an '
-                             'R(W)* object, writable only in the Pre-Op ESM '
-                             'state, so configure it via the CoE startup list '
-                             'or vendor software. Stored in NVRAM; re-calibrate '
-                             'after changing.')
+                             'FORWARD=0, BACKWARD=1. This is an R(W)* object, '
+                             'writable only in the Pre-Op ESM state, so the PLC '
+                             'applies a new setpoint only while the drive is in '
+                             'Pre-Op. Stored in NVRAM; re-calibrate after '
+                             'changing.')
     motor_load = Cpt(PytmcSignal, ':MOTOR_LOAD', io='i', kind='normal',
                      doc='Motor load')
     channel_type = Cpt(PytmcSignal, ':CHAN_TYPE', io='io', kind='config',
--- a/pcdsdevices/epics_motor.py
+++ b/pcdsdevices/epics_motor.py
@@ -1523,6 +1523,18 @@ class MotorDisabledError(Exception):
     pass
 
 
+def _set_long_name(signal, long_name):
+    """
+    Assign ``long_name`` to a component signal if it exists.
+
+    Subclasses may remove unsupported components by setting the corresponding
+    Component to None. This helper skips those so that ``__init__`` long-name
+    assignments do not raise AttributeError.
+    """
+    if signal is not None:
+        signal.long_name = long_name
+
+
 class SmarActOpenLoop(Device):
     """
     Class containing the open loop PVs used to control an un-encoded SmarAct
@@ -1569,16 +1581,16 @@ class SmarActOpenLoop(Device):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # Long name shenanigans
-        self.step_voltage.long_name = 'Step Voltage'
-        self.step_freq.long_name = 'Step Frequency'
-        self.jog_step_size.long_name = 'Jog Step Size'
-        self.jog_fwd.long_name = 'Jog Forward'
-        self.jog_rev.long_name = 'Jog Backward'
-        self.total_step_count.long_name = 'Total Step Count'
-        self.step_clear_cmd.long_name = 'Clear Step Count'
-        self.scan_move.long_name = 'Scan Voltage'
-        self.channel_temp.long_name = 'Channel Temp. (°C)'
-        self.module_temp.long_name = 'Module Temp. (°C)'
+        _set_long_name(self.step_voltage, 'Step Voltage')
+        _set_long_name(self.step_freq, 'Step Frequency')
+        _set_long_name(self.jog_step_size, 'Jog Step Size')
+        _set_long_name(self.jog_fwd, 'Jog Forward')
+        _set_long_name(self.jog_rev, 'Jog Backward')
+        _set_long_name(self.total_step_count, 'Total Step Count')
+        _set_long_name(self.step_clear_cmd, 'Clear Step Count')
+        _set_long_name(self.scan_move, 'Scan Voltage')
+        _set_long_name(self.channel_temp, 'Channel Temp. (°C)')
+        _set_long_name(self.module_temp, 'Module Temp. (°C)')
 
 
 class SmarActTipTilt(Device):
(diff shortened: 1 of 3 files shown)


pcdsdevices/ui/SmarAct.detailed.py
```diff
@@ -146,6 +147,126 @@ def fix_pvs(self):
...
+        Adapt the serial detailed screen for EtherCAT SmarAct stages.
+
+        The screen targets the serial MCS2 IOC. The EtherCAT DS402 stage,
+        detected here by its decoded ``channel_state`` device, exposes a
+        reduced object set with pytmc naming. Re-point the widgets whose PV
+        only changed name, and hide the widgets whose object has no DS402
+        backing so they do not sit disconnected.
+        """
+        if not hasattr(self.device, 'channel_state'):
+            return
```
ZLLentz: Nitpick/optional suggestion: is there a way we can make this more explicit? I appreciate that only the new class has `channel_state` but I'd prefer something more explicit e.g. an `isinstance` check

pcdsdevices/epics_motor.py
```diff
@@ -1724,6 +1736,198 @@ def __init__(self, prefix='', *, tip_pv, tilt_pv, **kwargs):
...
+    This combines the SmarAct-specific configuration and diagnostics with the
+    BeckhoffAxis motor record, PLC error handling, and homing. The DS402
+    interface exposes fewer objects than the serial MCS2 IOC, so unsupported
+    components (module temperature, default range, and the need-calibration
+    readback) are removed, and the remaining components are re-pointed to
+    their pytmc record names.
+    """
+    # DS402 has no separate module temperature, default range, or
+    # need-calibration readback
+    needs_calib = None
```
aberges-SLAC, line 1842: `:NEEDS_CALIB` is actually just a `calc` record that looks at the `HAS_ENCODER` and `IS_CALIBRATED` bits of the device channel state flags, if we wanted to keep compatibility at the IOC/pytmc level. I don't really see a need for it tbh and I agree with the decision here

pcdsdevices/epics_motor.py
```diff
@@ -1523,6 +1523,18 @@ class MotorDisabledError(Exception):
...
+    Assign ``long_name`` to a component signal if it exists.
+
+    Subclasses may remove unsupported components by setting the corresponding
+    Component to None. This helper skips those so that ``__init__`` long-name
+    assignments do not raise AttributeError.
+    """
+    if signal is not None:
+        signal.long_name = long_name
+
+
```
aberges-SLAC, line 1537: This is clever, I wish I thought of doing it this way originally.

Did you make sure this doesn't break the expert screen for `ioc-common-smaract` deployments?

ZLLentz (dismissed):
Leaving my approval- if Adam is happy I think we should merge

This is extremely thorough and well-documented. There is only a minimal amount of duplication and not in a way that will be a maintenance issue (it's probably worse to deduplicate).

I think the channel repointing you're doing in the screen class, while verbose, is better for maintenance than the alternatives (having two screen uis/pys, for example)

There are a few places where you make the ui fail safely- e.g. not crash when something isn't as we expect, e.g. missing widgets. While normally "fail loud" is better in e.g. hutch-python when data collection is involved, in a screen it's appropriate and correct to do it the way you did.

I left an optional suggestion that you can implement or ignore at your preference, it doesn't really matter that much.

ZLLentz:
Regarding the last checklist item:

> New/changed functions and methods are covered in the test suite where possible

I think it's correct to skip that for this PR. There's no meaningful test to do here without hardware.

pcdsdevices/epics_motor.py
```diff
@@ -1724,6 +1736,198 @@ def __init__(self, prefix='', *, tip_pv, tilt_pv, **kwargs):
...
+    diag_closed_loop_freq_max = Cpt(PytmcSignal, ':DIAG_CLF_MAX', io='i',
+                                    kind='normal',
+                                    doc='Diagnostic closed-loop frequency, '
+                                        'max (0x202A:1)')
+    diag_closed_loop_freq_avg = Cpt(PytmcSignal, ':DIAG_CLF_AVG', io='i',
+                                    kind='normal',
+                                    doc='Diagnostic closed-loop frequency, '
+                                        'avg (0x202A:2)')
+
+    # Calibrate command is a plain bo record, not a .PROC field
```
aberges-SLAC, line 1874: Calibrate is a `bo` field but it doesn't actually write anything out from the record, we're just hooking into the record processing. In `ioc-common-smaract` at the `proto` layer it is just sending the same command regardless of what is passed. 

That is why is important to not have it toggle back after processing or it will run the sequence twice, which is disastrous for the encoded tip-tilts

aberges-SLAC (commented):
This all looks great to me and I think you did an excellent job accommodating for the DS402 model. 

I think as long as we can test some hardware with the new version of the screen on something dependent on `ioc-common-smaract` records just to make sure things like `DO_CALIB` are proc'ing correctly— I remember this being weird in the past, which is why I needed to set to `.PROC` explicitly— then I think we'll be good to merge.

yannS2016:
> This all looks great to me and I think you did an excellent job accommodating for the DS402 model.
> 
> I think as long as we can test some hardware with the new version of the screen on something dependent on `ioc-common-smaract` records just to make sure things like `DO_CALIB` are proc'ing correctly— I remember this being weird in the past, which is why I needed to set to `.PROC` explicitly— then I think we'll be good to merge.

Yes, lets do that. i do have the hardware, just need to run an ioc based on 'ioc-common-smaract' do you have a sample i can use?

yannS2016:
Thansk evreryone for the thorough reviews. i'll coordinate with @aberges-SLAC to run a test against the serial interface to unsure that nothing is broken.

yannS2016:
> Leaving my approval- if Adam is happy I think we should merge
> 
> This is extremely thorough and well-documented. There is only a minimal amount of duplication and not in a way that will be a maintenance issue (it's probably worse to deduplicate).
> 
> I think the channel repointing you're doing in the screen class, while verbose, is better for maintenance than the alternatives (having two screen uis/pys, for example)
> 
> There are a few places where you make the ui fail safely- e.g. not crash when something isn't as we expect, e.g. missing widgets. While normally "fail loud" is better in e.g. hutch-python when data collection is involved, in a screen it's appropriate and correct to do it the way you did.
> 
> I left an optional suggestion that you can implement or ignore at your preference, it doesn't really matter that much.

Sure, i'll  look into it.

yannS2016:
> Thansk evreryone for the through reviews. i'll coordinate with @aberges-SLAC to run a test against the serial interface to unsure that nothing is broken.

pcdsdevices/ui/SmarAct.detailed.py
```diff
@@ -146,6 +147,126 @@ def fix_pvs(self):
...
+        # Safe direction (0x200C) is EtherCAT-only, so the serial .ui has no
+        # widget for it. Add a read-only row to the Configuration tab so
+        # operators can see the end-stop homing/calibration reference direction.
+        self._add_safe_direction_row(prefix)
+
+    def _add_safe_direction_row(self, prefix: str):
+        """
+        Append a read-only Safe Direction row to the Configuration tab grid.
+
+        ``safe_direction`` (SDO 0x200C) exists only on the EtherCAT DS402
```
aberges-SLAC, line 248: Really appreciate the documentation on the behavior nuances in this, btw

pcdsdevices/epics_motor.py
yannS2016: sure, for the plc interface, DO_CALIB would be the  calibration  command similar to the home command the PLC monitors to perfoms a calibration move. the plc acts on a rising edge. so the command can't be repeated twice.

KaushikMalapati:
@yannS2016 is this going to be used for the dream slits/sample paddle?

yannS2016:
> @yannS2016 is this going to be used for the dream slits/sample paddle?

yes!

KaushikMalapati:
@yannS2016 are you planning on updating tests/test_epics_motor.py for the new class or epics_motor._GetMotorClass or the Component Key + Python Class  table comment in the same file?

yannS2016:
> @yannS2016 are you planning on updating tests/test_epics_motor.py for the new class or epics_motor._GetMotorClass or the Component Key + Python Class table comment in the same file?

i could look into that, since exsting classes cover most usefull tests, I wasn't sure what new tests to add.

KaushikMalapati:
@yannS2016 you could do something like this to just have the new class put through the existing test cases
image
and this would be in epics_motor
image

both taken from my old branch used for the dream screens

yannS2016:
> @yannS2016 you could do something like this to just have the new class put through the existing test cases image and this would be in epics_motor image
> 
> both taken from my old branch used for the dream screens

Are you suggestting we use SmarActBeckhoff instead of SmarActEtehrCAT? also can we tell that all SmarActEtherCAT/Beckhoff motors  would have the key MMT? well maybe this is not an issue since it defaults to an EpicsMotor otherwise. sure i'll include it in the tests.

KaushikMalapati:
@yannS2016 I'm not suggesting a name change, just suggesting we add it to the tests and add a key if you know that all SmarActEtherCAT axes will use it (doesn't have to be MMT or we can not have a key if you think that's better)

yannS2016:
> @yannS2016 I'm not suggesting a name change, just suggesting we add it to the tests and add a key if you know that all SmarActEtherCAT axes will use it (doesn't have to be MMT or we can not have a key if you think that's better)

can't tell if all smaractethercat axes will have the MMT key. If limited to DREAM MNT it's okay, but i forsee this would be used beyond DREAM. Unless we want to enforce MMT, a key here could be misleading. well i'm not rigid on the class name, if you guys think SmarActBeckhoff is semantically better that's okay for me too.

KaushikMalapati:
I think EtherCAT is more correct than Beckhoff here because DS402 is not limited to Beckhoff. There's no reason to enforce MMT and I think I remember someone suggesting we move away from that anyways

ZLLentz:
Poking my head in here to triage because I thought this looked good- what's left is:

Most important, to be done prior to merge:
- Testing to make sure the serial smaract handling didn't break (I assume Adam is helping with this)

Maintenance, could be done now or could be documented to do later:
- Add the new class to the list for the existing many-class unit tests
- Add the new class to the `Motor` function (but maybe this isn't appropriate here?)

And then some smaller nitpicks, which can be done or forgotten.

Is this correct? I'm trying to be more clear about what is "required scope" in reviews because I think it's not so clear sometimes.

yannS2016:
> Poking my head in here to triage because I thought this looked good- what's left is:
> 
> Most important, to be done prior to merge:
> 
> * Testing to make sure the serial smaract handling didn't break (I assume Adam is helping with this)
> 
> Maintenance, could be done now or could be documented to do later:
> 
> * Add the new class to the list for the existing many-class unit tests
> * Add the new class to the `Motor` function (but maybe this isn't appropriate here?)
> 
> And then some smaller nitpicks, which can be done or forgotten.
> 
> Is this correct? I'm trying to be more clear about what is "required scope" in reviews because I think it's not so clear sometimes.

Yes that's correct, the serial smaract interface test were finalized, and it looks good. smaractethercat was added to the unit tests. as to adding the new class to the 'Motor' function probably not appriate at this stage.

aberges-SLAC (dismissed):
Leaving my approval after working with Yann earlier today on the hardware!

pcdsdevices/ui/SmarAct.detailed.py
yannS2016: Fixed the above.

ZLLentz (approved):
Looks good, thanks for doing the thorough testing!

I've confirmed that the new class is showing up in the test suite too:
```
pcdsdevices/tests/test_epics_motor.py::test_epics_motor_soft_limits[SmarActEtherCAT] PASSED [  7%]
```

aberges-SLAC (approved):

PR #1425, opened 2026-06-17
Added Kappa stages

jcotter814:
## Description
Added new Kappa motion stages to Kappa class, updated Kappa API, deleted obsolete class KappaXYZStage, changed sign of eta coordinate in kappa -> euler and euler -> kappa coordinate conversions per hutch scientist request, updated tests to accommodate changes to API and calculations.

A note/question - I rebased and now if this pull request gets merged, the log will have duplicate commits with different hashes... Is a git squash enough to correct this or should I backtrack and undo the rebase?

## Motivation and Context
Changes have been made for newly installed XPP Kappa and pcdsdevice class was modified to reflect these changes.

## How Has This Been Tested?
Kappa motion has been tested in hutch and verified to work. Test suite passes locally in pcds_conda environment.

## Where Has This Been Documented?
In the docstring of the class, here.

## Pre-merge checklist
- [X] Code works interactively
- [X] Code contains descriptive docstrings, including context and API
- [X] New/changed functions and methods are covered in the test suite where possible
- [X] Test suite passes locally

PR #1440, opened 2026-08-11
Restore xyzstage

carogee:
## Description

XYZStage was removed in b466a6cd ("Added Kappa stages #1425"), but is still used by XCS so restoring it in gon.py.

## Motivation and Context

XCS still uses this.

## How Has This Been Tested?

Tested in XCS hutch-python.

## Pre-merge checklist

### DIFF
--- a/pcdsdevices/gon.py
+++ b/pcdsdevices/gon.py
@@ -202,6 +202,38 @@ def Goniometer(**kwargs):
         return BaseGon(**kwargs)
 
 
+class XYZStage(BaseInterface, Device):
+    """
+    Sample XYZ stage.
+
+    Parameters
+    ----------
+    name : str
+        A name to refer to the device
+
+    prefix_x : str
+        The EPICS base PV of the sample-stage's x motor.
+
+    prefix_y : str
+        The EPICS base PV of the sample-stage's y motor.
+
+    prefix_z : str
+        The EPICS base PV of the sample-stage's z motor.
+    """
+
+    x = FCpt(IMS, '{self._prefix_x}', kind='normal')
+    y = FCpt(IMS, '{self._prefix_y}', kind='normal')
+    z = FCpt(IMS, '{self._prefix_z}', kind='normal')
+
+    tab_component_names = True
+
+    def __init__(self, *, name, prefix_x, prefix_y, prefix_z, **kwargs):
+        self._prefix_x = prefix_x
+        self._prefix_y = prefix_y
+        self._prefix_z = prefix_z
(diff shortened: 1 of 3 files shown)


KaushikMalapati:
@carogee can you run pre-commit and do you know why this was previously removed?

pcdsdevices/gon.py
```diff
@@ -202,6 +202,38 @@ def Goniometer(**kwargs):
         return BaseGon(**kwargs)
 
 
+class XYZStage(BaseInterface, Device):
```
KaushikMalapati: This used to inherit from GroupDevice instead of Device and also overrode format_status_info

ZLLentz:
> do you know why this was previously removed?

It was removed in #1425 because we didn't realize that it was not only used by the kappa
Since we thought it was kappa-only, it erroneously was deleted along with the old decommissioned kappa code

pcdsdevices/gon.py
ZLLentz: I think we ought to copy exactly the version that was deleted in #1425 

<details>

```
class XYZStage(BaseInterface, GroupDevice):
    """
    Sample XYZ stage.
    Parameters
    ----------
    name : str
        A name to refer to the device
    prefix_x : str
        The EPICS base PV of the sample-stage's x motor.
    prefix_y : str
        The EPICS base PV of the sample-stage's y motor.
    prefix_z : str
        The EPICS base PV of the sample-stage's z motor.
    """

    x = FCpt(IMS, '{self._prefix_x}', kind='normal')
    y = FCpt(IMS, '{self._prefix_y}', kind='normal')
    z = FCpt(IMS, '{self._prefix_z}', kind='normal')

    tab_component_names = True

    def __init__(self, *, name, prefix_x, prefix_y, prefix_z, **kwargs):
        self._prefix_x = prefix_x
        self._prefix_y = prefix_y
        self._prefix_z = prefix_z
        super().__init__('', name=name, **kwargs)

    def format_status_info(self, status_info):
        """Override status info handler to render the `XYZStage`."""
        x = get_status_float(status_info, 'x', 'position')
        y = get_status_float(status_info, 'y', 'position')
        z = get_status_float(status_info, 'z', 'position')
        units = get_status_value(status_info, 'x', 'user_setpoint', 'units')

        return f"""\
XYZStage
X, Y, Z: {x}, {y}, {z} [{units}]
"""
```

</details>

ZLLentz (approved):
Looks like a faithful recovery to me! Thanks for including the unit tests snippit too!

